PR #48, opened 2025-08-28
feat: logout and login

MiguelMedeiros:
Tasks:

- [x] logout
- [x] sign in with seed phrase
- [x] export to Pubky Ring
- [x] sign in with encrypted file
- [x] sign in with http relay
- [x] guard routes


catch-21:
Verified in 34e77c2cc87dbb14c8bd89fd647ee7b910551062
Tested sign in through every path and sign out. Redirect from root when signed in. Export to pubky-ring during signup flow.

PR #60, opened 2025-09-10
tests: init component snapshot testing

catch-21:
Introduce component snapshot testing to Franky.
https://vitest.dev/guide/snapshot

todo:
- [x] address hydration error with nested button
- [ ] address "without an `onChange` handler" error (not introduced in snapshot tests)
- [x] add snapshot tests to components with existing ui tests
- [x] add snapshot tests to remaining components
- [x] review tests with dev for usefulness
- [x] remove redundant non-snapshot UI tests such as specific css class checks
- [x] address missing `message` prop in `TestareaField` snapshot tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,13 @@
 
 # Franky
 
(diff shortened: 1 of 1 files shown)


MiguelMedeiros (approved):
ACK

PR #143, opened 2025-09-23
refactor: component `Button`

MiguelMedeiros:
## Changes
- Refactored Button component following project design patterns
- Extracted types to separate file (`Button.types.ts`)
- Added `dark` and `dark-outline` variants from Figma
- Removed unused `buttonVariants` export
- Enhanced tests for all 9 variants


catch-21:
Button colour looks different to design to me. have raised the following question why we take the approach of using transparency and oklch when figma is using hex, the values don't seem to match from my inspection.

The Button atom appears to be using `bg-brand/20` which is not totally opaque meaning the background will interfere with the brand colour, which is different depending on light/dark like `oklch(0.695 0.172 123.587)` that translates to `#88AD00` and `oklch(0.928 0.23 123.978)` that translates to `#C8FF03`, meanwhile Figma shows `#C8FF00`. Why this complexity and discrepancy?

PR #145, opened 2025-09-23
refactor: component `Badge`

MiguelMedeiros:
## Changes
- Refactored Badge component following project design patterns
- Extracted types to separate file (`Badge.types.ts`)
- Added `data-testid` and `data-variant` attributes
- Enhanced tests for all 4 variants (12 tests passing)
- Created comprehensive examples showcase (`Badge.examples.tsx`)


src/components/atoms/Badge/Badge.tsx
```diff
@@ -22,15 +24,28 @@ const badgeVariants = cva(
...
-  variant,
-  asChild = false,
-  ...props
-}: React.ComponentProps<'span'> & VariantProps<typeof badgeVariants> & { asChild?: boolean }) {
-  const Comp = asChild ? Slot : 'span';
+export interface BadgeProps extends React.HTMLAttributes<HTMLDivElement>, VariantProps<typeof badgeVariants> {
+  asChild?: boolean;
+}
+
+function Badge({ className, variant, asChild = false, ...props }: BadgeProps) {
```
infin1t3, line 31: Consider using the same approach here, like in other component MRs you opened. Using `const Badge = React.forwardRef<..`

src/components/atoms/Badge/Badge.tsx
```diff
@@ -22,15 +24,28 @@ const badgeVariants = cva(
...
+  const effectiveVariant = variant || 'default';
```
infin1t3, line 32: // Line 32: Unnecessary fallback
const effectiveVariant = variant || 'default';

CVA already has defaultVariants
```
defaultVariants: {
  variant: 'default',
}
```

src/components/atoms/Badge/Badge.tsx
```diff
@@ -22,15 +24,28 @@ const badgeVariants = cva(
...
+
+  const getTestId = () => {
```
infin1t3, line 34: Why do we need test logic? Also - this function is recreated on every render.

src/components/atoms/Badge/Badge.tsx
```diff
@@ -22,15 +24,28 @@ const badgeVariants = cva(
...
+    return `badge-${effectiveVariant}`;
+  };
+
+  const Comp = asChild ? Slot : 'div';
```
infin1t3, line 38: You use HTMLDiv interface for component type, but here you suggest it can be polymorphic

PR #163, opened 2025-09-24
fix: bg home mobile

flaviomoceri:
Screenshot 2025-09-24 alle 12 31 10


catch-21:
Verified in 1c41d10ec9ac84a5eadfe52ea97ae431ab4ddd7d

Smaller devices show less of this background, but most are larger and look ok. Probably not worth addressing.

Screenshot 2025-09-26 at 12 51 10

Screenshot 2025-09-26 at 12 52 29

PR #166, opened 2025-09-24
fix: home footer height

flaviomoceri:
Screenshot 2025-09-24 alle 13 52 22


MiguelMedeiros (approved):
ACK tested.

catch-21:
Tested on 6e02be58bb8702b55574fae1c118996817dc6da6

**Design**
Screenshot 2025-10-22 at 16 37 41


**Actual**
Screenshot 2025-10-22 at 16 37 35

The spacing below the buttons looks not enough to me. @infin1t3 can you explain to me, using tailwind, what would make this gap 68px.

PR #167, opened 2025-09-24
fix: add share button

flaviomoceri:


catch-21:
Tested work on mobile 1c41d10ec9ac84a5eadfe52ea97ae431ab4ddd7d
The share button is only meant to be on mobile, not desktop too: https://github.com/pubky/franky/issues/182

PR #171, opened 2025-09-25
bug: bootstrap errors during profile creation

tipogi:
Improves user experience during profile creation by adding proper error handling and user feedback when the bootstrap process fails due to indexing delays.


catch-21:
Verified in 824386e551cc49aae16689f238e08e80b6b1d1c2
No longer seeing rapid-fire requests for indexed profile nor the errors.

PR #187, opened 2025-09-29
fix: toaster UI

flaviomoceri:


MiguelMedeiros:
We will have to change this component to Sooner here is the docs:https://ui.shadcn.com/docs/components/sonner

MiguelMedeiros (changes requested):
We will have to change this component to Sooner here is the docs:https://ui.shadcn.com/docs/components/sonner

flaviomoceri:
Done

PR #190, opened 2025-09-29
Vj/fix avatar signup issue

infin1t3:


catch-21 (approved):
Appears to resolve the bug where redirect to `/` was happening after valid invite code used.

PR #196, opened 2025-09-30
bug: post models ids wrong

tipogi:
`post.details.id` is Crockford Base32 strings derived from timestamps. If two users post at the same time, the id will be the same. To avoid that, we need to use the `authorId:postId` format to ensure the id is unique.


afterburn (approved):

infin1t3 (approved):

MiguelMedeiros (commented):
ACK

PR #213, opened 2025-10-03
[DRAFT] feat: use homeserver for post tagging + initial setup application folder

afterburn:
DRAFT PR so we can discuss what @tipogi @MiguelMedeiros want to change about the controllers / application design, if anything


src/core/application/tag/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Tag } from './tag';
+export type { TCreateTagInput, TDeleteTagInput } from './tag.types';
```
MiguelMedeiros, line 2:
```suggestion
export * from './tag';
export * from './tag.types';
```
Can we export like that?

tipogi (approved):
LGTM 🚀

PR #214, opened 2025-10-03
fix: incorrect post timestamps

afterburn:
We currently show `user.indexed_at` instead of `post.indexed_at` as the timestamp


catch-21:
Verified on f5881d61ed3d26d3a1ea4afdcbfc270e14757803

Correct time value. We need to address the format, raised https://github.com/pubky/franky/issues/217

catch-21 (approved):

tipogi (approved):

PR #215, opened 2025-10-03
feat: nested reply link

afterburn:
Enable users to click a reply to open the post detail page with that reply as the root, allowing navigation into deeply nested threads.


catch-21 (dismissed):
Verified in f5881d61ed3d26d3a1ea4afdcbfc270e14757803
Is now possible to navigate to nested replies.

tipogi (dismissed):

MiguelMedeiros (approved):

MiguelMedeiros (approved):

PR #216, opened 2025-10-03
fix: remove hardcoded SN from header avatar

afterburn:
Replace hardcoded `SN` with the users initial, in future pr we should probably consolidate this logic somewhere/make a helper.


src/components/molecules/Header/Header.tsx
MiguelMedeiros, line 1: Molecules should never have liveQuerys.
This should be moved into an organism.

MiguelMedeiros (approved):

PR #227, opened 2025-10-04
fix: header mobile

flaviomoceri:


catch-21 (approved):
Onboarding and sign in header now shows on mobile and desktop

PR #230, opened 2025-10-05
feat: tag stream model

tipogi:


catch-21:
Verified on e70ae1acde02b338073ab88a999ea7214d09b50a

Confirm `user_streams` is created on sign up with correct values from nexus.

PR #241, opened 2025-10-06
Hide onboarding share button on desktop

SHAcollision:
## Summary
- hide the PublicKeyCard share button on non-touch devices so onboarding desktop UI no longer shows it
- extend the PublicKeyCard tests to cover touch vs non-touch rendering and silence console noise from the share error path

## Testing
- npm test -- PublicKeyCard

------
https://chatgpt.com/codex/tasks/task_e_68e3db74c8988329aba8a1357fe179de


MiguelMedeiros (approved):

catch-21:
This approach works and updates dynamically when detecting touch or non-touch device, however we may find there's some devices that i'll show the share button on that are not mobile. It might be best to detect based on screen size rather than whether it is a touch device.

catch-21:
Raised https://github.com/pubky/franky/issues/265 to track this proposed improvement

PR #246, opened 2025-10-07
chore: eliminate network dependency for fonts at build time

SHAcollision:
Local pre-commit hooks randomly fail due to transient network errors. It seems the `build` check requires network to download fonts. I eliminated the network dependency:

- Updated the root container to load Inter Tight via `next/font/local`, pointing at the new checked-in WOFF2 assets so builds no longer hit Google Fonts at compile time.
- Swapped the Vitest font mock to target `next/font/local`, keeping tests aligned with the new font-loading approach.


SHAcollision:
> @SHAcollision seems to me that the fonts are not matching mb missing more .woff2 font variations?
> 
> image image

I've not been able to see these differences :thinking: Maybe because google fonts are anyway failing always for me.

I tried adding ALL the google fonts. But that made the frontend always pre-fetch +10 files on first load, not a great solution.

I opted for:
- Registered local Inter Tight variable fonts using CSS system with fallbacks to eliminate build-time network calls while preserving rendering quality. That means I could remove the `next/fonts/google` dependency)
- Added a root `head` component that preconnects to Google Fonts and serves their stylesheet at runtime so the hosted files from google remain the primary source when available: ensures it renders the same as before. But if google files cannot be found, it serves them locally from `/public`.
- Simplified the root container and test bootstrap now that `next/font` is unnecessary, relying on global CSS for typography.

PR #248, opened 2025-10-07
feat: component post

MiguelMedeiros:


src/core/services/local/stream/users/users.ts
```diff
@@ -0,0 +1,51 @@
...
+      const undefinedUsers = filteredUsers.reduce((acc, user, index) => {
+        if (user === undefined) {
+          acc.push(users[index]);
+        }
+        return acc;
+      }, [] as string[]);
+      // remove duplicates
+      const uniqueUndefinedUsers = [...new Set(undefinedUsers)];
+
+      const usersData = await Core.NexusUserStreamService.fetchByIds({ user_ids: uniqueUndefinedUsers });
```
tipogi: That one is using an external service, instead of return `void` we could return the `userIds` and fetch in the application layer

src/core/services/local/stream/posts/posts.ts
```diff
@@ -0,0 +1,156 @@
...
+      const posts: Core.NexusPost[] = nexusPosts;
+
+      // No posts available -> signal no update
+      if (!posts || posts.length === 0) {
+        return null;
+      }
+
+      // Persist to local database
+      const users = posts.map((post) => post.details.author);
+      await Core.LocalStreamUsersService.persistUserStream(users);
```
tipogi: Not sure if this is the place to do that one. It feels more that we should do it in the `application` layer that task

src/core/services/local/bootstrap/bootstrap.ts
```diff
@@ -0,0 +1,61 @@
...
+
+      Libs.Logger.debug('Bootstrap data persisted successfully', {
+        users: data.users.length,
+        posts: data.posts.length,
+        streams: {
+          timeline: data.list.stream.length,
+          influencers: data.list.influencers.length,
+          recommended: data.list.recommended.length,
+          hotTags: data.list.hot_tags.length,
+        },
```
tipogi: do we really need that one, we can find that info in the network tab

src/core/application/profile/profile.ts
```diff
@@ -1,20 +1,24 @@
...
   }
 
-  static async create({ profile, url, pubky }: TCreateProfileInput) {
+  static async create({ profile, url, pubky }: Core.TCreateProfileInput) {
     try {
       await Core.HomeserverService.request(Core.HomeserverAction.PUT, url, profile.toJson());
-      // Initialise authentication store
+      // Initialize authentication store
       Core.useAuthStore.getState().setCurrentUserPubky(pubky);
       Core.useAuthStore.getState().setAuthenticated(true);
```
tipogi: if we give the responsibility to edit the store to the `controllers`, that ones should be moved to its controller

src/core/application/bootstrap/bootstrap.ts
```diff
@@ -0,0 +1,29 @@
+import * as Core from '@/core';
+import * as Libs from '@/libs';
+
+export class BootstrapApplication {
+  static async read(pubky: Core.Pubky) {
+    const data = await Core.NexusBootstrapService.fetch(pubky);
+    await Core.LocalBootstrapService.persistBootstrap(data);
```
tipogi: as we comment, we could call to each `service` to persist data

src/components/organisms/PostContent/PostContent.tsx
```diff
@@ -1,17 +1,28 @@
...
   postId: string;
+  className?: string;
 }
 
-export function PostContent({ postId }: PostContentProps) {
-  const postDetails = useLiveQuery(() => Core.db.post_details.get(postId).then((details) => details), [postId]);
+export function PostContent({ postId, className }: PostContentOrganismProps) {
+  // Fetch post details for content
+  const postDetails = useLiveQuery(async () => {
+    return await Core.PostDetailsModel.findById(postId);
```
tipogi: if we access to model directly, we do it in all. If not, we need to fix because in `PostHeader` we use 
```js
...
// in that case the function name would be, 
return await Core.PostController.getDetails({ postId });
// would be, getDetails
return await Core.ProfileController.read({ userId: userId as Core.Pubky });
```
Not sure also, what is the best way for dexie to be more performant and also which is the best way to subscribe to a row

src/components/organisms/PostHeader/PostHeader.tsx
```diff
@@ -0,0 +1,58 @@
...
+            @{Libs.formatPublicKey({ key: userId, length: 8 })}
+          </span>
+        </div>
+      </div>
+      {Libs.timeAgo(new Date(postDetails.indexed_at)) && (
+        <div className="flex items-center gap-1 text-xs font-medium tracking-[0.075em] text-muted-foreground">
+          <Libs.Clock className="size-4 text-muted-foreground" />
+          <span>{Libs.timeAgo(new Date(postDetails.indexed_at))}</span>
+        </div>
+      )}
```
tipogi: future: can we avoid that to render here and render in the post content organism as an absolute position to avoid that subscription?

tipogi (changes requested):

PR #260, opened 2025-10-08
fix: profile avatar scaling

### LINKED ISSUE #148: [Bug]: Profile avatar image doesn't retain scaled dimensions
### Describe the bug

When adding an avatar to profile (during sign up) the image gets stretched or squashed. It should be possible to scale it to get a precise thumbnail of the uploaded image file as we do in pubky-app.

### Reproduce

1. Follow sign up to 'Create your profile' page
2. Click 'Choose file' button
3. Select an image file from local storage (ideally one that is not perfectly square)
4. Observe image is not the same dimensions as the original, it's stretched or squashed

### Screenshots / Recording

SVG
Image

Image

-----

Jpeg

Image

Image

### Web Browser

_No response_

### Pubky App Version

34e77c2cc87dbb14c8bd89fd647ee7b910551062

### Console log output

```shell

```

SHAcollision:
fixes #148 

- Added a `DialogCropImage` molecule that uses react-easy-crop to let users crop and zoom a square avatar within a dedicated modal. 
- Implemented a reusable `cropImageToBlob` helper and exposed it through the shared libs barrel so clients can persist cropped blobs with consistent quality settings. 
- Wired the new modal into the create-profile flow, managing pending previews/cleanup, improving validation recovery, and updating the test harness to exercise the cropping workflow.

Screenshot from 2025-10-08 16-04-21
Screenshot from 2025-10-08 16-04-44
Screenshot from 2025-10-08 16-04-56

- todo: tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "qrcode.react": "^4.2.0",
         "react": "^19.0.0",
         "react-dom": "^19.0.0",
+        "react-easy-crop": "^5.5.3",
         "sonner": "^2.0.6",
         "tailwind-merge": "^3.3.0",
         "zod": "^3.25.36",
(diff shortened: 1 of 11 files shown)


catch-21:
This is how it looks in pubky-app. Very similar but need to use mouse wheel to scale. Maybe we don't need the descriptive text?

Screenshot 2025-10-08 at 15 04 02

PR #271, opened 2025-10-09
fix: confirm recovery phrase does not trunkate words

SHAcollision:
Follow up of #261 and comment https://github.com/pubky/franky/pull/261#issuecomment-3382750362

- Updated the readonly recovery word slot container to position the index badge absolutely and stretch the slot to full width, aligning layout with the editable version.
- Trimmed input padding and forced flex growth for readonly slots so longer recovery words remain fully visible while preserving error/success coloring.

image


catch-21:
verified in fa15d906d8d8238738b6aa9be3c1e9dee77dd3c4

PR #273, opened 2025-10-09
fix: regression header is not visible on small screens

### LINKED ISSUE #270: [Bug]: Regression: no header or background on smaller viewports
### Describe the bug

Did not fully find the source of this regression, but it is recent. It might be  #239

When the viewport is smaller than 1024px, the header and background is not shown.

### Reproduce

1. Open franky
2. make the viewport smaller than 1024

### Screenshots / Recording

Regular:

Image

Regression:

Image

### Web Browser

_No response_

### Pubky App Version

_No response_

### Console log output

```shell

```

SHAcollision:
fixes #270

Mostly fixes it but this is likely a **suboptimal solution**

- Wrapped the global header content in a full-width sticky <header> so its gradient and layout remain visible across onboarding and other routes.
- Updated the header unit tests and snapshots to assert the new structure and styling classes introduced by the header wrapper.
- Hid the desktop header navigation group behind the lg breakpoint so the feed only shows the mobile toolbar on small screens. 
- Let HeaderContainer accept an optional className so the sticky gradient wrapper can share styling logic while allowing conditional visibility tweaks.
- Hide the desktop header on small screens when a user is signed in outside onboarding, preserving the onboarding header experience.
- Expanded header unit tests to cover custom class merging and the new mobile visibility rules.

### DIFF
--- a/src/components/molecules/Header/Header.test.tsx.snap
+++ b/src/components/molecules/Header/Header.test.tsx.snap
@@ -35,7 +35,7 @@ exports[`Header Components - Snapshots > matches snapshot for HeaderButtonSignIn
 
 exports[`Header Components - Snapshots > matches snapshot for HeaderContainer 1`] = `
 <header
-  class="sticky top-0 z-20 w-full bg-gradient-to-b from-[var(--background)]/95 to-[var(--transparent)] backdrop-blur-sm"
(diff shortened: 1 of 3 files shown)


catch-21:
Coverage added for E2E test in https://github.com/pubky/franky/pull/315

PR #274, opened 2025-10-09
fix: indexdb detect version mismatch

### LINKED ISSUE #194: [Bug]: IndexedDB doesn't detect version mismatch
### Describe the bug

When IndexedDB instance 'franky' is a different version than the one set with `NEXT_PUBLIC_DB_VERSION`, accounting for [transformation](https://github.com/pubky/franky/blob/b09fbd532748789adb31b2be209b427582dcb3e4/src/libs/env/env.ts#L18-L23), then the DB is not recreated even though we have logic in place to do so, see [in initilisation](https://github.com/pubky/franky/blob/b09fbd532748789adb31b2be209b427582dcb3e4/src/core/database/franky/franky.ts#L81).

We need to first decide whether we even need versioning. If so, define rules around it, how and when versions increment, and when Franky should detect and handle them.

### Reproduce

1. Sign up or sign in
2. Open dev tools to observe IndexedDB 'franky' database version (probably `10` by default after transformation from `1`)
3. Delete some entry in the database
4. Sign out
5. Open database manually with a different version by running the following in console `indexedDB.open("franky", 69);`
6. Sign in
7. Observe 'franky' database remains version 6, no console log like "Database version mismatch" and entry deleted from step 3 is still missing

### Screenshots / Recording

_No response_

### Web Browser

_No response_

### Pubky App Version

824386e551cc49aae16689f238e08e80b6b1d1c2

### Console log output

```shell

```

SHAcollision:
fixes #194

reckless PR needs an in-depth review by @MiguelMedeiros because I don't fully understand all of the changes.

- Added an IndexedDB version discovery helper, guarded initialization when the API is unavailable, and centralized database recreation logic to enforce schema version mismatches.

- Updated the Franky database initialization tests to mock the new helper behavior and cover both mismatch and indeterminate version scenarios.

- Added native IndexedDB helper utilities and an end-to-end integration test that mirrors the manual reproduction steps to confirm mismatched versions trigger database recreation and remove legacy stores.

- Normalized stored IndexedDB versions during initialization so Dexie’s internal multiplier no longer triggers false mismatches and added raw-version metadata when recreating the Franky database.

- Deferred importing the core database in the Vitest setup until after environment variables are applied, ensuring tests see the intended schema version.

### DIFF
--- a/src/core/database/franky/franky.test.ts
+++ b/src/core/database/franky/franky.test.ts
@@ -132,7 +132,9 @@ describe('Database Initialization', () => {
       ([message]) => typeof message === 'string' && message.startsWith('Database version mismatch'),
     );
 
-    expect(mismatchLog?.[0]).toBe(`Database version mismatch. Current: ${injectedVersion}, Expected: ${DB_VERSION}`);
+    // The normalized version will be injectedVersion / 10
+    const normalizedVersion = injectedVersion / 10;
+    expect(mismatchLog?.[0]).toBe(`Database version mismatch. Current: ${normalizedVersion}, Expected: ${DB_VERSION}`);
 
     expect(reproductionDb.verno).toBe(DB_VERSION);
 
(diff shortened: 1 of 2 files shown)


catch-21:
for QA see https://github.com/pubky/franky/issues/194#issuecomment-3404679193

PR #276, opened 2025-10-09
feat: follow event

tipogi:


catch-21:
will test once integrated with UI

PR #289, opened 2025-10-11
fix: update DialogConfirmBackup tests and add snapshot support

MiguelMedeiros:
Resolves https://github.com/pubky/franky/issues/221

### Changes:
- Removed mock implementations for atoms and libs, using actual implementations instead.
- Added a snapshot test for the DialogConfirmBackup component.
- Updated the DialogConfirmBackup component to include a new class for the 'Done' button.
- Cleaned up the test structure for better readability and maintainability.


catch-21:
Looks like `libs` was already addressed but importing real `atoms` is something we should do for all tests.

PR #290, opened 2025-10-11
fix: secret key always renewed when creating new account

MiguelMedeiros:
Resolves https://github.com/pubky/franky/issues/144


src/components/molecules/Install/Install.test.tsx
```diff
@@ -1,148 +1,59 @@
...
 
-// Mock Next.js router
-const mockPush = vi.fn();
-vi.mock('next/navigation', () => ({
-  useRouter: () => ({
-    push: mockPush,
-  }),
-}));
+// Helper function to normalize Radix UI IDs for snapshot testing
+const normalizeRadixIds = (html: string): string => {
```
catch-21: This already exists in https://github.com/pubky/franky/blob/6e02be58bb8702b55574fae1c118996817dc6da6/src/libs/utils/utils.ts#L73

PR #303, opened 2025-10-15
chore: storybook docs use background colour

catch-21:


catch-21:
Cool. If we want the background colour used for the docs background then can merge this.

**Before**
Screenshot 2025-10-16 at 06 06 46

**After**
Screenshot 2025-10-16 at 06 06 35

PR #313, opened 2025-10-16
feat: delete account

afterburn:
Data layer specific code for the delete account operation

### DIFF
--- a/src/core/services/local/user/user.ts
+++ b/src/core/services/local/user/user.ts
@@ -1,3 +1,0 @@
-export class LocalUserService {
-  // User service methods will be added here
-}
--- a/src/core/application/profile/profile.delete.test.ts
+++ b/src/core/application/profile/profile.delete.test.ts
@@ -14,12 +14,10 @@ vi.mock('@/core/services/homeserver', () => ({
   },
 }));
 
-// Mock Local.Profile service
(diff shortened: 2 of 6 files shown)


src/core/services/local/user/user.ts
```diff
@@ -0,0 +1,59 @@
...
+            // Post tables
+            Core.PostCountsModel.table.clear(),
+            Core.PostDetailsModel.table.clear(),
+            Core.PostRelationshipsModel.table.clear(),
+            Core.PostTagsModel.table.clear(),
+            Core.PostTtlModel.table.clear(),
+            // Stream tables
+            Core.PostStreamModel.table.clear(),
+            Core.UserStreamModel.table.clear(),
+            Core.TagStreamModel.table.clear(),
```
tipogi: maybe create in the abstract class that function of `clear()`and like this we do not reference `.table`?

afterburn:
@tipogi resolved your feedback:
- Moved delete account to profile system instead of user system.
- Added `clear()` to base models so we no longer have to call `.table.clear()`.
- Temporarily passing Infinity to `homeserver.list` so we make to delete all files instead of only first `500` until `homeserver.list` supports pagination/cursor.

tipogi (dismissed):

tipogi (approved):

PR #314, opened 2025-10-16
feat: download user data

afterburn:
Data layer specific code for downloading user data


afterburn:
@tipogi resolved your feedback:
- Moved download data method from user to profile system.
- Temporarily pass Infinity to homeserver list until it support pagination/cursor.

tipogi (approved):
LGTM

catch-21:
Additional test added in https://github.com/pubky/franky/pull/383
Will test further once integrated with UI

PR #316, opened 2025-10-17
fix: very long file name on recovery

### LINKED ISSUE #293: [Bug]: Long recovery file name causes fields and buttons to stretch off the screen
### Describe the bug

Similar to https://github.com/pubky/franky/issues/223, but instead of horizontal scroll the input fields and buttons go off off the screen to the right.   It might be best to truncate the text in the filename field.

### Reproduce

1. On mobile 
2. Go to sign in with recovery file
3. Use one with long filename
4. Observe input fields and buttons go off the right of the screen

### Screenshots / Recording

Image

### Web Browser

_No response_

### Pubky App Version

316efc3d80ed13045dccec11de98cb3b48e7fc1c

### Console log output

```shell

```

SHAcollision:
fixes #293

image

- Added a formatSelectedFileName helper that trims selected recovery filenames to 25 characters (preserving their extensions) and reused it when rendering the encrypted file picker. 

- Updated the dialog regression test to assert the truncated filename output and enforce the length constraint for long names

### DIFF
--- a/src/components/organisms/DialogRestoreEncryptedFile/DialogRestoreEncryptedFile.tsx
+++ b/src/components/organisms/DialogRestoreEncryptedFile/DialogRestoreEncryptedFile.tsx
@@ -8,28 +8,6 @@ import * as Core from '@/core';
 import * as Libs from '@/libs';
 import * as Hooks from '@/hooks';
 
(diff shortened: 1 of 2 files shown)


MiguelMedeiros (approved):
Tested ACK.

PR #329, opened 2025-10-26
doc: cursor core context + adr

tipogi:


MiguelMedeiros:
> Depends on what we are trying to do.
> 
> If we want to use the files just as documentation for an AI IDE (Cursor, Windsurf, Claude Code), we can just put them in any folder and write them in `.md` format. Then, we can reference those files in `AGENTS.md` and it will work. In that case, we don't have to put them in `.cursor` folder, we can make a new folder and call it `docs`
> 
> If we want to have actual Cursor rules, those files have to be in `.cursor/rules` and they have to be in `.mdc` format. Otherwise, Cursor will not recognize them as rules and will treat them as regular docs.
> 
> Source: https://cursor.com/docs/context/rules

We should use `cursor` folder with their pattern since all of the team mainly uses Cursor to code. But we can use `AGENTS.md` file as an index pointing to `.cursor/rules` folder so any other AI-IDE can use it as a context. 

Win win, IMHO.

MiguelMedeiros (approved):

PR #330, opened 2025-10-29
feat: notifications

tipogi:
## Overview
This PR implements a notification system that persists notification data in `IndexedDB` to show in user notification page. The system provides real-time notification state management with efficient polling capabilities.
### What's New
- **Notification Store**: New Zustand store with `{lastRead: number, unread: number}` state
- **IndexedDB Model**: New notification model for storing and querying notification data locally
- **Application Layer**: Polling service for initial state retrieval and incremental updates with API integration
- **Controller Layer**: Bootstrap integration and hook-ready methods for notification state management

### DIFF
--- a/src/core/models/notification/notification.ts
+++ b/src/core/models/notification/notification.ts
@@ -1,9 +1,8 @@
 import { Table } from 'dexie';
 import * as Core from '@/core';
-import { FlatNotification, NotificationType } from './notification.types';
 import * as Libs from '@/libs';
-
-const DEFAULT_LIMIT = 20;
+import * as Config from '@/config';
+import { FlatNotification, NotificationType } from './notification.types';
 
 // Primary key: auto-incrementing integer (++id) managed by Dexie
 export class NotificationModel {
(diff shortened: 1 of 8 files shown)


src/core/models/notification/notification.ts
```diff
@@ -0,0 +1,72 @@
+import { Table } from 'dexie';
+import * as Core from '@/core';
+import { FlatNotification, NotificationType } from './notification.types';
+import * as Libs from '@/libs';
+
+const DEFAULT_LIMIT = 20;
```
MiguelMedeiros: this should be a config variable. mb we should only have one limit set up (30).

src/core/services/polling/notifications/.gitkeep
MiguelMedeiros: Are you going to do pooling notification service in another PR?

MiguelMedeiros (approved):
ACK

catch-21:
Breaks E2E onboarding test due to expecting `last_read` to exit in homeserver after profile creation.
We could write a low value there to prevent edge-cases with its absence later on when viewing notifications.

Will test notifications fully once integrated with UI

PR #340, opened 2025-11-02
feat: bookmarks

tipogi:


catch-21:
Will test bookmarks once integrated with UI

PR #341, opened 2025-11-03
feat(app): reply dialog - ui

### LINKED ISSUE #401: [UI] Reply dialog
Implement UI for reply dialog

Figma:
https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=21607-105112&t=1m0HhjsXcOuEEsl4-4

infin1t3:
Resolves #401 

image
image


src/components/organisms/DialogReplyInput/DialogReplyInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const { replyContent, setReplyContent, handleReplySubmit } = Hooks.usePostReply({ postId, onSuccess });
+  const { ref: containerRef } = Hooks.useElementHeight();
+  const currentUserId = Core.useAuthStore((state) => state.selectCurrentUserPubky());
+
+  const handleKeyDown = (e: React.KeyboardEvent) => {
+    if (e.key === 'Enter' && !e.shiftKey) {
+      e.preventDefault();
+      void handleReplySubmit();
+    }
+  };
```
secondl1ght: I noticed we duplicate this function across the codebase a lot, but it looks like @catch-21 made a hook to extract this into a reuseable function here `src/hooks/useEnterSubmit/useEnterSubmit.ts`. So we should use that instead (updating all areas of the code can probably be done in a follow-up task though).

src/components/organisms/DialogReply/DialogReply.tsx
```diff
@@ -0,0 +1,33 @@
+'use client';
```
secondl1ght: This component is missing tests, but our convention is that all components should be tested.

secondl1ght (changes requested):
I left some inline comments in the code, and I will add the rest here in this comment:

1. There is a white color on the bottom curve of the thread connector line that doesn't match the design. Also the dashed border colors are different.

DESIGN
image

CODE 
image

2. Too much top spacing on response card (should be same as bottom spacing).

DESIGN
image

CODE
image

3. Spacing below add tag button incorrect on mobile.

DESIGN
image

CODE
image

4. In the design there is a character counter which keeps track of the 2000 character limit, this is missing in the code.

DESIGN
image

CODE
image

5. The dialog breaks when replying to a post with long content, but it should handle this using y-axis overflow to scroll.

image

6. There are warnings in the console that should be addressed about the missing ARIA description attribute.

image

7. Needs better validation on the add tag input, I can add 2 of the same tag which should not be possible.

image

8. After hitting enter when adding a tag, the entire dialog is focused (see white border). Also I have to click the add button again to add another tag, but the functionality on pubky-app is that you can add multiple tags in a row by typing and hitting enter again.

image

9. The component breaks when entering a long string.

image

10. I am not sure if this was supposed to be included in the scope of this PR, but none of the buttons in the reply dialog do anything when clicked (ex. the emoji button and others). If not we should make sure we have more tasks to track these missing pieces. 

11. Again not sure if this feature is supposed to be working in this PR, but when I try to reply to a post I get a bunch of errors (however it looks like this is attempting to make a real call via the `Core.PostController`). The tags are also not being applied to the post.

src/hooks/usePostReply/usePostReply.ts
```diff
@@ -0,0 +1,50 @@
...
+    try {
+      await Core.PostController.create({
+        parentPostId: postId,
+        content: replyContent.trim(),
+        authorId: currentUserId,
+      });
+      setReplyContent('');
+      onSuccess?.();
+    } catch (error) {
+      console.error('Failed to submit reply:', error);
```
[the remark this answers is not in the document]
infin1t3: The focus of my task was on the UI, therefore I just wrapped existing logic into a hook without making any changes to it like error handling. I don't mind doing it, just thought to leave it for the next sweep when we start adding and testing connectivity.

PR #359, opened 2025-11-03
bug: last_read not found after sign up

tipogi:


src/core/application/bootstrap/bootstrap.ts
```diff
@@ -18,6 +18,10 @@ export class BootstrapApplication {
       Core.NexusBootstrapService.fetch(params.pubky),
       this.fetchNotifications(params),
     ]);
+    if (!data) {
+      // TODO: Maybe in the UI, we should redirect or show some special message to the user.
```
catch-21: Created https://github.com/pubky/franky/issues/370 to track this. Worth referencing this issue in the comment.

catch-21 (approved):

PR #360, opened 2025-11-04
feat: stream posts

MiguelMedeiros:
- also returned an empty return for last_read notification if the request failed.

### DIFF
--- a/src/core/controllers/stream/posts/post.test.ts
+++ b/src/core/controllers/stream/posts/post.test.ts
@@ -1,0 +1,172 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import * as Core from '@/core';
+import * as Config from '@/config';
+import { StreamPostsController } from './posts';
+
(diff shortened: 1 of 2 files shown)


src/components/organisms/Timeline/Timeline.tsx
```diff
@@ -6,26 +6,37 @@ import * as Atoms from '@/atoms';
...
+  const getTimestampFromPostId = async (postId: string): Promise<number | undefined> => {
+    try {
+      const postDetails = await Core.PostDetailsModel.findById(postId);
+      return postDetails?.indexed_at;
+    } catch (error) {
+      console.error('Failed to get timestamp from post ID', { postId, error });
+      return undefined;
+    }
+  };
+
```
tipogi: I think, if we cannot direct access to the `timestamp` in that component, it does not make sense to call to the model. anti-pattern. We should get the timestamp in the application layer

PR #361, opened 2025-11-04
fix: replace build with lint check in pre-commit hook

MiguelMedeiros:
## Purpose

Optimize the pre-commit hook by replacing the full build with a lint check, making commits faster and more efficient.

## Changes

- Replaced `npm run build` with `npm run lint` in the pre-commit hook
- Kept format check as the first validation step
- Added proper error handling for the lint check

## Motivation

Running a full build on the pre-commit hook was making commits very slow, especially in a Next.js project where builds can take several seconds. A lint check is sufficient to ensure code quality before committing, while the build can be delegated to the CI/CD pipeline.

## Pre-commit Validations

The hook now only runs:
1. **Format check** - Verifies code formatting
2. **Lint check** - Validates linting rules and potential errors

## How to test

1. Make a change to any `.ts` or `.tsx` file
2. Try to commit
3. Verify that the hook executes quickly (< 5 seconds instead of 30s+)
4. Confirm that linting errors are still detected

## Notes

- The full build still runs in CI/CD
- This doesn't compromise code quality, it just makes development more agile

### DIFF
--- a/src/components/organisms/SinglePostUserDetails/SinglePostUserDetails.test.tsx.snap
+++ b/src/components/organisms/SinglePostUserDetails/SinglePostUserDetails.test.tsx.snap
@@ -78,7 +78,7 @@ exports[`SinglePostUserDetails - Snapshots > matches snapshot with different pos
               data-size="sm"
               data-testid="typography"
             >
-              over 1 year ago
+              almost 2 years ago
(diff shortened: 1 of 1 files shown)


tipogi (approved):
nice!!

PR #363, opened 2025-11-05
feat: hot tags

afterburn:
This PR sets up the core part for hot tags, the active users and trending posts will be addressed in separate prs


src/core/services/nexus/hot/hot.ts
```diff
@@ -0,0 +1,31 @@
...
+      const url = Core.tagApi.hot(params);
+      const response = await Core.queryNexus<Core.NexusHotTag[]>(url);
+
+      Libs.Logger.debug('Hot tags fetched successfully', { count: response?.length ?? 0 });
+
+      return response || [];
+    } catch (error) {
+      Libs.Logger.error('Failed to fetch hot tags from Nexus', error);
+      throw error;
+    }
```
tipogi: If you do not have other mayor reason, I would avoid `try/catch` statement. `queryNexus` function does internally and after, it goes to application layer that error

src/core/services/local/hot/hot.ts
tipogi: missing the `hot.test.ts`

tipogi (changes requested):
Missing the persisting in indexdb

tipogi (approved):
LGTM 🚀

PR #384, opened 2025-11-10
feat: initial setup User streams

afterburn:
In this PR I created two new systems for followers and following streams. I made them into standalone controllers, applications and services for modularity (as opposed to combining them into profile or something similar).

### DIFF
--- a/src/core/services/nexus/stream/users/userStream.api.ts
+++ b/src/core/services/nexus/stream/users/userStream.api.ts
@@ -38,34 +38,34 @@ function buildUserStreamUrl(
 export const userStreamApi = {
   // Sources requiring user_id
   followers: (params: Core.TUserStreamWithUserIdParams) =>
-    buildUserStreamUrl(params, Core.UserStreamSource.FOLLOWERS, Core.USER_STREAM_PREFIX.USERS),
+    buildUserStreamUrl(params, Core.UserStreamSource.FOLLOWERS, Core.USER_STREAM_PREFIX.USER_IDS),
 
   following: (params: Core.TUserStreamWithUserIdParams) =>
-    buildUserStreamUrl(params, Core.UserStreamSource.FOLLOWING, Core.USER_STREAM_PREFIX.USERS),
+    buildUserStreamUrl(params, Core.UserStreamSource.FOLLOWING, Core.USER_STREAM_PREFIX.USER_IDS),
 
   friends: (params: Core.TUserStreamWithUserIdParams) =>
-    buildUserStreamUrl(params, Core.UserStreamSource.FRIENDS, Core.USER_STREAM_PREFIX.USERS),
+    buildUserStreamUrl(params, Core.UserStreamSource.FRIENDS, Core.USER_STREAM_PREFIX.USER_IDS),
 
   muted: (params: Core.TUserStreamWithUserIdParams) =>
-    buildUserStreamUrl(params, Core.UserStreamSource.MUTED, Core.USER_STREAM_PREFIX.USERS),
+    buildUserStreamUrl(params, Core.UserStreamSource.MUTED, Core.USER_STREAM_PREFIX.USER_IDS),
 
   recommended: (params: Core.TUserStreamWithUserIdParams) =>
-    buildUserStreamUrl(params, Core.UserStreamSource.RECOMMENDED, Core.USER_STREAM_PREFIX.USERS),
+    buildUserStreamUrl(params, Core.UserStreamSource.RECOMMENDED, Core.USER_STREAM_PREFIX.USER_IDS),
 
   // Influencers with additional parameters
   influencers: (params: Core.TUserStreamInfluencersParams) =>
-    buildUserStreamUrl(params, Core.UserStreamSource.INFLUENCERS, Core.USER_STREAM_PREFIX.USERS),
+    buildUserStreamUrl(params, Core.UserStreamSource.INFLUENCERS, Core.USER_STREAM_PREFIX.USER_IDS),
 
   // Post replies requiring author_id and post_id
   postReplies: (params: Core.TUserStreamPostRepliesParams) =>
-    buildUserStreamUrl(params, Core.UserStreamSource.POST_REPLIES, Core.USER_STREAM_PREFIX.USERS),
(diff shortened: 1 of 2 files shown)


afterburn:
@tipogi I combined the followers/following into a single user streams system, using the `userId:[reach]` stream id format we discussed yesterday.

If this is the right design let me know, then I will also use it for the remainder of user streams.

src/core/controllers/stream/users/users.ts
```diff
@@ -0,0 +1,51 @@
...
+  }: Core.TReadUserStreamChunkParams): Promise<Core.TReadUserStreamChunkResponse> {
+    const {
+      nextPageIds,
+      cacheMissUserIds,
+      skip: nextSkip,
+    } = await Core.UserStreamApplication.getOrFetchStreamSlice({
+      streamId,
+      user_id,
+      skip,
+      limit,
```
afterburn: `limit: Config.NEXUS_USERS_PER_PAGE`

src/core/controllers/stream/users/users.ts
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @param streamId - User stream identifier
+   * @param user_id - ID of the user whose stream is being fetched
+   * @param skip - Number of users to skip (for pagination)
+   * @param limit - Number of users to return
+   * @returns Next page of user IDs and pagination offset
+   */
+  static async getOrFetchStreamSlice({
+    streamId,
+    user_id,
```
afterburn: Depending on whether we are on /profile (current user) dont send user_id, if we are viewing someone else's profile then do include user_id

src/core/controllers/stream/users/users.ts
```diff
@@ -0,0 +1,51 @@
...
+    skip,
+    limit = Config.NEXUS_USERS_PER_PAGE,
+  }: Core.TReadUserStreamChunkParams): Promise<Core.TReadUserStreamChunkResponse> {
+    const {
+      nextPageIds,
+      cacheMissUserIds,
+      skip: nextSkip,
+    } = await Core.UserStreamApplication.getOrFetchStreamSlice({
```
afterburn: If there is no user_id, use the viewerId instead.

src/core/application/stream/users/users.ts
```diff
@@ -0,0 +1,163 @@
...
+   * @param streamId - User stream identifier (e.g., 'followers:today:all', 'following:today:all')
+   * @param user_id - ID of the user whose stream is being fetched
+   * @param skip - Number of users to skip (for pagination)
+   * @param limit - Number of users to return
+   * @returns Next page of user IDs, cache miss IDs, and pagination offset
+   */
+  static async getOrFetchStreamSlice({
+    streamId,
+    user_id,
+    skip = 0,
```
afterburn: Remove default value for skip, because we always get it from controller

src/core/controllers/stream/users/users.ts
```diff
@@ -0,0 +1,49 @@
...
+   *
+   * @param streamId - Composite user stream identifier (userId:reach) e.g., 'user123:followers'
+   * @param skip - Number of users to skip (for pagination)
+   * @param limit - Number of users to return
+   * @returns Next page of user IDs and pagination offset
+   */
+  static async getOrFetchStreamSlice({
+    streamId,
+    skip,
+    limit,
```
tipogi: the UI does not need to pass that field, we pass internally from the config params

src/core/controllers/stream/users/users.ts
tipogi: missing that definition

src/core/services/local/stream/users/users.ts
```diff
@@ -1,29 +1,81 @@
...
+      userRelationships.push([userId, user.relationship]);
+
+      // Convert tags to NexusTag format
+      const nexusTags = user.tags.map((tag) => ({
+        label: tag.label,
+        taggers: tag.taggers,
+        taggers_count: tag.taggers_count,
+        relationship: tag.relationship,
+      }));
+      userTags.push([userId, nexusTags]);
```
tipogi: `Tags` already in `NexusTag` format, right?
```js
userTags.push([userId, user.tags]);
```

src/core/models/stream/user/userStream.types.ts
```diff
@@ -9,8 +11,10 @@ export enum UserStreamTypes {
   // Active users in the UI. We get randomly, preview=true param active in nexus
   TODAY_INFLUENCERS_ALL = 'influencers:today:all',
   RECOMMENDED = 'recommended:all:all',
```
tipogi: add here a TODO task because we need to add all the possible cases

src/core/application/stream/users/users.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    // Extract user IDs
+    const userIds = nexusUsers.map((user) => user.details.id);
+
+    // Persist full user data to IndexedDB
+    await Core.LocalStreamUsersService.persistUsers(nexusUsers);
+
+    // Upsert stream (append to existing or create new)
+    const updatedStream = cachedStream ? [...cachedStream.stream, ...userIds] : userIds;
+    await Core.LocalStreamUsersService.upsert({ streamId, stream: updatedStream });
```
tipogi: what if `userIds` already has the users that we have persisted in the cache, `cachedStream`. Before upsert, do the filtering of the users

src/core/application/stream/users/users.ts
```diff
@@ -0,0 +1,147 @@
...
+    const { url, body } = Core.userStreamApi.usersByIds({
+      user_ids: cacheMissUserIds,
+      viewer_id: viewerId,
+    });
+
+    const userBatch = await Core.queryNexus<Core.NexusUser[]>(url, 'POST', JSON.stringify(body));
+    if (userBatch) {
+      await Core.LocalStreamUsersService.persistUsers(userBatch);
+    }
+  }
```
tipogi: since the controller does not handle errors and as we are not wrapping with `try/catch` that function, might be better to wrapp and log the error

tipogi (approved):

PR #397, opened 2025-11-12
test: more core test coverage

catch-21:
Adding more tests around work done in **feat: stream posts** #360

-----

**Note** this test highlights duplicates in the new chuck of streamed posts are not filtered out, ~is this expected?~ have added comment explaining why franky should not need to handle this.

https://github.com/pubky/franky/blob/af11d08ab089df664b7319182d296b01d2506b5e/src/core/services/local/stream/posts/posts.test.ts#L634-L670


src/core/application/bootstrap/bootstrap.test.ts
```diff
@@ -333,6 +337,115 @@ describe('BootstrapApplication', () => {
...
+
+      expect(mocks.persistUsers).toHaveBeenCalledWith([]);
+      expect(mocks.persistPosts).toHaveBeenCalledWith(bootstrapData.posts);
+      expect(mocks.upsertPostsStream).toHaveBeenCalledWith({
+        streamId: Core.PostStreamTypes.TIMELINE_ALL_ALL,
+        stream: ['post-1'],
+      });
+      expect(result).toEqual({ unread: 0, lastRead: MOCK_LAST_READ });
+    });
+
```
tipogi: that cases are quite impossible because nexus first fetches all the posts + post_replies to add in `post` field and after, it searches all the authors of the posts to add in the `users` field. Before get that response, we might get an error

tipogi (approved):
merge it

src/core/application/bootstrap/bootstrap.test.ts
catch-21: Good to know. In this case, should Franky reject such a scenario?  If not I'm tempted to leave the tests in with a comment explaining what you say, that is is unrealistic, although not technically impossible with some unexpected change in future.

PR #408, opened 2025-11-14
feat(merge): post replies timeline

MiguelMedeiros:


src/core/services/nexus/stream/posts/postStream.api.ts
```diff
@@ -58,6 +58,10 @@ export const postStreamApi = {
   replies: (params: Core.TStreamPostRepliesParams) =>
     buildPostStreamUrl(params, Core.StreamSource.REPLIES, Core.STREAM_PREFIX.POSTS),
 
+  // Alias for replies (for backwards compatibility)
+  postReplies: (params: Core.TStreamPostRepliesParams) =>
+    buildPostStreamUrl(params, Core.StreamSource.REPLIES, Core.STREAM_PREFIX.POSTS),
+
```
MiguelMedeiros: @tipogi is this key the right one?

`postReplies` or `post_replies`

src/core/services/nexus/stream/posts/postStream.api.ts
```diff
@@ -66,6 +70,10 @@ export const postStreamApi = {
   author_replies: (params: Core.TStreamAuthorRepliesParams) =>
     buildPostStreamUrl(params, Core.StreamSource.AUTHOR_REPLIES, Core.STREAM_PREFIX.POSTS),
 
+  // Alias for author_replies (for backwards compatibility)
+  authorReplies: (params: Core.TStreamAuthorRepliesParams) =>
+    buildPostStreamUrl(params, Core.StreamSource.AUTHOR_REPLIES, Core.STREAM_PREFIX.POSTS),
+
```
MiguelMedeiros: @tipogi same here: `authorReplies` or `author_replies` ?

src/core/services/local/post/post.ts
```diff
@@ -168,6 +168,64 @@ export class LocalPostService {
...
+      const counts = await Core.PostCountsModel.findById(postId);
+      return counts ?? ({ id: postId, tags: 0, unique_tags: 0, replies: 0, reposts: 0 } as Core.PostCountsModelSchema);
+    } catch (error) {
+      Libs.Logger.error('Failed to get post counts', { postId, error });
+      throw Libs.createDatabaseError(Libs.DatabaseErrorType.QUERY_FAILED, 'Failed to get post counts', 500, {
+        error,
+        postId,
+      });
+    }
+  }
```
tipogi: dont do `try/catch` here, let bubble up the error. One day we will centralise all the errors in the application layer, so we will handle there

src/components/organisms/Timeline/PostReplies/PostReplies.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const replyIds = useLiveQuery(
+    async () => {
+      // Only fetch if there are replies
+      if (!postCounts?.replies || postCounts.replies === 0) {
+        return [];
+      }
+
+      try {
+        // Use controller to get first 3 replies
+        return await Core.PostController.getFirstReplies({ postId, limit: 3 });
```
tipogi: why don't you use the controller function with streamId, `post_replies:[pubky]:[postId]`, I do not see the need for that

src/core/controllers/post/post.ts
```diff
@@ -78,4 +78,25 @@ export class PostController {
...
+
+  /**
+   * Get post counts for a specific post
+   * @param params - Parameters object
+   * @param params.postId - ID of the post to get counts for
+   * @returns Post counts (with default values if not found)
+   */
+  static async getPostCounts({ postId }: { postId: string }): Promise<Core.PostCountsModelSchema> {
+    return await Core.LocalPostService.getPostCounts(postId);
+  }
```
tipogi: Not sure if that function we should add here. I see more in `controllers/stream/posts` controller as it is a util function to render a replies. The `controllers/posts` I will focus more for single actions related with posts as creation, deletion, bookmark or the need it utils, do you agree?

src/core/services/nexus/stream/posts/postStream.api.ts
MiguelMedeiros: https://github.com/pubky/franky/pull/408/commits/5131dbb9d365c9f83b0acde8adf60795f60faaf7

src/components/organisms/Timeline/PostReplies/PostReplies.tsx
MiguelMedeiros: https://github.com/pubky/franky/pull/408/commits/012140afa597c0def745eab944f80fa39f0c7b5d

src/components/organisms/Timeline/PostReplies/PostReplies.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * Renders replies for a specific post in the timeline with thread connectors (flat structure, 1 level).
+ */
+interface TimelinePostRepliesProps {
+  postId: string;
+  onPostClick: (postId: string) => void;
+}
+
+export function TimelinePostReplies({ postId, onPostClick }: TimelinePostRepliesProps) {
+  // Watch for changes in post_counts to trigger refetch when replies count changes
+  const postCounts = useLiveQuery(() => Core.PostController.getPostCounts({ postId }), [postId]);
```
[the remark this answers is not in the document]
tipogi: so this is what I said. This component already has a *useLiveQuery* hook which it will force the re-render, when there is a new posts counts and there is not need to add in `TimelinePosts` component that
```js
line 28 const postCounts = useLiveQuery(() => Core.PostController.getPostCounts({ postId }), [postId]);
line 29 const hasReplies = postCounts?.replies && postCounts.replies > 0 ? true : false;
and
line 35 {hasReplies &&
```
in this component then, we could use `useEffect` hook and
```js
const [replyIds, setReplyIds] = useState<string[]>([]);
  // Watch for changes in post_counts to trigger refetch when replies count changes
  const postCounts = useLiveQuery(() => Core.PostController.getPostCounts({ postId }), [postId]);

  useEffect(() => {
    if (!postCounts?.replies || postCounts.replies < 1) {
      setReplyIds([]);
      return;
    }

    const repliesCount = postCounts.replies;

    async function fetchReplies() {
      try {
        const response = await Core.StreamPostsController.getOrFetchStreamSlice({
          streamId: `${Core.StreamSource.REPLIES}:${postId}`,
          streamTail: 0,
          lastPostId: undefined,
          limit: repliesCount > 3 ? 3 : repliesCount,
        });
        setReplyIds(response.nextPageIds);
      } catch (error) {
        // Silently handle errors - don't show replies if there's an issue
        Libs.Logger.error('Failed to fetch post replies:', error);
        setReplyIds([]);
      }
    }

    fetchReplies();
  }, [postCounts?.replies]);

  const hasReplies = replyIds && replyIds.length > 0;

  // Don't render anything if there are no replies
  if (!hasReplies) {
    return null;
  }
  ...
  ```
The `limit` param is changed because there was a bug. If the limit is smaller than 3, it goes to nexus because the controller could not find in the cache an array with `limit` length. Imagine, the post parent has 2 replies but we go to cache and we cannot find 3 posts because it only has 2 replies. So, we do not get enough posts and it goes to nexus. Limiting in that way, we avoid that query in all the replies.

Now, still there is a bug in the post creation, PUT action, *core* side (will be fixed in the future PR that updates all kind of streams in post creation). If we add a new reply, it has to also persist in the postReplies stream and now it does not. Thats why always, when the counter increments, we get one post less

Issue #396, opened 2025-11-12
Implement PostContent parsing for attachments, links, and tags


MiguelMedeiros:
### PostContent Parsing Implementation

We need to implement the new **PostContent** feature to handle parsing of all supported content types, including attachments, embeds, and tags, consistent with the old app and updated Figma design.

---

#### Parsing Requirements

**Attachments**
- [ ] Single image  
- [ ] Multiple images (carousel / grid layout)  
- [ ] Single video  
- [ ] Multiple videos  
- [ ] Generic file attachments (PDF, ZIP, etc.)

**Link Embeds**
- [ ] YouTube  
- [ ] Vimeo  
- [ ] Twitch  
- [ ] Generic website previews  
- [ ] Tweets (X posts)

**Tags**
- [ ] Parse and render `@mentions` and `#hashtags`  
- [ ] Ensure parsing is safe and follows shadcn_ui text styles

---

#### References
- Legacy logic: [`_Parsing.tsx`](https://github.com/pubky/pubky-app/blob/a4f2b5a8455c6311f578874bf38387276a3fb446/src/components/Content/_Parsing.tsx#L40)  
- Design spec (Figma): [shadcn_ui-PUBKY](https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=19386-25234&m=dev)

---

#### Testing
- Use **Staging** to reproduce and verify all edge cases:
  - Multiple attachments  
  - Mixed media (image + video)  
  - Broken links or invalid embeds  
- Ensure rendering and interactions match the Figma design.

---

#### Notes
- Keep parsing logic modular and extendable for future content types.  
- Follow existing **shadcn_ui** component patterns and style tokens.

catch-21:
Imo we keep this issue as is, deprioritised for now, reference it in backlog tracking, and revive it when we're ready to continue.

PR #410, opened 2025-11-16
feat: youtube link embeds in post content

### LINKED ISSUE #409: YouTube
See parent task for detailed description.

Parsing Requirements > Link Embeds > YouTube

secondl1ght:
Closes #409 (sub-issue of #396)

After taking a closer look at the legacy code I found lots of issues, so I chose to rewrite this from scratch starting from first principles rather than refactor the old logic. Now we have a much simpler, readable, maintainable, etc. approach, while still retaining the same functionality as the legacy app. 

This PR includes the parse and render of YouTube links as outlined in the parent task. It is easily extensible, so once this PR is approved I will submit a follow-up PR to include the rest of the link embed support. Or I can do one PR per embed type if we want to keep each one tightly scoped. 

One note is that I did not see any link embeds designed in the Figma file, so I left a question for @aldertnl about this. So depending on his response there may be another small style PR to add some simple interactions like hover state for example.

image

image


src/components/organisms/PostContent/PostContent.tsx
```diff
@@ -21,10 +23,12 @@ export function PostContent({ postId, className }: PostContentOrganismProps) {
   }
 
   return (
-    <div className={Libs.cn('flex flex-col', className)}>
+    <Atoms.Container className={Utils.cn('gap-3', className)}>
```
MiguelMedeiros: Keep consistent with other components in the codebase

```suggestion
    <Atoms.Container className={Libs.cn('gap-3', className)}>
```

src/components/organisms/PostContent/PostContent.tsx
```diff
@@ -2,7 +2,9 @@
 
 import { useLiveQuery } from 'dexie-react-hooks';
 import * as Core from '@/core';
-import * as Libs from '@/libs';
+import * as Atoms from '@/atoms';
+import * as Molecules from '@/molecules';
+import * as Utils from '@/libs/utils';
```
MiguelMedeiros:
```suggestion
import * as Libs from '@/libs';
```

src/components/molecules/PostLinkEmbed/PostLinkEmbed.tsx
```diff
@@ -0,0 +1,73 @@
...
+import * as Atoms from '@/atoms';
+import { useMemo } from 'react';
+
+export type PostLinkEmbedProps = {
+  content: string;
+};
+
+type LinkEmbed = 'none' | 'youtube';
+
+const parseContentForLinkEmbed = (content: string): { type: LinkEmbed; url: string } => {
```
MiguelMedeiros: At the top of the file or in a types file. Then the function signature becomes cleaner.

```suggestion
type EmbedType = 'youtube' | 'vimeo' | 'twitch' | 'twitter' | 'generic';
type LinkEmbed = { type: EmbedType | 'none'; url: string };

const parseContentForLinkEmbed = (content: string): LinkEmbed => {
```

src/components/molecules/PostLinkEmbed/PostLinkEmbed.tsx
```diff
@@ -0,0 +1,73 @@
...
+  try {
+    // Simple URL pattern for social media links
+    const urlRegex =
+      /(https?:\/\/[a-zA-Z0-9-._~:/?#[\]@!$&'()*+,;=%]+)|(?:www\.)?([a-zA-Z0-9][a-zA-Z0-9-]*\.)+[a-zA-Z]{2,}(?:\/[^\s]*)*/g;
```
MiguelMedeiros: Problems:
- ReDoS vulnerability: This regex has nested quantifiers that could cause catastrophic backtracking with malicious input
- Overly permissive: Matches many things that aren't valid URLs
- Edge case: The regex allows URLs without protocols AND with protocols in the same pattern, making it complex.

Better approach: Use browser's built-in URL parser or a well-tested library

```suggestion
    // Match potential URLs more simply
    const urlRegex = /(?:https?:\/\/|www\.)[\S]+/gi;
```

src/components/organisms/PostContent/PostContent.tsx
secondl1ght: I thought we talked about this on our last Franky call based on my comment in Slack:

image

We decided it would be better to not import the entire `Libs` if you only need to use a small subset of it in your file. Or are we not making that change until after the initial release?

src/components/molecules/PostLinkEmbed/PostLinkEmbed.tsx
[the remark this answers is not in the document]
secondl1ght: OK my thinking for the naming was that the component only returns one embed, but yes it does support any type of link. If I were doing a component for each type, it would have been `PostYouTubeLinkEmbed` for example, but I can update to `PostLinksEmbeds`. 👍🏼

src/components/molecules/PostLinkEmbed/PostLinkEmbed.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const hostname = parsedUrl.hostname.toLowerCase();
+
+    switch (hostname) {
+      case 'youtube.com':
+      case 'www.youtube.com':
+      case 'youtu.be':
+      case 'm.youtube.com':
+      case 'www.youtube-nocookie.com':
+      case 'youtube-nocookie.com': {
+        const youtubeIdRegex = /^.*(youtu.be\/|v\/|u\/\w\/|embed\/|watch\?v=|&v=)([^#&?]*).*/;
```
MiguelMedeiros: **Problem**: This regex is duplicated from legacy code and has some issues:
- `u\/\w\/` pattern is outdated (user videos feature was deprecated)
- `.*` at start and end is inefficient
- Not handling playlist IDs or other parameters

**Better approach**: More focused regex with named groups

```typescript
const extractYouTubeId = (url: string): string | null => {
  // Handle different YouTube URL formats
  const patterns = [
    // Standard watch: youtube.com/watch?v=VIDEO_ID
    /(?:youtube\.com\/watch\?v=)([a-zA-Z0-9_-]{11})/,
    // Short URL: youtu.be/VIDEO_ID
    /(?:youtu\.be\/)([a-zA-Z0-9_-]{11})/,
    // Embed: youtube.com/embed/* or youtube-nocookie.com/embed/*
    /(?:youtube(?:-nocookie)?\.com\/embed\/)([a-zA-Z0-9_-]{11})/,
    // Shorts: youtube.com/shorts/VIDEO_ID
    /(?:youtube\.com\/shorts\/)([a-zA-Z0-9_-]{11})/,
    // Live streams: youtube.com/live/VIDEO_ID
    /(?:youtube\.com\/live\/)([a-zA-Z0-9_-]{11})/,
    // Old embed: youtube.com/v/VIDEO_ID (legacy)
    /(?:youtube\.com\/v\/)([a-zA-Z0-9_-]{11})/,
  ];
  
  for (const pattern of patterns) {
    const id = url.match(pattern)?.[1];
    if (id?.length === 11) return id;
  }
  
  return null;
};
```

src/components/molecules/PostLinkEmbed/PostLinkEmbed.tsx
secondl1ght: OK this is the same regex that is used the legacy app, I thought we wanted to keep the same functionality as there so that is why I used this parsing method (but improved other parts of the code). If we want to rethink all functionality, I agree that we can look into better options here. Was trying to follow the guidance of "consistent with the old app" and not change any core functionality, for example which types of links are handled in the new app vs old app.

src/components/molecules/PostLinkEmbed/PostLinkEmbed.tsx
MiguelMedeiros: The component has try-catch in parsing but doesn't have an error boundary for render errors. If the iframe fails to load or there's a CSP issue, it fails silently. We can also ask @aldertnl for a better UI/UX for failing embeds. Consider:

```typescript
export const PostLinkEmbed = ({ content }: PostLinkEmbedProps) => {
  const linkEmbed = useMemo(() => parseContentForLinkEmbed(content), [content]);

  if (linkEmbed.type === 'none') return null;

  return (
    <Atoms.Container>
      <ErrorBoundary fallback={<div className="text-muted-foreground text-sm">Failed to load embed</div>}>
        {linkEmbed.type === 'youtube' && (
          <iframe
            width="100%"
            height="315"
            src={linkEmbed.url}
            title="YouTube video player"
            allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share"
            allowFullScreen
            loading="lazy"
            className="rounded-md"
          />
        )}
      </ErrorBoundary>
    </Atoms.Container>
  );
};
```

src/components/molecules/PostLinkEmbed/PostLinkEmbed.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+      const iframe = screen.getByTitle('YouTube video player');
+      expect(iframe).toBeInTheDocument();
+      expect(iframe).toHaveAttribute('src', 'https://www.youtube-nocookie.com/embed/dQw4w9WgXcQ');
+    });
+
+    it('does not render embed for invalid YouTube video ID', () => {
+      render(<PostLinkEmbed content="Invalid: https://www.youtube.com/watch?v=invalid" />);
+
+      expect(screen.queryByTitle('YouTube video player')).not.toBeInTheDocument();
+    });
```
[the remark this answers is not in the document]
secondl1ght: OK this is another example where I left the legacy validation to have the same functionality in both apps, but I know now that I have the greenlight to make any changes necessary in the rewrite. So this is a good suggestion and one I considered handling to improve from old feature. 👍🏼

src/components/molecules/PostLinkEmbed/PostLinkEmbed.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export const PostLinkEmbed = ({ content }: PostLinkEmbedProps) => {
+  const linkEmbed = useMemo(() => parseContentForLinkEmbed(content), [content]);
+
+  if (linkEmbed.type === 'none') return null;
+
+  return (
+    <Atoms.Container>
+      {linkEmbed.type === 'youtube' && (
+        <iframe
```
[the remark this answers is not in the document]
secondl1ght: Yes good point, I remember thinking about this but then must have not circled back to it!

src/components/molecules/PostLinkEmbed/PostLinkEmbed.test.tsx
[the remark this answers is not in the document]
secondl1ght: Very nice!! I am still getting familiar with Vitest and this is awesome functionality.

src/components/molecules/PostLinkEmbeds/PostLinkEmbeds.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+    it('handles YouTube URL with trailing punctuation', () => {
+      render(<PostLinkEmbeds content="Check this out: https://www.youtube.com/watch?v=dQw4w9WgXcQ!" />);
+
+      const iframe = screen.getByTestId('YouTube video player');
+      expect(iframe).toBeInTheDocument();
+      expect(iframe).toHaveAttribute('src', 'https://www.youtube-nocookie.com/embed/dQw4w9WgXcQ');
+    });
+
+    it('handles YouTube URL with additional parameters', () => {
+      render(<PostLinkEmbeds content="Timestamped: https://www.youtube.com/watch?v=dQw4w9WgXcQ&t=123s" />);
```
catch-21: Does additional params also work for `youtu.be` ? This is the commonly 'shared' url format from YT site.

src/components/molecules/PostLinkEmbeds/PostLinkEmbeds.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const hostname = parsedUrl.hostname.toLowerCase();
+
+    switch (hostname) {
+      case 'youtube.com':
+      case 'www.youtube.com':
+      case 'youtu.be':
+      case 'm.youtube.com':
+      case 'www.youtube-nocookie.com':
+      case 'youtube-nocookie.com': {
+        const youtubeIdRegex = /^.*(youtu.be\/|v\/|u\/\w\/|embed\/|watch\?v=|&v=)([^#&?]*).*/;
```
secondl1ght: Thanks for the code suggestions btw! 🙏🏼

PR #413, opened 2025-11-17
feat: click to copy recovery phrase during backup

catch-21:
addresses https://github.com/pubky/franky/issues/142

It adds click/tap to copy space-delimited 12-word phrase **without** including the phrase in toast description.
Retains toast description for other copy actions by default.

https://github.com/user-attachments/assets/7b45897f-cc5c-460e-8630-a01b9342dd7c


src/hooks/useCopyToClipboard/useCopyToClipboard.tsx
```diff
@@ -28,7 +32,9 @@ export function useCopyToClipboard(options: UseCopyToClipboardOptions = {}) {
 
         const toastInstance = Molecules.toast({
           title: successTitle,
-          description: text,
+          ...(successDescription !== undefined
+            ? successDescription !== '' && { description: successDescription }
```
catch-21: **Note:**
Use of new `successDescription` overrides existing default use of the `text` value for toast description.

For recovery phrase I define `successDescription = ''` to show no toast description.

To be more explicit, we should change default to use no description and if we want a description then `successDescription` must be assigned a value.

PR #423, opened 2025-11-19
feat(PostContent): handle music youtube videos

secondl1ght:
Task tracked in https://github.com/pubky/franky/issues/396

image


catch-21:
Verified in fd32df95052c0deda8af823d419efea4ec4157f7
Works well with other text and with timestamps.

PR #425, opened 2025-11-20
feat: Bookmarks data layer

afterburn:
Did a first pass on the bookmarks core system, leaving this in draft for an initial review by @tipogi and to have something to look at while discussing.

### DIFF
--- a/src/core/services/nexus/stream/posts/postStream.test.ts
+++ b/src/core/services/nexus/stream/posts/postStream.test.ts
@@ -391,7 +391,8 @@ describe('createPostStreamParams', () => {
       const streamTail = 1234567890;
       const result = createPostStreamParams(Core.PostStreamTypes.TIMELINE_BOOKMARKS_ALL, streamTail, 20, mockViewerId);
 
-      expect(result.params.start).toBe(streamTail);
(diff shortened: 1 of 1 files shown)


src/core/services/local/bookmark/bookmark.ts
```diff
@@ -0,0 +1,132 @@
...
+   *
+   * Note: We remove from all possible stream types without checking post details
+   * for efficiency. The removeFromStream operation is idempotent.
+   *
+   * @param postId - Composite post ID
+   */
+  private static async removeFromBookmarkStreams(postId: string): Promise<void> {
+    // Remove from all bookmark stream types in parallel
+    await Promise.all(
+      Object.values(BOOKMARK_STREAMS).map((streamId) =>
```
tipogi: Just delete only the related bookmark streams: `all` and `kind`. In the function, pass as an argument the `kind` value

tipogi (changes requested):

tipogi (dismissed):
nice!

catch-21:
Thorough test coverage.
I'd like to see franky support updates in homeserver not made by this instance of the client. For another time.

PR #441, opened 2025-11-24
fix: refactor Header component to use Controllers and fix avatar fallback logic

MiguelMedeiros:
## 🎯 Summary

This PR refactors the `HeaderSignIn` component to use the proper layered architecture (Controllers instead of direct database access) and fixes the avatar initial fallback logic to ensure a default "U" is displayed when user has no name set.

## 🐛 Problem

The `HeaderSignIn` component had two issues:

1. **Architecture violation**: Directly accessing the database (`Core.db.user_details.get()`) instead of using the `ProfileController`
2. **Missing avatar URL logic**: Not using `FileController.getAvatarUrl()` for proper avatar image URLs
3. **Broken fallback logic**: When `extractInitials()` returned an empty string for users without names, the fallback "U" wasn't being applied
4. **Test failures**: 17 tests failing due to missing controller mocks

### Component Changes

**`src/components/organisms/Header/Header.tsx`**

- ✅ Replaced `Core.db.user_details.get()` with `Core.ProfileController.read()` (follows layered architecture)
- ✅ Replaced `userDetails?.image` with `Core.FileController.getAvatarUrl()` for proper CDN URLs
- ✅ Added fallback logic: `const avatarInitial = Libs.extractInitials({ name: userDetails?.name || '' }) || 'U';`
- ✅ Removed unnecessary comment and cleaned up code structure

### Test Changes

**`src/components/organisms/Header/Header.test.tsx`**

- ✅ Added `ProfileController` mock with `read` method
- ✅ Added `FileController` mock with `getAvatarUrl` method
- ✅ Ensures tests properly mock the controllers used by the component

**`src/components/molecules/Header/Header.test.tsx`**

- ✅ Added `ProfileController` mock with `read` method
- ✅ Added `FileController` mock with `getAvatarUrl` method
- ✅ Fixes 3 failing tests for avatar initial fallback cases

### Before

- ❌ 2 test files failed (17 failing tests)
- ❌ Tests expecting "U" fallback for empty names were failing

### After

- ✅ All 267 test files passed
- ✅ 3,306 tests passed (4 skipped)
- ✅ No linter errors
- ✅ Avatar fallback logic works correctly for:
  - `undefined` names
  - Empty string names (`''`)
  - Whitespace-only names (`'   '`)

## 📋 Architecture Compliance

This change aligns with the project's layered architecture:

```
UI → Controllers → Application → Services → Database
```

**Before**: UI → Database ❌ (violated architecture)  
**After**: UI → Controllers ✅ (correct architecture)

## 📝 Additional Notes

- The avatar fallback "U" represents "User" as a default initial
- The change maintains backward compatibility while fixing the architecture violation
- No breaking changes to the component's public API
- Follows the project's ADR-0004 (Layering and Dependency Rules)

### DIFF
--- a/src/core/controllers/user/user.ts
+++ b/src/core/controllers/user/user.ts
@@ -16,8 +16,7 @@ export class UserController {
    * This is a read-only operation that queries the local cache
    */
   static async getCounts({ userId }: Core.TReadProfileParams): Promise<Core.NexusUserCounts | null> {
-    const counts = await Core.UserApplication.counts({ userId });
-    return counts;
+    return await Core.UserApplication.counts({ userId });
   }
 
   static async follow(eventType: Core.HomeserverAction, { follower, followee }: Core.TFollowParams) {
--- a/src/core/controllers/user/user.test.ts
+++ b/src/core/controllers/user/user.test.ts
@@ -78,7 +78,7 @@ describe('UserController', () => {
     it('should return null when user counts not found', async () => {
       const userId = 'non-existent-user';
 
-      vi.spyOn(Core.UserApplication, 'counts').mockResolvedValue(undefined);
+      vi.spyOn(Core.UserApplication, 'counts').mockResolvedValue(null);
 
       const result = await UserController.getCounts({ userId });
 


src/core/controllers/user/user.ts
```diff
@@ -7,16 +7,17 @@ export class UserController {
...
 
   /**
    * Get user counts from local database
    * This is a read-only operation that queries the local cache
    */
-  static async getCounts(userId: string): Promise<Core.UserCountsModelSchema | null> {
-    return await Core.UserCountsModel.findById(userId);
+  static async getCounts({ userId }: Core.TReadProfileParams): Promise<Core.NexusUserCounts | null> {
+    const counts = await Core.UserApplication.counts({ userId });
+    return counts ?? null;
```
tipogi: we return `null` already in the application

src/core/services/local/profile/profile.ts
```diff
@@ -2,6 +2,49 @@ import * as Core from '@/core';
...
 export class LocalProfileService {
+  private constructor() {} // Prevent instantiation
+
+  /**
+   * Retrieves user details from local database.
+   * @param userId - The user ID to fetch details for
+   * @returns Promise resolving to user details or null if not found
+   */
+  static async details({ userId }: Core.TReadProfileParams): Promise<Core.NexusUserDetails | null> {
+    return (await Core.UserDetailsModel.findById(userId)) ?? null;
```
tipogi: `findById` already returns a _null_ value, when we do not find the related record. The same with others

tipogi (changes requested):
few changes

PR #452, opened 2025-11-26
feat: Core - Modify streams when user follows/unfollows

afterburn:
@tipogi was this what we had in mind for updating user streams when someone follows/unfollows, or should I take it in a different direction?


src/core/services/local/follow/follow.ts
```diff
@@ -115,4 +154,66 @@ export class LocalFollowService {
...
     }
   }
+
+  /**
+   * Invalidate timeline streams by clearing them from cache
+   * Forces fresh fetch from Nexus on next load
+   *
+   * @param includeFriends - Whether to also invalidate friends timelines
+   */
+  private static async invalidateTimelineStreams(includeFriends: boolean): Promise<void> {
```
tipogi: That is it is ok if the route is not `/home` but what about if we are in that route and we `follow`/`unfollow` a user. This will clear up the stream that we are rendering and might break the timeline of the cache. @MiguelMedeiros

src/core/services/local/follow/follow.ts
```diff
@@ -1,9 +1,38 @@
...
+const FRIENDS_TIMELINE_STREAMS = [
+  Core.PostStreamTypes.TIMELINE_FRIENDS_ALL,
+  Core.PostStreamTypes.TIMELINE_FRIENDS_SHORT,
+  Core.PostStreamTypes.TIMELINE_FRIENDS_LONG,
+  Core.PostStreamTypes.TIMELINE_FRIENDS_IMAGE,
+  Core.PostStreamTypes.TIMELINE_FRIENDS_VIDEO,
+  Core.PostStreamTypes.TIMELINE_FRIENDS_LINK,
+  Core.PostStreamTypes.TIMELINE_FRIENDS_FILE,
+] as const;
+
```
tipogi: could be added in `follow.contants.ts`

src/core/services/local/follow/follow.ts
```diff
@@ -65,7 +74,11 @@ export class LocalFollowService {
     }
   }
 
-  static async delete({ follower, followee }: Core.TFollowParams) {
+  static async delete({
+    follower,
+    followee,
+    activeStreamId,
+  }: Core.TFollowParams & { activeStreamId?: Core.PostStreamTypes | null }) {
```
tipogi: same

tipogi (changes requested):

PR #468, opened 2025-11-27
fix: filter out invalid urls for og-metadata to avoid unnecessary DNS resolution failures

catch-21:
Was seeing this when it attempted to use `https://www.` so added more validation filters.

```
DNS resolution failed: Error: queryA ENOTFOUND www
    at ignore-listed frames {
  errno: undefined,
  code: 'ENOTFOUND',
  syscall: 'queryA',
  hostname: 'www'
}
```


MiguelMedeiros:
> Another option would be to just use the same `linkify` package we are using to extract the URL from the post content, I am sure it has more validation built-in than we could think of by creating our own.

@secondl1ght interesting thought, but actually I'd go the other way.

We're only using linkify for simple URL matching, something we can do with straightforward regex/validation like @catch-21 did here.

This PR shows we don't really need linkify at all. Maybe worth a follow-up to remove that dependency entirely and keep our URL handling lean and custom.

For now, this approach is solid. ✅

MiguelMedeiros (approved):

secondl1ght:
@MiguelMedeiros we originally used a regex in the previous app and brought that over to the initial version of Franky, but then in this comment one of the suggested options to improve was to use a well-tested library (https://github.com/pubky/franky/pull/410#discussion_r2532047709) and that's when we added `linkify`. I am OK with either option though in this scenario. 👍🏼

catch-21:
Related: https://github.com/pubky/franky/issues/483

PR #473, opened 2025-11-29
feat(PostContent): hashtags

secondl1ght:
Sub-task of #396 

NOTE: I also found a bug in the `Link` Atom during development, passing a `className` prop would result in the component classes being overwritten entirely instead of merging via `cn`. This is because of this line (https://github.com/pubky/franky/blob/4111138f6c5184553d10adb427e2c42419c07a35/src/components/atoms/Link/Link.tsx#L75) where `className` was included in `...props` and overwriting `className={linkClassName}`. It was a bit of a tricky bug to figure out why it was not working as expected, but it is fixed now. (We may want to review other areas where `Atoms.Link` is being used to make sure it was developed expecting the default classes to be present on the component, because now they will be applied).

image

```md
Add hashtag parsing and icon support to PostText

Implement hashtag detection in PostText component with custom rendering
via new PostHashtags molecule. Hashtags matching specific keywords
(#bitcoin, #synonym, #tether, #pubky, #bitkit, #blocktank) display
corresponding brand icons.

- Add PostHashtags molecule with icon mapping for brand hashtags
- Create remarkHashtags plugin to parse hashtag syntax in markdown
- Update Link component to properly destructure className prop
- Add comprehensive test coverage for hashtag rendering and edge cases
- Include brand icon components (Synonym, Blocktank, Bitkit, etc.)
```


MiguelMedeiros (dismissed):

MiguelMedeiros:
Nice work!

PR #480, opened 2025-12-01
feat: Core - Set up other notification types

afterburn:


src/core/controllers/notification/notification.ts
```diff
@@ -51,13 +52,15 @@ export class NotificationController {
    * @returns Promise resolving to notifications and next timestamp for pagination
    */
   static async getOrFetchNotifications({
+    types,
```
tipogi: not sure if this is an UI param or an internal list (don't know yet where we could save) that the users chooses in the settings page. In that case, controller should get that data and pass to application layer. The types that we get would be the ones that we should exclude. Notification filtering by type, is not ready yet in nexus...

PR #489, opened 2025-12-02
feat: Human Proof Section

SeverinAlexB:
Adds
- Human Proof section replacing invitecodes
- SMS verification to signup
- LN payment to signup
- Removed the previous invite code page
- Added new signup qr code for Pubky Ring
- Updated onboarding progress bar

### DIFF
--- a/src/components/organisms/HumanPhoneInput/HumanPhoneInput.test.tsx
+++ b/src/components/organisms/HumanPhoneInput/HumanPhoneInput.test.tsx
@@ -1,7 +1,8 @@
 import { describe, it, expect } from 'vitest';
 import { render } from '@testing-library/react';
 
-import { HumanPhoneInput, parsePhoneNumber } from './HumanPhoneInput';
+import { HumanPhoneInput } from './HumanPhoneInput';
+import { parsePhoneNumber } from '@/libs';
 
 describe('HumanPhoneInput', () => {
   it('matches snapshot', () => {


src/components/organisms/HumanLightningPayment/HumanLightningPayment.tsx
```diff
@@ -0,0 +1,141 @@
...
+export const HumanLightningPayment = ({ onBack, onSuccess }: HumanLightningPaymentProps) => {
+  const [invoice, setInvoice] = useState('');
+  const rate = useBtcRate();
+  const [isLoading, setIsLoading] = useState(true);
+  const { toast } = Molecules.useToast();
+
+  const requestLightningInvoice = React.useCallback(async () => {
+    try {
+      setIsLoading(true);
+      const invoice = await Homegate.requestLightningInvoice();
```
tipogi: *Homegate* is a `service` as you define but the organisms does not have access to the services. The only way to access to the `services` is through, `controller` and after `application`. So we are missing, a homegate controller and application. In that case, you will call to `HomegateController.requestLightningInvoice()`.

src/components/organisms/HumanSelection/HumanSelection.tsx
tipogi: This is a *molecule*, it does not have any interaction with data layer or/and external services. It is just an static component

src/components/molecules/Backup/Backup.tsx
```diff
@@ -5,12 +5,31 @@ import { useRouter } from 'next/navigation';
...
-    router.push(App.ONBOARDING_ROUTES.HOMESERVER);
+  const [loading, setLoading] = useState(false);
+  const { toast } = useToast();
+  const { pubky, secretKey, inviteCode } = Core.useOnboardingStore();
+  const onHandleContinueButton = async () => {
+    setLoading(true);
+    try {
+      const keypair = { pubky, secretKey };
+      const signupToken = inviteCode;
+      await Core.AuthController.signUp({ keypair, signupToken });
```
tipogi: Adding a controller, that molecule now is organism

PR #490, opened 2025-12-02
bump: pubky-app-specs: 0.4.1

tipogi:


tipogi:
@catch-21 feel free to push on top, if you see something wrong

PR #504, opened 2025-12-03
Core - Initial setup Search

afterburn:
@tipogi I made this PR a draft to get some early feedback from you. I added a search system which reuses existing services. LMK if you want me to take this into a different direction.


afterburn:
> I would add the search apis in its related domain controller

@tipogi Okay, ill refactor to that approach then. I personally feel more for having all search related functionality in a single system, but your way will work too.

src/core/controllers/search/search.types.ts
tipogi: I think that one is not in use anymore

src/core/controllers/search/search.ts
```diff
@@ -0,0 +1,29 @@
...
+import { type TPrefixSearchParams } from '@/core/services/nexus';
+
+export class SearchController {
+  private constructor() {}
+
+  /**
+   * Search users by ID prefix (pubky)
+   * @returns Array of user IDs (pubkeys) matching the search prefix
+   */
+  static async getUsersById(params: TPrefixSearchParams): Promise<string[]> {
```
tipogi: Shall we create a common return type as `SearchResult`. IDK where it should be the best to type, maybe in the local service?

PR #545, opened 2025-12-08
feat: Client-side muted post filtering

afterburn:
This PR serves as a starting point for the problem described below:

Nexus doesn't know who muted who (for privacy reasons) and shouldn't know. Instead the client filters muted users' posts. However, in such a setup requesting 10 posts might only return 3 (if 7 of them were from muted users) resulting in a janky UI and loading experience.

This PR solves the problem by introducing a queue:
  - Fetching posts from Nexus, filtering out muted users, adding the remaining posts to a queue
  - Keep fetching until queue has enough posts/matches limit (max 5 iterations), each time filtering muted users' posts.
  - Return the requested amount of posts from the queue (dequeuing them) once limit is reached, and saving the remaining for next pagination request.

See TLDraw for more info:
https://www.tldraw.com/f/zfIaeNSAmM_2GrIUs_tj8?d=v-1213.-251.3086.1736.FKRH66j8CJrNurh9TpvNl
  
  Notes
  - Queue is stored in IndexedDB so it persists across refreshes. If we don't need persistence, we could use Zustand instead.
  - This PR does not handle immediate mute/unmute UI updates. Frontend team can decide on approach.


src/core/application/stream/posts/post.ts
```diff
@@ -77,19 +82,77 @@ export class PostStreamApplication {
...
+    const mutedUserIds = new Set(mutedStream?.stream ?? []);
+
+    let isFirstFetch = true;
+    const result = await postStreamQueue.collect(streamId, {
+      limit,
+      cursor: streamTail,
+      filter: (posts) => MuteFilter.filterPosts(posts, mutedUserIds),
+      fetch: async (cursor) => {
+        // First fetch checks cache because we might be able to reuse leftover posts from previous fetch, subsequent fetches go directly to Nexus
+        const result = isFirstFetch
```
tipogi: can you destruct that object 😇, `{ posts, cacheMissIds, timestamp }`

tipogi (approved):
Nice! NIT comments, if not ready to merge!

afterburn:
Hey @MiguelMedeiros, to answer your questions:
1) The PR description is old, @tipogi and I discussed and concluded we didnt need the complexity from storing the queue in IndexedDB to persist across refreshes, thats why I moved it to an in-memory map. On a refresh the queue is thrown away and we just fetch as per usual. No risk of duplicates since the collect method dedupes using a seen Set before adding posts to the result. There shouldn't be gaps either since the post stream cache still lives in indexeddb, the queue is just for holding overflow posts between pagination calls.

2) Yep the service internally calls the model. `LocalPostService.readDetails` just wraps `PostDetailsModel.findById`, so it is testing the right code path.

3) After merging master into this branch just now I noticed the test has been added back!

PR #559, opened 2025-12-09
feat: Setup PWA

afterburn:
Copied the manifest and images over from legacy app into franky. Added some optional fields to manifest.json. Please LMK if you want me to change anything @MiguelMedeiros @catch-21

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
         "react-easy-crop": "5.5.6",
         "react-markdown": "10.1.0",
         "react-syntax-highlighter": "16.1.0",
-        "react-tweet": "3.2.2",
+        "react-tweet": "3.3.0",
         "remark-gfm": "4.0.1",
         "serwist": "^9.2.3",
         "sonner": "2.0.7",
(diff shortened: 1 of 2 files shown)


MiguelMedeiros:
@afterburn , there is missing Service Worker (Critical for PWA)

The manifest alone doesn't make a full PWA. Without a service worker, the app:
- Won't pass PWA installability checks on some browsers
- Won't work offline
- Won't show the install prompt in all cases

Consider adding a service worker using `next-pwa` or similar. Even a minimal one enables the install banner.

public/manifest.json
```diff
@@ -0,0 +1,105 @@
+{
+  "name": "Pubky App",
+  "short_name": "Pubky",
```
MiguelMedeiros: The `id` field is recommended by Chrome for uniquely identifying your PWA across updates. Without it, changing `start_url` or `scope` can create duplicate installations.

```suggestion
  "short_name": "Pubky",
  "id": "/",
```

secondl1ght:
Just adding the specific commit where I configured this for my demo (https://github.com/secondl1ght/local-first-demo/commit/a5cfab3324b9eb2f588723fec7133d944ee46606), but maybe double-check the docs as well to see if anything has changed since then. 👍🏼

catch-21:
Tested on c7168a47429e52f03c02d510f0f1932d159fcb1f

Proven can install using: MacOS Chrome, Android Brave and iPhone Safari.

Raised https://github.com/pubky/franky/issues/607

PR #576, opened 2025-12-10
feat(app): dialog new post

### LINKED ISSUE #568: New post dialog
Implement New Post dialog

Figma:
https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=25225-82617&t=LBABqoN1KPsrebtg-4

infin1t3:
Resolves #568 

https://github.com/user-attachments/assets/b00aef7c-0b7d-4abe-9f52-cdcfca5b9908

### DIFF
--- a/src/components/organisms/DialogBackupPhrase/DialogBackupPhrase.test.tsx
+++ b/src/components/organisms/DialogBackupPhrase/DialogBackupPhrase.test.tsx
@@ -27,12 +27,16 @@ vi.mock('next/image', () => ({
 }));
 
 // Mock stores
-vi.mock('@/core', () => ({
-  useOnboardingStore: () => ({
-    secretKey: 'mock-secret-key',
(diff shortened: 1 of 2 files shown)


src/components/organisms/DialogConfirmDiscard/DialogConfirmDiscard.tsx
```diff
@@ -0,0 +1,38 @@
+'use client';
```
infin1t3: We don't have a design for this cause in new version we should have drafts.

Used pattern from Pubky app.

catch-21:
Tested on 742664928318c3f231b6bb807634518f930f8a7c
Seems to function well and matches design. Only issue is that posts are not added optimistically (unsure if this is intentional for some reason), see https://github.com/pubky/franky/issues/618
(still to review tests)

PR #579, opened 2025-12-11
feat(app): add external link confirmation dialog and checkbox component

### LINKED ISSUE #578: Implement a Dialog for checking before navigating to links
-- TBI ---

infin1t3:
Resolves #578 

https://github.com/user-attachments/assets/a203b311-d7b8-4bcd-ba30-d4b2504017ca

### DIFF
--- a/src/components/organisms/LanguageSelector/LanguageSelector.test.tsx
+++ b/src/components/organisms/LanguageSelector/LanguageSelector.test.tsx
@@ -1,0 +1,121 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, fireEvent } from '@testing-library/react';
+import { LanguageSelector } from './LanguageSelector';
+
+// Mock settings store
+const mockSetLanguage = vi.fn();
+const mockUseSettingsStore = vi.fn();
+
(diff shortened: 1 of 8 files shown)


src/components/molecules/ProfilePageLinks/ProfilePageLinks.tsx
```diff
@@ -1,54 +1,91 @@
...
       })) || [],
     [links],
   );
 
+  const handleLinkClick = useCallback(
+    (url: string, e: React.MouseEvent) => {
+      e.preventDefault();
+
+      // mailto: links bypass check
+      if (url.startsWith('mailto:')) {
```
infin1t3: Did this intentionally - links don't navigate to external sites; they open the user's email client
Lower risk than web links (no redirects, no JavaScript execution)

catch-21:
Tested on 742664928318c3f231b6bb807634518f930f8a7c
This functions as intended, by only displaying confirmation dialog when navigating from a profile link. It can be configured via settings.

One comment i have is that is also displays when navigating to another url of the same domain. For example, one could link to their business pubky profile. Raised https://github.com/pubky/franky/issues/670

(still need to review tests)

PR #588, opened 2025-12-12
tests: update e2e-test.yml

catch-21:


catch-21:
Passing https://github.com/pubky/franky/actions/runs/20163451515/job/57881091835

PR #617, opened 2025-12-16
feat: core - post moderation

afterburn:
Exposes moderation state on posts so UI can blur flagged content.

**What it does:**
- Detects if a post has been tagged as moderated (by checking for specific tags from the moderator account)
- Adds `is_moderated` and `is_blurred` fields to post details on read (enrich post with 2 new fields)
- `is_blurred` considers: moderation status + user's per-post preference + global setting
- Stores user's per-post unblur choices in local DB

**Usage:**
- All display paths (bootstrap, feed pagination, single post) use `usePostDetails` → `PostApplication.getPostDetails`
- Whether post comes from Nexus fetch or IndexedDB cache, `is_moderated` and `is_blurred` fields are computed on read and returned through `Core.EnrichedPostDetails`.
- UI can use fields to blur, and can call `ModerationController.setPostBlur(<postId>, false)` when user wants to view post.

**Future optimization**
We currently do 2 queries per post to a) get tags, and b) get the moderation data from the dexie tables. Instead we could:
1. Pass tags when we get moderation state, so we no longer have to read from db.
2. Store is_moderated, is_blurred fields in DB, requires modifications to PostDetails model.

### DIFF
--- a/src/core/application/moderation/moderation.ts
+++ b/src/core/application/moderation/moderation.ts
@@ -14,7 +14,7 @@ export class ModerationApplication {
   ): Promise<Core.EnrichedPostDetails[]> {
     if (posts.length === 0) return [];
 
-    const records = await Core.ModerationModel.findByIds(posts.map((p) => p.id));
+    const records = await Core.LocalModerationService.getModerationRecords(posts.map((p) => p.id));
     const recordMap = new Map(records.map((r) => [r.id, r]));
 
     return posts.map((post) => {
--- a/src/core/services/local/moderation/moderation.ts
+++ b/src/core/services/local/moderation/moderation.ts
@@ -34,4 +34,18 @@ export class LocalModerationService {
       });
     }
   }
+
+  /**
+   * Gets the moderation records for multiple posts.
+   */
+  static async getModerationRecords(postIds: string[]): Promise<Core.ModerationModelSchema[]> {
+    try {
+      return await Core.ModerationModel.findByIds(postIds);
+    } catch (error) {
+      throw Libs.createDatabaseError(Libs.DatabaseErrorType.QUERY_FAILED, 'Failed to get moderation records', 500, {
+        error,
+        postIds,
+      });
+    }
+  }
 }


tipogi:
@secondl1ght what do you think about that impl?

src/core/controllers/moderation/moderation.ts
```diff
@@ -0,0 +1,17 @@
+import * as Core from '@/core';
+
+export class ModerationController {
+  private constructor() {}
+
+  static async setPostBlur({ postId, blur }: { postId: string; blur: boolean }): Promise<void> {
+    return Core.ModerationApplication.setBlur(postId, blur);
+  }
+
+  static async enrichPost({ post }: { post: Core.PostDetailsModelSchema }): Promise<Core.EnrichedPostDetails> {
```
secondl1ght: I don't think these enrich Controllers will ever be used, because the normal post details Controller has already been enriched with moderation right?

src/core/application/moderation/moderation.ts
```diff
@@ -4,35 +4,26 @@ import { shouldBlur } from './moderation.utils';
...
-    const isBlurDisabledGlobally = !Core.useSettingsStore.getState().privacy.blurCensored;
-    const record = await Core.LocalModerationService.getModerationRecord(post.id);
+  static async enrichPostsWithModeration(
+    posts: Core.PostDetailsModelSchema[],
+    isBlurDisabledGlobally: boolean,
+  ): Promise<Core.EnrichedPostDetails[]> {
+    if (posts.length === 0) return [];
 
-    if (!record) {
+    const records = await Core.ModerationModel.findByIds(posts.map((p) => p.id));
```
tipogi: application layer, just calls to the `service` layer, we are accessing to the `model`.

tipogi (changes requested):
ready to merge, after the requested change 😉

tipogi (approved):
Merge it! 🚀

secondl1ght:
OK so looks like we decided to do the profile moderation in a follow-up PR?

PR #631, opened 2025-12-17
fix: max char input validation for edit profile name and bio

catch-21:
Fixes https://github.com/pubky/franky/issues/630

Add input validation for Edit Profile 'name' and 'bio' fields to highlight when maximum character count, as defined in pubky-app-specs, has been exceeded.

If we'd rather prevent text entry when maximum char count is reached then we should instead use `Molecules.InputField` `maxLength`.

Screenshot 2025-12-17 at 19 06 03

### DIFF
--- a/src/components/molecules/Header/Header.test.tsx
+++ b/src/components/molecules/Header/Header.test.tsx
@@ -123,11 +123,15 @@ vi.mock('@/libs', async (importOriginal) => {
 });
 
 // Mock the config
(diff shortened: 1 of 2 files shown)


MiguelMedeiros (dismissed):

MiguelMedeiros (approved):

catch-21:
Verified in b6edc20b21d17cd76eca4086f4e6a22ad6340ba5

PR #651, opened 2025-12-23
style: better handling of word breaks in posts

### LINKED ISSUE #541: [Bug]: Words in post can wrap
### Describe the bug

It is possible for any words, including parsed pubky that displayed username like `@abcde`, to wrap over lines even when they are relatively short. It would be easier to read and look better if words are always shown on a single line.

It is acceptable to wrap very long words (consider hyphenating). That applies to both desktop and mobile.

### Reproduce

1. Create a post of a paragraph of text with a profile reference part way through
2. Observe multiple words including parsed text can wrap over two lines

### Screenshots / Recording

I tried viewing this exact post in pubky-app (desktop) with various widths and I don't think this is possible for any words to wrap.

Image

-----

This wrapping of long usernames can happen in pubky-app, but generally words do not wrap there.
Image

### Web Browser

_No response_

### Pubky App Version

8c58e2f86b962edca703e43043b32132df59d513

### Console log output

```shell

```

secondl1ght:
### Summary

This PR improves the text wrapping behavior in the `PostText` component to provide better handling of long words and URLs while maintaining text readability.

### Changes

- **Updated CSS classes for better word wrapping**
  - Replaced `break-all` with `wrap-anywhere` for smarter line breaking that only breaks words when necessary
  - Added `hyphens-auto` to enable automatic hyphenation for cleaner word breaks

### Why?

The previous `break-all` behavior would aggressively break words at any character, which could result in awkward mid-word breaks even when unnecessary. The new approach:

- `wrap-anywhere` - Only breaks words when they would overflow, providing a better reading experience
- `hyphens-auto` - Adds proper hyphenation when words do need to break, making the breaks more natural and readable

### Files Changed

| File | Changes |
|------|---------|
| `src/components/molecules/PostText/PostText.tsx` | Updated text wrapping CSS classes |

---

### Related

Closes #541


MiguelMedeiros (approved):
**Congrats!** This PR passed the review with **ZERO issues found** - that's rare and deserves recognition! 🎉

PR #655, opened 2025-12-30
fix: dialog overlay dismiss on mobile

### LINKED ISSUE #582: Dismiss media carousel on outside tap mobile
da63ff51f2a8ee5d0c3edc83afcb0a5fe9ea3cc5

On mobile, it in intuitive to be able to dismiss the carousel by tapping outside of the area. Currently, need to find and tap the close button. Would be good to be a bit more flexible here for easier and quicker UX.

https://github.com/user-attachments/assets/71f29dd2-4401-4b58-ab58-f0e6c209dbe2

_Originally posted by @catch-21 in https://github.com/pubky/franky/issues/537#issuecomment-3641757895_

secondl1ght:
### Summary

This PR fixes the dialog overlay dismiss behavior on mobile devices. The overlay click-to-close functionality was working correctly on desktop but not on mobile due to internal Radix UI logic with `stopPropagation`. This change ensures consistent behavior across all devices.

#### `src/components/atoms/Dialog/Dialog.tsx`

- **Overlay click behavior**: Added programmatic trigger to click the dialog close button when the overlay is clicked, resolving Radix's internal mobile handling issues when using `stopPropagation`
- **Close button rendering**: Refactored the `DialogClose` component to always render in the DOM but conditionally apply visibility:
  - When `showCloseButton` is `true`: Shows the styled close button in the top-right corner
  - When `showCloseButton` is `false`: The close button is hidden via `hidden` class but remains in the DOM

#### `src/components/atoms/Dialog/Dialog.test.tsx`

- ✅ Added test: **closes dialog when overlay is clicked** - Verifies that clicking the overlay triggers `onOpenChange(false)`
- ✅ Added test: **renders close button with hidden class when showCloseButton is false** - Ensures the close button is always in the DOM with proper visibility handling
- ✅ Updated snapshots

### Why This Change?

Radix UI's dialog overlay has inconsistent close behavior on mobile devices when used with `stopPropagation`. By ensuring the `DialogClose` component is always present in the DOM (even if visually hidden), we can programmatically trigger it when the overlay is clicked. This provides a reliable cross-platform solution that works on both desktop and mobile.

### Technical Details

- Uses `document.getElementById('dialog-close-btn')?.click()` pattern to trigger the close action
- The `DialogClose` component from Radix handles the actual dialog state management
- `cn()` utility used for conditional class application
- Tests use `userEvent` for realistic user interaction simulation

---

### Related

Closes #582

### DIFF
--- a/cypress/e2e/onboarding.cy.ts
+++ b/cypress/e2e/onboarding.cy.ts
@@ -29,7 +29,7 @@ describe('Onboarding', () => {
     cy.get('#backup-btn').should('exist').click();
     // confirm backup dialog can be shown and closed
     cy.get('#backup-dialog-title').should('exist');
-    cy.get('#dialog-close-btn').click();
+    cy.get('[data-testid="dialog-close"]').click();
     // confirm backup done button can be clicked
     cy.get('#backup-done-btn').click();
     // confirm backup confirmation dialog can be shown and dismissed
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -71,7 +71,7 @@ Cypress.Commands.add(
 
       // either skip phrase confirmation or confirm it
       if (backup?.includes(BackupType.RecoveryPhraseWithoutConfirmation)) {
-        cy.get('#dialog-close-btn').click();
+        cy.get('[data-testid="dialog-close"]').click();
       } else if (backup?.includes(BackupType.RecoveryPhraseWithConfirmation)) {
         cy.get(`@recoveryPhrase-${profileName}`).then((recoveryPhrase) => {
           confirmRecoveryPhrase(recoveryPhrase.toString());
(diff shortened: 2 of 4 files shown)


MiguelMedeiros (dismissed):
Hey! Made some changes to the Dialog component in your PR:

1. **Replaced `getElementById` with `ref`** - Before, there was a hardcoded `id="dialog-close-btn"` on the close button and it used `getElementById` to click it when clicking the overlay. The problem is that if multiple dialogs are open at the same time (like a confirmation dialog inside another one), it could close the wrong dialog. Now it uses refs which is safer and keeps everything within React's component tree.

2. **Fixed accessibility warnings** - Added `aria-describedby={undefined}` to DialogContent to get rid of that annoying "Missing Description" warning in the console.

3. **Updated e2e tests** - Since I removed the ID, I added `data-testid="dialog-close"` to the component and updated the Cypress tests (`onboarding.cy.ts` and `commands.ts`) to use `[data-testid="dialog-close"]` instead of `#dialog-close-btn`.

Updated the snapshots as well to reflect the changes. 👍

cypress/e2e/onboarding.cy.ts
```diff
@@ -29,7 +29,7 @@ describe('Onboarding', () => {
     cy.get('#backup-btn').should('exist').click();
     // confirm backup dialog can be shown and closed
     cy.get('#backup-dialog-title').should('exist');
-    cy.get('#dialog-close-btn').click();
+    cy.get('[data-testid="dialog-close"]').click();
```
catch-21: We should use `data-cy` instead

secondl1ght: Saw you moved this to Done, will you include this fix in one of your follow-up QA PRs?

PR #420, opened 2025-11-19
feat(ProfilePage): integrate TimelinePosts for user posts and replies

MiguelMedeiros:
## 🎯 Overview

This PR implements the **Posts** and **Replies** tabs for the user profile page with infinite scroll pagination, and fixes the post count display issue in the profile sidebar.

https://github.com/user-attachments/assets/17dda42d-e4be-43f2-9180-ea55b6e2e05c

### 1. **Profile Page Posts & Replies** 
- ✅ Implemented `ProfilePagePosts` to display user's posts using `author:{userId}` stream
- ✅ Implemented `ProfilePageReplies` to display user's replies using `author_replies:{userId}` stream
- ✅ Both components use infinite scroll pagination
- ✅ Full test coverage with unit tests

### 2. **Code Reuse & DRY**
- ✅ Refactored `TimelinePosts` to accept optional `streamId` prop
- ✅ Removed ~460 lines of duplicated code by reusing `TimelinePosts` component
- ✅ `ProfilePagePosts` and `ProfilePageReplies` are now lightweight wrappers (20 lines each)

### 3. **Fixed Post Count Display Bug**
- 🐛 **Problem**: Profile sidebar showed incorrect post counts (e.g., 27 posts when user only had replies)
- ✅ **Solution**: UI now calculates actual posts as `totalPosts - replies` in `useProfileStats`
- ✅ Works correctly for all scenarios:
  - User with only posts: displays correctly
  - User with only replies: shows 0 posts, N replies
  - User with both: shows correct counts for each

### Component Architecture

```typescript
// Timeline with filters (Home page)
<TimelinePosts /> 

// Profile Posts tab
<TimelinePosts streamId="author:userId" />

// Profile Replies tab
<TimelinePosts streamId="author_replies:userId" />
```

### Stream IDs

- **Posts**: `author:{userId}` - fetches only non-reply posts from user
- **Replies**: `author_replies:{userId}` - fetches only replies from user

### Post Count Fix

**Before:**
```typescript
posts: userCounts?.posts ?? 0  // ❌ Includes replies in count
```

**After:**
```typescript
const totalPosts = userCounts?.posts ?? 0;
const repliesCount = userCounts?.replies ?? 0;
const actualPostsCount = Math.max(0, totalPosts - repliesCount);
// ✅ Correctly shows only non-reply posts
```

## 🧪 Testing

- ✅ All existing tests passing (2933 tests)
- ✅ Added comprehensive tests for `ProfilePagePosts`
[...]

Issue #123, opened 2025-09-22
[Bug]: Loading /onboarding/ shows 'Not Found' page instead of redirect


catch-21:
### Describe the bug

When loading `/onboarding` or any non-existent sub like `/onboarding/any` a "Could not find requested resource" message is displayed with no button to navigate back (other than the logo). It should redirect to a useful page such as `/` or `/onboarding/install`

### Reproduce

1. Go to `/onboarding`
2. Observe "Not Found page" , see screenshot

### Screenshots / Recording

<img width="845" height="469" alt="Image" src="https://github.com/user-attachments/assets/0bed6472-e164-444a-830c-3805c2c4f5f9" />

### Pubky App Version

4dc08e4ec30c515a464601534cbe77c0e08e210d

### Console log output

```shell

```

PR #668, opened 2025-12-31
refactor: hashtag validation to allow start/only numbers

### LINKED ISSUE #485: [Bug]: Highlight hashtags that are numbers or symbols
### Describe the bug

https://github.com/pubky/franky/pull/473 has added hashtag handling, although it doesn't allow hashtags that start with a number or symbol such as `#?` or `#420`. Pubky-app does recognise these.

I expect we should allow hashtags to be any value, although there are length (20) and character (no spaces, commas, or colons) rules defined in pubky-app-specs for tags to consider.

### Reproduce

1. View a post that uses hashtags starting with symbol or number
2. Observe the hashtags are not highlights green

### Screenshots / Recording

Image

### Web Browser

_No response_

### Pubky App Version

45dec8800547a4b8163387998e6f88db90441f8d

### Console log output

```shell

```

secondl1ght:
# 🏷️ Allow more numeric hashtags

## Summary

Updates the hashtag regex pattern to allow hashtags that start with or contain only numbers (e.g., `#420`, `#1984`, `#12rules`). Previously, hashtags were required to start with a letter, which restricted some legitimate use cases. Underscore handling was also improved.

### `PostText.utils.ts`
- **Updated regex pattern** from `#[a-zA-Z][a-zA-Z0-9_]*` to `#[a-zA-Z0-9]+(?:_[a-zA-Z0-9]+)*`
- **Improved underscore validation**: Underscores are now only allowed between alphanumeric characters (prevents `#_test`, `#test_`, `#test__more`)
- Updated comment to reflect new pattern behavior

### `PostText.utils.test.ts`
- Renamed test "requires hashtag to start with a letter" → "allows hashtag to start with a letter"
- Changed test "does not match hashtags starting with a number" → "allows hashtags starting with a number" (now expects match)
- Changed test "does not match hash followed by only numbers" → "matches hash followed by only numbers" (now expects match)
- Updated "handles mixed valid and invalid hashtag patterns" test to expect `#123numeric` as valid
- **Added new underscore validation tests:**
  - `allows multiple underscores separating words` - verifies `#one_two_three` works
  - `does not match hashtags starting with underscore` - verifies `#_invalid` doesn't match
  - `stops at trailing underscore` - verifies `#hello_` only matches `#hello`
  - `stops at double underscore` - verifies `#hello__world` only matches `#hello`

### `PostText.test.tsx`
- Changed test "does not parse hashtag starting with number" → "parses hashtag starting with number" (now expects hashtag to render)

## Hashtag Pattern Behavior

| Pattern | Before | After |
|---------|--------|-------|
| `#test` | ✅ | ✅ |
| `#123` | ❌ | ✅ |
| `#1test` | ❌ | ✅ |
| `#test1` | ✅ | ✅ |
| `#hello_world` | ✅ | ✅ |
| `#_test` | ✅ | ❌ |
| `#test_` | ✅ | ❌ |
| `#test__more` | ✅ | ❌ |

## Related

Closes #485 

## Notes

- The stricter underscore validation is a side effect improvement that prevents malformed hashtags
- All 211 tests pass (117 in utils, 94 in component)

### DIFF
--- a/src/components/molecules/PostHeaderUserInfo/PostHeaderUserInfo.test.tsx
+++ b/src/components/molecules/PostHeaderUserInfo/PostHeaderUserInfo.test.tsx
@@ -1,13 +1,73 @@
 import { render, screen } from '@testing-library/react';
-import { describe, it, expect, vi } from 'vitest';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { PostHeaderUserInfo } from './PostHeaderUserInfo';
 import * as Libs from '@/libs';
+
+// Mock hooks
+const mockUseUserProfile = vi.fn();
+const mockUseIsFollowing = vi.fn();
+const mockUseFollowUser = vi.fn();
+const mockUseAuthStore = vi.fn();
+const mockUseProfileStats = vi.fn();
+const mockUseProfileConnections = vi.fn();
+const mockUseCurrentUserProfile = vi.fn();
+
+vi.mock('@/hooks', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@/hooks')>();
+  return {
+    ...actual,
+    useUserProfile: () => mockUseUserProfile(),
+    useIsFollowing: () => mockUseIsFollowing(),
+    useFollowUser: () => mockUseFollowUser(),
+    useProfileStats: () => mockUseProfileStats(),
+    useProfileConnections: () => mockUseProfileConnections(),
+    useCurrentUserProfile: () => mockUseCurrentUserProfile(),
+    CONNECTION_TYPE: {
+      FOLLOWERS: 'followers',
+      FOLLOWING: 'following',
+      FRIENDS: 'friends',
+    },
+  };
+});
+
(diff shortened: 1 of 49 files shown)


catch-21:
It looks like tags with underscores is working, which is good but differs to your claim in the table above.
However, hyphens don't work. The trend of using hyphens to separate words means we should probably support them here, and add test for it.

Screenshot 2026-01-01 at 15 28 47

secondl1ght:
> It looks like tags with underscores is working, which is good but differs to your claim in the table above. However, hyphens don't work. The trend of using hyphens to separate words means we should probably support them here, and add test for it.
> 
> Screenshot 2026-01-01 at 15 28 47

How do the underscores differ from the table?

I was following Twitter's approach for hashtags which doesn't allow hyphens (or only numbers), but I think we can support hyphens without issue. We should stop there for symbol support though, as mentioned in the comment you reference in the parent issue and the Slack discussion about this we decided against it. 👍🏼

MiguelMedeiros (approved):
LFG!

Issue #695, opened 2026-01-04
[Bug]: Own following list can show users with 'Follow' button


catch-21:
### Bug Description

When viewing your own profile's 'Following' list it is possible to have some users listed with the 'Follow' button even though they have not been unfollowed. This appears to only happen when logging out to switch user profiles, as is done in notification E2E tests. Never the less, the button state should always reflect the correct state.

### Steps to Reproduce

1. Create two profiles
2. Profile 1 follows profile 2
3. Profile 1 signs out
5. Profile 2 clicks profile 1 username in notification for new follow to navigate to their profile page
6. Profile 2 clicks 'Follow' button to follow profile 1
7. Profile 2 signs out
8. Profile 1 views own 'Following' list on profile page
9. Observe 'profile 2' has 'Follow' button

### Affected Area

Feed / Timeline

### Suspected Layer (if known)

None

### Frequency

Always (100%)

### Expected Behavior

Users listed in own profile's 'Following' list should have button with text 'Following' if they have not yet been unfollowed.

### Screenshots / Recording

<img width="987" height="500" alt="Image" src="https://github.com/user-attachments/assets/07bbdbe7-b4bc-4927-8b04-b7971bff691d" />

### Console Log Output

```shell

```

PR #706, opened 2026-01-05
fix: #695 - Fix cached relationships bug

afterburn:
Users fetched during notification processing were missing viewerId, causing relationships to be cached without the viewer's perspective and showing incorrect Follow button state.


tipogi (approved):
nice

Issue #526, opened 2025-12-05
[Bug]: Avatar thumbnail remains in header after removing avatar from profile


catch-21:
### Describe the bug

When you remove your profile picture the thumbnail does immediately update on your profile page but not in the header.

### Reproduce

1. Go to your profile page (that already has an avatar)
2. Edit your profile to remove the avatar
3. Save profile edit
4. Observe header thumbnail remains as the removed avatar image

### Pubky App Version

dacd0a6a6aea518ae8202978b8848efa2c3d861a

### Console log output

```shell

```

catch-21:
Verified in 8f316597784c7ec821d49d4406d815ff7575c460

Avatar no longer goes black on removal.

PR #728, opened 2026-01-07
fix: avatar not immediately shown on own profile page after adding

### LINKED ISSUE #524: [Bug]: Avatar not immediately shown on own profile page after adding
### Describe the bug

After editing profile to add an avatar it is not immediately applied on the profile page or header thumbnail. Reloading the page applies it.

Also, if you had a different avatar previously this may be loaded instead of the new avatar (see second video). I suspect both cases are related to cache and will have same fix. **If this isn't fixed same time then we need to open separate issue.**

### Reproduce

1. Go to own edit profile page
2. Add an avatar
3. Save profile edit
4. Observe avatar placeholder is replaced with black
5. Observe header thumbnail doesn't update

### Screenshots / Recording

https://github.com/user-attachments/assets/13652c01-1ba5-403c-80f7-32c730f0dc53

https://github.com/user-attachments/assets/090394c9-f237-42e8-b920-12f89e5fccec

### Web Browser

_No response_

### Pubky App Version

dacd0a6a6aea518ae8202978b8848efa2c3d861a

### Console log output

```shell

```

SHAcollision:
Fixes #524 and #526. Autonomous fix by Antigravity.

## Description
This PR fixes the issue where the user's avatar would not update immediately on the profile page and header after uploading a new image. The issue was caused by the browser caching the avatar URL, as the URL remained constant even when the underlying file changed.

To fix this, I implemented a cache-busting mechanism using the `indexed_at` timestamp of the user profile.

## Changes
-   **API**: Updated `filesApi.getAvatarUrl` to accept an optional `version` parameter.
-   **Application/Controller**: Propagated the `version` parameter through `FileApplication` and `FileController`
-   **Local Service**: Updated `LocalProfileService.updateDetails` to refresh the `indexed_at` timestamp on every profile update.
-   **UI**: Updated `useUserProfile` and `HeaderSignIn`  to pass the `indexed_at` timestamp as a version query parameter (`?v=<timestamp>`) to the avatar URL.

### Manual Verification
I verified the fix by signing up a new user, uploading an initial avatar, and then updating it to a different image. In both cases, the avatar updated immediately without a page reload.

**Initial Upload (Clean Run):**
profile_page_after_avatar_save_1767780027523

**Update Avatar:**
profile_after_second_avatar_save_no_reload_1767780223566

### Tests
-   Added unit tests for `filesApi.getAvatarUrl` version parameter.
-   Updated existing tests in `useUserProfile.test.tsx` to support the new signature.
-   Ran `npm test` successfully.

### DIFF
--- a/src/components/organisms/PostMain/PostMain.test.tsx
+++ b/src/components/organisms/PostMain/PostMain.test.tsx
@@ -113,16 +113,23 @@ vi.mock('@/organisms', () => ({
     taggedKind,
     showCount: _showCount,
     showInput: _showInput,
+    showAddButton: _showAddButton,
     addMode: _addMode,
   }: {
     taggedId: string;
     taggedKind: unknown;
     showCount?: boolean;
     showInput?: boolean;
+    showAddButton?: boolean;
     addMode?: boolean;
     [key: string]: unknown;
   }) => (
-    <div data-testid="clickable-tags-list" data-tagged-id={taggedId} data-tagged-kind={String(taggedKind)}>
+    <div
+      data-testid="clickable-tags-list"
+      data-tagged-id={taggedId}
+      data-tagged-kind={String(taggedKind)}
+      data-show-add-button={String(_showAddButton)}
+    >
       ClickableTagsList {taggedId}
     </div>
   ),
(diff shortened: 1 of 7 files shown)


MiguelMedeiros (approved):

catch-21:
Some problems with this one. See QA notes in issues.

catch-21:
I think the problem is to do with nexus not having the image immediately available. Franky needs to retry to get it, or not overwrite the one locally added in cache, or similar.

Issue #627, opened 2025-12-17
[Bug]: 'See new posts' button inserts older posts above newer posts in feed


catch-21:
### Describe the bug

It is possible for the 'See new post' button to insert posts above ones that have since been created, this means the feed is no longer sorted by 'Recent'.

### Reproduce

1. View feed with 'Recent' sorting
2. Create new post from other instance (or wait for one to come in)
3. Wait for 'See new post' button to appear
4. Create some new posts so they optimistically show in the feed
6. Click 'See new post button;
7. Observe the older posts are inserted at top of feed, above the ones that were created since

### Screenshots / Recording

<img width="694" height="680" alt="Image" src="https://github.com/user-attachments/assets/c7afe3e7-7f4e-4f13-b64f-e47b95273f9f" />

### Pubky App Version

bc4e0516dc366a6a3b8b9a86422b54c606694166

### Console log output

```shell

```

PR #730, opened 2026-01-07
fix: #627 - Preserve chronological order when pressing See new posts button

afterburn:
Posts were inserted without sorting, causing older posts to appear above newer user-created ones as described in #627. This PR fixes it in both core (database merge) and UI hook (prepend) because each maintains separate state. The database for persistence, the hook for optimistic updates that haven't synced yet.

### DIFF
--- a/src/hooks/useStreamPagination/useStreamPagination.ts
+++ b/src/hooks/useStreamPagination/useStreamPagination.ts
@@ -51,6 +51,9 @@ export function useStreamPagination({
         let result: Core.TReadPostStreamChunkResponse;
 
         if (isInitialLoad) {
+          // Sanity check: if the first post in the cached stream is older than configured max age, clear the stream cache
+          await Core.StreamPostsController.clearStaleStreamCache({ streamId });
+
           const cachedLastPostTimestamp = await Core.StreamPostsController.getCachedLastPostTimestamp({ streamId });
           setStreamTail(cachedLastPostTimestamp);
 
--- a/src/config/nexus.ts
+++ b/src/config/nexus.ts
@@ -5,3 +5,4 @@ export const NEXUS_NOTIFICATIONS_LIMIT = 30;
 export const NEXUS_POSTS_PER_PAGE = 10; // Number of posts to fetch per page in streams
 export const NEXUS_USERS_PER_PAGE = 10; // Number of users to fetch per page in streams
 export const CDN_URL = Env.NEXT_PUBLIC_CDN_URL;
+export const STREAM_CACHE_MAX_AGE_MS = Env.NEXT_PUBLIC_STREAM_CACHE_MAX_AGE_MS; // Maximum age for stream cache before considered stale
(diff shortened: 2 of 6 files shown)


src/core/services/local/stream/posts/posts.ts
```diff
@@ -152,7 +152,15 @@ export class LocalStreamPostsService {
...
     const combinedStream = [...unreadPostStream.stream, ...uniqueExistingPosts];
 
-    await Core.PostStreamModel.upsert(streamId, combinedStream);
+    // Sort by timestamp (indexed_at) in descending order (most recent first)
+    const posts = await Core.PostDetailsModel.findByIdsPreserveOrder(combinedStream);
+    const postTimestamps = combinedStream.map((postId, index) => ({
+      postId,
+      timestamp: posts[index]?.indexed_at || 0,
+    }));
+    const sortedStream = postTimestamps.sort((a, b) => b.timestamp - a.timestamp).map((item) => item.postId);
```
MiguelMedeiros: **🔄 DRY Violation (Low Priority)**

This sorting logic is duplicated in `useStreamPagination.ts`. Consider extracting to a shared utility:

```typescript
// In Core - e.g., src/core/utils/sorting.ts
export async function sortPostIdsByTimestamp(postIds: string[]): Promise<string[]> {
  const posts = await PostDetailsModel.findByIdsPreserveOrder(postIds);
  return postIds
    .map((postId, index) => ({ postId, timestamp: posts[index]?.indexed_at || 0 }))
    .sort((a, b) => b.timestamp - a.timestamp)
    .map((item) => item.postId);
}
```

Then use it in both places:
```typescript
const sortedStream = await Core.sortPostIdsByTimestamp(combinedStream);
```

src/hooks/useStreamPagination/useStreamPagination.ts
```diff
@@ -164,9 +164,19 @@ export function useStreamPagination({
       return;
     }
 
-    const updatedPostIds = [...newIds, ...postIdsRef.current];
-    postIdsRef.current = updatedPostIds;
-    setPostIds(updatedPostIds);
+    // Combine new and existing posts
+    const allIds = [...newIds, ...postIdsRef.current];
+
```
MiguelMedeiros: **🏃 Potential Race Condition (Low Priority)**

This reads from `postIdsRef.current`, does async work, then writes back. Rapid consecutive calls could cause race conditions:

```
Call 1: Read postIdsRef → [A, B, C]
Call 2: Read postIdsRef → [A, B, C]  (before Call 1 writes)
Call 1: Fetch timestamps...
Call 2: Fetch timestamps...
Call 1: Write postIdsRef → [X, A, B, C]
Call 2: Write postIdsRef → [Y, A, B, C]  ← overwrites Call 1's changes!
```

**Risk:** Low in practice (users rarely click "See new posts" rapidly), but worth noting for future reference.

Issue #731, opened 2026-01-07
[Bug]: Some urls such as /hot and /settings redirect to /home


catch-21:
### Bug Description

When using browser address bar to navigate to url `franky.pubky.app/hot` or `/settings` or `/bookmarks` I get redirected to `/home`. The same for reloading the page when viewing Hot, Settings or Bookmarks.

### Steps to Reproduce

1. Visit some page in the app that is not Home (e.g. Hot)
2. Reload the page
3. Observe directed to Home

### Affected Area

Navigation / Routing

### Suspected Layer (if known)

None

### Frequency

Always (100%)

### Expected Behavior

It should be possible to load any of the main pages (Hot, Settings, etc.) within the app without forced redirection to `/home`.

### Screenshots / Recording

https://github.com/user-attachments/assets/a445297f-fb24-4f50-8c8a-2e16d6e45517

### Console Log Output

```shell

```

PR #737, opened 2026-01-08
fix: #731 - Prevent premature redirect while session is being restored

afterburn:
Fixed a bug where reloading on `/hot`, `/settings`, or `/bookmarks` would redirect users to `/home`. The issue was a race condition during session restoration. On page reload, `sessionExport` (the exported string from sdk) loads from localStorage before session (actual auth object) is recreated, causing the route guard to briefly see an unauthenticated state and trigger a redirect. Added `isSessionRestorePending` check to `useAuthStatus` that keeps `isLoading` true while credentials exist but haven't been restored yet.


tipogi (approved):
TOP

PR #741, opened 2026-01-08
fix: tag labels get truncated

### LINKED ISSUE #736: [Bug]: Post tag value can be truncated
### Bug Description

When viewing a post (in feed or other list) tags with long values are truncated. This is not the case in any design or in the existing pubky-app.

### Steps to Reproduce

1. Tag a post with a long value (wide character like 'M' x20)
2. Observe it is truncated and appended with `...` when viewed in the post

### Affected Area

Tags

UI / Component (visual issue)

### Frequency

Always (100%)

### Expected Behavior

Tag value should always be visible in full when viewing on a post.
Although it will take up more space we should try to always fit a few in before hiding them to be revealed in a list when 'tag' action button is clicked.
Pubky-app always shows 3on desktop and mobile regardless of size - we could do this. Or we could aim to be my dynamic and fit as many as we can with the available space.

### Screenshots / Recording

**Desktop**

Image

**Mobile**

Image

### Console Log Output

```shell

```

### Browser & Version

_No response_

### Device

_No response_

### Additional Context

_No response_

SHAcollision:
Fixes #736

Tag labels are not truncated. The max number of tax and text to show is parametrized, maximum 3 tags of 20 chars each. We can tune these values as we find best later on.

image
image

### DIFF
--- a/src/components/atoms/Dialog/Dialog.test.tsx.snap
+++ b/src/components/atoms/Dialog/Dialog.test.tsx.snap
@@ -12,7 +12,7 @@ exports[`Dialog - Snapshots > matches snapshot for DialogContent with close butt
 >
   <div
     aria-labelledby="radix-_r_0_"
-    class="relative z-50 grid duration-200 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 m-4 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95 gap-6 rounded-lg border bg-background p-6 shadow-lg sm:rounded-xl sm:p-8"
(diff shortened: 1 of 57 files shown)


MiguelMedeiros (approved):
LGTM! ✅ The implementation correctly aligns with the design spec and pubky-app behavior.

**One consideration for follow-up:** With multiple long tags (up to 20 chars each) + the add button + tag input on smaller screens, there may be layout overflow scenarios.

Should we open a design discussion/issue to define overflow handling options?
- Wrap to multiple lines (current behavior after this PR)
- Max visible tags with "+N more" indicator
- Responsive limits (fewer visible tags on mobile)
- CSS text-overflow with tooltip on hover

Not blocking this PR - just flagging for design team awareness. 👍

PR #745, opened 2026-01-08
feat: clear stale streams

tipogi:


MiguelMedeiros (approved):
## Review v2

✅ **Previous feedback addressed!**

The sentinel value handling fix in commit `5c0bf0f` is exactly what was needed:

```typescript
// Now correctly handles both sentinel values
if (headTimestamp === Core.SKIP_FETCH_NEW_POSTS || headTimestamp === Core.FORCE_FETCH_NEW_POSTS) {
  return;
}
```

The improved comment explains the "why" clearly - great documentation.

**Optional:** Consider adding unit tests for the stale detection logic in `PostStreamApplication` (see my previous comment for examples), but this doesn't block merge.

Ship it! 🚀

Issue #688, opened 2026-01-02
[Bug]: Viewing repost that embeds an older post can show without post details and be forever 'Loading'


catch-21:
### Describe the bug

When viewing a new repost of an old post it can show "Loading header..." and "Loading content..." for the embedded post. I suspect this is because the old post has not been stored in local db because this doesn't happen when a recent post has been reposted.

### Reproduce

1. Be viewing the feed in franky (A)
2. From a separate instance of pubky-app or franky (B) that does not share db with instance from step 1, find an old post that has not been viewed by franky instance A
3. Wait for 'see new post' button to appear at top of feed for franky instance A
4. Click on 'see new post' button
5. Observe new repost is shown with embedded post that has "Loading header..." and "Loading content..." and doesn't ever load the post details

### Screenshots / Recording

<img width="763" height="337" alt="Image" src="https://github.com/user-attachments/assets/7ff0632e-6e4b-470d-8fbd-f9d93ffd1d07" />

-----

To demonstrate the bug, here a repost is created using a separate instance of pubky-app so that it comes in as a new repost in this instance of franky. Part way through the video it shows how it correctly displays in pubky-app to compare with franky.

https://github.com/user-attachments/assets/d0d70776-6a70-4c91-a310-658db833b964

### Pubky App Version

92100e42b3142b88a89fc7c4c726d14e34d136c4

### Console log output

```shell

```

PR #749, opened 2026-01-09
fix: #688 - Fix edge case in fetching original posts for reposts served from cache

afterburn:
he fix in #718 ensured original posts are fetched when reposts are retrieved from Nexus (cache miss). However, when reposts are served from cache (cache hit), `fetchMissingPostsFromNexus` is never called, so original posts that were evicted from cache aren't re-fetched. This causes repost content to not render when the repost is cached but its original post is missing.

Added logic after fetching stream posts that looks up relationships via `LocalPostService.readRelationshipsByIds`, identifies reposts, and fetches any missing original posts via `fetchOriginalPostsByUris`. This ensures original post content is always available regardless of whether reposts came from cache or Nexus.

Note: There's a related edge case in the TTL system where `forceRefreshPostsByIds` doesn't fetch original posts for reposts, and PostPreviewCard doesn't subscribe original posts to TTL tracking. This will be addressed in a separate PR as this one is aimed to ONLY correct the missed scenario in #718.


src/core/services/local/post/post.ts
```diff
@@ -63,6 +63,28 @@ export class LocalPostService {
...
+   * Reads post relationships for multiple posts
+   *
+   * @param postIds - Array of composite post IDs (author:postId)
+   * @returns Array of post relationships (undefined entries for posts not found)
+   *
+   * @throws {DatabaseError} When database operations fail
+   */
+  static async readRelationshipsByIds(postIds: string[]): Promise<(Core.PostRelationshipsModelSchema | undefined)[]> {
+    try {
+      return await Core.PostRelationshipsModel.table.bulkGet(postIds);
```
tipogi: Services shouldn’t access the tables directly — they should go through the model functions. In this case, `findByIdsPreserveOrder` is the right one to use.

MiguelMedeiros (approved):

PR #787, opened 2026-01-13
fix: e2e tests

catch-21:
Use of double quotes in the string used to compose the `.env` broke the static analysis.
This PR escapes the double quotes.


MiguelMedeiros (dismissed):
ACK

Image

https://github.com/pubky/franky/actions/runs/20974284155

PR #792, opened 2026-01-14
fix(tags): allow adding existing tag to increment counter

### LINKED ISSUE #782: [Bug]: Adding tag to post with input field doesn't work if someone else already added the same tag
### Bug Description

When adding a tag to a post by typing and hitting Enter, if the tag already exists then it is not added and counter does not increment.

### Steps to Reproduce

1. Find a post in feed that has been tagged by someone else
2. Type the same tag value into the tag input field and press Enter to submit
3. Observe tag is not added and the input field clears

### Affected Area

Tags

UI / Component (visual issue)

### Frequency

Always (100%)

### Expected Behavior

When adding a tag, if you were not the one to add it then it should be added and the existing tag counter should increment.

### Screenshots / Recording

https://github.com/user-attachments/assets/9cb89510-ae20-46b9-9350-88724a1f12c2

https://github.com/user-attachments/assets/8e195fdd-c45d-4a6e-b545-34649678da49

### Console Log Output

```shell

```

### Browser & Version

_No response_

### Device

_No response_

### Additional Context

_No response_

Orlandohub:
Fix #782 bug where adding a tag that someone else already added was blocked instead of incrementing the counter.

- Fix ClickableTagsList to filter existingTags by viewer's tags only
- Add data-cy attributes to tag components for E2E testing
- Add E2E tests for tag functionality including bug fix verification

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -72,8 +72,10 @@ NEXT_PUBLIC_CDN_URL=https://nexus.staging.pubky.app/static
 NEXT_PUBLIC_TESTNET=false
 NEXT_PUBLIC_PKARR_RELAYS=https://pkarr.pubky.app
 NEXT_PUBLIC_DEFAULT_HTTP_RELAY=https://httprelay.staging.pubky.app
-NEXT_PUBLIC_HOMESERVER_ADMIN_URL=http://localhost:6288/generate_signup_token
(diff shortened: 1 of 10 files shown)


MiguelMedeiros (approved):
ACK

PR #821, opened 2026-01-15
feat: bump pubky-sdk and use `pubky<z32>` prefixed displays

SHAcollision:
This PR
- Bumps `@synonymdev/pubky` to final release of `0.6.0`
- Uses the new public key display formats with `.toString()` to match design keys `pubky{z32}`
- Removes `pk:`
- Parses mentions in comments with `pubky...` and `pk:` (for retrocompatibility)

All transport / database / routes / api calls should remain using the `z32` strings. Display for humans use `.toString()` except for shorten keys like `AAAA...ZZZZ` where the pubky prefix is not shown.


SHAcollision:
Pushed a fix to the greptile comments

PR #917, opened 2026-01-20
fix: disable "Export recovery phrase " button in onboarding backup

### LINKED ISSUE #848: [Bug]: Onboarding / Homegate. Export recovery phrase QR will fail.
### Bug Description

If you signup a new user with SMS / LN, opt into "Create keys in browser" and then try to export the secret to Pubky Ring via QR, Pubky Ring will throw an error because the user IS NOT YET signed up to any homeserver.

Maybe the easiest fix: just remove the "Export Recovery phrase" via QR option on the invitationless flow. If the user has just opted out of Pubky Ring for user signup, we can safely assume he doesn't want to use RIng just on the next step. Too convoluted.

### Steps to Reproduce

1. Create a new user via SMS or LN.
2. Click on "Create Keys in browser" instead of "Continue with Pubky Ring"
Image
3. Click "Export Recovery phrase"
4. Scan with Ring and click "Save"
5. You see an error because the user does not have yet a valid Homeserver user and Ring cannot signup the user without invitation code.

Image

### Affected Area

Authentication / Login

None

### Frequency

Always (100%)

### Expected Behavior

Either or both:
1. Do not offer QR export right here (not really needed)
2. Ring should gracefully handle an import of a user secret that does not have a homeserver yet @coreyphillips 

### Screenshots / Recording

_No response_

### Console Log Output

```shell

```

### Browser & Version

_No response_

### Device

_No response_

### Additional Context

_No response_

tipogi:
fix #848


MiguelMedeiros (approved):
✅ **APPROVED**

This PR fixes issue #848 by removing the "Export recovery phrase" QR export option from the onboarding backup flow. The bug occurs when users who opted for "Create keys in browser" try to export their recovery phrase to Pubky Ring - the export fails because the user isn't yet signed up to a homeserver.

Issue #762, opened 2026-01-12
[Bug]: Clicking on username or avatar notification should navigate to profile


catch-21:
### Bug Description

Clicking on the profile name or avatar in a notification navigates to the post associated with the notification rather than the profile.

### Steps to Reproduce

1. Go to own profile page
2. Find a notification relating to a post (reply, tag)
3. Hover mouse over their profile name
4. Observe username does not underline (bad)
5. Click
6. Observe navigation goes to associated post instead of their profile page (bad)

### Affected Area

Profile / User

### Suspected Layer (if known)

None

### Frequency

Always (100%)

### Expected Behavior

Clicking in the notification list item area, it should be possible to navigate to their profile by clicking on their username or avatar.

### Console Log Output

```shell

```

PR #944, opened 2026-01-21
fix: #762 - Clicking on username or avatar notification should navigate to profile

afterburn:


MiguelMedeiros (approved):
✅ **APPROVED**

Splits notification text into separate clickable links for username and action text. Avatar and username now navigate to user profile, while action text and timestamp/icon navigate to the notification target (post or profile depending on type).

Issue #538, opened 2025-12-08
[Bug]: Unable to sign out if own profile pubky cannot be resolved


catch-21:
### Describe the bug

If franky cannot resolve your own pubky then it is not possible to proceed to sign out. Perhaps we should clear local state and sign out even if homeserver cannot be contacted.

This is probably unlikely to happen IRL, although still possible if the client is unable to contact homeserver to terminate session.

### Reproduce

Although unrealistic, the easiest way to reproduce this is to reconfigure franky from staging to testnet whilst your browser session is still signed in:

1. Be signed in on staging environment
2. Stop franky server
3. Change .env to local env
4. Re-build and start franky server
5. Navigate to profile page
6. Click Sign Out
7. Observe franky does not proceed to `/logout`

### Screenshots / Recording

https://github.com/user-attachments/assets/fd7e02f8-c775-4388-a694-c42e118a2ed8

### Pubky App Version

8c58e2f86b962edca703e43043b32132df59d513

### Console log output

```shell

```

PR #950, opened 2026-01-21
fix: #538 - Ensure signout succeeds when profile resolve fails

afterburn:
Users couldn't sign out if their profile pubky couldn't be resolved. This happened because `PubkySpecsSingleton.reset()` was called after `HomeserverService.logout()` in `AuthApplication.logout()`. When the homeserver logout failed (i.e. due to pubky resolution failure), the reset was never called.

In this PR I moved `PubkySpecsSingleton.reset()` from `AuthApplication.logout()` to `AuthController.logout()` where it runs unconditionally after the try/catch block. This ensures the singleton is always reset regardless of whether the homeserver logout succeeds or fails.


MiguelMedeiros (approved):
## ✅ APPROVED

Fixes issue #538 where users couldn't sign out if their profile pubky couldn't be resolved. The fix moves `PubkySpecsSingleton.reset()` from `AuthApplication.logout()` to `AuthController.logout()`, placing it after the try/catch block to ensure it runs unconditionally regardless of homeserver logout success/failure.

Issue #265, opened 2025-10-08
[Bug]: Detect mobile device with display size instead of touch device for dynamic Share button


catch-21:
### Describe the bug

https://github.com/pubky/franky/pull/241 has addressed the original problem in https://github.com/pubky/franky/issues/182 by detecting for touch device rather than display size. To align with the rest of the app where Tailwind is used to handle responsive UI, such as [here](https://github.com/pubky/franky/blob/7366d73d2b5e6c520013e0466c5dbfa30fc53193/src/components/molecules/Scan/Scan.tsx#L56-L106) in Scan.tsx, we should take a similar approach here to avoid false-positives such as desktops with touch capability.

### Reproduce

1. Onboard as new user
2. Observe share button is displayed for non-touch device only, even if that is desktop screen size (use dev tools)

### Pubky App Version

7366d73d2b5e6c520013e0466c5dbfa30fc53193

### Console log output

```shell

```

PR #955, opened 2026-01-21
fix: #265 - Share button visibility on mobile

afterburn:


MiguelMedeiros (approved):
✅ **APPROVED**

This PR fixes issue #265 by changing the share button visibility logic from JavaScript-based API detection to CSS-based responsive design. Previously, the button was conditionally rendered based on `isWebShareSupported()`. Now it's always rendered with `md:hidden` class, and when Web Share API is unavailable, it gracefully falls back to clipboard copy.

PR #959, opened 2026-01-22
fix: memoize PostText to prevent re-renders on TTL refresh

tipogi:
### Problem

Every time the TTL (Time-To-Live) coordinator refreshed stale posts from IndexedDB, **all `PostText` components would re-render**, even when the actual `content` hadn't changed.

**Root cause chain:**

1. `TtlCoordinator` detects stale posts and calls `forceRefreshPostsByIds`
2. Fresh data is fetched from Nexus and persisted to IndexedDB via `LocalStreamPostsService.persistPosts`
3. `useLiveQuery` in `usePostDetails` hook detects the database change
4. A **new `postDetails` object reference** is returned (even if content is identical)
5. Parent components re-render, passing the same `content` string to `PostText`
6. `PostText` re-renders and re-parses the Markdown unnecessarily

### Solution

Wrapped `PostText` with `React.memo`. Since all props are primitive types (strings/booleans), React's default shallow comparison prevents re-renders when props haven't actually changed.

| Scenario               | Before                              | After                                    |
|------------------------|-------------------------------------|------------------------------------------|
| Parent re-renders      | `PostText` always re-renders        | `memo` compares props, skips if same     |
| TTL refresh occurs     | Markdown parsed every time          | Markdown only parsed when content changes |

### DIFF
--- a/src/components/organisms/PostContentBase/PostContentBase.test.tsx
+++ b/src/components/organisms/PostContentBase/PostContentBase.test.tsx
@@ -203,7 +203,10 @@ describe('PostContentBase - Snapshots', () => {
     vi.clearAllMocks();
     const actualMolecules = await vi.importActual<typeof import('@/molecules')>('@/molecules');
     // Replace the mock implementations with real ones for snapshots
-    vi.mocked(Molecules.PostText).mockImplementation(actualMolecules.PostText);
+    // PostText is wrapped with React.memo(), so we need to access the underlying function via .type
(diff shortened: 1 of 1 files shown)


MiguelMedeiros (approved):
## ✅ APPROVED

This PR wraps the `PostText` component with `React.memo` to prevent unnecessary re-renders when the TTL coordinator refreshes stale posts from IndexedDB. Since all props are primitives (strings/booleans), React's default shallow comparison effectively prevents re-renders when content hasn't actually changed.

PR #1073, opened 2026-01-27
fix: instant avatar preview after profile upload

tipogi:
## Summary

Fixes avatar upload race condition where the old avatar gets permanently cached.

### Problem

When users upload a new avatar, there's a race condition between the client and the indexer:

1. User uploads new avatar
2. Client immediately requests the avatar with an updated timestamp
3. **But the indexer hasn't processed the new image yet**
4. The old avatar is returned and gets cached with the new timestamp
5. This stale image is now **persisted forever** because the cache key never changes

### Solution

Introduce a local-first avatar display pattern using an in-memory Zustand store for blob URLs:
```
User selects file
      ↓
URL.createObjectURL(file) → "blob:https://app.com/3f2c..."
      ↓
Store in Zustand: { localAvatarUrl: "blob:..." }
      ↓
All avatar components read from store
      ↓
avatarSrc = localAvatarUrl ?? cdnAvatarUrl
      ↓
<img src={avatarSrc} /> → Shows instantly!
```

### Key Improvements

- **Fixes the race condition** - Local blob URL is used until user refresh the page
- **Instant feedback** - Avatar changes appear immediately across all components
- **No extra network requests** - Uses browser's in-memory blob storage
- **Scoped subscriptions** - Only the current user's avatar subscribes to local store changes
- **Memory safe** - Blob URLs are revoked on logout and when replaced
- **Extensible** - Store architecture supports post `attachments` in the future

### DIFF
--- a/src/components/organisms/AvatarWithFallback/AvatarWithFallback.utils.ts
+++ b/src/components/organisms/AvatarWithFallback/AvatarWithFallback.utils.ts
@@ -33,8 +33,6 @@ export function extractUserIdFromAvatarUrl(avatarUrl: string | undefined | null)
   const userIdWithParams = avatarUrl.slice(expectedPrefix.length);
   const userId = userIdWithParams.split('?')[0]; // Remove query params like ?v=12345
 
-  console.log('userId', userId);
-
   if (!USER_ID_PATTERN.test(userId)) return null;
 
   return userId;


secondl1ght:
If this one gets approved and merged it looks like @tipogi set it up with https://github.com/pubky/franky/issues/656 in mind, so we can use the same solution on that one. 👍🏼

MiguelMedeiros:
@greptile

PR #1104, opened 2026-01-29
feat: add API-based tag suggestions to all tag input fields

afterburn:
Adds a tag suggestions dropdown that fetches recommendations from the Nexus API when users type 3+ characters in any tag input field. The feature is enabled across all tag input contexts including post creation, replies, reposts, articles, profile tagged pages, and post detail panels.

Clicking a suggestion immediately adds the tag without requiring an additional Enter press. Already-added tags are automatically excluded from the suggestions list.


package-lock.json
MiguelMedeiros: remove this peer props from package-lock.json

PR #1125, opened 2026-01-30
tests: bump pubky-stack for E2E tests

catch-21:


MiguelMedeiros:
@greptile

MiguelMedeiros:
Hey @catch-21! 👋

I noticed the Docker image tags in `e2e-test.yml` were updated to use the new format:
- `local/pubky-nexus:latest` (line 109)
- `local/homeserver:latest` (line 118)

However, `cache-docker-images.yml` still uses the old format:
- `nexusd:latest` (line 39)
- `homeserver:latest` (line 48)

This means the cache workflow won't produce images usable by the e2e workflow since the tags don't match.

Looking at the `pubky/pubky-docker` docker-compose.yml, I see the image naming pattern uses `${REGISTRY}/image-name:tag`, which explains why we need the `local/` prefix when `REGISTRY=local` is set.

**Was this intentional, or should `cache-docker-images.yml` be updated to match?**

If it needs updating, here are the suggested changes:

```yaml
# Line 39: Change from
tags: nexusd:latest
# To
tags: local/pubky-nexus:latest

# Line 48: Change from
tags: homeserver:latest
# To
tags: local/homeserver:latest
```

Let me know! 🙏

MiguelMedeiros (approved):

PR #1143, opened 2026-01-31
feat(tags): load all taggers when expanding tag list

### LINKED ISSUE #1067: [Bug]: tag avatar list not loading all users
### Bug Description

Currently the expanded tag list with Avatars row has a maximum of 5 "taggers" even when the tag count is visually higher.

Same applies to when we expand the taggers by clicking the Avatar row, it's only visible up to 5 taggers

### Steps to Reproduce

1. go to Miguel [post](http://localhost:3000/post/owoickyrwfakar33zdpmwonswhkysxbmi7xntdqwrmt8mr4f64xo/00344AAPBNDFG)
2. Look at the right tag list, specifically `homegate`
3. it's visible 16 on the tag count but only 4 avatars +1
4. click the avatar row, only 5 taggers are visible

### Affected Area

Tags

Data not loading (network/cache)

### Frequency

Always (100%)

### Expected Behavior

the number of avatars and taggers should match the tag count number.

Figma design:
@https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=25556-122264&m=dev

### Screenshots / Recording

Image

### Console Log Output

```shell

```

### Browser & Version

_No response_

### Device

_No response_

### Additional Context

_No response_

Orlandohub:
### Summary

When expanding a tag's tagger list, only a limited subset of taggers was displayed. This PR adds the ability to fetch and display all taggers for a tag when the user expands the list.

### Changes

- New usePostTaggers hook - Fetches paginated taggers on demand when a tag is expanded
- New fetchTaggers API endpoint - Added to PostController and NexusPostService to support tagger pagination per tag label
- Updated WhoTaggedExpandedList - Now receives tagger IDs and uses useBulkUserAvatars for user details with fallback support
- Updated TaggedList - Orchestrates tagger fetching when a tag is expanded, uses ref pattern to prevent unnecessary re-fetches
- Added widthMode prop to PostTagsPanel - Allows fit or full width behavior for different layouts

https://github.com/user-attachments/assets/efd55a9b-eca4-4a28-88f6-0f8c5828733e

### Implementation Notes

- Taggers are fetched lazily only when user expands a specific tag
- Initial taggers from the tag response are used immediately while additional taggers load
- Loading state is shown while fetching remaining taggers
- Results are cached per tag label to avoid re-fetching

### Test Plan

- [x] Navigate to a post with tags
- [x]  Click on a tag's avatar group to expand the tagger list
- [x] Verify all taggers load (not just the initial subset)
- [x] Verify loading indicator appears while fetching
- [x] Collapse and re-expand - should not re-fetch
- [x] Expand a different tag - should fetch that tag's taggers
 
 Closes #1067


MiguelMedeiros:
@greptile

src/components/molecules/WhoTaggedExpandedList/WhoTaggedExpandedList.tsx
```diff
@@ -1,11 +1,13 @@
 'use client';
 
+import { useMemo } from 'react';
```
infin1t3: Check do you need this `useMemo`

infin1t3 (dismissed):
Approved with a minor comment:

src/hooks/usePostTaggers/usePostTaggers.ts - you can remove useMemo/useCallback here since we use react-compiler now. Not a blocker for merge.

Would also like for this PR to be verified by @MiguelMedeiros before merging just to make sure I didn't miss anything

Issue #29, opened 2025-08-11
feat: custom favicon dark/light


flaviomoceri:
Add this favicon:
![Image](https://github.com/user-attachments/assets/36498d73-b0f7-43a4-aae0-86e56c4d17cf)

PR #1146, opened 2026-01-31
fix: emoji truncation in post link previews

secondl1ght:
### Summary

This PR fixes an issue where emojis in post previews could be broken/corrupted when truncating the metadata description for social sharing and SEO purposes. The previous implementation used simple string slicing (`slice(0, 100)`), which can split multi-byte Unicode characters (like emojis) in the middle, resulting in garbled text.

### Changes

- **Replaced simple string slicing with `Intl.Segmenter`**: Now uses grapheme cluster segmentation to ensure emojis and other multi-codepoint characters are kept intact when truncating
- **Increased truncation limit from 100 to 200 graphemes**: Provides more context in post previews while still keeping descriptions reasonable for meta tags
- **Added reusable `Intl.Segmenter` instance**: Created at module scope to avoid instantiation overhead on every request, with documentation explaining locale choice

### Technical Details

The `Intl.Segmenter` API with `{ granularity: 'grapheme' }` follows Unicode Standard Annex #29 rules, which are language-agnostic. This ensures that:
- Emojis (including compound emojis with ZWJ sequences like 👨‍👩‍👧‍👦) stay intact
- Accented characters and other combining sequences are preserved
- The truncation is visually consistent across all Unicode text

### Testing

No test files were modified in this PR.

---


secondl1ght:
@greptile just testing if you detect this.

infin1t3 (approved):

PR #1160, opened 2026-02-02
feat: add global snapshot serializer for Radix UI ID normalization

### LINKED ISSUE #1101: [Feature]: Radix ID normalization on snapshot tests
### Problem Statement

Right now there's an issue which makes snapshot tests inconsistent which is introduced by Radix non normalized ids.

**Example**:

- A new PR is open, with new snapshot tests where the tested component uses something like `const headerId = React.useId();`
- When we run the update snapshots it will generate the new snapshots for this test and then correctly pass
- This will generate a new ID, on every new snapshot update command and therefore break the tests
- The bad part is, this only happens after the new code being merged with master and therefore breaking on other devs new branches

We've been fighting this by applying the following workaround:

```
const normalisedContainer = normaliseRadixIds(container);
expect(normalisedContainer.firstChild).toMatchSnapshot(); 
```

Which works, but only after the fact, when snapshots break. So now it means someone else needs to apply `normaliseRadixIds` utility to the new component tests.

### Proposed Solution

Implement a Global Snapshot Serializer.
This way we would find the radix generated IDs via a REGEX like:
`const RADIX_ID_REGEX = /\b(radix-)?_r_[a-z0-9]+_?\b/gi;` and then replace it with our normalized value.

Finally, delete all `normaliseRadixIds` instances.

### Feature Area

New Feature Area

### Feature Type

Enhancement (improve existing feature)

### Acceptance Criteria

- [ ] Every time snapshot updates happen, the tests should not break due to Radix IDs inconsistency

### Design / Mockups

_No response_

### Alternatives Considered

_No response_

### Suggested Priority

### Additional Context

_No response_

Orlandohub:
Implement automatic Radix ID normalization via a global Vitest snapshot serializer, eliminating the need for manual `normaliseRadixIds()` calls in individual test files.

Changes:
- Add snapshot serializer to `src/config/test.ts` that replaces dynamic Radix IDs (e.g., radix-_r_abc123) with 'radix-normalized'
- Remove `normaliseRadixIds` utility function from src/libs/utils/utils.ts
- Remove associated tests from `src/libs/utils/utils.test.ts`
- Update 24 test files to remove `normaliseRadixIds` imports and usage
- Update `.cursor/rules/component-testing.mdc` documentation

This ensures snapshot tests are consistent across test runs regardless of test execution order, without requiring developers to remember to wrap containers in a normalization function.

Closes #1101


.cursor/rules/component-testing.mdc
```diff
@@ -310,69 +310,30 @@ This document outlines the essential rules and patterns for creating unit tests
...
+
+   This means you can write snapshot tests normally without worrying about Radix ID instability:
+
    ```typescript
    describe('MyComponent - Snapshots', () => {
      it('matches snapshot for default state', () => {
        const { container } = render(<MyRadixComponent />);
-       // ❌ This WILL fail randomly due to dynamic Radix IDs
-       expect(container).toMatchSnapshot();
+       expect(container.firstChild).toMatchSnapshot();
```
infin1t3: Why do we sometimes use `expect(container.firstChild).toMatchSnapshot()` and sometimes `expect(container).toMatchSnapshot()` - I can see both in components.

src/libs/utils/utils.test.ts
```diff
@@ -557,125 +556,6 @@ describe('Utils', () => {
     });
   });
 
-  describe('normaliseRadixIds', () => {
```
infin1t3: Nitpick: You might wanna add test for the serializer itself.

infin1t3 (changes requested):
Would like to for @catch-21 to review this as well, in case I missed anything.

src/libs/utils/utils.test.ts
catch-21: ➕

PR #1205, opened 2026-02-03
fix(ui): notification action text color

MiguelMedeiros:
Summary
- ensure notification action links and spans use the default foreground color so they remain readable
- add the autogenerated Franky environment manifest that defines recommended npm commands

Testing
- Not run (not requested)


MiguelMedeiros:
@greptile

PR #1275, opened 2026-02-10
tests: feed e2e

catch-21:
Several new E2E tests covering feed filtering and sorting.


catch-21:
https://github.com/pubky/pubky-app/actions/runs/21903060580

PR #1283, opened 2026-02-12
fix(ui): fix modals and keyboard in mobile - 1189 & 1194

### LINKED ISSUE #1189: [UI]: Mobile: iOS Safari shows up/down arrows above keyboard
### Description

With Safari on iOS mobile. When inputing text into New Post dialog, there's a weird bar with up/down arrows placed above the on screen keyboard.

### Expected Behaviour

We don't need this bar. I think it is for navigating forms. Also, it may not be possible to remove, see https://discussions.apple.com/thread/256174563?answerId=261654908022&sortBy=rank#261654908022

### Steps to Reproduce

1. On iOS Safari
2. Tap 'New Post' button in bottom left
3. With New Post text field in focus, observe button bar above soft keyboard with an up down arrow and checkmark/done button

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

### LINKED ISSUE #1194: [UI]: Mobile: Footer navigation bar hides behind keyboard
### Description

On mobile, the footer navigation bar (with buttons 'home', 'search', etc.)  remains pinned to the bottom of the screen hidden when the soft keyboard shows.

Need to confirm whether this is just where there is space, e.g. typing into the search bar of search page, or also when (New Post, Reply etc.) modals are displayed where we may need all the space we have, see [Figma](https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq?node-id=28420-82819#1616876978).

### Expected Behaviour

Footer navigation bar should be sticky on top of the soft keyboard when keyboard is active.

### Steps to Reproduce

1. Tap some input field to show the soft keyboard
2. Observe the footer navigation bar is hidden behind

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

Nirzhuk:
- implement keyboard offset handling in MobileFooter and DialogNewPost components
- Added `useKeyboardOffset` hook to calculate the offset needed to keep UI elements above the keyboard.
- Updated `MobileFooter` to apply a transform based on keyboard visibility and offset.
- Enhanced `DialogNewPost` to adjust its position when the keyboard is visible, including an offset adjustment option.
- Added tests for keyboard visibility and offset behavior in both components. 

Fix #1189
Fix #1194


MiguelMedeiros:
Hey @Nirzhuk can you fix conflicts?

PR #1337, opened 2026-02-17
bug: add missing backup window in onboarding

tipogi:
## Summary

After signing up, users were stuck in a redirect loop because the **backup page** was not included in the allowed routes for authenticated users who haven't created a profile yet. The route guard would immediately redirect them away from `/onboarding/backup` back to `/onboarding/profile`, skipping the backup step entirely.

### Changes

- **Fix redirect loop**: Added `ONBOARDING_ROUTES.BACKUP` to `NEEDS_PROFILE_CREATION_ROUTES.allowedRoutes` so users can access the backup page after sign-up
- **Fix continue button icon**: Changed the continue button icon from `LogIn` to `ArrowRight` for visual consistency with the back button (`ArrowLeft`)
- **Fix ProfileNavigation layout**: Added conditional styling to right-align the continue button when the back button is hidden

### Snapshot test fix

The `LogIn` → `ArrowRight` icon change caused **9 snapshot test failures** in `ButtonsNavigation.test.tsx`. The root cause was that the test file was mocking `@/components/ui` with plain HTML stubs instead of using real `@/atoms` components, so the old snapshots never captured actual icon SVGs. The fix:

- Removed unnecessary mocks (`@/libs`, `@/components/ui`) — tests now use real component and icon implementations per testing guidelines
- Added missing test coverage for `hiddenBackButton`, `hiddenContinueButton`, `loadingContinueButton`, and `id` props
- Regenerated snapshots with real Shadcn components and lucide-react SVG icons, which will properly catch icon or component regressions going forward

### DIFF
--- a/src/components/molecules/ButtonsNavigation/ButtonsNavigation.test.tsx
+++ b/src/components/molecules/ButtonsNavigation/ButtonsNavigation.test.tsx
@@ -2,38 +2,6 @@ import { describe, it, expect, vi } from 'vitest';
 import { render, screen, fireEvent } from '@testing-library/react';
 import { ButtonsNavigation } from './ButtonsNavigation';
 
-// Mock libs - use actual utility functions and icons from lucide-react
-vi.mock('@/libs', async () => {
-  const actual = await vi.importActual('@/libs');
-  return { ...actual };
-});
-
-// Mock UI components
-vi.mock('@/components/ui', () => ({
-  Button: ({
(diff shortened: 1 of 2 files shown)


catch-21:
E2E tests run: https://github.com/pubky/pubky-app/actions/runs/22094005011/job/63846668312

MiguelMedeiros (approved):

PR #1349, opened 2026-02-18
fix(notification): decouple polling cursor from read/unread boundary

tipogi:
## Summary

Introduces a dedicated `lastPolledTimestamp` polling cursor, fully decoupled from the `lastRead` read/unread boundary, to fix a bug where marking notifications as read could cause subsequent polls to silently skip unfetched notifications. Also refactors `persistAndGetUnreadCount` into separate `bulkSave` and `countUnreadSince` operations so the unread count is always derived from the complete IndexedDB state.

## Problem

`lastRead` was serving a dual purpose: it was used as the Nexus API polling cursor (`end` parameter) **and** as the read/unread boundary. When a user marked notifications as read, `lastRead` advanced forward, which also moved the polling cursor. Any notifications that arrived between the old and new `lastRead` values were then skipped on the next poll — resulting in silently missed notifications.

Additionally, the unread count was calculated only from the current fetch batch via a combined `persistAndGetUnreadCount` call, rather than from the full set of persisted notifications. This meant the count could drift out of sync with the actual stored data.

## Solution

Introduce a new `lastPolledTimestamp` field that is solely responsible for tracking the Nexus polling position:

- **`lastPolledTimestamp`** is used as the `end` parameter for Nexus queries. It advances to `newestTimestamp + 1` after each successful fetch to avoid re-fetching the same notification.
- **`lastRead`** is used exclusively as the read/unread boundary. It is passed to a new `countUnreadSince()` method that queries IndexedDB for all notifications with a timestamp strictly greater than `lastRead`.

The persistence logic was also split: `bulkSave` handles writing to IndexedDB, and `countUnreadSince` performs a dedicated count query (`countNewerThan`) against the full table. This ensures the unread count is always accurate regardless of fetch batch size.

This separation guarantees that marking notifications as read never causes the polling cursor to skip over unfetched notifications.

### DIFF
--- a/src/core/application/bootstrap/bootstrap.ts
+++ b/src/core/application/bootstrap/bootstrap.ts
@@ -155,10 +155,11 @@ export class BootstrapApplication {
       notifications: notificationList,
       viewerId: pubky,
     });
-    await Core.LocalNotificationService.bulkSave({ flatNotifications });
-    const unread = await Core.LocalNotificationService.countUnreadSince(userLastRead);
-    // We add 1 to the newest timestamp to avoid fetching the same notification again
-    const lastPolledTimestamp = notificationList.length > 0 ? notificationList[0].timestamp + 1 : undefined;
-    return { unread, lastRead: userLastRead, lastPolledTimestamp };
+    const { unread, nextPollCursor } = await Core.NotificationApplication.persistAndSummarize({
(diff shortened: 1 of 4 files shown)


tipogi:
@greptile second round, go!

PR #1351, opened 2026-02-18
feat(og-metadata): add handling for upstream 403 responses with fallback metadata

### LINKED ISSUE #1232: [Bug]: Some link previews are not showing due to 403
### Description

Creating a post with links such as https://ra.co/events/2363466 do not show a preview, the response from og-metadata is 403. Whilst in the same instance of the app other links such as https://bsky.app/profile/washingtonstatestandard.com/post/3me5lkysxak2a can be used.

### Expected Behaviour

The https://ra.co link is working on whatsapp so there must be a preview pubky-app can use.

### Steps to Reproduce

1. Create a new post using url https://ra.co/events/2363466
2. Observer no link preview is shown (and dev tools show 403 for the network response)

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

Nirzhuk:
- Implemented a new function to build fallback metadata when upstream requests return a 403 status.
- Updated the GET request handler to return fallback metadata instead of an error response in case of a 403.
- Added a test case to verify the fallback behavior when a 403 status is encountered.
https://github.com/pubky/pubky-app/issues/1232
fix #1232

### DIFF
--- a/src/components/molecules/HotTagCard/HotTagCard.test.tsx.snap
+++ b/src/components/molecules/HotTagCard/HotTagCard.test.tsx.snap
@@ -7,11 +7,11 @@ exports[`HotTagCard - Snapshots > matches snapshot with default props 1`] = `
   style="background: linear-gradient(90deg, rgba(5, 5, 10, 0.7) 0%, rgba(5, 5, 10, 0.7) 100%), rgb(255, 153, 0);"
 >
   <div
-    class="flex flex-col gap-2.5 px-6"
+    class="flex w-full flex-col gap-2.5 px-6"
     data-testid="container"
   >
     <div
-      class="flex items-center gap-3"
+      class="flex w-full items-center gap-3"
       data-testid="container"
     >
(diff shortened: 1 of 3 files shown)


src/app/api/og-metadata/route.test.ts
```diff
@@ -1050,4 +1050,28 @@ describe('API Route: /api/og-metadata', () => {
...
+        headers: new Headers({ 'content-type': 'text/html' }),
+      });
+
+      const request = createRequest('https://ra.co/events/2363466');
+      const response = await GET(request);
+      const data = await response.json();
+
+      expect(response.status).toBe(200);
+      expect(data).toEqual({
+        url: 'https://ra.co/events/2363466',
```
infin1t3: This string (31 chars) is under the 40-char truncation limit, so the assertion passes. Consider adding a test with a longer URL to verify truncation behavior in the fallback path.

PR #1379, opened 2026-02-24
bug: migrate custom feeds from auto-increment to string IDs

tipogi:
- **ID system migration** — Feed IDs are now HashId-derived strings instead of Dexie auto-increment integers. This means the feed ID is deterministic and derived from the feed's configuration, so changing config fields (tags, reach, sort, content, layout) produces a new ID while renaming keeps the same one.
- **Application layer rewrite** — `FeedApplication.persist()` now uses the `meta.id` from the normalized `FeedResult` as the primary key. When an update changes config fields and the HashId changes, it handles the delete-old + create-new cycle automatically. Removed `String()` wraps that were bridging numeric IDs to the URI builder.
- **Service layer simplification** — `LocalFeedService.createOrUpdate()` no longer branches on `id === 0` for auto-increment. It's now a plain upsert since the caller always provides the string ID upfront.
- **Controller update flow** — `commitUpdate` now normalizes the merged params to derive a new HashId. If the ID differs from the existing feed, the persist layer handles the swap transparently. Added `existingId` tracking through the update path.
- Add the ability to fetch and persist all custom feeds from the homeserver. On bootstrap, the app lists all feed URIs under the user's homeserver `/feeds/` directory, fetches each payload, and validates it through `PubkySpecsBuilder.createFeed` to produce a normalized `FeedModelSchema`. Invalid or malformed feeds are silently skipped with a warning log, ensuring a single bad entry doesn't block the rest of the feed list from loading.
- 🚨 **Database version bump required** — The feed table schema changed from `++id` (auto-increment) to `&id, created_at` (string primary key + indexed timestamp). This is a breaking schema change that requires a Dexie database version bump and a migration to convert existing numeric feed IDs to their HashId-derived string equivalents. Without the version bump, existing users will hit a schema mismatch error on app load. 
- Increment in one the `NEXT_PUBLIC_DB_VERSION` so existing clients trigger DB version mismatch/recreation and pick up the new schema. CC @catch-21

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -33,6 +33,11 @@ on:
         required: false
         type: string
 
+      commit_sha:
+        description: 'Commit SHA to checkout'
+        required: false
+        type: string
+
   workflow_call:
     outputs:
       image_tag:
@@ -62,6 +67,11 @@ on:
         required: false
         type: string
 
+      commit_sha:
+        description: 'Commit SHA to checkout'
+        required: false
+        type: string
+
 env:
   registry: 'europe-west6-docker.pkg.dev/infra-464608/synonym-private-repo'
 
@@ -76,6 +86,8 @@ jobs:
     steps:
       - name: Checkout this repository
         uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.commit_sha || github.sha }}
 
       - name: Set HEAD commit hash
         id: sha_tag
@@ -111,6 +123,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.commit_sha || github.sha }}
 
       - name: Login to registry
         uses: pubky/ci-workflows/.github/actions/docker/registry_login@main
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -51,9 +51,14 @@ jobs:
           else
             PR_NUMBER="${{ github.event.pull_request.number }}"
           fi
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
+            COMMIT_SHA="${{ github.event.pull_request.head.sha }}"
+          else
+            COMMIT_SHA="${{ github.sha }}"
+          fi
           echo "pr_number=${PR_NUMBER}" >> $GITHUB_OUTPUT
           echo "service_name=${{ env.SERVICE_BASE_NAME }}-pr-${PR_NUMBER}" >> $GITHUB_OUTPUT
-          echo "short_sha=$(echo ${{ github.sha }} | cut -c1-7)" >> $GITHUB_OUTPUT
+          echo "short_sha=$(echo ${COMMIT_SHA} | cut -c1-7)" >> $GITHUB_OUTPUT
           echo "platforms=linux/amd64" >> $GITHUB_OUTPUT
 
   build-preview:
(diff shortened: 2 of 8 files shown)


src/core/application/feed/feed.ts
```diff
@@ -63,6 +118,95 @@ export class FeedApplication {
...
+      return {
+        id: feedMeta.id,
+        name: feed.name,
+        tags: tags ?? [],
+        reach,
+        sort,
+        content: content ?? null,
+        layout,
+        created_at: Number(feed.created_at),
+        updated_at: Number(feed.created_at),
```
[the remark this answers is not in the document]
secondl1ght: @tipogi is this something you want to change?

src/core/application/feed/feed.types.ts
```diff
@@ -45,6 +45,23 @@ export type TFeedPersistInput = {
...
+}
+
+export interface HomeserverFeedJson {
+  name: string;
+  feed: {
+    tags?: string[];
+    reach: string;
+    layout: string;
+    sort: string;
+    content: string | null;
```
secondl1ght: Shouldn't these be numbers?  In the pubky app specs node module the enum values are numbers:

```ts
/**
 * Enum representing the layout of the feed.
 */
export enum PubkyAppFeedLayout {
  Columns = 0,
  Wide = 1,
  Visual = 2,
}
/**
 * Enum representing the reach of the feed.
 */
export enum PubkyAppFeedReach {
  Following = 0,
  Followers = 1,
  Friends = 2,
  All = 3,
}
/**
 * Enum representing the sort order of the feed.
 */
export enum PubkyAppFeedSort {
  Recent = 0,
  Popularity = 1,
}
/**
 * Represents the type of pubky-app posted data
 * Used primarily to best display the content in UI
 */
export enum PubkyAppPostKind {
  Short = 0,
  Long = 1,
  Image = 2,
  Video = 3,
  Link = 4,
  File = 5,
}
```

package-lock.json
```diff
@@ -224,7 +224,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
```
secondl1ght: This file can be removed from the PR because it is unrelated.

tipogi: 1816adc

src/core/application/feed/feed.types.ts
tipogi: `createFeed` requests strings so we keep as it is from now.

secondl1ght (approved):
Let's GO!!! 🚀

src/core/models/feed/feed.schema.ts
```diff
@@ -13,7 +13,6 @@ export interface FeedModelSchema {
 }
 
 // Schema design rationale:
-// - ++id: Auto-incrementing primary key (like notifications)
-// - created_at: For sorting feeds by creation time
-// - updated_at: For tracking when feed was last modified
-export const feedTableSchema = '++id, name, created_at, updated_at';
+// - &id: Primary key (string, HashId-derived)
```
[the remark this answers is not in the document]
tipogi: Lets see what happens in staging, but bumping the version, I think it clears all the db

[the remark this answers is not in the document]
secondl1ght: Yes but this PR does not bump the version, so where will it be bumped?

PR #1381, opened 2026-02-24
fix(auth): improve mobile deeplink flow and remove store fallback

### LINKED ISSUE #1380: [Bug]: Upon attempting to Sign In Pubky Ring directs me to Play Store as opposed to Pubky Ring on device
### Description

I am using Graphene (Pixel 6), I don't have Google Store services enabled, when I attempt to sign in via mobile Pubky Ring directs me to the Google Play Store as opposed to the pubky ring instance I already have running, which prevents me from signing up with Pubky Ring.

### Expected Behaviour

App should check if I have Pubky Ring installed and if I do, use that as opposed to directing me to Playstore.

### Steps to Reproduce

go to pubky.app
try and sign up (pubky ring should already be installed)

### Screenshots

Image
https://github.com/user-attachments/assets/7406a4ff-6707-4ee2-ac73-11c3cae642e4

### Console Log

```shell

```

### Browser & Version

Brave

infin1t3:
## Summary

- Introduces `useMobileAuth` hook to unify mobile authorization flow across iOS and Android
- Uses direct deeplink navigation (`window.location.href`)
- Shows "Opening Pubky Ring..." feedback while attempting app handoff

## Known limitations

This is a heuristic-based solution. Universal Links (iOS) / App Links (Android) would provide a more reliable long-term fix by letting the OS handle routing before the browser gets involved.

Closes #1380

### DIFF
--- a/cypress/e2e/notifications.cy.ts
+++ b/cypress/e2e/notifications.cy.ts
@@ -7,10 +7,15 @@ import {
   clickFollowButton,
   checkLatestNotification,
   addProfileTags,
-  waitForNotificationDotsToDisappear,
-  causeLastReadToBeUpdated,
+  causeNotificationsToBeRead,
 } from '../support/profile';
-import { BackupType, CheckForNewPosts, HasBackedUp, WaitForNewPosts } from '../support/types/enums';
+import {
+  BackupType,
+  CheckForNewPosts,
+  HasBackedUp,
+  LatestNotificationReadState,
+  WaitForNewPosts,
+} from '../support/types/enums';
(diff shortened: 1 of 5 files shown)


catch-21:
On android with brave I never see the Play Store redirect. There can be a 'Authorization was not completed' message but mostly works for me.

two success in a row

https://github.com/user-attachments/assets/c1601237-47a7-4eb6-874c-a7168dbed1d1

-----

with error

https://github.com/user-attachments/assets/735cf7f8-7b41-4e1c-8d20-1b9cae5b20a8

Issue #1136, opened 2026-01-30
[Bug]: Spontaneous sign out


catch-21:
### Bug Description

A few times now I've experienced seemingly _random_ sign out whilst navigating. One time, before sign out I was unable to tag or untag a post, so I knew something was up, probably the app did not think I have a session. I still have valid session cookie in browser and homeserver probably did not invalidate it.

Console log error: `[ERROR] Failed to restore session from persisted export`

The console error contains details "No HTTPS endpoints found in PKARR record" which comes from the sdk https://github.com/pubky/pubky-core/blob/0ec564d746236f46553ef56888a3aef9c1096f55/pubky-sdk/src/client/http_targets/wasm.rs#L105

Interestingly, I have cancelled requests that could have originated from the `restoreSession` function call. Could that have been caused by the successful 304 'Not Modified' response and result in an error back from sdk that triggered the sign out? Potential race condition?

<img width="898" height="476" alt="Image" src="https://github.com/user-attachments/assets/87659a35-0aed-4bfe-9dbb-5f9324b8ddd9" />

### Steps to Reproduce

Not sure how to reproduce. It had happened twice on https://pubky.app this week.

### Affected Area

Authentication / Login

### Suspected Layer (if known)

None

### Frequency

Always (100%)

### Expected Behavior

User only gets automatically signed out if they have genuinely lost their session (e.g. expired cookie or homeserver has invalidated it).

### Console Log Output

```shell
2411-f9dc6ae33defb43d.js:86 AppError: Pkarr operation failed: Pkarr record is malformed or missing required data: No HTTPS endpoints found in PKARR record for `_pubky.oojdj73roegeh7r5h59s91qpn8g54j9u8j3q8fdguwy9hb1gy9ao` (original URL: https://_pubky.oojdj73roegeh7r5h59s91qpn8g54j9u8j3q8fdguwy9hb1gy9ao/session)
    at N (2411-f9dc6ae33defb43d.js:86:225281)
    at Object.auth (2411-f9dc6ae33defb43d.js:86:225645)
    at 2411-f9dc6ae33defb43d.js:68:26003
    at aN (2411-f9dc6ae33defb43d.js:68:26144)
    at aT (2411-f9dc6ae33defb43d.js:68:26512)
    at aF.restoreSession (2411-f9dc6ae33defb43d.js:68:34711)
    at async 2411-f9dc6ae33defb43d.js:1:280657
[...]

Issue #1383, opened 2026-02-24
[Bug]: Loading the SignIn in different Tabs breaks the Signin flow


SeverinAlexB:
### Description

When you load the QR signin page in two tabs, it breaks the signin in the second tab.

What is happening in the background: Somehow, loading tab2 makes tab1 listen to the tab2 http relay link while keeping the old QR code

### Expected Behaviour

Tabs shouldn't influence each other.

### Steps to Reproduce

See video. Open two signup tabs at the same time.

### Screenshots

https://github.com/user-attachments/assets/8b9f77c2-3626-4f83-ae9d-3e8e64975417

### Console Log

```shell

```

PR #1384, opened 2026-02-25
fix: remove serwist defaultCache to prevent other origin services such as pkar, homeserver and http-relay from being cached by service worker

catch-21:
Hoping this will fix https://github.com/pubky/pubky-app/issues/1136 but hard to reproduce so still uncertain.

The theory is that the promises made by via the service worker for each of the requests made to each of the two pkarr relays hit the Serwit cache and concurrently succeed but also cancel each other resulting in total failure. The full report is internally: https://synonymworkspace.slack.com/archives/C031F2DC1JM/p1771924740369439?thread_ts=1771869915.915639&cid=C031F2DC1JM

### DIFF
--- a/src/sw.ts
+++ b/src/sw.ts
@@ -1,6 +1,5 @@
 /// <reference lib="webworker" />
 
-import { defaultCache } from '@serwist/next/worker';
 import type { PrecacheEntry, SerwistGlobalConfig } from 'serwist';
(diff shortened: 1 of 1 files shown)


catch-21:
This also fixes https://github.com/pubky/pubky-app/issues/1383

PR #1388, opened 2026-02-25
feat: Utilise Prelude's SDK for sending client browser signals for fraud prevention

### LINKED ISSUE #1230: [Feature]: Call Prelude at beginning of SMS verification process
### Problem Statement

We should be sending as much user environment data to Prelude's Sms verification as possible. This increases our confidence in unblocking countries with higher fraud rates.

### Proposed Solution

Prelude's SDK captures a load of info about the user's browser. It seems fairly straightforward to setup - [docs here](https://docs.prelude.so/introduction/frontend-sdks/web#frontend-web-sdk-integration).

Homegate's `/sms_verification/send_code`  takes an option `dispatch_id` param already.

Ping @86667 for an API key

### Suggested Priority

The sooner the better in order to open up SMS onboarding flow for users in ~80 countries.

86667:
Resolves https://github.com/pubky/pubky-app/issues/1230

Also added a few basic Sms verification unit tests.


src/core/services/homegate/homegate.ts
```diff
@@ -71,6 +73,27 @@ function parseLnVerificationStatus(json: TRawApiResponse): TLnVerificationStatus
...
+  private static async getPreludeDispatchId(): Promise<string | undefined> {
+    const sdkKey = Env.NEXT_PUBLIC_PRELUDE_SDK_KEY;
+    if (!sdkKey) {
+      return undefined;
+    }
+
+    try {
+      return await Promise.race([
+        dispatchSignals(sdkKey),
+        new Promise<undefined>((resolve) => setTimeout(() => resolve(undefined), Env.NEXT_PUBLIC_PRELUDE_SDK_TIMEOUT_MS)),
```
86667: Is there a preferred method for doing this?

.env.example
```diff
@@ -70,6 +70,12 @@ NEXT_PUBLIC_DEFAULT_HTTP_RELAY=https://httprelay.staging.pubky.app
 NEXT_PUBLIC_EXCHANGE_RATE_API=https://api1.blocktank.to/api/fx/rates/btc
 NEXT_PUBLIC_HOMEGATE_URL=https://homegate.staging.pubky.app
 
+# Prelude SDK key for sending client browser signals for SMS fraud prevention.
+# Optional. If not set we still use Prelude but without providing all available client signals.
+# NEXT_PUBLIC_PRELUDE_SDK_KEY=your_prelude_sdk_key_here
```
secondl1ght, line 77: Is this a private API key? If so then this env var should not be prefixed with `NEXT_PUBLIC` as that will expose the secret.

src/core/services/homegate/homegate.ts
```diff
@@ -101,14 +126,23 @@ export class HomegateService {
   /**
    * Sends a SMS code to the user. This only errors on network errors.
    * Any phone number is valid to avoid user enumeration.
+   * Automatically dispatches Prelude signals for fraud prevention if configured.
    * @param phoneNumber - The phone number to send the SMS code to.
    * @returns The result of the SMS code send request.
    */
   static async sendSmsCode(phoneNumber: string): Promise<TSendSmsCodeResult> {
     const url = homegateApi.sendSmsCode();
+    const body: Record<string, string> = { phoneNumber };
```
secondl1ght: Just a nit but this could be typed as `const body: { phoneNumber: string; dispatchId?: string } = { phoneNumber };`.

src/core/services/homegate/homegate.ts
```diff
@@ -71,6 +73,29 @@ function parseLnVerificationStatus(json: TRawApiResponse): TLnVerificationStatus
...
+
+    try {
+      return await Promise.race([
+        dispatchSignals(sdkKey),
+        new Promise<undefined>((resolve) =>
+          setTimeout(() => resolve(undefined), Env.NEXT_PUBLIC_PRELUDE_SDK_TIMEOUT_MS),
+        ),
+      ]);
+    } catch (error) {
+      console.warn('Prelude dispatchSignals failed:', error);
```
secondl1ght: We have a `Logger` class for handling errors that can be used.

src/libs/env/env.ts
```diff
@@ -323,6 +334,8 @@ function parseEnv(): z.infer<typeof envSchema> {
     NEXT_PUBLIC_MODERATED_TAGS: process.env.NEXT_PUBLIC_MODERATED_TAGS,
     NEXT_PUBLIC_EXCHANGE_RATE_API: process.env.NEXT_PUBLIC_EXCHANGE_RATE_API,
     NEXT_PUBLIC_HOMEGATE_URL: process.env.NEXT_PUBLIC_HOMEGATE_URL,
+    NEXT_PUBLIC_PRELUDE_SDK_KEY: process.env.NEXT_PUBLIC_PRELUDE_SDK_KEY,
```
secondl1ght, line 339: You'll also need to add this env var to the docker build via `.github/workflows/build-docker-image.yml` and `Dockerfile`.

secondl1ght (commented):
To set the new environment variables in production please coordinate with @catch-21 @SpontaneousOverthrow.

src/core/services/homegate/homegate.ts
secondl1ght: Probably would be best to clear the timeout, but it shouldn't affect anything in practice. If we used this pattern more I would say to extract it into a re-useable function, however this is only the second time it is used in the codebase.

secondl1ght: Also `dispatchSignals` will still complete even after the timeout, but I assume that is intentional?

86667: > Also dispatchSignals will still complete even after the timeout, but I assume that is intentional?

Yes, that is desired.

.env.example
86667: It is public. Updated some of the docs to make that clearer

secondl1ght (approved):

secondl1ght:
@86667 after approving I noticed this new package says that it supports an older node version than what we use.

```bash
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: '@prelude.so/js-sdk@0.4.0',
npm warn EBADENGINE   required: { node: '22.x' },
npm warn EBADENGINE   current: { node: 'v24.14.0', npm: '11.9.0' }
npm warn EBADENGINE }
```

Also it looks like the version in this PR is now several versions behind https://www.npmjs.com/package/@prelude.so/js-sdk. So maybe the node issue is already resolved. I am going to put in a new PR to bump the version now and assign you as the reviewer.

PR #1437, opened 2026-03-03
feat(posts): add side tags layout to PostMain for wide feed and replies

### LINKED ISSUE #1434: [Bug]: Feed post tags on wide layout not correctly placed
### Description

On wide layout mode the feed post tags and taggers list are being displayed under the post text.

### Expected Behaviour

Instead they should be displayed on the right side of the post, like we do on the individual post page.

Implement this design from Figma.
@https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=20324-52336&m=dev

### Steps to Reproduce

1. Login to pubky app
2. Click on wide, under left column Layout section
3. Confirm tags being rendered under post text and not on the right side

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

Orlandohub:
## Summary
- fix #1434
- Add a `tagsLayout` prop to `PostMain` enabling a two-column grid layout (tags expanded on the right) matching `SinglePostCard`, applied in the wide home feed and for all replies on the single post page
- Fix thread connector height issue caused by circular dependency between `useElementHeight` and `PostThreadConnector` by using absolute positioning
- Fixed tag input focus

## Changes
- Add `PostMain.types.ts` with `TagsLayout` type (`'inline' | 'side'`) and `PostMainProps` interface, re-exported from barrel
- Update `PostMain` to conditionally render a responsive two-column grid (`lg:grid-cols-3`) when `tagsLayout='side'`, with tags always visible on desktop right column and inline on mobile
- Thread `tagsLayout` prop through `TimelineFeed` → `TimelinePosts` → `PostMain` / `TimelinePostReplies` → `PostMain`, derived from the home store layout setting
- Pass `tagsLayout="side"` explicitly in `SinglePostContent` for all reply `PostMain` instances
- Fix thread connector to use absolute positioning (`absolute left-0 top-0 bottom-0`) to decouple it from flex layout height calculations, preventing inflated reply card heights

## Test plan
- [ ] Open home feed in **wide layout** — posts should display tags expanded in a right column on desktop, inline on mobile
- [ ] Switch home feed to **compact layout** — posts should use the original inline tag toggle behavior
- [ ] Navigate to a **single post page** — all replies should display the two-column side tags layout
- [ ] Click on a post in wide feed view — should navigate to the single post page
- [ ] Click on a reply in the single post page — should navigate to that reply's post page
- [ ] Verify thread connectors on replies render correctly without inflating card height

## Notes
- The `handleFooterClick` (`e.stopPropagation()`) is scoped only to the tags panel and actions bar containers, so clicks on post content/header still bubble up for navigation
- Custom feed layouts are also respected via `useCustomFeed` hook


Orlandohub:
@greptile-apps

PR #1494, opened 2026-03-12
feat: Using shards for running vitest

SpontaneousOverthrow:


.github/workflows/test.yml
```diff
@@ -11,10 +12,18 @@ on:
   # Enable manual trigger
   workflow_dispatch:
 
+concurrency:
+    group: ${{ github.workflow }}-${{ github.ref }}
```
catch-21: NIT:: over-indented, may not affect workflow parsing

.github/workflows/test.yml
```diff
@@ -24,18 +33,70 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: '20'
-          cache: 'npm'
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 'latest'
```
catch-21: pin to version (1.3.11) to prevent breaking update between merge queue runs

.github/workflows/test.yml
SpontaneousOverthrow: we want to cancel current workflows if the new commit appears.
for instance you made a pr, the workflows are triggered. Then you realise that you made a mistake and push a quick fix, without that setting you should wait for workflows to finish and only after the workflows for the new commit run. This this settings if there are running workflows, the new commit cancel them and runs the new one.
Also the group can't be omitted.

catch-21: I agree that we cancel in progress. I was just saying the yml is over-indented for 'group' and 'cancel-in-progress' 😃

catch-21 (dismissed):

catch-21 (approved):

PR #1525, opened 2026-03-18
feat(DialogReply): implement auto-focus and scroll behavior for reply dialog

### LINKED ISSUE #996: [Bug]: Mobile: Missing auto scroll down and open soft keyboard when replying to post
### Bug Description

On mobile, when post's reply button is tapped the reply input field should be scrolled down to and the soft keyboard should be displayed.

### Steps to Reproduce

1. On mobile, find a tall post in the feed (with embedded X is a good example)
2. Tap reply button for the post
3. Observe dialog shows but the reply input field is not entirely in view, and the soft keyboard is not displayed so the user needs to tap the input field

### Affected Area

Post (create/edit/delete)

UI / Component (visual issue)

### Frequency

Always (100%)

### Expected Behavior

Scroll the post input fields into view and auto-show the soft keyboard.

### Screenshots / Recording

**Feedback from design team**

Image

**Video of behaviour**

https://github.com/user-attachments/assets/3aa857ef-ca3c-47c5-8389-e20a10ff6ff5

### Console Log Output

```shell

```

### Browser & Version

_No response_

### Device

_No response_

### Additional Context

_No response_

Nirzhuk:
- Added a new `onOpenAutoFocus` prop to the `DialogContent` component to handle focusing and scrolling the reply textarea when the dialog opens.
- Introduced a `scrollReplyTextareaIntoView` function to ensure the textarea is visible and focused appropriately.
- Updated tests to verify the new focus and scroll behavior when the dialog is opened.
fix #996


Nirzhuk:
@greptileai review it again

src/components/organisms/DialogReply/DialogReply.utils.ts
```diff
@@ -0,0 +1,34 @@
+const REPLY_TEXTAREA_SELECTOR = '[data-cy="reply-post-input"] [data-slot="textarea"]';
```
Taewa: @Nirzhuk 
Using Cypress selector (`data-cy=...`) might not be a good idea.

src/components/organisms/DialogReply/DialogReply.tsx
```diff
@@ -15,6 +17,25 @@ export function DialogReply({ postId, open, onOpenChangeAction }: DialogReplyPro
...
     });
 
+  React.useEffect(() => {
+    if (!open || typeof window === 'undefined') return;
+
+    const scrollAuto = () => scrollReplyTextareaIntoDialog('auto');
+    const scrollSmooth = () => scrollReplyTextareaIntoDialog('smooth');
+
+    const initialTimeoutId = window.setTimeout(scrollAuto, 0);
+    const settleTimeoutId = window.setTimeout(scrollSmooth, 200);
```
Taewa: @Nirzhuk 
The `200`ms is for waiting for the dialog component open?

src/components/organisms/DialogReply/DialogReply.tsx
```diff
@@ -1,12 +1,14 @@
 'use client';
 
+import * as React from 'react';
```
Taewa: @Nirzhuk 
I am not 100% sure but I think our new code convention is `import { useEffect } from 'react'`. But actually I can't find related description from our ADR.

src/components/organisms/DialogReply/DialogReply.utils.ts
```diff
@@ -0,0 +1,34 @@
...
+const OPEN_DIALOG_REPLY_TEXTAREA_SELECTOR = `[data-slot="dialog-content"][data-state="open"] ${REPLY_TEXTAREA_SELECTOR}`;
+const DIALOG_CONTENT_SELECTOR = '[data-slot="dialog-content"]';
+
+export function getReplyTextarea(): HTMLTextAreaElement | null {
```
Taewa: @Nirzhuk 
Actually in the `PostInput` component, there is `textareaRef`, so this textarea select logic is not needed. However, if you touch `PostInput`, it'd impact other components that use `PostInput`.
From what I know, `DialogReply` is the only `Dialog` that needs autoscroll. So I understand you don't want to touch where impacts everywhere.
In this case, could you please add `TODO` or `NOTE` that mentions this could be implemented in the `PostInput` in the future when the autoscroll is needed in other place? (Ex: `DialogEditPost` might needs it).
And if it's implemented in the `PostInput`, it'd be simpler since it already has `textareaRef`.

Taewa: @Nirzhuk 
Nit: `export` needed in `getReplyTextarea()`?

src/components/organisms/DialogReply/DialogReply.tsx
```diff
@@ -15,6 +17,25 @@ export function DialogReply({ postId, open, onOpenChangeAction }: DialogReplyPro
...
+    if (!open || typeof window === 'undefined') return;
+
+    const scrollAuto = () => scrollReplyTextareaIntoDialog('auto');
+    const scrollSmooth = () => scrollReplyTextareaIntoDialog('smooth');
+
+    const initialTimeoutId = window.setTimeout(scrollAuto, 0);
+    const settleTimeoutId = window.setTimeout(scrollSmooth, 200);
+
+    const viewport = window.visualViewport;
+    viewport?.addEventListener('resize', scrollSmooth);
```
Taewa: The window resize event should auto scroll to the textarea, right? But the auto scroll doesn't work when I tested it. Can you tell me how to test it please?

src/components/organisms/DialogReply/DialogReply.utils.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (typeof document === 'undefined') return null;
+
+  return (
+    document.querySelector<HTMLTextAreaElement>(OPEN_DIALOG_REPLY_TEXTAREA_SELECTOR) ??
+    document.querySelector<HTMLTextAreaElement>(REPLY_TEXTAREA_SELECTOR)
+  );
+}
+
+export function scrollReplyTextareaIntoDialog(behavior: ScrollBehavior = 'auto') {
```
Taewa, line 14: @Nirzhuk 
It deserves some unit tests?

Taewa (commented):
I left a couple of comments

src/components/organisms/DialogReply/DialogReply.tsx
Nirzhuk: Yes that's correct, to sync both animations.

src/components/organisms/DialogReply/DialogReply.utils.ts
Nirzhuk: changed to id, done :D

src/components/organisms/DialogReply/DialogReply.utils.ts
Nirzhuk: no, removed

src/components/organisms/DialogReply/DialogReply.tsx
Nirzhuk: This is only testeable in phone when you rotate the screen.
If you dont have it, this is testeable making the website smaller vertically when you're in a post bigger than the screen.

This works perfectly for me.

src/components/organisms/DialogReply/DialogReply.utils.ts
Nirzhuk: This is a dom interactive function, the test could be too mocky to be useful

Taewa (approved):

PR #1558, opened 2026-03-23
feat(ui): add success toasts for tag add/remove operations

### LINKED ISSUE #1112: [Bug]: No confirmation shown when successfully tagging
### Bug Description

When tagging a post with a new tag it's not obvious whether it was successful because the tag is not immediately shown, which may be correct because we want to show the most popular tags in the limited space (often limited to top 3).

The failure scenario is already covered by an error toast.

### Steps to Reproduce

1. Find a post with more than 3 tags, each with more than 1 tagger
2. Add a new tag to the post
3. Input field clears and no indication of success is shown

### Affected Area

Feed / Timeline

None

### Frequency

Always (100%)

### Expected Behavior

We need some indication of a successful add and remove. A toast message makes the most sense to me because other write events show a toast. Imo we should show whilst tagging profiles too.

### Screenshots / Recording

https://github.com/user-attachments/assets/33f11c9d-f70b-4565-b97f-70303e59b93c

### Console Log Output

```shell

```

### Browser & Version

_No response_

### Device

_No response_

### Additional Context

_No response_

Orlandohub:
## Summary
- fix #1112
- Add success toast notifications when users add or remove tags, providing positive feedback alongside the existing error toasts
- Internationalize all tag-related toast strings via next-intl across 9 locales

## Changes
- Add success toasts to `handleTagAdd` and `handleTagToggle` in both `usePostTags` and `useTagged` hooks
- Replace hardcoded English toast strings with `useTranslations('toast.tags')` calls
- Add `added`, `addedDesc`, `removed`, `removedDesc` translation keys to all 9 locale files (ar, de, en, es, fr, it, ja, pt-BR, zh)
- Add test coverage for success and error toast assertions in both hook test files

## Test plan
- [ ] Add a tag to a post → success toast appears with correct label
- [ ] Remove a tag from a post → success toast appears with correct label
- [ ] Add a tag to a user profile → success toast appears
- [ ] Remove a tag from a user profile → success toast appears
- [ ] Switch locale → toast messages display in the selected language
- [ ] Unit tests pass: `npx vitest run src/hooks/usePostTags src/hooks/useTagged`

## Notes
- Error toasts for failed operations were already present and are now also i18n-ized
- No UI or styling changes — only toast content and hook logic


secondl1ght (approved):
Tested:

https://github.com/user-attachments/assets/8fec94b7-bfb2-4108-af33-564b9207b6d3

While testing I happened to notice a couple other unrelated things:

1. It is hard to read the toast due to its transparency if the elements behind it have a strong color difference.
2. Zero count tags are still shown.

Maybe both of those are already known/expected but worth pointing out for when @catch-21 takes a look later.

Also because this is an opinionated UX change, maybe @aldertnl should advise if he agrees with this toast behavior? At least for the visible tags I found the current border highlighting and counter to be enough indication of the user action, but having the toast as well doesn't hurt. Another option would be to always append any new tag to the visible row. These are ideas--but the PR resolves what is described in the related issue. 👍🏼

PR #1578, opened 2026-03-25
fix: scope pull to refresh feed to timeline comp

### LINKED ISSUE #1555: [Bug]: Cannot scroll up custom feeds on mobile
### Description

On mobile, with many custom feeds, it's not possible to scroll back up the list. It appears to be triggering the feed reload action.

### Expected Behaviour

Is possible to scroll up and down the list of custom feeds.

### Steps to Reproduce

1. On mobile
2. Create many custom feeds so that the list is longer than the display height
3. Scroll down the list by touch-dragging up
4. Attempt to scroll back up by touch-dragging down
5. Observe list does not scroll down, it invokes the feed reload instead

### Screenshots

https://github.com/user-attachments/assets/cb6195bd-5cb4-4ad3-ba7e-b2c7ec0cf677

### Console Log

```shell

```

### Browser & Version

_No response_

secondl1ght:
## 🐛 Fix: Scope pull-to-refresh to feed container

Fixes #1555

### Problem

The `usePullToRefresh` hook attached touch event listeners to `window` globally, meaning **any** touch-drag anywhere on the page could trigger pull-to-refresh — including touches on dialogs, sidebar menus, drawer overlays, and custom feed lists. The only guards were `disabled` and `isAtTop()` (`window.scrollY <= 0`), neither of which were aware of overlays or non-feed scroll containers.

This was first noticed when scrolling up through a long list of custom feeds on mobile, which triggered the feed reload instead of scrolling. However, the bug existed before custom feeds — it affected any overlay or scrollable panel open on top of the timeline.

### Solution

Scoped pull-to-refresh touch events from `window` to a required `containerRef` element, so only touches originating inside the timeline feed trigger the interaction.

### Changes

- **`usePullToRefresh.ts`** — Changed event listeners from `window` to `containerRef.current`. The hook now early-returns if the container element isn't mounted yet.
- **`usePullToRefresh.types.ts`** — Added required `containerRef: React.RefObject<HTMLElement | null>` to `UsePullToRefreshOptions`. This is intentionally non-optional to prevent accidental global pull-to-refresh.
- **`TimelineFeedContent.tsx`** — Wrapped feed content in an `Atoms.Container` with a ref, matching `ContentLayout`'s main content area classes (`min-w-0 flex-1 gap-6 lg:overflow-hidden`) to preserve existing flex-col spacing.
- **`usePullToRefresh.test.tsx`** — Refactored all tests to dispatch touch events on a container element. Added two new tests:
  - Touches on `window` (outside container) are ignored
  - Touches on sibling elements are ignored
- **`TimelineFeedContent.test.tsx`** — Updated assertions to verify `containerRef` is passed. Snapshots updated.
- **`TimelineFeed.test.tsx`** — Snapshots updated.

### Related

- Closes #1555

### Screenshots

[Screencast from 2026-03-24 22-35-53.webm](https://github.com/user-attachments/assets/912e6f95-7fc6-4241-b2d0-4186d2632c4d)

### Notes

- The `containerRef` is required (not optional) by design — this prevents anyone from accidentally re-introducing a global pull-to-refresh in the future.
- The `PullToRefreshIndicator` uses `absolute` positioning with `pointer-events-none`, so it works correctly inside the container without interfering with touch events.
- The `Atoms.Container` wrapper classes match `ContentLayout`'s main content area to preserve the same flex-col `gap-6` spacing that children previously inherited as direct descendants.

---


infin1t3 (approved):
https://github.com/user-attachments/assets/dbe030b7-30c8-48cc-9a7d-43d67f728b2e

Verified on `f6dc05e3c0a98c63abf285308caeaecdff985783` - lgtm.

PR #1579, opened 2026-03-25
fix(search): validate tag input before applying search

### LINKED ISSUE #1262: [Bug]: Can apply search for tag using invalid value length
### Description

When I search for a long value it attempts to search for it as a tag and then remembers this value for future even though the value is invalid due to it's length. It can also yields results, which is unexpected.

### Expected Behaviour

Shouldn't be able to search for tags longer than 20 chars because this is max tag value size.
Shouldn't have any results.

### Steps to Reproduce

1. Search value `pk:abcdefghijklmnopqrstuvwxyz`
2. Observe tag search is performed

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

Orlandohub:
## Summary
- fix #1262
- Searching with invalid strings (e.g. `pk:abcdefghijklmnopqrstuvwxyz`) no longer produces broken tag filters that return random posts
- The colon in `pk:` was corrupting stream ID parsing since `:` is the component delimiter — validation now rejects all banned tag characters on Enter
- Invalid input is preserved with a toast message explaining tag constraints, instead of being silently discarded

## Changes
- Add `isValidTagLabel` utility in `src/libs/utils/utils.ts` that checks length (max 20, grapheme-aware) and banned characters (`:`, `,`, whitespace) using existing `sanitizeTagInput` and `TAG_MAX_LENGTH`
- Gate `handleEnter` in `SearchInput.tsx` — reject invalid tags with a localized toast and return `false` to preserve the input
- Update `useSearchInput` so `onEnter` can return `false` to prevent the input from being cleared
- Add `searchQuery.length <= TAG_MAX_LENGTH` guard in `useSearchAutocomplete.ts` to skip the `tags/by_prefix` API call for over-length queries (user search still proceeds)
- Add `invalidTag` translation key across all 9 locale files
- Add 8 unit tests for `isValidTagLabel` and 2 boundary tests for the autocomplete length guard

## Test plan
- [ ] Type `pk:abcdefghijklmnopqrstuvwxyz` in search bar and press Enter — input stays, toast appears, no tag is added
- [ ] Type a string longer than 20 characters and press Enter — input stays, toast appears
- [ ] Type a valid short tag (e.g. `bitcoin`) and press Enter — tag is added, input clears, search executes normally
- [ ] Type `pk:abc` — autocomplete should show user suggestions but no tag suggestions
- [ ] Click a user suggestion from autocomplete — still navigates to their profile
- [ ] Click a tag suggestion from autocomplete — still adds it as a search tag

## Notes
- Autocomplete while typing is unchanged — user searches by name and ID still work for long/prefixed queries; only the tag-by-prefix call is gated
- `handleTagClick` (from autocomplete/hot-tags) is not gated since those tags are already validated server-side
- `onEnter` return type changed from `void` to `boolean | void` — backward compatible, existing consumers that return nothing still clear the input


catch-21:
LGTM for the most part. My only request is that we don't clear the search field when >20 chars and hitting enter.  Because, for example, if you type a whole pubky or prefix that is this long then hitting enter clears the search input and need to re-input it.

infin1t3 (approved):
https://github.com/user-attachments/assets/3dd14fea-5898-4882-a317-6febdde37466

Tested on `2f7b88b75c4bd6419c2f9386f70d3ae62b66fdf6` - LGTM.

PR #1584, opened 2026-03-25
feat(ui): add skeleton loaders for all remaining loading states

### LINKED ISSUE #1446: [Feature]:  Finish adding skeletons for loading states across the app
### Problem Statement

All loading states across the app are supposed to use skeletons. It looks like some of this was completed in #1315, but there are still areas missing, for example avatars, feed, etc.

### Proposed Solution

Finish adding skeletons.

### Suggested Priority

Orlandohub:
## Summary
- fix #1446
- Replace all non-skeleton loading states (text placeholders, `Atoms.Spinner`, raw `animate-pulse` divs) with proper `Atoms.Skeleton`-based skeleton loaders across 18 components
- Follows the existing skeleton architecture: colocated `*.skeleton.tsx` for single-owner components, standalone folder for reusable skeletons

## Changes
- Add 16 new skeleton files: `PostCardSkeleton` (reusable), `PostHeader.skeleton`, `PostContentBase.skeleton`, `PostActionsBar.skeleton`, `PostPageHeader.skeleton`, `SinglePostContent.skeleton`, `NotificationsContainer.skeleton`, `ProfileTagged.skeleton`, `SinglePostParticipants.skeleton`, `EditProfileForm.skeleton`, `MutedUsersList.skeleton`, `GenericPreview.skeleton`, `ShareTarget.skeleton`, `ProfileMenuActionsContent.skeleton`, `PostMenuActionsContent.skeleton`
- Update `TimelineLoading`/`TimelineLoadingMore` to render post card skeletons instead of "Loading posts..." text
- Replace `Atoms.Spinner` load-more indicators in `ProfileFollowing`, `ProfileFollowers`, `ProfileFriends`, `WhoToFollowPageMain` with `FullUserListItemSkeleton`
- Replace raw `animate-pulse` in `PostPageHeader` and `MarkdownEditor` with `Atoms.Skeleton`
- Replace inline `"Loading..."` text in `VisualTimelinePosts` tile overlay with inline skeleton

## Test plan
- [ ] Navigate to the feed and verify skeleton loaders appear briefly during initial load instead of "Loading posts..." text
- [ ] Open a single post page and verify the post header, content, and actions bar show skeletons while loading
- [ ] Visit the notifications page and verify notification item skeletons appear on initial load and when loading more
- [ ] Visit a user profile's Tagged tab and verify tag pill skeletons appear during loading
- [ ] Visit a user profile's Following/Followers/Friends tabs and scroll to trigger load-more skeletons
- [ ] Open the Edit Profile form and verify the form skeleton appears while user data loads
- [ ] Visit Settings > Muted Users and verify user row skeletons appear during loading
- [ ] Open a post's "..." menu and a profile's "..." menu to verify menu item skeletons
- [ ] Verify the Who to Follow page shows skeletons when loading more users

## Notes
- Button-level spinners for user-initiated actions (follow/unfollow, submit, bookmark toggle) are intentionally unchanged — spinners are the correct UX pattern for action feedback
- All skeleton counts use constants or props per `docs/skeleton-architecture.md`, never hardcoded literals


Orlandohub:
@greptile-apps review

secondl1ght:
@Orlandohub just a couple conflicts here.

secondl1ght:
@greptileai

secondl1ght (commented):
Skeleton for posts does not match the real elements:

Image

Image

1. Action buttons on left instead of right
2. Four action buttons instead of five
3. Missing tag button skeleton
4. Shapes/sizes are different

Was the idea to just have a rough similarity for the loading states instead of matching the elements? I won't go through every one to verify the differences, but this is one obvious example.

secondl1ght (commented):
The timeline feed jumps up after the loading state, there is no reason why the loading state should have more top margin than the real feed. Skeleton transitions should be as seamless as possible with the minimum amount of layout shift.

https://github.com/user-attachments/assets/98df48b5-a355-4e58-9561-4600502b59cd

src/components/molecules/PostLinkEmbeds/Providers/Generic/GenericPreview.skeleton.tsx
```diff
@@ -0,0 +1,13 @@
+import * as Atoms from '@/atoms';
+
+export function GenericPreviewSkeleton() {
+  return (
+    <Atoms.Container className="justify-between gap-6 rounded-md bg-muted p-6 lg:flex-row">
```
secondl1ght: This skeleton just renders a solid grey box because the container has the same background color as the skeletons.

src/components/organisms/PostContentBase/PostContentBase.skeleton.tsx
```diff
@@ -0,0 +1,11 @@
+import * as Atoms from '@/atoms';
+
+export function PostContentBaseSkeleton() {
+  return (
+    <Atoms.Container className="min-w-0 gap-3" overrideDefaults>
```
secondl1ght: This gap is not doing anything because there is `overrideDefaults`.

```suggestion
    <Atoms.Container className="min-w-0 gap-y-1 flex flex-col" overrideDefaults>
```

BEFORE
Image

AFTER
Image

secondl1ght (commented):
An observation--there are a few ADR violations, maybe some are valid but maybe the ADR document itself is overly restrictive. We can consider making changes to the doc if during implementation we find certain rules do not apply well in practice.

src/components/organisms/PostCardSkeleton/PostCardSkeleton.tsx
```diff
@@ -0,0 +1,16 @@
+import * as Atoms from '@/atoms';
```
secondl1ght, line 1: Missing tests for this component.

src/libs/env/env.ts
```diff
@@ -290,6 +290,7 @@ function parseEnv(): z.infer<typeof envSchema> {
     NEXT_PUBLIC_DB_NAME: process.env.NEXT_PUBLIC_DB_NAME,
     NEXT_PUBLIC_DB_VERSION: process.env.NEXT_PUBLIC_DB_VERSION,
     NEXT_PUBLIC_DEBUG_MODE: process.env.NEXT_PUBLIC_DEBUG_MODE,
+
```
secondl1ght: Was this committed on purpose?

src/config/test.ts
```diff
@@ -20,6 +20,7 @@ import enMessages from '../../messages/en.json';
 
 process.env.NEXT_PUBLIC_DB_VERSION = '1';
 process.env.NEXT_PUBLIC_DEBUG_MODE = 'false';
+
```
secondl1ght: Was this committed on purpose?

secondl1ght (commented):
We are missing skeletons for user avatars, the smiley face fallback should only appear if there truly is no avatar available. Currently on refresh or initial page load the smiley faces appear before a user avatar is shown. The flow should be:

Skeleton > Avatar > Fallback

secondl1ght (commented):
Dead Translation Keys (12 keys to remove)

| Key | Value |
|---|---|
| `common.loading` | `"Loading..."` |
| `common.loadingActions` | `"Loading actions..."` |
| `common.loadingHeader` | `"Loading header..."` |
| `common.loadingPost` | `"Loading post..."` |
| `common.loadingContent` | `"Loading content..."` |
| `common.loadingReplies` | `"Loading replies..."` |
| `common.loadingParticipants` | `"Loading participants..."` |
| `common.loadingTags` | `"Loading tags..."` |
| `common.loadingTaggers` | `"Loading taggers..."` |
| `common.loadingPosts` | `"Loading posts..."` |
| `common.loadingMorePosts` | `"Loading more posts..."` |
| `common.loadingPreview` | `"Loading preview..."` |

**Still used:** `common.loadingMoreTags` (in `TaggedList.tsx`), `common.loadingBookmark` (in `PostActionsBar.tsx` as an aria-label).

secondl1ght (commented):
Were these components skipped from becoming skeletons for a reason?

1. **`TaggedList.tsx`** — renders `t('loadingMoreTags')` as visible text during pagination
2. **`ProfilePageFilterBar.tsx`** — uses `<Atoms.Spinner>` as a count placeholder while tab counts load
3. **`InputField.tsx`** — has a hardcoded `'Loading...'` default prop (also an i18n issue)

secondl1ght:
FYI @catch-21 some E2E tests may also need to be updated as a result of this PR, for example: https://github.com/pubky/pubky-app/blob/1bc83c043cb4aa9ace881f39b3ce5c3561c191c8/cypress/support/posts.ts#L443

src/libs/env/env.ts
Orlandohub: nope, reverted

src/config/test.ts
Orlandohub: nope, reverted

Orlandohub:
@secondl1ght 
fixed almost all requests with the exception to the following:

> We are missing skeletons for user avatars, the smiley face fallback should only appear if there truly is no avatar available. Currently on refresh or initial page load the smiley faces appear before a user avatar is shown. The flow should be:
> 
> Skeleton > Avatar > Fallback

This is not the desired behavior from the design team. When loading the avatar it should always be the smiley face has default, no skeleton, because the smiley face is the de facto avatar loader. Now, when there's components like the taggers list, where the avatar is part of composition which needs to load together, in that case there will be a skeleton loader for the avatar section of the component which will be replaced by smiley face or profile picture once the full component loads as one.

> Were these components skipped from becoming skeletons for a reason?
> 
> TaggedList.tsx — renders t('loadingMoreTags') as visible text during pagination
> ProfilePageFilterBar.tsx — uses <Atoms.Spinner> as a count placeholder while tab counts load
> InputField.tsx — has a hardcoded 'Loading...' default prop (also an i18n issue)

TaggedList has been refactor but the other two do not make sense to have a skeleton loader. the profile page loader spinner is for a number which will only have a loading state when undefined, (which I believe is never, I think Vlada made it as a fail safe fallback) and I think a skeleton loader for a number does not look appealing. Finally the InputField, it's a very specific one, inside the input field loader which is only used in one place for the onboarding `PublicKeyCard` when the pubkey is generated. I do not thing it will look great to have a skeleton loader inside an input field

secondl1ght (commented):
@Orlandohub can you resolve conflicts to pull latest from dev? I am seeing a few bugs but want to see if they are resolved first.

EDIT: Tested locally and the bugs were unrelated.

secondl1ght:
@greptileai

src/components/molecules/MarkdownEditor/MarkdownEditor.tsx
```diff
@@ -5,11 +5,14 @@ import { forwardRef } from 'react';
 import { type MDXEditorMethods, type MDXEditorProps } from '@mdxeditor/editor';
 import * as Atoms from '@/atoms';
 
-// This is the only place InitializedMDXEditor is imported directly.
 const Editor = dynamic(() => import('./InitializedMDXEditor'), {
-  // Make sure we turn SSR off
   ssr: false,
-  loading: () => <Atoms.Container className="h-22 animate-pulse rounded-md bg-card" />,
```
secondl1ght, line 12: I had this set to the same height as the element (88px) so that there was no layout shift after loading, but now it is shorter. I guess it doesn't have to be perfect, or maybe we can look into that lib someone shared on Slack recently where the skeletons are automatically generated to the right dimensions.

src/hooks/usePostTags/usePostTags.constants.ts
```diff
@@ -1 +1 @@
-export const TAGS_PER_PAGE = 10;
+export const TAGS_PER_PAGE = 3;
```
secondl1ght, line 1: Why was this changed? It seems outside the scope of skeletons, we only want 3 tags now instead of 10?

secondl1ght:
@greptileai I wanted you to do a final re-review, not sell us some of your products. Did you do the review and found nothing? If so then say that.

secondl1ght (dismissed):
Left a couple more comments but approving, specifically this one https://github.com/pubky/pubky-app/pull/1584#discussion_r3061599162 if it was unintentional then it should be reverted before merging. 

I'll do a follow-up PR for the translation issue in `InputField` and request your review on that.

src/hooks/usePostTags/usePostTags.constants.ts
Orlandohub: This was done because while fixing the post skeleton ui I've noticed the tags per page set to 10 were to much, making the UI jump. Also, Since I've implemented this code, UI changed and now there's a max 3 tags being displayed, which if you want to see more you need to click a button.

src/components/molecules/MarkdownEditor/MarkdownEditor.tsx
Orlandohub: I've updated this by using hardcoded px value so there's no jumps and all works with the previous 88px total height

https://github.com/user-attachments/assets/ffe1b361-6800-4576-a77e-d95fa4a55b8f

secondl1ght (approved):

PR #1585, opened 2026-03-25
feat: update SinglePost layout with filters and columns

### LINKED ISSUE #367: [Bug]: Post view should display columns
### Describe the bug

When viewing a post it is standalone without the columns (Reach, Sort, Layout, Content, Participants, Feedback). It should display within the same columns used for feed page. Only without columns during wide view.

### Reproduce

1. View feed with columns
2. Click on any post in the feed
3. Observe post is displayed without columns

### Screenshots / Recording

**Design**

Image

**Actual**

Image

### Web Browser

_No response_

### Pubky App Version

1125a965b4bef1e1e271e5165e47af6847eeb10e

### Console log output

```shell

```

Nirzhuk:
- Introduced SinglePostDrawer, SinglePostLeftSidebar, and SinglePostSidebar components to improve the SinglePost page structure.
- Updated ContentLayout to accept new props for left and right sidebar and drawer content.
- Enhanced SinglePostContent and UserListItem for better layout and user experience.
- Added tests for new components and updated existing tests to ensure functionality.

This refactor aims to streamline the SinglePost page and improve the overall user interface.


src/components/organisms/SinglePostSidebar/SinglePostSidebar.tsx
```diff
@@ -0,0 +1,14 @@
+import * as Atoms from '@/atoms';
+import * as Organisms from '@/organisms';
```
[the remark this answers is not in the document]
Nirzhuk: @greptile-apps this is someething happen everywhere. I know it's bad we're working to remove it.

src/components/organisms/SinglePostLeftSidebar/SinglePostLeftSidebar.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return <SinglePostFilters />;
+}
+
+/**
+ * SinglePostLeftDrawer
+ *
+ * Left drawer for SinglePost page (tablet/mobile).
+ */
+export function SinglePostLeftDrawer() {
+  return <SinglePostFilters />;
```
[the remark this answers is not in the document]
Nirzhuk: @greptile-apps it's okey due the integration separation of components even we have identical components.

Orlandohub (changes requested):
## Left column not aligned with post title
Image

## Mobile Breadcrumb vertical spacing to card

### Actual - (24px)
Image

### Design
Image

## Tags not applied correctly on column mode

### Actual
Image

### Design
Image

Nirzhuk:
@greptileai review it again please

Nirzhuk:
@greptileai review it again please

Orlandohub (dismissed):

Orlandohub (changes requested):
## SinglePostLeftSidebar reinvents HomeFeedSidebar instead of reusing it

The existing HomeFeedSidebar is already designed for reuse — it's parameterized with hideReachFilter, hideLayoutFilter, allowVisualLayout, and a variant prop. The Bookmarks page reuses it directly

**Problem**:
This duplicates the filter composition logic and breaks away from the established pattern. If the filter set or order changes, you now have two places to update. 

**Low effort solution:**
extend HomeFeedSidebar with enabledFilters and reuse it.

**Best solution:**
extract the inner filter composition into something generic, for example FeedFiltersPanel or SidebarFiltersPanel, and have HomeFeedSidebar become a thin feed-flavored wrapper around it.

## feedVariant={TIMELINE_FEED_VARIANT.BOOKMARKS} is semantically wrong

The SinglePost page is not a bookmarks feed
`feedVariant={TIMELINE_FEED_VARIANT.BOOKMARKS}`

**Problem**
This is borrowing the bookmarks variant purely to get a specific ContentLayout behavior (likely how the layout resolution works). If the bookmarks variant behavior ever changes, the post page silently breaks.

**Possible solution**
Don't pass feedVariant at all and keep the logic in ContentLayout but replace feedVariant with a union prop like:

```ts
type LayoutPolicy =
  | { kind: 'feed'; variant: TimelineFeedVariant }
  | { kind: 'page'; allowWide: boolean; allowVisual: boolean };
```

##  SinglePostCard is effectively a fork of PostMain

The tag panel logic, footer branching, refs, reply/repost dialogs, and wide-vs-inline rendering are largely duplicated.

**Solution**
Use PostMain directly, or through a very thin SinglePostMain wrapper to preserve naming and keep post-page wiring isolated.

PostMain already has the right abstraction boundary — it takes tagsLayout as a prop. The SinglePost template could:

1. Read layout from the store at the **template level** (where layout decisions belong)
2. Map layout === 'wide' → tagsLayout="side" / else → tagsLayout="inline"
3. Pass that to PostMain (or a shared variant of it)

## SinglePostCard reads global layout store

```
  const { layout } = Core.useHomeStore();
  // ...
  const isWideLayout = layout === 'wide';
```

**Problem**
this organism is now coupled to page-level layout policy and duplicates logic that already exists elsewhere

That decision belongs higher up, because it is page/template policy, not card-local behavior.

**Solution**
The better API is not isWideLayout, but the same abstraction PostMain.tsx (line 22) already uses: tagsLayout="inline" | "side"

## SinglePostSidebar and SinglePostDrawer are near-identical

SinglePostSidebar = Participants + FeedbackCard. SinglePostDrawer = Participants only. Same props type, same wrapper. These could easily be a single component:

```
export const SinglePostRightPanel = ({ postId, showFeedback = true }: Props) => (
  <Container overrideDefaults className="flex flex-col gap-6">
    <SinglePostParticipants postId={postId} />
    {showFeedback && <FeedbackCard />}
  </Container>
);
```
This mirrors how the feed uses HomeFeedContent as a shared base for both sidebar and drawer.

Nirzhuk:
@Orlandohub 

> SinglePostLeftSidebar reinvents HomeFeedSidebar instead of reusing it

This first iteration doesnt have filters in purpose because we dont have the backend for it, but we miss the design part which is what we're fixing right now. I believe if we does this change, we might need patch again because the filters are going change for sure, unique, because it's not going to be a copy 1:1. Right now creating this files seems like a waste but I believe will be used as scafolding.

> feedVariant={TIMELINE_FEED_VARIANT.BOOKMARKS} is semantically wrong. 

I believe the correct fix is just going to remove that yeah, this needs to be improved when the filters can work with backend. 

> SinglePostCard reads global layout store

Which in the core is the same but w're not doing an extreme prop drilling of 5 components to do the same effect.

> SinglePostCard is effectively a fork of PostMain

This is actually dangerous right now, because we can bring this we dont want or changing the whole thing just for the sake of "smelling" code which this is a small bug fixing, not a whole refactoring a very important part of the website. This needs its own time, not treating it like a 5 minute change.

> SinglePostSidebar and SinglePostDrawer are near-identical

This change in my opinion is not cleaner, you're creating something with a conditional when 50% of the usage doesn't need it, it's the same type of smell code. The other case it's true is not perfect, but it's isolated per case, which makes sense.

src/components/organisms/SinglePostCard/SinglePostCard.tsx
```diff
@@ -103,29 +106,47 @@ export function SinglePostCard({ postId, className }: SinglePostCardProps) {
                 {/* Spacer to push actions bar to bottom */}
                 <Atoms.Container overrideDefaults className="flex-1" />
 
-                {/* Tags on mobile - always visible */}
-                <Organisms.PostTagsPanel
-                  ref={mobileTagsPanelRef}
```
infin1t3, line 108: Given you removed this line from `mobileTagsPanelRef`, ref declaration on line 28 and focus declaration on line 34 are now dead code.

**Fix:** Either pass ref={mobileTagsPanelRef} to the inline PostTagsPanel when tagsExpanded is true, or remove both the ref declaration and the .focus() call.

infin1t3 (changes requested):

src/components/organisms/SinglePostCard/SinglePostCard.tsx
Nirzhuk: I will remove them, good catch.

src/components/organisms/SinglePostCard/SinglePostCard.tsx
```diff
@@ -21,19 +21,16 @@ import type { SinglePostCardProps } from './SinglePostCard.types';
  */
```
infin1t3, line 21: This component is now effectively a fork of `PostMain`. Given that this is not a new component introduced in this PR and the fact that this PR already has a lot of changes - I would recommend you to make a refactoring ticket to address this in a follow-up PR right after this one is merged - and to use `PostMain` or `PostMain` with a thin wrapper directly

src/components/organisms/SinglePostLeftSidebar/SinglePostLeftSidebar.tsx
```diff
@@ -0,0 +1,43 @@
+'use client';
```
infin1t3, line 1: This component looks visually exactly the same as `HomeFeedSidebar`.

Spoke to @Nirzhuk about it and I will try to see does it make sense to have just one component instead - or the logic is to complex and using one component would be a mess and having this separation with simple encapsulation is actually better. Will report back my findings.

src/components/organisms/SinglePostCard/SinglePostCard.tsx
```diff
@@ -103,29 +105,47 @@ export function SinglePostCard({ postId, className }: SinglePostCardProps) {
...
+                <Atoms.Container
+                  onClick={handleFooterClick}
+                  className={Libs.cn(
+                    'flex-col items-start gap-2 md:flex-row md:justify-between md:gap-4',
+                    tagsExpanded ? 'md:items-end' : 'md:items-start',
+                  )}
+                >
+                  {!isWideLayout &&
+                    (tagsExpanded ? (
+                      <Organisms.PostTagsPanel
```
infin1t3: This is a verbatim duplicate of a code block 65-85. Consider extracting a local `tagsSection` variable above the return, then use {tagsSection} in both branches.

src/components/organisms/SinglePostDrawer/SinglePostDrawer.types.ts
```diff
@@ -0,0 +1,3 @@
+export interface SinglePostDrawerProps {
```
infin1t3: I would inline this if you use it only in one component - no need for a file.

src/components/organisms/SinglePostDrawer/SinglePostDrawer.tsx
```diff
@@ -0,0 +1,12 @@
+import { Container } from '@/atoms/Container';
```
infin1t3: SinglePostDrawer = SinglePostParticipants. SinglePostSidebar = SinglePostParticipants + FeedbackCard. 
Same Container wrapper, same className, same types. 

Consider a single component with `showFeedback?: boolean` prop - should reduce code and complexity with a simple prop.

```
export const SinglePostRightPanel = ({ postId, showFeedback = true }: Props) => (
  <Container overrideDefaults className="flex flex-col gap-6">
    <SinglePostParticipants postId={postId} />
    {showFeedback && <FeedbackCard />}
  </Container>
);
```

infin1t3: Consider do you need 
```
 SinglePostDrawer/index.ts, SinglePostLeftSidebar/index.ts, SinglePostSidebar/index.ts
```
given you use these only in sibling modules - or you can just delete index.ts files.

src/components/templates/Post/SinglePost/SinglePost.tsx
```diff
@@ -1,37 +1,37 @@
...
  * - Main post card (FULL WIDTH) with tags panel in two-column layout
- * - Below: Two columns with Replies timeline (larger) and Participants sidebar (smaller)
+ * - Below:  Replies timeline
  *
  * This template uses a FIXED layout that doesn't change based on user preferences.
  * All hook logic is delegated to the SinglePostContent organism.
  */
 export function SinglePost({ postId }: SinglePostProps) {
   return (
     <>
```
infin1t3: I think you don't need this Fragment anymore.

infin1t3 (dismissed):

src/components/organisms/SinglePostCard/SinglePostCard.tsx
Nirzhuk: Created https://github.com/pubky/pubky-app/issues/1635
✅

src/components/organisms/SinglePostDrawer/SinglePostDrawer.types.ts
Nirzhuk: Which are the rules about this, to know better for the next one, I feel if I dont create the file someone can ask me to create the file and if this a feeling more than a rule, I prefer to go as standard as possible.
Because there's other cases with 1 line interfaces.

src/components/organisms/SinglePostDrawer/SinglePostDrawer.tsx
Nirzhuk: I don't like this solution ngl, because whatever this component changes we need to test 2 things, the other way we have enough insolation will only test the proper component. But I'm not going to fight more for this, code changed.
This solution it's a bad pattern in general which when things grows it's a pain in the ass to maintain. (this case is maybe small, but bigger components with the same pattern have this problem.


This been said: **change have been update it.**

src/components/organisms/SinglePostDrawer/SinglePostDrawer.types.ts
infin1t3: Will add this to our AI code reviewer

src/components/organisms/SinglePostLeftSidebar/SinglePostLeftSidebar.tsx
infin1t3: Ok, so after spending some time with this component here is what I think.

The challenge is the fact that the current component - `HomeFeedSidebar` is intended to serve as a Left Sidebar for some Timeline pages - Home, Search, Bookmarks.

Where this newly introduced component `SinglePostLeftSidebar` is intended to serve as a Left Sidebar for the Single Post view.

Visually they line up; internally there’s extra stuff on the timeline side — feedVariant is the obvious one. On Single Post we don’t pass it from the template, and with layoutOnly most of that path doesn’t affect what you see, but the hook still runs with the default — so it’s fairer to say we don’t surface feedVariant there, not that it literally doesn’t exist in the stack.

So: unifying both makes sense and doesn’t, and it’s not only taste — there’s also maintenance vs branching. Duplicating two components risks drift when order/spacing/Figma changes; one component with flags risks conditionals like the feedDerived / reachProps blocks — that’s the real “con”, and the alternative “con” is two files that have to stay in sync.

Whatever we pick, we should expect to revisit if requirements split - that’s normal, not a failure of the first approach.

If I had to unify this now, I would do it somehow like this - https://github.com/pubky/pubky-app/tree/fix/unify-single-post-left-sidebar-with-home-feed-filters - creating a third component that holds all the logic and that is used by both sidebars.

Pros - single source of truth.

Cons - well this lol
```
const feedDerived = layoutOnly
    ? {
        resolvedContent: undefined,
        disabledContentTabs: [],
        showVisualLayout: false,
      }
    : {
        resolvedContent: resolveVisualFeedContent({
          content,
          variant: feedVariant,
          isVisualActive,
        }),
        disabledContentTabs: isVisualActive ? VISUAL_DISABLED_CONTENT : [],
        showVisualLayout: allowVisualLayout && !isPhoneViewport,
      };

  const reachProps = layoutOnly
    ? { selectedTab: undefined, defaultSelectedTab: undefined, disabled: true as const }
    : { selectedTab: reach, onTabChange: setReach };

  const sortProps = layoutOnly
    ? { selectedTab: undefined, defaultSelectedTab: undefined, disabled: true as const }
    : { selectedTab: sort, onTabChange: setSort };

  const contentProps = layoutOnly
    ? { selectedTab: undefined, defaultSelectedTab: undefined, disabled: true as const }
    : {
        selectedTab: feedDerived.resolvedContent,
        onTabChange: setContent,
        disabledTabs: feedDerived.disabledContentTabs,
      };
```

In any case if you guys are also not sure what to do here, let's share on #franky with the team so everyone else can chime in.

cc @Nirzhuk @Orlandohub

Orlandohub (dismissed):
After a lengthy review we agreed @Nirzhuk will open new tickers for the side bars and post cards which should tackle concerns raised during this PR review.

src/components/organisms/SinglePostLeftSidebar/SinglePostLeftSidebar.tsx
Nirzhuk: https://github.com/pubky/pubky-app/issues/1648
 Seems a good idea but let's not overscope this bug and give the love it needs the refactor

Orlandohub (approved):

PR #1586, opened 2026-03-25
fix(ui): enable username truncation to nested reply column

### LINKED ISSUE #1566: [Bug]: Long username on nested reply is not truncated
### Description

Nested replies don't apply truncation to long usernames like top level posts and replies do.

### Expected Behaviour

Username should be truncated on nested replies as they are for posts and top level replies.

### Steps to Reproduce

1. On mobile or browser with narrow viewport
2. Find a post that has been replied to, and that reply has been replied to by someone with a long username
3. Observe the long username pushes the reply's age indicator off to the right and the end of the username goes off the screen

### Screenshots

https://github.com/user-attachments/assets/4645e68e-ef1a-4d4b-b5df-392348837d0c

https://github.com/user-attachments/assets/b5dbb67b-17b8-4e82-8c45-a7d77806be30

### Console Log

```shell

```

### Browser & Version

_No response_

Orlandohub:
## Summary
- fix #1566
- Long usernames in nested sub-replies were not truncating, pushing action buttons (tags, bookmark, repost) off-screen
- Root cause: the sub-reply flex column in `ReplyWithNested` was missing `min-w-0`, breaking the CSS min-width constraint chain

## Changes
- Add `min-w-0` to the nested sub-reply `Container` in `ReplyWithNested.tsx` (both `isLastReply` and non-last branches)
- Add test verifying the truncation constraint class is applied to the nested column
- Update snapshot

## Test plan
- [ ] Open a post with deeply nested replies where a user has a very long username
- [ ] Verify the username truncates with ellipsis and action buttons remain visible
- [ ] Verify top-level replies still truncate correctly
- [ ] `vitest run ReplyWithNested` — 11 tests pass

## Notes
- This is a single `min-w-0` addition — the flex truncation chain was intact from `PostMain` down through `PostHeaderUserInfo`, but broken at the `ReplyWithNested` nesting container level

### DIFF
--- a/src/components/organisms/ReplyWithNested/ReplyWithNested.tsx
+++ b/src/components/organisms/ReplyWithNested/ReplyWithNested.tsx
@@ -79,7 +79,7 @@ export function ReplyWithNested({
               {!isLastReply && <Atoms.Container overrideDefaults className="w-3 shrink-0 border-l border-border" />}
 
               {/* Indented sub-replies */}
-              <Atoms.Container overrideDefaults className={isLastReply ? 'ml-6 flex-1' : 'ml-3 flex-1'}>
+              <Atoms.Container overrideDefaults className={isLastReply ? 'ml-6 min-w-0 flex-1' : 'ml-3 min-w-0 flex-1'}>
                 {nestedReplyIds.map((nestedId, index) => {
                   const isLastNested = index === nestedReplyIds.length - 1 && !hasMoreReplies;
                   return (
--- a/src/components/organisms/ReplyWithNested/ReplyWithNested.tsx
+++ b/src/components/organisms/ReplyWithNested/ReplyWithNested.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import { useEffect, useState } from 'react';
+import * as Libs from '@/libs';
 import * as Atoms from '@/atoms';
 import * as Molecules from '@/molecules';
 import * as Organisms from '@/organisms';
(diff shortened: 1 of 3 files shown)


src/components/organisms/ReplyWithNested/ReplyWithNested.tsx
```diff
@@ -79,7 +79,7 @@ export function ReplyWithNested({
               {!isLastReply && <Atoms.Container overrideDefaults className="w-3 shrink-0 border-l border-border" />}
 
               {/* Indented sub-replies */}
-              <Atoms.Container overrideDefaults className={isLastReply ? 'ml-6 flex-1' : 'ml-3 flex-1'}>
+              <Atoms.Container overrideDefaults className={isLastReply ? 'ml-6 min-w-0 flex-1' : 'ml-3 min-w-0 flex-1'}>
```
infin1t3: Instead of duplicating, why don't you use `cn`?

```
className={cn('min-w-0 flex-1', isLastReply ? 'ml-6' : 'ml-3')}
```

infin1t3 (changes requested):

infin1t3 (approved):
https://github.com/user-attachments/assets/36d97ffe-c3b7-4f8d-a064-1ad8fccf4d39

Tested on `56bbf4aaa83fb7324be2c2c53979b221bcdf251d` - LGTM.

PR #1591, opened 2026-03-26
fix(auth): handle logout when only persisted session export exists

### LINKED ISSUE #1127: [Bug]: Cannot sign out using /logout url
### Bug Description

When navigating to `/logout` the "See you soon!" page is shown, then attempting to got to 'Homepage' or 'Sign back in' takes you back to your feed. You never left!
This should behave the same as clicking 'Sign out' button, it could be a tricky to make you think you've signed out when you haven't.

### Steps to Reproduce

1. Be signed in
2. Navigate to `/logout` in browser url bar
3. Observe "See you soon!' page
4. Click 'Sign back in'
5. Observe you're back at your feed and still have a session

### Affected Area

Authentication / Login

Data not loading (network/cache)

### Frequency

Always (100%)

### Expected Behavior

When navigating to `/logout` we should perform the same session end and local storage cleanup as when clicking 'Sign out' button to avoid any confusion or risk of exploitation.

### Screenshots / Recording

_No response_

### Console Log Output

```shell

```

### Browser & Version

_No response_

### Device

_No response_

### Additional Context

_No response_

Orlandohub:
## Summary
- fix #1127
- Users navigating directly to `/logout` on a fresh page load could not sign out because the live session hadn't been restored yet — only the serialized `sessionExport` existed. The controller now restores the persisted session before performing a real homeserver sign-out.
- The Logout page now shows loading/error/success states with an inline retry instead of silently failing.

## Changes
- `AuthController.logout()` detects when only `sessionExport` is present and calls `restorePersistedSession()` before attempting homeserver logout
- Rewrote `Logout.tsx` to manage a `LogoutViewState` state machine (`idle` → `loading` → `success`/`error`) with proper hydration guards
- Added error state with retry button and homepage navigation using existing `ButtonsNavigation` molecule
- Added `loadingTitle`, `loadingSubtitle`, `errorTitle`, `errorSubtitle`, and `retry` i18n keys across all 9 locale files
- Added component tests (`Logout.test.tsx`) and two new controller test cases for session-restore-before-logout

**Successful Logout**
https://github.com/user-attachments/assets/2153ed1c-58eb-436e-94ee-e13ec055bd83

**Unsuccessful Logout**
https://github.com/user-attachments/assets/cbc19fa7-b769-4b26-ba1a-cce39b94b037

## Test plan
- [ ] Navigate directly to `/logout` while authenticated — should show loading spinner, then success
- [ ] Navigate to `/logout` when already signed out — should show success immediately
- [ ] Simulate network failure during logout — should show error state with Retry button
- [ ] Click Retry from error state — should re-attempt and reach success
- [ ] Click Homepage from error state — should navigate to `/`
- [ ] Run `npx vitest run src/components/templates/Auth/Logout/Logout.test.tsx`
- [ ] Run `npx vitest run src/core/controllers/auth/auth.test.ts`

## Notes
- Non-English locale files have the new keys in English only — translations should be added as a follow-up
- When `restorePersistedSession()` fails, the controller returns early without running local cleanup (the session is already broken; cleanup would double-clear on retry)


src/components/templates/Auth/Logout/Logout.tsx
```diff
@@ -1,23 +1,122 @@
...
 export function Logout() {
-  // Reset the isLoggingOut flag when arriving at the logout page
-  // This ensures the flag is cleared after successful navigation
+  const router = useRouter();
+  const t = useTranslations('logout');
+  const onboardingHasHydrated = Core.useOnboardingStore((state) => state.hasHydrated);
+  const authHasHydrated = Core.useAuthStore((state) => state.hasHydrated);
+  const session = Core.useAuthStore((state) => state.session);
+  const sessionExport = Core.useAuthStore((state) => state.sessionExport);
+  const isLoggingOut = Core.useAuthStore((state) => state.isLoggingOut);
```
Nirzhuk, line 34: why dont get everything with destructuring?

Orlandohub: This is intentional and a Zustand performance best practice. Each selector subscribes to only one slice of the store, so the component only re-renders when that specific value changes.

If you destructured them in a single call:
```
const { authHasHydrated, session, sessionExport, isLoggingOut } = Core.useAuthStore((state) => ({
  authHasHydrated: state.hasHydrated,
  session: state.session,
  sessionExport: state.sessionExport,
  isLoggingOut: state.isLoggingOut,
}));
```
This would create a new object reference on every store update (even unrelated fields), causing the component to re-render on any store change — because {} !== {} by reference.

Nirzhuk (approved):

PR #1608, opened 2026-03-31
fix(notifications): resolve pubky mentions to user names in notification previews (#487)

### LINKED ISSUE #487: [Bug]: Notification with post content should parse user pubky
### Describe the bug

When a notification previews beginning of post content, if that is a user's pubky then the pubky value is displayed. It should parse user's pubky and show their user name, as we do for posts.

### Reproduce

1. With a reply made to you quoting someone
2. View own notifications on profile page
3. Observe pubky value is displayed instead of user's profile name

### Screenshots / Recording

Image

### Web Browser

_No response_

### Pubky App Version

45dec8800547a4b8163387998e6f88db90441f8d

### Console log output

```shell

```

Taewa:
## Description
 - Add `resolvePubkyToNames` helper that replaces `pk:/pubky` in notification post previews with display names
  - Resolve mentions async when loading notification post content in NotificationItem
  - Add unit tests for `resolvePubkyToNames` (empty input, both prefixes, multiple mentions, deduplication, fallback on failure)
  - Extract shared mention regex source (`PUBKY_IDENTIFIER_WITH_PREFIX_SOURCE`) to Identity class, reuse in `remarkMentions`

## Result
Screenshot 2026-03-31 at 12 25 13

### DIFF
--- a/src/components/molecules/PostText/PostText.utils.ts
+++ b/src/components/molecules/PostText/PostText.utils.ts
@@ -1,6 +1,7 @@
 import type { Root, Paragraph, Text, Link, PhrasingContent, Parent, RootContent } from 'mdast';
 import { ReactNode } from 'react';
 import { visit } from 'unist-util-visit';
+import { Identity } from '@/libs';
 import { TRUNCATION_LIMIT } from './PostText.constants';
 
 // We assign full code blocks without a language specified as plaintext (ex. ```...```)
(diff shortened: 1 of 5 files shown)


secondl1ght (commented):
@Taewa just one thing, can we prefix the mentions with `@` please? This is how they are displayed in posts, and makes it clear that the text following is a username.

Taewa:
Added `@` on username.

Screenshot 2026-04-03 at 11 16 11

secondl1ght (approved):

PR #1622, opened 2026-04-01
tests: fix e2e

catch-21:
Update E2E tests inline with recent changes in app. Also enabled 'can post from new post'.


catch-21:
E2E tests passing: https://github.com/pubky/pubky-app/actions/runs/23854850592 (feed tests are insisting on running with broken nexus https://github.com/pubky/pubky-nexus/issues/592, these tests pass locally for me)

PR #1624, opened 2026-04-01
tests: fix failing feed test in CI

catch-21:
Mostly hardening tests around virtuoso, which can cause posts to more aggressively not be mounted in the DOM as they scroll out of view.


catch-21:
All E2E tests passing: https://github.com/pubky/pubky-app/actions/runs/23870783405/job/69602465001

PR #1625, opened 2026-04-01
fix(users): invalidate recommended stream cache on follow

infin1t3:
Clear the recommended:all:all user stream from IndexedDB when a user is followed so the Who to Follow card shows fresh suggestions on next mount.

Made-with: Cursor


infin1t3:
@greptileai review

infin1t3:
Will be implemented in some of the next sprints, closing for now until we finish barrels.

PR #1630, opened 2026-04-02
fix(twitter-video): temporary workaround for twitter video play

### LINKED ISSUE #1171: [Bug]: Embedded X (twitter) videos don't play
### Bug Description

Links to X posts that contain a video cannot be played. The video request is always 403.

### Steps to Reproduce

1. Create a post with X video link, e.g. https://x.com/itsolelehmann/status/2018311011678167166?s=20
2. Click on play button for video
3. Observe it loads and then doesn't play

### Affected Area

Post (create/edit/delete)

None

### Frequency

Always (100%)

### Expected Behavior

Embedded X videos can be played

### Screenshots / Recording

https://github.com/user-attachments/assets/59f85f04-655f-463e-9124-de93c5b8be8b

Image

### Console Log Output

```shell

```

### Browser & Version

_No response_

### Device

_No response_

### Additional Context

_No response_

Taewa:
## NOTE: This is a temporary workaround for [vercel/react-tweet#212](https://github.com/vercel/react-tweet/issues/212), not an official fix. Please review the pros & cons below before deciding whether to merge. We should remove this patch once an upstream fix is released
---------------- 
### The Problem
  Twitter's CDN (`video.twimg.com`) started blocking video requests that include a `Referer` header from external sites,
  returning **403 Forbidden**. This breaks all embedded tweet videos via `react-tweet`.

  This is a known issue: [vercel/react-tweet#212](https://github.com/vercel/react-tweet/issues/212)

  ### Our Fix
  We patched `react-tweet` (via `patch-package`) to change the video play behavior:

  1. User sees the poster image (thumbnail) — **no download happens**
  2. User clicks play → `fetch(videoUrl, { referrerPolicy: 'no-referrer' })` downloads the video **without** a `Referer`
  header
  3. The response is converted to a local blob URL
  4. The `<video>` element plays from the blob URL

  The patch lives in `patches/react-tweet+3.3.0.patch` and is auto-applied on `npm install` via `patch-package`.

  ### Pros
  - **Works** — videos play without 403 errors
  - **Targeted** — only affects Twitter video requests, no side effects on other parts of the app
  - **No wasted data** — video is only downloaded when the user clicks play, not on page load
  - **Reproducible** — `patch-package` ensures every developer and CI gets the fix
  - **Minimal change** — only the video component is patched, all other `react-tweet` behavior is untouched

  ### Cons
  - **No streaming** — downloads the entire video before playback starts (noticeable delay on longer videos)
  - **Memory usage** — full video is held in browser memory as a blob while the component is mounted
  - **Maintenance** — patch must be updated if `react-tweet` upgrades

  ### Security Hardening
  The following measures were added to address resource abuse, memory leaks, and error handling:

  - **`res.ok` check** — Non-200 responses (e.g. 403 HTML error pages) are rejected immediately instead of being converted to
  a blob and fed to the video player
  - **`AbortController`** — In-flight fetches are cancelled on component unmount or if the user re-clicks, preventing wasted
  bandwidth and orphaned requests
  - **Blob URL cleanup** — Previous blob URLs are revoked via `URL.revokeObjectURL()` before creating new ones, preventing
  memory leaks on re-renders or retries
  - **Trusted URL source** — Video URLs originate from Twitter's syndication API (resolved internally by `react-tweet`), not
  from user input, limiting the attack surface for malicious URLs
  - **CORS enforcement** — `fetch()` uses CORS mode by default, meaning only servers that explicitly allow cross-origin access
   can be fetched

  ### Alternatives Considered
  | Approach | Why we didn't use it |
  |---|---|
  | `referrerPolicy` on `<video>` | Browser ignores it — not a supported attribute on video elements |
  | `Referrer-Policy` header in Next.js config | Works, but strips referer from **all** requests on the page, not just Twitter
   videos |
  | Proxy via API route | Most robust, but adds server load and latency |
  | Preload on mount via `useEffect` | Works, but wastes bandwidth on infinite scroll pages where users may never click play |
  | Wait for upstream fix | Issue is open, no merged PR yet |

 ### Result

https://github.com/user-attachments/assets/df0ff8cc-2b3e-4556-b95e-733e7f595ac3


Taewa:
Closed due to https://github.com/pubky/pubky-app/pull/1633

Issue #1603, opened 2026-03-27
[Bug]: Disabled deleted notifications still showing


SeverinAlexB:
### Description

I disabled the deleted post notifications but I still get them. Example: https://pubky.app/post/ihgjy51fdnaingcp8rum1omfzd6p8bhm7usune41grd97dho5cwy/0034VMC9CJAQ0

I think this was a reply to one of my posts

### Expected Behaviour

Should not get any deleted post notifications anymore

### Steps to Reproduce

1. Disable deleted notifications
2. Other user deletes post

### Screenshots

<img width="1660" height="1338" alt="Image" src="https://github.com/user-attachments/assets/58520675-2cdf-4ea3-92c7-7b2421e8ec43" />

<img width="2068" height="960" alt="Image" src="https://github.com/user-attachments/assets/e6765bae-b611-4304-9a61-2832884a77c6" />

### Console Log

```shell

```

catch-21:
Thanks. This appears to not be applied when toggling settings or on settings recovery on sign in.
I'll add a long overdue E2E test for this too.

catch-21:
skipped E2E test added in https://github.com/pubky/pubky-app/pull/1639/changes/c3bf26ae3f798b2d205a64170fb5f18af08b1807

catch-21:
Enabled E2E test in https://github.com/pubky/pubky-app/pull/1790

catch-21:
Verified fix in c897b64c9448fdb73542c54c73aade12429f11f7

PR #1639, opened 2026-04-03
tests: add 'being mentioned' notification e2e tests

catch-21:
This PR introduces tests for both enabled and disabled notification settings when profile is mentioned.
Disabled is skipped due to bug https://github.com/pubky/pubky-app/issues/1603


Taewa (approved):
LGTM

PR #1644, opened 2026-04-06
Idea: add dynamic variables

SpontaneousOverthrow:
Problem: currently we have dynamic variables, like `NEXT_PUBLIC_DEFAULT_HTTP_RELAY`, `NEXT_PUBLIC_HOMEGATE_URL` and etc, they are different across all envs(staging, prod, testnet). That's why we need to have several docker images for each of the envs.

This is a draft for one of the possible solutions, we use `mock` values when we building the image, and use `entrypoint` script to change this values to the correct ones based on the env.

Wanna hear your thoughts on that solution, after the discussion I will implement the one which we all agree on.


src/libs/env/env.ts
```diff
@@ -374,17 +366,8 @@ function parseEnv(): z.infer<typeof envSchema> {
...
-    }
-    // Validate each relay is a valid URL
-    for (const relay of relays) {
-      if (typeof relay !== 'string') {
-        throw new Error('Each relay must be a string');
-      }
-      new URL(relay);
-    }
+    const relays = [];
+    relays.push(val);
```
catch-21, line 370: this change appears breaks the format, see failing 'should match test config value' test

src/libs/env/env.ts
```diff
@@ -36,9 +36,9 @@ const envSchema = z.object({
   // For production, override these with your production URLs.
 
   /** Main API endpoint */
-  NEXT_PUBLIC_NEXUS_URL: z.string().url().default('https://nexus.staging.pubky.app'),
+  NEXT_PUBLIC_NEXUS_URL: z.string().default('https://nexus.staging.pubky.app'),
```
catch-21, line 39: Removal of `url()` format checks conflicts with the `docs/environments.md`, if we do make this change then need to update expectations there.

src/libs/env/env.ts
SpontaneousOverthrow: yes, also this solution support only only relay
this is a dirty solution to check if the approach is working. I didn't create the clean one, because currently not sure if this approach(use the mock variables and after replace them with the actual ones on container start) is the correct one.

SpontaneousOverthrow: we need to research the possible solutions and came up with the one we will use

src/libs/env/env.ts
catch-21: could try `NEXT_PUBLIC_NEXUS_URL: z.ipv4().or(z.url()).default('https://nexus.staging.pubky.app')`

PR #1646, opened 2026-04-06
fix(timeline): hide soft-deleted posts from main feed

### LINKED ISSUE #1541: [Bug]: Soft-deleted post stub visible in main feed
### Description

When a post is soft-deleted (i.e. deleted but it has replies, reposts, or tags), a "This post has been deleted by its author" stub is shown in the main feed timeline. This stub is intended only for reply thread views where the deleted parent provides context for child replies — it shouldn't appear as a standalone card in the main feed.

Related: #667 (comment by @BitcoinErrorLog with screenshot)

### Expected Behaviour

Soft-deleted posts should be completely hidden in the main feed (timeline). The "This post has been deleted by its author" stub should only appear in reply thread views where it serves as context for replies underneath it.

### Steps to Reproduce

1. Create a post
2. Have another user reply to it, repost it, or tag it (so it becomes "linked")
3. Delete the post — it is now soft-deleted because it has connections
4. Observe the main feed: the deleted post appears as a stub card saying "This post has been deleted by its author"

infin1t3:
Extract `TimelinePostItem` so `usePostDetails` runs at component level (not inside Virtuoso `itemContent`). Gated rows: render nothing until local post details exist, then hide rows whose content is `[DELETED]`.

Unit tests cover deleted vs loading (no `postDetails`) vs visible posts.

Resolves: #1541

### DIFF
--- a/src/components/organisms/Timeline/Posts/Posts.tsx
+++ b/src/components/organisms/Timeline/Posts/Posts.tsx
@@ -28,9 +28,9 @@ const virtuosoComponents = { Footer: TimelineVirtuosoFooter };
 /**
  * Extracted from Virtuoso's `itemContent` callback so hooks are valid: hooks must run
  * only at the top level of a React component (or custom hook), not inside plain
- * render callbacks. Here we need `usePostDetails` / `usePostNavigation` to hide
- * soft-deleted feed items once details resolve (stream filter is fail-open when
- * details are not cached yet).
+ * render callbacks. `usePostDetails` gates each row: render nothing until details exist,
+ * then hide soft-deleted items (`[DELETED]`). The stream list itself can still contain
+ * ids before details are cached; rows stay empty until local-first details resolve.
  */
 function TimelinePostItem({ postId, tagsLayout }: { postId: string; tagsLayout?: TagsLayout }) {
   const { navigateToPost } = Hooks.usePostNavigation();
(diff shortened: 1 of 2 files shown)


src/components/organisms/Timeline/Posts/Posts.tsx
```diff
@@ -24,6 +25,33 @@ interface TimelinePostsProps {
...
+ * render callbacks. Here we need `usePostDetails` / `usePostNavigation` to hide
+ * soft-deleted feed items once details resolve (stream filter is fail-open when
+ * details are not cached yet).
+ */
+function TimelinePostItem({ postId, tagsLayout }: { postId: string; tagsLayout?: TagsLayout }) {
+  const { navigateToPost } = Hooks.usePostNavigation();
+  const { postDetails } = Hooks.usePostDetails(postId);
+  const isDeleted = Libs.isPostDeleted(postDetails?.content);
+
+  if (isDeleted) return null;
```
secondl1ght:
```suggestion
  if (!postDetails || isDeleted) return null;
```

Otherwise the component will start rendering but then be removed if it is found to be deleted after.

secondl1ght (approved):

secondl1ght:
@catch-21 FYI these changes were reverted after this was merged due to a bug that was found. The new PR with #1656 handles the issue much better anyway, because it filters deleted posts closer to the source.

PR #1656, opened 2026-04-07
fix(timeline): filter deleted posts upstream to avoid Virtuoso zero-size warnings

infin1t3:
Adding a second-pass filter in the Controller after the cache-miss fetch to close this gap at the source

https://github.com/user-attachments/assets/c3563a3f-1709-4035-b272-c8da431b9b1c

Resolves: #1646


secondl1ght (changes requested):
## Deleted-post filtering: move upstream into the Controller

### Issues

**1. Unnecessary re-trigger after removal**
When the effect finds deleted posts and calls `removePosts`, that updates `rawPostIds` state, which re-triggers the same effect. The second run finds nothing and stops — so it's not infinite — but it's a wasted async round-trip to IndexedDB every time a deletion is found.

**2. Full-array rescan on every pagination**
The effect depends on `rawPostIds`, which changes every time `loadMore` appends new posts. Each time, `filterDeletedPosts` is called on **all** accumulated IDs, not just the new page. This gets increasingly expensive as the user scrolls.

**3. Brief UI flash of deleted posts**
Between the initial render and the async `filterDeletedPosts` resolution, deleted posts are visible. `PostMain` still has its own `isDeleted` check so it renders a `<PostDeleted />` placeholder rather than full content — but users will see that placeholder card appear and then disappear as `removePosts` removes it from the list.

### Proposed alternative

The pipeline already filters deleted posts during `postStreamQueue.collect()`, but uses fail-open semantics (posts without cached details are kept). Right after that, `StreamPostsController.getOrFetchStreamSlice` calls `fetchMissingPostsFromNexus` to resolve those cache misses — and then immediately returns `nextPageIds` **without re-checking**. That's the gap.

Adding a second-pass filter in the Controller after the cache-miss fetch closes this gap at the source:

```ts
    if (cacheMissPostIds.length > 0) {
      await Core.PostStreamApplication.fetchMissingPostsFromNexus({
        cacheMissPostIds,
        viewerId,
      });
      // Second-pass: cache-miss details are now resolved,
      // re-filter to catch posts that were fail-open in the first pass
      const validIds = await Core.PostStreamApplication.filterDeletedPosts(nextPageIds);
      return { nextPageIds: validIds, timestamp, reachedEnd };
    }
    return { nextPageIds, timestamp, reachedEnd };
```

This eliminates all three issues: it runs once per fetch (not per render), only checks the current page (not all accumulated IDs), and deleted posts never reach the UI. It also keeps filtering logic in the Controller layer rather than leaking it into the component.

The `useEffect` in `TimelineFeedContent` and its associated tests can then be removed. Real-time deletions (post deleted while the user is viewing the timeline) are already handled by `PostMain`'s live Dexie query → `isPostDeleted` → `<PostDeleted />`, which is the right layer for that.

secondl1ght (approved):

PR #1662, opened 2026-04-08
fix(timeline): auto-paginate visual feed through non-media pages

### LINKED ISSUE #1661: [Bug]: Visual layout feed stalls with empty grid when stream pages contain no media posts
### Description

When the homepage is set to **Recent** sort + **Visual** layout (with content filter `all`), the feed shows an empty grid and never loads posts. The visual grid only renders tiles for posts with image/video attachments, but the underlying stream (`timeline:all:all`) returns all post types. When a page of posts has no media content:

1. `useVisualFeedTiles` produces 0 tiles → 0 grid rows
2. `react-virtuoso`'s `endReached` callback never fires because the data array length didn't change (or is 0)
3. `hasMore` remains `true`, so the "no posts" empty state doesn't appear
4. The feed is stuck: no loading indicator, no empty state, no way to load more

This also affects **mid-scroll pagination**: if the user scrolls to the bottom and `loadMore` fetches a page of text-only posts, no new rows are added. Virtuoso won't fire `endReached` again for the same data length, causing the feed to stall until the user switches to columns layout, loads more posts there, and switches back.

### Expected Behaviour

The visual feed should automatically paginate through pages that contain no visual media (text, links, files) until either:
- Posts with image/video attachments are found and rendered as tiles
- The stream is fully exhausted, at which point the filtered empty state should display

### Steps to Reproduce

1. Go to the homepage
2. Set sort to **Recent** and layout to **Visual**
3. If the most recent posts in the stream are text-only (no images/videos), observe an empty feed with no loading or empty state
4. Alternatively: scroll to the bottom of a working visual feed until `loadMore` fetches a page that is entirely text posts — the feed stops loading more

Orlandohub:
## Summary
- fix #1661
- Visual layout feed stalls when stream pages contain only text/link posts because Virtuoso's `endReached` never re-fires for unchanged data length
- Affects both the initial load (empty grid) and mid-scroll pagination (feed stops loading at the bottom)

## Changes
- Add a `stableRowCountRef` in `VisualTimelinePosts` that tracks the grid row count at each idle state
- Add an effect that auto-triggers `loadMore` when a completed fetch didn't produce new visual rows and the stream has more pages
- Reset the ref on stream/filter changes (`loading` becomes true) to prevent stale comparisons

## Test plan
- [x] Set sort to **Recent**, layout to **Visual**, content to **All** — feed should load and display media posts even if the first pages are text-only
- [x] Scroll to the bottom of the visual feed — if the next page has no media, the feed should auto-load subsequent pages until media appears or the stream ends
- [x] Switch content filter or sort — verify the feed resets and loads correctly
- [x] Verify the filtered empty state still shows when the entire stream has no media posts

## Notes
- The root cause is that `react-virtuoso` only fires `endReached` when the user scrolls past rendered items; with 0 new rows the callback never re-triggers
- The fix is conservative: it only fires when idle, non-errored, with no pending tile probes, bounded by the stream's own `hasMore` / `reachedEnd` signal

### DIFF
--- a/src/components/organisms/Timeline/Feed/TimelineFeed/VisualTimelinePosts.tsx
+++ b/src/components/organisms/Timeline/Feed/TimelineFeed/VisualTimelinePosts.tsx
@@ -302,6 +302,30 @@ export function VisualTimelinePosts({
   const { navigateToPost } = Hooks.usePostNavigation();
   const { rows, hasPendingTiles } = useVisualFeedTiles({ postIds, hasMore });
 
+  // The visual grid only renders posts with image/video attachments. When a page of
+  // posts contains no media, the row count stays unchanged and Virtuoso won't fire
+  // endReached again (the data length didn't change). Track the row count from the
+  // last stable state so we can auto-paginate through media-less pages until visual
+  // tiles appear or the stream is exhausted.
+  const stableRowCountRef = React.useRef(0);
+
+  React.useEffect(() => {
+    if (loading) {
+      stableRowCountRef.current = 0;
+      return;
+    }
+
+    if (loadingMore || error || !hasMore || hasPendingTiles || postIds.length === 0) {
+      return;
+    }
+
+    if (rows.length <= stableRowCountRef.current) {
+      void loadMore();
+    }
+
+    stableRowCountRef.current = rows.length;
+  }, [loading, loadingMore, error, hasMore, postIds.length, rows.length, hasPendingTiles, loadMore]);
+
   const showFilteredEmptyState =
     !loading && !error && postIds.length > 0 && rows.length === 0 && !hasMore && !loadingMore && !hasPendingTiles;
 


catch-21:
Verified in 4d94943ddc4358fea4c6b48dcd91cf543d2a4c05

Loads first time on staging and prod for me. Proof of prod with and without muted users:

https://github.com/user-attachments/assets/c95a47c9-e8ab-4270-9e1f-10c66b8a720b

infin1t3 (approved):
https://github.com/user-attachments/assets/13b93d23-921f-422e-87dd-d9392e2145f7

Tested at `4d94943ddc4358fea4c6b48dcd91cf543d2a4c05` - LGTM.

catch-21:
It can work on global feed although I found two new bugs:
1. When results is 3 or less, none of them show (e.g. with a fresh DB as is used in E2E tests, or when applying a tag filter yeilding 3 results). I think I have a fix for it in https://github.com/pubky/pubky-app/pull/1674
2. Often the global feed shows only 2 results when there are more, appears to impact "All" content filter often and "Image" filter less often. Can be reproduced by reloading the page, no cache clear required.

PR #1671, opened 2026-04-09
Fix/visual feed pagination settlement

infin1t3:


catch-21:
Unfortunately, doesn't fix the bug and E2E test still fails.  However, this approach  works for me and E2E test passes: https://github.com/pubky/pubky-app/pull/1674

PR #1674, opened 2026-04-09
fix: visual layout not showing all posts

### LINKED ISSUE #1677: [Bug]: Visual layout often only shows 2 posts
### Description

When viewing the Visual layout 'global' feed (with no tag filters and Content "All") I often only see the first two posts and no more load. Switching to Content "Images" correctly shows more posts. Adjusting the height of the page triggers more posts to load as they should have done initially.

I believe this is due to Virtusos not firing endReached on initial render as is done for loadMore.

### Expected Behaviour

Visual layout always shows all posts with images and videos on initial load.

### Steps to Reproduce

1. On staging, reload the page or fresh sign in
2. Click 'Visual' Layout
3. Observe only two posts are shown in a single row and no more posts load

### Screenshots

https://github.com/user-attachments/assets/baca3c5f-702b-4eeb-9329-1f852a4791ea

https://github.com/user-attachments/assets/d485e004-fee3-49d8-9ae0-a38fc4280d44

### Console Log

```shell

```

### Browser & Version

_No response_

### LINKED ISSUE #1678: [Bug]: Visual layout can show no results when tag filter is applied
### Description

When viewing a Visual layout in combination with a tag filer that yeilds few results (3 or less), OR, when using a fresh backend stack with emtpy homeserver and nexus databases (as is done with E2E testing) and viewing the global feed with Visual layout, then a blank page with no posts is displayed.

I believe this is due to `VISUAL_PENDING_TAIL_LIMIT = 3` , where results are 3 or less it doesn't show any at all. But we introduced this for optimisation so let's find a way to keep it.

### Expected Behaviour

Visual layout is capable of showing less than 4 results.

### Steps to Reproduce

1. On staging
2. Create a custom feed for visual layout and `zyx` tag
3. View custom feed results (it may well show the correct posts)
4. Return to global feed
5. Reload browser with cleared cache (similar to a fresh sign in)
6. View custom feed
7. Observe empty page and no posts

It can also be reproduced by running E2E test 'can view image-only posts in visual layout' in CI or with clean local DB.

### Screenshots

https://github.com/user-attachments/assets/cdad9d34-cc80-44cc-9b54-a5ea4545b3ca

### Console Log

```shell

```

### Browser & Version

_No response_

catch-21:
Fixes https://github.com/pubky/pubky-app/issues/1677 & https://github.com/pubky/pubky-app/issues/1678

This PR addresses the above bugs by ensuring that pending size category for each image are waited on and that it never fails to assign a fallback size. Also that loading of more posts is triggered on initial load, not just when resizing the viewport.

### DIFF
--- a/src/components/organisms/Timeline/Feed/TimelineFeed/useVisualFeedTiles.ts
+++ b/src/components/organisms/Timeline/Feed/TimelineFeed/useVisualFeedTiles.ts
@@ -279,17 +279,21 @@ export function useVisualFeedTiles({ postIds, hasMore }: { postIds: string[]; ha
   }, [missingFileUris, missingFileUrisKey]);
 
   const tiles = (snapshot?.tiles ?? []).map(resolveTileProbeState);
-  const pendingOverflowFallbackIds = React.useMemo(() => getVisualPendingOverflowFallbackIds(tiles), [tiles]);
+  const pendingOverflowFallbackIds = React.useMemo(
+    () => getVisualPendingOverflowFallbackIds(tiles, hasMore ? undefined : 0),
+    [tiles, hasMore],
+  );
   const pendingOverflowFallbackIdSet = React.useMemo(
     () => new Set(pendingOverflowFallbackIds),
     [pendingOverflowFallbackIds],
   );
 
   React.useEffect(() => {
+    if (!hasMore) return;
     pendingOverflowFallbackIds.forEach((tileId) => {
       setVisualTilePreferredSizeFallback(tileId, 'medium');
     });
-  }, [pendingOverflowFallbackIds]);
+  }, [pendingOverflowFallbackIds, hasMore]);
 
   const stabilizedTiles = React.useMemo(() => {
     return tiles.map((tile) => {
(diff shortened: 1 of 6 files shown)


catch-21:
https://github.com/pubky/pubky-app/actions/runs/24189409761/job/70602947317
All 'feed' E2E tests now passing. The failing 'posts' tests were already failing and require seprate investigation.

src/components/organisms/Timeline/Feed/TimelineFeed/useVisualFeedTiles.ts
```diff
@@ -279,17 +279,21 @@ export function useVisualFeedTiles({ postIds, hasMore }: { postIds: string[]; ha
   }, [missingFileUris, missingFileUrisKey]);
 
   const tiles = (snapshot?.tiles ?? []).map(resolveTileProbeState);
-  const pendingOverflowFallbackIds = React.useMemo(() => getVisualPendingOverflowFallbackIds(tiles), [tiles]);
+  const pendingOverflowFallbackIds = React.useMemo(
+    () => getVisualPendingOverflowFallbackIds(tiles, hasMore ? undefined : 0),
```
Orlandohub, line 283: **Improvement opportunity:**

You don't need those useMemo calls anymore. The React Compiler automatically memoizes values and dependencies, so hand-written useMemo is redundant here.

That said, leaving the existing useMemo calls in place is also harmless — the compiler treats them as hints and respects them. So this is a cleanup opportunity, not a correctness issue. If you're touching this file anyway it's worth simplifying, but there's no urgency to go remove every useMemo across the codebase.

Orlandohub (approved):
LGTM

src/components/organisms/Timeline/Feed/TimelineFeed/useVisualFeedTiles.ts
catch-21: Noted. I'm going to leave as is for now for simplicity. Created task to track your feedback https://github.com/pubky/pubky-app/issues/1687

PR #1675, opened 2026-04-09
fix(ui): wide layout does not match design

### LINKED ISSUE #1628: [Bug]: UI: Wide Layout implementation does not match design
### Description

The current Wide Layout implementation for posts and feed does not match the design. It currently uses the incorrect spacing, font size, and size of elements. 

Image

### Expected Behaviour

Wide layout is designed to provide a more immersive experience for content consumption. The visual appearance and feel of content in a Wide layout post differ intentionally from those in a Column layout post. This difference is achieved through varying spacing, font size, and avatar size, which create a breathing effect and shorten sentences, enhancing readability.

**Wide layout posts:**
- avatar should be 64px * 64px
- username should be text-2xl 
- paragraphs should be text-xl
- spacing for the both left and right side of the card should be spacing 12 (48px)
- automatically making the space between left and right side of card 2 * 48px = 96px.

Image

### Steps to Reproduce

1. Go to pubky.app feed
2. Click on 'Wide' layout option in sidebar
3. Compare the layout to the design in Figma

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

Chrome

Orlandohub:
## Summary

- Closes #1628
- Wide-layout posts now match the Figma spec end-to-end: 64px avatars, `text-2xl` usernames, `text-xl` body copy, and 48px (`p-12`) horizontal card padding
- Wide layout cascades into nested replies, the single-post page, **and the inline post / reply input boxes** so every post surface stays consistent
- Column layout, mobile layout, and nested previews remain unaffected

### Wide-layout design specs

- Thread `textClassName` prop through `PostContent` → `PostContentBase` → `PostText` so wide-mode surfaces can override the default body text size
- Pass `size="large"` to `PostHeader` and `textClassName="text-xl leading-7"` to `PostContent` in wide mode from both `PostMain` and `SinglePostCard`
- Apply `p-12` padding on `CardContent` in wide mode (vs default `p-6`)
- Align `PostHeaderUserInfo` typography with the spec (`text-2xl leading-8` / `gap-4` for `large`)
- Split wide post layouts into separate left and right sections so the centre spacing matches the design without changing mobile or columns behaviour

### Reply-tree alignment via shared context

`PostMain` previously read `tagsLayout` from a prop that was prop-drilled through the timeline, while the reply tree below it (rendered recursively via `ThreadTree` → `ReplyWithNested` → `PostMain`) had no way to know which layout the surface was in. Root posts in the feed rendered correctly in wide mode, but their nested replies fell back to the inline layout — visible mismatch on the same screen.

To fix this end-to-end without re-introducing prop-drilling through a recursive component tree, the PR introduces a single source of truth at the surface entry point:

- New `PostMainLayoutProvider` / `usePostMainLayout` context in `src/components/organisms/PostMain/PostMainLayout.tsx`. The surface (`SinglePostContent` for the single-post page, `TimelineFeedContent` for every feed variant) wraps its subtree once with the provider; `PostMain` and `SinglePostCard` consume the layout via context.
- New `WIDE_POST_LAYOUT_CLASSES` constant exported from the same module — `shell`, `leftColumn`, `rightColumn`, and `bodyText` are now defined in one place and consumed by both `PostMain` and `SinglePostCard`. This keeps the wide-layout magic numbers (`lg:w-96`, `lg:p-12`, `text-xl leading-7`, etc.) from drifting between the two surfaces. (Interim seam — a real `WidePostLayout` component will land in the upcoming post-component unification pass.)
- Direct path-mapped imports `@/organisms/PostMain/PostMainLayout` (not the `@/organisms` barrel) — keeps the no-barrel-reexports rule honoured and avoids a `PostMain → @/organisms → PostMain` import cycle.
- Helper `getTagsLayoutForSurfaceLayout(layout)` replaces the duplicated `effectiveLayout === LAYOUT.WIDE ? 'side' : 'inline'` mapping that was inlined six times in `TimelineFeed.tsx`.

### Wide layout for input boxes (PostInput + QuickReply)

The same provider now also covers the inline input boxes on every feed surface, so the "What's on your mind?" composer at the top of the feed and the QuickReply at the end of every reply thread match the post-card spec in wide mode:

- Lift the `PostMainLayoutProvider` in `TimelineFeedContent` to wrap the entire feed content (`children` slot + `TimelinePosts`) so the `PostInput` rendered as `TimelineFeed` children inherits the active tags layout. `QuickReply` already lives below the provider via `ThreadTree`.
- Both inputs read `usePostMainLayout()` and `useIsMobile()` to derive the wide flag, mirroring `PostMain`'s cascade — provider value, mobile fallback to inline.
- In wide mode: outer dashed-border container switches to `p-12`, the avatar uses `size="xl"` (64px), `PostHeader` is rendered at `size="large"`, and the textarea picks up `text-xl leading-7` via the shared `WIDE_POST_LAYOUT_CLASSES.bodyText` constant — no new magic numbers.
- Dialog-mounted PostInputs (`DialogNewPost`, `DialogReply`, `ShareTarget`) are intentionally untouched: they live outside the surface provider and keep their existing dialog-scaled spacing.

### Cleanup that fell out of the refactor

- Drop the `tagsLayout` prop from `PostMain`, `TimelinePosts`, and `TimelinePostReplies` — context-only flow, single source of truth
- Remove dead code in `SinglePostCard`'s inline branch: orphaned `<PostTagsPanel>` (with its dangling `desktopTagsPanelRef` binding), redundant 1-column grid wrapper, and five unreachable `isWideLayout ? : ` ternaries that the previous wide-vs-columns split had left behind
- Consolidate the wide repost-header spacing from two siblings (`px-12 pt-12` on the wrapper + `pt-6` on the body) into a single wrapper (`px-12 pt-12 pb-6`) so the gap intent lives in one place

## Test plan

- All affected component tests pass — 438 tests across 27 files (PostMain, SinglePostCard, SinglePostContent, ReplyWithNested, TimelinePosts, TimelinePostReplies, TimelineFeed, TimelineFeedContent, PostContent, PostContentBase, PostText, PostInput, QuickReply, DialogReply, plus all Feed templates)
- `PostMain.test.tsx` adds a regression test for the wide + repost-header combination (`px-12 pt-12 pb-6` wrapper renders when `showRepostHeader && tagsLayout === 'side'`)
- `PostMain.test.tsx` and `SinglePostCard.test.tsx` switch all side-layout assertions from prop drilling to wrapping in `PostMainLayoutProvider`
- `ReplyWithNested.test.tsx` adds a regression test confirming the entire nested reply subtree inherits `tagsLayout="side"` from the thread context
- `SinglePostCard.test.tsx` adds a `falls back to inline layout when rendered without a PostMainLayoutProvider` regression
- `PostInput.test.tsx` and `QuickReply.test.tsx` add three tests each for the wide-layout cascade: no-provider fallback (inline), wrapped in `tagsLayout="side"` (wide padding, large header / xl avatar, `text-xl leading-7` textarea), and mobile-with-side-provider (still falls back to inline)
- Snapshots updated where the lifted provider changed nesting; rendered DOM is otherwise unchanged
- Lint, format, and Prettier checks all pass; no new `tsc --noEmit` errors introduced (pre-existing test-mock typing issues are unrelated)

## Notes

- `textClassName` is intentionally not forwarded to article or blurred-content branches — articles have their own heading typography system
- The `xl` avatar size (`h-16 w-16` = 64px) already existed in the Avatar variants; this PR just activates it in wide mode
- The duplication between `PostMain`'s and `SinglePostCard`'s wide JSX is now scoped to a single class constant; full component extraction (a `WidePostLayout` shared shell) is deferred to the broader post-component unification follow-up


aldertnl (dismissed):
LGTM. Maybe we can fix the padding and styling for post creation input field and replies in a separate PR

Image

Image

Orlandohub:
@greptile-apps review

aldertnl (dismissed):
LGTM

infin1t3 (approved):
Given Aldert already confirmed the visual part, I just focused on the code. LGTM - approved.

PR #1683, opened 2026-04-10
fix: use translation instead of hardcoded loading string

secondl1ght:
### Summary

Replaced the hardcoded `'Loading...'` default value for the `loadingText` prop in the `InputField` molecule with a proper `next-intl` translation using the existing `common.loading` key from `en.json`.

### Changes

- **`src/components/molecules/InputField/InputField.tsx`**
  - Added `'use client'` directive (required for the `useTranslations` hook, consistent with other molecules)
  - Imported `useTranslations` from `next-intl`
  - Removed hardcoded `'Loading...'` default from the `loadingText` prop
  - Added `useTranslations('common')` hook and resolved loading text via `loadingText ?? t('loading')`
  - Callers can still override `loadingText` explicitly if needed

### No Test Changes Required ✅

The global `next-intl` mock in `src/config/test.ts` already uses the actual `en.json` messages, so all 12 existing tests pass without modification. Snapshots updated cleanly since the translated value matches the previous hardcoded string.

---

### Related

- Follow-up from #1584.

### Screenshots

- _N/A — no visual changes_

### Notes

- _Add any additional notes here_

---


Orlandohub (approved):
LGTM

PR #1686, opened 2026-04-10
fix(ui): enable Visual layout mode on Search results page

### LINKED ISSUE #1598: [Bug]: Visual Mode missing from Search returns
### Description

User has no way to actively filter Visual mode

### Expected Behaviour

1. When viewing Visual mode, and using the Search Bar, I should be shown a filtered view in the same mode.
2. Instead, I am taken to the Search return page, in Columns view, where the Visual link is missing from the sidebar.

User should be able to actively filter the Visual feed using tags.

User should be able to convert a Search Return from Column mode to Visual Mode while keeping the current tag filters intact, by clicking Visual in the sidebar.

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

Chrome MacOS

Orlandohub:
Supersedes #1657 (same changes; head branch renamed to `bug-ui/1598-visual-mode-search-returns`).

## Summary

Fixes #1598 — Visual mode was missing from the Search results page. Users could not filter search results using Visual layout, and navigating from Visual mode on Home to Search would silently fall back to Columns.

### Root cause

Three blocking points prevented Visual mode on Search:

1. `TIMELINE_FEED_VARIANT.SEARCH` was not in `VISUAL_SUPPORTED_FEED_VARIANTS`, so `resolveFeedLayout` always fell back to Columns
2. The Search template did not pass `feedVariant` to `ContentLayout`, so the wide shell layout never activated
3. The Search template did not pass `allowVisualLayout` to sidebar/drawer components, so the Visual tab was hidden

### Fix

* **`useFeedLayoutResolution.ts`** — Added `SEARCH` to `VISUAL_SUPPORTED_FEED_VARIANTS`
* **`Search.tsx`** — Pass `feedVariant={TIMELINE_FEED_VARIANT.SEARCH}` to `ContentLayout` and `allowVisualLayout` to all three sidebar/drawer components
* **`TimelineFeed.tsx`** — Added visual content resolution + sync in `SearchTimelineFeed` (same pattern as Home/Bookmarks), coercing unsupported content types (short/long/links/files) to `all` when Visual is active
* **`TimelineFeedVisual.helpers.ts`** — Added `SEARCH` to `VISUAL_INTERACTIVE_CONTENT_VARIANTS` to enable content coercion
* **`useSearchStreamId.ts`** — Added optional `contentOverride` parameter so `SearchTimelineFeed` can pass the resolved content directly, avoiding a one-render lag with unsupported content types
* **`useFeedLayoutResolution.test.ts`** — Updated test that used SEARCH as an "unsupported" variant (now uses HOT), added new test confirming SEARCH supports Visual


secondl1ght (approved):
Verified via the preview URL and appears to work as expected. 

I found an unrelated bug during testing though. The Wide layout on the Search page is currently broken, this is on production right now. You can see it at the end of my demo video. I will put in a another PR to fix this. 

https://github.com/user-attachments/assets/af8c4a46-e27e-408a-a990-b0bf2bfceb3e

PR #1688, opened 2026-04-10
fix(timeline): remove Virtuoso and restore infinite scroll lists

infin1t3:
Remove Virtuoso and restore infinite scroll lists


catch-21:
https://github.com/pubky/pubky-app/actions/runs/24252496587
All but 1 E2E test are passing. That's acceptable because the failing test needs updating to handle new Column layout on post page.

catch-21:
LGTM. Can scroll down a long way and back up without flickers or jumping around. Can use pubky logo to get back to top reliably. Can use Visual layout with tag filter to yeild less than 3 results.

https://github.com/user-attachments/assets/e1b3b806-c862-44d2-9d54-dc92eb59381d

https://github.com/user-attachments/assets/e4bd6d8f-7153-492e-9510-bd1f179e8cc5

Orlandohub (approved):

PR #1472, opened 2026-03-06
feat: add visual feed layout

Orlandohub:
## Summary
- fix #1449
- Adds a new "Visual" feed layout that renders posts as an image/video tile grid, giving users a media-centric browsing experience alongside the existing Columns and Wide layouts

## Changes
- Introduced `VisualTimelinePosts` component with masonry-style tile rendering for image/video posts, including auto-playing video previews and hover overlays with post actions
- Added `useVisualFeedTiles` hook to compute tile layout (column spans, aspect ratios, row grouping) from post attachment metadata
- Created `useFeedLayoutResolution` hook to resolve the effective layout based on feed variant and viewport size (visual is disabled on phone viewports)
- Extended `FilterLayout` with an optional "Visual" tab and `FilterContent` with disabled-tab support to lock content filters when visual mode is active
- Updated `ContentLayout` and `HomeFeedSidebar` to propagate `feedVariant` and handle the visual layout's wide-shell sidebar behavior
- Added i18n translations for the "visual" label across all 9 locales
- Added comprehensive tests for all new components, hooks, and helper utilities

## Test plan
- [ ] Switch to the Visual layout on Home, Search, and Bookmarks feeds — verify tile grid renders correctly
- [ ] Confirm video tiles auto-play muted for ≤5s and pause when scrolled out of view
- [ ] Verify content filter tabs (Short, Long, Links, Files) are disabled when Visual is active
- [ ] Resize to phone viewport and confirm Visual layout falls back to Columns
- [ ] Run existing and new unit tests: `npm test`

## Notes
- Visual layout is only available on feeds that opt in via `allowVisualLayout` / `feedVariant` — Hot feed currently excluded
- On phone viewports (`< md`), the visual layout automatically downgrades to Columns to preserve readability

PR #1690, opened 2026-04-11
fix: wide layout on search page

secondl1ght:
### Summary

The Wide layout option has been broken on the Search page since it was first created. Users could see and click the "Wide" option in the filter sidebar, but it would immediately snap back to "Columns" due to `useLayoutReset()` forcibly resetting the layout on every mount.

### Root Cause

The Search page was created as a placeholder in commit `485c690e` (Dec 2025) with `useLayoutReset()` and a comment: *"Reset to column layout on mount (this page doesn't support wide)"*. At the time Search was just static placeholder content.

As Search matured into a full feed page, the reset was never removed. The visual feed layout PR (#1472) briefly fixed it in one commit (`741791d0`) but accidentally re-added it a few hours later in a follow-up fixup (`ab5620c2` — "Restore layout reset for Hot pages") that swept Search in along with Hot. No released version ever had Wide working on Search.

### Changes

**`src/components/templates/Feed/Search/Search.tsx`**
- ❌ Removed `Hooks.useLayoutReset()` — this was resetting `wide` → `columns` on every mount

**`src/components/templates/Feed/Search/Search.test.tsx`**
- Removed `mockUseLayoutReset` and its associated test
- Added test verifying `SEARCH` feedVariant is correctly passed to `ContentLayout`
- Updated `ContentLayout` mock to capture and expose the `feedVariant` prop

### Screenshots

image

---

### DIFF
--- a/src/components/templates/Feed/Search/Search.test.tsx
+++ b/src/components/templates/Feed/Search/Search.test.tsx
@@ -2,12 +2,10 @@ import { describe, expect, it, vi, beforeEach } from 'vitest';
 import { render, screen } from '@testing-library/react';
 import { Search } from './Search';
 
-const mockUseLayoutReset = vi.fn();
 const mockUseIsMobile = vi.fn(() => false);
 const mockUseSearchTags = vi.fn(() => ['pubky']);
 
 vi.mock('@/hooks', () => ({
-  useLayoutReset: () => mockUseLayoutReset(),
   useIsMobile: () => mockUseIsMobile(),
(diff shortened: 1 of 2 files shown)


Orlandohub (approved):
Good catch!!

LGTM

catch-21:
Verified in fa2806d

https://github.com/user-attachments/assets/46cdf59e-ff0b-4c83-a8b8-61a75ecc7095

Issue #1669, opened 2026-04-09
[Feature]: Notification Task 2: Filter notification list in Controller


Taewa:
### Problem Statement

`NotificationController.getOrFetchNotifications()` returns all notifications from IndexedDB regardless of user preferences.
Disabled notification types still appear in the list.

### Proposed Solution

Read NotificationPreferences from `useSettingsStore` in the controller and apply the shared filter helper before
returning results to the `useNotifications` hook. Update controller tests.

### Suggested Priority

None

Issue #1668, opened 2026-04-09
[Feature]: Notification Task 1: Create shared notification preference filter helper


Taewa:
### Problem Statement

No reusable utility exists to filter notifications by user preferences. The filtering logic will be needed in multiple controller methods.

### Proposed Solution

Create a pure function that takes `FlatNotification[]` and `NotificationPreferences`, maps each `NotificationType` to its preference key, and returns only enabled notifications. Place it alongside the controller in `src/core/controllers/notification/`. 
Add unit tests.

### Suggested Priority

None

Issue #1667, opened 2026-04-09
[Feature]: notification filtering by setting


Taewa:
### Problem Statement

Notification preferences (Settings > Notifications) are stored and synced to the homeserver but never consulted for filtering. 
The backend returns all notifications unconditionally, and the frontend only filters by muted users — not by notification type. 
Toggling a notification type off (e.g., "deleted post") has no effect on either the notification list or the unread badge count (both in `/profile` page).

### Proposed Solution

- Store all notifications in IndexedDB regardless of preferences (preserves data if user re-enables a type)
- Apply filtering in `NotificationController` — reads preferences from `useSettingsStore`, passes them to the helper
- Filter both code paths using the same helper:
  - getOrFetchNotifications() — notification list
  - fetchNotifications() — unread badge count
  - Controller layer chosen to respect ADR-0004 layering (Application layer stays store-free)

Impacted Areas
- NotificationController (`src/core/controllers/notification/notification.ts`)
- Notification list page (`useNotifications` hook consumers)
- Unread badge count (`MobileFooter`, `HeaderSignIn`)

### Suggested Priority

None

Taewa:
Turn out this issue https://github.com/pubky/pubky-app/issues/1603 is not a bug but the feature is not yet implemented.

PR #1693, opened 2026-04-12
Feat/1672 filter notification badge count

### LINKED ISSUE #1672: [Feature]: Notification Task 3: Filter unread badge count by notification preferences
### Problem Statement

Notification preferences (Settings > Notifications) are not applied to the unread badge count.
The badge in `MobileFooter`, `HeaderSignIn`, and `ProfilePageFilterBar` includes disabled notification types.
This also affects app startup (bootstrap) where the initial badge count is unfiltered.

### Proposed Solution

Pass user notification preferences through the existing layers (Controller → Application → Service → Model) to count only enabled notification types as unread.
- Controller reads preferences from the settings store and converts to allowed types
- Application/Service/Model count unread using only those allowed types via an efficient IndexedDB query
- Bootstrap resolves final preferences (remote settings win over local) before computing the initial badge count
- Existing pure filter helpers (`filterByPreferences`, type mapping) moved to `NotificationNormalizer` (Pipes layer) for shared access

### Suggested Priority

None

Taewa:
## Note
  This is part of the notification filtering feature (#1667). The full feature is split into
  sequential PRs:
  - Task 1: Shared preference filter helper (#1668) ✅
  - Task 2: Filter notification list (#1669) ✅
  - **Task 3: Filter unread badge count (#1672) ← this PR**
  - Task 4: Immediate badge update on preference toggle (upcoming)

## Summary
  - Filter unread badge count by user notification preferences across all layers (Controller →
  Application → Service → Model)
  - Add efficient IndexedDB query (`countFilteredNewerThan`) that counts only enabled notification
   types
  - Move shared filter helpers (`filterByPreferences`, `toEnabledTypes`, type mapping) to
  `NotificationNormalizer` (Pipes layer)
  - Bootstrap resolves final preferences (remote settings win over local) before computing initial
   badge count

### 1. Notification filter by settings
https://github.com/user-attachments/assets/d3181aa0-67e1-4db2-9df4-6cd31d5f35a3

### 2. Notification setting should remain after signin (bootstrap)

https://github.com/user-attachments/assets/32a7b20d-8e2a-4653-9e92-f5ef84b8c8c8


Taewa:
Note that "Dead code — countUnreadSince is no longer called in production" is now removed.

src/core/pipes/notification/notification.normalizer.ts
```diff
@@ -2,6 +2,24 @@ import { LastReadResult } from 'pubky-app-specs';
...
 import { Err, ValidationErrorCode, ErrorService } from '@/libs';
 import { getBusinessKey } from '@/core/models/notification/notification.helpers';
+import type { NotificationPreferences } from '@/core/stores/settings/settings.types';
+import { NotificationType } from '@/core/models/notification/notification.types';
+import type { FlatNotification } from '@/core/models/notification/notification.types';
+
+// Maps each NotificationType enum to its corresponding NotificationPreferences key.
+// Needed because NotificationType values don't match the preference key names directly.
+// exported for tests
+export const NOTIFICATION_TYPE_TO_PREFERENCE_KEY: Record<NotificationType, keyof NotificationPreferences> = {
```
Orlandohub: This feels like it would be better placed at `notification.constants.ts` similar to `/models/post/details/postDetails.constants.ts`

src/core/services/local/notification/notification.ts
```diff
@@ -24,12 +26,14 @@ export class LocalNotificationService {
   }
 
   /**
-   * Counts unread notifications by querying IndexedDB for entries newer than lastRead.
+   * Counts unread notifications filtered by allowed types.
```
Orlandohub: Should we keep the "for entries newer than lastRead."?

src/core/services/local/notification/notification.ts
```diff
@@ -1,4 +1,6 @@
 import * as Core from '@/core';
+import { NotificationType } from '@/core/models/notification/notification.types';
+import { NotificationModel } from '@/core/models/notification/notification';
```
Orlandohub, line 3: Well done with the new type of imports instead of barrels!!

src/core/application/notification/notification.types.ts
```diff
@@ -4,12 +4,13 @@ export type TNotificationApplicationNotificationsParams = {
...
   lastRead: number;
+  allowedTypes: Core.NotificationType[];
 };
 
 /**
- * Result from fetchNotifications containing unread count and the next poll cursor.
+ * Result from fetchNotifications containing the filtered unread count and next poll cursor.
  *
- * @property unread - Number of unread notifications (those newer than lastRead)
+ * @property unread - Number of unread notifications filtered by user preferences
```
Orlandohub: is it not true that we still keep "(those newer than lastRead)" besides the filtered preferences?

src/core/application/notification/notification.ts
```diff
@@ -44,6 +46,18 @@ export class NotificationApplication {
...
 
+  /**
+   * Counts unread notifications filtered by allowed types.
+   * Used by Controllers to compute the preference-filtered badge count.
+   *
+   * @param lastRead - Timestamp of the last read notification
+   * @param allowedTypes - Notification types to include in the count
+   * @returns Promise resolving to the filtered unread count
+   */
+  static async countFilteredUnreadSince(lastRead: number, allowedTypes: Core.NotificationType[]): Promise<number> {
```
Orlandohub, line 57: This one was created but it's just a passthrough for `Core.LocalNotificationService.countFilteredUnreadSince` and is not being used anywhere. What's the idea here?

Orlandohub (changes requested):
I've left comments on what I found, plus one pointer from the AI reviewer:

### Asymmetric filtering pattern between fetchNotifications and getOrFetchNotifications
After this PR, the two flows handle preferences differently:

- fetchNotifications (badge count): controller computes allowedTypes via NotificationNormalizer.toEnabledTypes and passes it into the application; filtering happens in the model query.
- getOrFetchNotifications (list): application returns unfiltered data; controller post-filters via NotificationNormalizer.filterByPreferences.

This split is defensible — count needs SQL-level filtering for efficiency, list can JS-filter the already-loaded array — but the two shapes drift apart. A more uniform pattern would be: application always receives allowedTypes and is responsible for filtering (controller is a thin wrapper that converts preferences → allowedTypes via the pipe, never filters itself). This would also remove the controller's import of NotificationNormalizer for the list path.

Not a blocker, but worth considering — Task 4 may make this asymmetry harder to refactor later.

### Inconsistent matcher in `bootstrap.test.ts`

Two assertions on the same mock use different shapes:
```
// assertCommonCalls
expect(mocks.countFilteredUnreadSince).toHaveBeenCalledWith(MOCK_LAST_READ, expect.any(Object));
// "should throw error when countFilteredUnreadSince fails"
expect(mocks.countFilteredUnreadSince).toHaveBeenCalledWith(MOCK_LAST_READ, expect.any(Array));
```
Both pass for an array, but the intent is NotificationType[] — both should use expect.any(Array) (or, better, expect.arrayContaining(Object.values(Core.NotificationType)) to actually verify content).

src/core/application/notification/notification.ts
Taewa: @Orlandohub 
It's used by `SettingsController.recalculateUnreadBadge()` (There is up coming PR, the Task 4, commit [e1aa00e](https://github.com/pubky/pubky-app/pull/1709/changes/e1aa00efef330ef6122a1cc4536a377619fa7489)). 
The passthrough exists because Controllers cannot call Services directly per ADR-0004 — the dependency flow is Controllers → Application → Services. 
Even though it's a thin wrapper, skipping Application would violate the layer boundary.

src/core/pipes/notification/notification.normalizer.ts
Taewa: Make sense!

src/core/services/local/notification/notification.ts
Taewa: Yep you are right

src/core/application/notification/notification.types.ts
Taewa: It is. I put it back! Thanks!

Taewa:
@Orlandohub 
For the asymmetric topic you commented:

Thanks, this is a good callout.
For the unread count path, pushing filtering down makes clear sense because it enables DB-level filtering. For the list path, the current behavior is just a JS filter over already-loaded data, so moving that into Application would mostly improve symmetry, not efficiency or behavior.

Given that, I’d prefer to keep this PR as-is and revisit unifying the two paths if Task 4 or later changes make the asymmetry painful in practice. If we do refactor it, I agree the cleaner long-term shape would be for the Controller to derive `allowedTypes` and for Application to consistently own the filtering.

Orlandohub (approved):

PR #1696, opened 2026-04-13
refactor(Dialog): improvement in the DialogReply utils to make it from PostInput

### LINKED ISSUE #1653: [Bug]: Mobile: Missing auto scroll down and open soft keyboard when reposting
### Description

Same as https://github.com/pubky/pubky-app/issues/996 fixed in https://github.com/pubky/pubky-app/pull/1525 but for reposts.

On mobile, when post's repost button is tapped the input field should be scrolled into focus and the soft keyboard should be displayed.

### Expected Behaviour

On mobile, when post's repost button is tapped the input field should be scrolled into focus and the soft keyboard should be displayed.

### Steps to Reproduce

1. On mobile, find a tall post in the feed (with embedded X is a good example)
2. Tap repost button for the post
3. Observe dialog shows but the input field is not entirely in view, and the soft keyboard is not displayed so the user needs to tap the input field

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

Nirzhuk:
fix #1653 

- Moved scrolling functionality from DialogReply and DialogRepost to a new utility function in PostInput.utils.
- Updated DialogReply and DialogRepost components to use the new scrollDialogTextareaIntoDialog function.
- Removed the now redundant DialogReply.utils file.

This change improves code reusability and maintains consistent scrolling behavior across dialog components.


Taewa:
@Nirzhuk 
Could you please add a test video or screenshots that can help to verify?

Taewa (approved):
@Nirzhuk 
Approved and it reminds our [discussion](https://github.com/pubky/pubky-app/pull/1525#discussion_r3011046403) and `PostInput.utils.ts` can be a good target for VRT soon :)

PR #1665, opened 2026-04-08
Hotfix 1.2.1

catch-21:
To fix:
- https://github.com/pubky/pubky-app/issues/1661
- https://github.com/pubky/pubky-app/issues/1659
- https://github.com/pubky/pubky-app/issues/1610
- https://github.com/pubky/pubky-app/issues/1677
- https://github.com/pubky/pubky-app/issues/1678

removes Virtuoso in https://github.com/pubky/pubky-app/pull/1688

PR #1707, opened 2026-04-21
Hotfix 1.2.1

catch-21:
same as https://github.com/pubky/pubky-app/pull/1665 but with correct commit removing Virtuoso.


catch-21:
Passing E2E: https://github.com/pubky/pubky-app/actions/runs/24717078682

PR #1708, opened 2026-04-21
fix(posts): apply simple repost tags to original post

### LINKED ISSUE #1240: [Bug]: Tags are not applied to reposts without content
### Description

When reposting a post, the tags added in the Repost dialog are only applied if the repost has content and.or attachment, not if there is no content. This is because reposts without content show the tags for the original post in the feed.

### Expected Behaviour

Either we need the tags to be applied to the original post when there is no content and to the repost if there is content and/or attachment, which is inconsistent and could be confusing, or, we disable the tag field unless there is content and/or attachment.

### Steps to Reproduce

1. Find post or repost with no content that already has tags
2. Repost via post or existing repost with no content
3. Observe existing tags are not listed in the Repost dialog (good because new tags would be associated with the repost event)
4. Add a tag
5. With no content or attachments, repost
6. Observe repost in feed does not show the new tag, only tags from the original post

### Screenshots

Repost from post
https://github.com/user-attachments/assets/a74b662d-fbe5-494e-8e35-304eb862159f

Repost from existing repost that has no content
https://github.com/user-attachments/assets/ea2b894b-9cfc-4cae-b178-3d19a7fbfdec

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
Simple reposts (no text or attachments) attach user tags to the embedded original; quote reposts keep tags on the new post. Adds resolveTagTargetCompositeIdForPostCreate and controller tests.

https://github.com/user-attachments/assets/9b3e05c7-7c84-45df-8feb-d45f54a6a67e

Resolves: #1240


src/core/controllers/post/post.ts
```diff
@@ -151,10 +151,18 @@ export class PostController {
     const { id: postId } = meta;
 
     if (tags) {
+      const tagTargetCompositeId = Core.resolveTagTargetCompositeIdForPostCreate({
```
Taewa, line 154: @infin1t3 
Are you keeping the `Core...` since the barrel will be removed at once?

src/core/controllers/post/post.test.ts
```diff
@@ -478,6 +479,71 @@ describe('PostController', () => {
         bodyJson: expect.any(Object),
       });
     });
+
+    it('should apply tags to the original post for a simple repost', async () => {
+      await setupExistingPost();
+      const ApplicationModule = await import('@/core/application');
```
Taewa, line 491: I think the Application layer is out of the concern of this test (it's the post controller test). So it's unnecessary from my point of view.

src/core/controllers/post/post.test.ts
```diff
@@ -478,6 +479,71 @@ describe('PostController', () => {
...
+      const tagCommitSpy = vi.spyOn(ApplicationModule.TagApplication, 'commitCreate');
+
+      const { PostController } = await import('./post');
```
Taewa, line 494: I noticed that there are approximately 50 dynamic imports and this is also unnecessary unless we have to.
The main purpose of this test file is `PostController` and it can be simply done by `import { PostController } from './post';` at the top of the test file.

Otherwise, do we need the dynamic import?

src/core/controllers/post/post.test.ts
```diff
@@ -478,6 +479,71 @@ describe('PostController', () => {
...
+        originalPostId: testData.fullPostId,
+        tags: ['bitcoin'],
+      });
+
+      expect(tagCommitSpy).toHaveBeenCalled();
+      const { tagList } = tagCommitSpy.mock.calls[0][0];
+      expect(tagList[0]?.taggedId).toBe(testData.fullPostId);
+      expect(tagList[0]?.label).toBe('bitcoin');
+
+      tagCommitSpy.mockRestore();
```
Taewa: I see a repeated pattern `tagCommitSpy.mockRestore()`
Isn't it better to put in `afterEach`?

src/core/pipes/post/post.kind.ts
```diff
@@ -58,3 +67,30 @@ export const inferPostKindForCreate = ({ content, attachments, isArticle }: TInf
...
+  attachments,
+  isArticle,
+}: TResolveTagTargetCompositeIdParams): string => {
+  const newCompositeId = `${authorId}:${newPostId}`;
+  if (!originalPostId) {
+    return newCompositeId;
+  }
+
+  const hasAttachments = (attachments?.length ?? 0) > 0;
+  const isSimpleRepost = !content.trim() && !hasAttachments && !isArticle;
```
Taewa: I tested with an article post.
1. Repost an article post with no quote, no attachment. And add a tag.
2. Because of this rule `!isArticle`, I expected the new tag should be displayed on the shared post (new post)
3. But the new tag is added on the original one

I guess it's not intended, right?

src/core/controllers/post/post.ts
infin1t3: I'm not sure I understand? If you are referring to the fact that we are using whole Core namespace through the barrel import - we already use it in multiple places in this controller so I just followed the existing pattern.

Barrel imports will be handled in a separate PR.

src/core/controllers/post/post.test.ts
infin1t3: True, but we use application module here just to verify the controller forwards the correct taggedId to TagApplication. Since PostController.commitCreate doesn't return the target, spying on that call is the only way to assert the routing. Pure logic is already covered in post.kind.test.ts; these tests prove the controller wires it correctly.

src/core/controllers/post/post.test.ts
infin1t3: It is an issue but, as you noticed, there are approximately 50 dynamic imports in that file already so I was just following existing pattern. Happy to clean them all up in a follow-up PR to keep this diff focused on the bug fix.

src/core/pipes/post/post.kind.ts
infin1t3: It is. The flow is good, it's just the name is confusing.

**isArticle** flag in our codebase refers to the NEW post being authored (long-form editor),not the kind of the original.

On the other hand, I think that we can not create an article from a repost dialog so I will just remove this to avoid confusion.

src/core/controllers/post/post.ts
Taewa: Personally, I stop using barrel import pattern since we decided to deprecate. 
So my comment was more like "why do we use a pattern that we are about to be deprecated?" and you answered that there will be a separated PR for removing barrel. Therefore no problem.

src/core/controllers/post/post.test.ts
Taewa: I see two things here
1. dynamic import `ApplicationModule`: I think you follow the existing pattern as you mentioned below comment so maybe refactor can happen in another PR.

2. The flow is (A)`PostController.commitCreate` -> (B)`PostApplication.commitCreate` -> (C) `TagApplication.commitCreate`. 
And your code tests (C) in (A) so I thought maybe it's too deep in a unit test level. I am not saying it's bad but just I am thinking testing (A) -> (B) would be enough.
And test (B) -> (C) in `PostApplication.commitCreate` test file.

src/core/controllers/post/post.test.ts
Taewa: Maybe could you leave somewhere `TODO`, so in the future reviewers know that it's about to be refactored? :)

infin1t3: 👍

Taewa (approved):
Tested on deployed app. 

1. Repost with adding a tag **without** quote (no text) => the new tag displayed in the original post
2. Repost with adding a tag **with** quote => the new tag displayed in the new post
3. Tested with an "article", it behaves the same as 1,2

PR #1711, opened 2026-04-22
test: fix test typings and enforce typecheck in CI

### LINKED ISSUE #640: Fix TS errors in test files and include check in CI flow
Test files (`.test.tsx`, `.test.ts`) are **never type-checked** because:
1. Vitest doesn't type-check - it only transpiles TypeScript for speed
2. `next build` excludes test files - they're not part of production bundle
3. No `tsc --noEmit` in CI - there was no standalone type-check step (but this script is mentioned in the `AGENTS.md` file even though it does not exist)

infin1t3:
Add npm run typecheck to pre-commit and the Code Quality workflow. Update test mocks and assertions to match current types.

Resolves #640

Made-with: Cursor


infin1t3:
@greptileai review again

catch-21:
This appears to address https://github.com/pubky/pubky-app/issues/640

I see quite a few `as unknown` casting in the test which effectively bypasses the type check for like whether mocks conform with the interface. We should never do `as unknown as X` or `as any` unless it is a deliberate invalid input or when using an external type such as with the pubky client sdk.

catch-21:
@infin1t3 I've pushed a mega commit 134bc6234aa32bc03c6f741ef1e128e1bdf66803 to remove all uses of `as unknown as ` and `as any` from tests, banning it entirely to guide us better towards type safety. The rule now is to always be explicit when using invalid or opaque types, and events such as keyboard input use factory model. There are docs and cursor rules are updated too so this may not be a burden to adhere to going forwards.

I treat this as an experiment. Please be critical if we don't like the complexity added or impact to readability here. There is also the option of splitting this PR into two or more stages.

Orlandohub (approved):
I would recommend adding something along these lines to the `AGENTS.md` file:

```
## Quality gates

Run before pushing (also enforced in `pre-commit` and the `Code Quality` workflow):

- `npm run format:check` — Prettier
- `npm run lint` — ESLint (incl. test-only rules in `eslint.config.mjs`)
- `npm run typecheck` — `tsc --noEmit`, includes `*.test.{ts,tsx}`

Tests must type-check. In test files `as any` and `as unknown as T` are banned by ESLint — route every cast through a named helper in `src/test-utils` (`asInvalid`, `asOpaque`, `mockAuthStore`, `mockSession`, `mockKeyboardEvent`, …). See `docs/component-testing.md` and `src/test-utils/README.md`.
```

secondl1ght (commented):
Gave this a quick review since I created the task, and since we now typecheck everything at the commit stage, this also has the benefit that errors will now be caught before the build fails. 👍🏼

PR #1632, opened 2026-04-03
feat: hot page mobile tab navigation

secondl1ght:
## 🐞 Fix: Add missing navigation tabs on Hot page for mobile

Resolves https://github.com/pubky/pubky-app/issues/1001

### Summary

On mobile, the Hot page was missing the **Tags**, **Users**, and **Posts** navigation tabs along the top. This PR introduces a new `HotMobileMenu` component that provides a sticky tab navigation bar visible only on mobile (`< lg` breakpoint), allowing users to toggle between the three content sections.

#### 🆕 New Component — `HotMobileMenu`
- **`HotMobileMenu.tsx`** — A mobile-only sticky tab bar with three tabs (Tags, Users, Posts), each rendered with an icon and translated label. Uses the same positioning pattern as `ProfilePageMobileMenu` with negative margin overrides (`-mx-6`, `-mt-6`) to stretch full-width edge-to-edge and sit flush against the header.
- **`HotMobileMenu.types.ts`** — Defines the `HotSection` enum (`TAGS`, `USERS`, `POSTS`) and the `HotMobileMenuItem` / `HotMobileMenuProps` interfaces.
- **`index.ts`** — Barrel export for the new component.

#### 🔧 Modified — `Hot` Template
- Integrates `HotMobileMenu` into the Hot page layout.
- Uses `useState` with `HotSection` to track the active tab on mobile.
- Leverages `useIsMobile` hook to conditionally apply `hidden` CSS class to inactive sections, preserving component state and avoiding refetches on tab switch.
- Passes `hasGradientBackground={false}` to `ContentLayout` for a solid mobile header background.

#### 🔧 Modified — `ContentLayout`
- Added `hasGradientBackground` prop to `ContentLayoutProps` and forwarded it to `MobileHeader`, enabling control over whether the mobile header uses a gradient or solid background.

#### 🌐 i18n — Translation Files
- Added `tags`, `users`, and `posts` translation keys to the `hot` namespace across all **9 locale files**: `ar`, `de`, `en`, `es`, `fr`, `it`, `ja`, `pt-BR`, `zh`.

### 🧪 Testing

Tests have been created or updated for the following files:
- `src/components/molecules/HotMobileMenu/HotMobileMenu.test.tsx` _(new)_
- `src/components/molecules/HotMobileMenu/HotMobileMenu.test.tsx.snap` _(new)_
- `src/components/templates/Feed/Hot/Hot.test.tsx` _(updated)_
[...]

PR #1734, opened 2026-04-24
refactor: unify mobile tab menus into shared component

secondl1ght:
# Unify Mobile Tab Bars into a Shared `MobileTabBar` Molecule

## 🎯 Summary

Three structurally identical mobile tab-bar components (`ProfilePageMobileMenu`, `SettingsMobileMenu`, and the newly added `HotMobileMenu`) existed because this repeated pattern was not initially made into a reusable component. This PR extracts a single shared **`MobileTabBar`** molecule that owns the visual + a11y contract, and migrates all three call sites onto it.

Delivered as a **follow-up PR targeting the original `HotMobileMenu` PR branch** so reviewers can evaluate the unification in isolation from the Hot-page feature work.

### New shared component
- `src/components/molecules/MobileTabBar/MobileTabBar.tsx`
- `src/components/molecules/MobileTabBar/MobileTabBar.types.ts`
- `src/components/molecules/MobileTabBar/MobileTabBar.test.tsx` (+ snapshot)
- `src/components/molecules/MobileTabBar/index.ts`
- Registered in `src/components/molecules/index.ts`

**API highlights:**
- Per-item `isActive` + `onSelect` (not a global `activeKey`) — keeps `requireAuth` wrapping, `pathname` comparisons, and `setState` setters natural in consumers.
- `showLabels` toggles icon-only vs icon + label.
- `position` toggles between `sticky top-(--header-height-mobile)` and `fixed top-(--header-height-mobile) right-0 left-0`.
- `className` is merged last via `cn(...)` so consumers can append margin overrides without fighting defaults.
- `data-testid` passthrough + `aria-label` default (with optional override) + `aria-current="page"` on active.

### Migrated consumers (thin adapters)
- **`HotMobileMenu`** → `<MobileTabBar showLabels position="sticky" className="-mx-6 -mt-6 mb-6" ... />`. Keeps `HOT_MOBILE_MENU_ITEMS`, `HotSection`, and `hot.*` i18n.
- **`ProfilePageMobileMenu`** → `<MobileTabBar position="sticky" ... />`. Preserves `PROFILE_MENU_ITEMS` export, `isOwnProfile` filtering, `requireAuth`-wrapped clicks, and raw English labels.
- **`SettingsMobileMenu`** → `<MobileTabBar position="fixed" ... />`. Keeps `useTranslations('settings.menu')`, `SETTINGS_MOBILE_ITEMS`, and `usePathname`/`useRouter` selection.

## ⚠️ Behavioral change (intentional, scoped)

**Settings z-index consolidation:** `z-(--z-sticky-header)` (20) → `z-(--z-mobile-menu)` (30), which is the variable documented for mobile navigation and the value already used by Profile + Hot. This was a silent drift that this refactor corrects.

No other behavior changes:
- Settings remains `fixed`; Profile + Hot remain `sticky`.
- Hot keeps its `-mx-6 -mt-6 mb-6` layout overrides via `className` passthrough.
- Profile `requireAuth` and `isOwnProfile` filtering are preserved.
- i18n namespaces are unchanged.
- `lg:hidden` still hides all three on desktop.

## 🧹 Drive-by fix

Removed a stale literal string (`'data-testid="profile-page-mobile-menu-item"'`) that was being passed into `cn(...)` as a className in the old `ProfilePageMobileMenu` render block. It had no effect but was technically invalid class input.

## 🧪 Tests

- **`MobileTabBar`** — 18 tests: item rendering, icon-only vs `showLabels`, sticky + fixed positioning, always-applied root classes, active vs inactive states (border, icon color, `aria-current`), click dispatch, `aria-label` default + override, `className` merge, `data-testid` passthrough, button class toggling.
- **`HotMobileMenu`** — 9 tests, trimmed to behavior-level assertions (label rendering, `-mx-6 -mt-6 mb-6` passthrough, sticky, `data-testid`, `aria-current`, click wiring). Snapshot regenerated.
- **`ProfilePageMobileMenu`** — 8 tests. Added explicit coverage for `isOwnProfile={false}` filtering and `requireAuth`-gated click dispatch via `vi.mock('@/hooks', ...)`. Snapshot regenerated.
- **`SettingsMobileMenu`** — 5 tests. Added explicit `z-(--z-mobile-menu)` assertion + negative assertion against the old `z-(--z-sticky-header)`, plus a `fixed` positioning assertion. Snapshot regenerated.
- **Consumer suites** — `templates/Feed/Hot` (17) and `templates/Settings` (40) all pass with zero prop changes.

**Totals:** 97 targeted tests ✅ · `yarn tsc --noEmit` ✅ · `yarn lint` ✅

## 📦 File inventory

**New**
- `src/components/molecules/MobileTabBar/MobileTabBar.tsx`
- `src/components/molecules/MobileTabBar/MobileTabBar.types.ts`
- `src/components/molecules/MobileTabBar/MobileTabBar.test.tsx`
- `src/components/molecules/MobileTabBar/MobileTabBar.test.tsx.snap`
- `src/components/molecules/MobileTabBar/index.ts`

**Modified**
- `src/components/molecules/index.ts`
- `src/components/molecules/HotMobileMenu/HotMobileMenu.tsx`
- `src/components/molecules/HotMobileMenu/HotMobileMenu.test.tsx`
- `src/components/molecules/HotMobileMenu/HotMobileMenu.test.tsx.snap`
- `src/components/molecules/ProfilePageMobileMenu/ProfilePageMobileMenu.tsx`
- `src/components/molecules/ProfilePageMobileMenu/ProfilePageMobileMenu.test.tsx`
- `src/components/molecules/ProfilePageMobileMenu/ProfilePageMobileMenu.test.tsx.snap`
- `src/components/molecules/Settings/SettingsMobileMenu/SettingsMobileMenu.tsx`
- `src/components/molecules/Settings/SettingsMobileMenu/SettingsMobileMenu.test.tsx`
- `src/components/molecules/Settings/SettingsMobileMenu/SettingsMobileMenu.test.tsx.snap`

## 🚧 Deferred / follow-ups

- **Phase 5 (optional):** investigate whether Settings can move from `fixed` to `sticky` (would eliminate the last positioning fork). Deferred to keep this PR's behavior change surface minimal.
- **Future i18n sweep:** `PROFILE_MENU_ITEMS` still uses raw English labels while Hot and Settings use translation keys. Out of scope here.

## Related

This PR was created to directly address this review comment https://github.com/pubky/pubky-app/pull/1632#pullrequestreview-4075361705.

---

### DIFF
--- a/src/components/molecules/MobileTabBar/index.ts
+++ b/src/components/molecules/MobileTabBar/index.ts
@@ -1,0 +1,2 @@
+export * from './MobileTabBar';
+export * from './MobileTabBar.types';
--- a/src/components/molecules/ProfilePageMobileMenu/ProfilePageMobileMenu.tsx
+++ b/src/components/molecules/ProfilePageMobileMenu/ProfilePageMobileMenu.tsx
@@ -1,10 +1,10 @@
 'use client';
 
 import * as React from 'react';
-import * as Atoms from '@/atoms';
 import * as Libs from '@/libs';
 import * as Hooks from '@/hooks';
 import * as Types from '@/app/profile/types';
+import { MobileTabBar, type MobileTabBarItem } from '../MobileTabBar';
 
 export interface ProfileMenuItem {
   icon: React.ComponentType<{ size?: number; className?: string }>;
@@ -41,6 +41,15 @@ export interface ProfilePageMobileMenuProps {
   isOwnProfile?: boolean;
 }
 
+/**
+ * Mobile navigation menu for profile pages.
+ * Delegates rendering to the shared `MobileTabBar` molecule.
+ *
+ * Owns profile-specific semantics:
+ * - Filters items via `isOwnProfile` (e.g. hides Notifications on other users).
+ * - Wraps click handlers with `requireAuth` for unauthenticated users.
+ * - Uses raw English labels (no i18n sweep yet).
+ */
 export function ProfilePageMobileMenu({
   activePage,
   onPageChangeAction,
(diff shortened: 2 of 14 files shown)


src/components/molecules/ProfilePageMobileMenu/ProfilePageMobileMenu.tsx
Orlandohub, line 61: Maybe we could use this opportunity to also clean up this unnecessary `useMemo`?

src/components/molecules/Settings/SettingsMobileMenu/SettingsMobileMenu.test.tsx
Orlandohub, line 1: Settings is the only consumer whose selection is derived from usePathname and whose click handler performs a side effect (router.push). Hot and Profile both gained parallel coverage for their click wiring in this PR (onSectionChange, requireAuth) — Settings is now the odd one out.

I would suggest adding two tests mocking next/navigation — one asserting router.push is called with the clicked item's path, one asserting the active item when usePathname matches.

src/components/molecules/MobileTabBar/index.ts
Orlandohub, line 1: Technically conflicts with the no-barrel-reexports rule in .greptile/config.json, whose stated exception is narrow (src/components/atoms/index.ts)

Orlandohub (approved):
I'll approve since I do not deem those comments as critical but nice to have instead

src/components/molecules/ProfilePageMobileMenu/ProfilePageMobileMenu.tsx
secondl1ght: I think this could be yet another tech debt task to add to the list, and do it all in one sweep. :P

src/components/molecules/MobileTabBar/index.ts
secondl1ght: I have not been fighting with AI on barrels because it always uses the existing code for examples and the barrels are everywhere. But 100% agree that after the barrel cleanup we should never let another one into the codebase again. :D

PR #1735, opened 2026-04-24
fix(useProfileConnections): ensure it handles pagination and updates the following list accurately when users are unfollowed

### LINKED ISSUE #1234: [Bug]: Profile page Following list can be incomplete
### Description

After signing in (to ensure a fresh db), I don't see a complete 'Following' list in my profile page. Some users are missing.

### Expected Behaviour

All listed (Folowers, Following & Friends) are complete at all times.

### Steps to Reproduce

1. Sign in
2. Navigate to profile page
3. Click 'Followers'
4. Scroll down to see all your followers
5. Click 'Friends'
6. Scroll down to see all your friends
7. Click 'Following'
8. Scroll down
9. Observe not all users are listed (see my video for specific example)

### Screenshots

I have 'Geist' and 'Gabreil' as followers and friends but they are not listed in my 'Following'

https://github.com/user-attachments/assets/ed3fc471-65fc-4aad-b37e-fb22bbb44ce4

### Console Log

```shell

```

### Browser & Version

_No response_

Nirzhuk:
- Enhanced the `useProfileConnections` hook to ensure it handles pagination and updates the following list accurately when users are unfollowed.
- Updated related tests to reflect changes in the caching mechanism for the following list.

This ensures a robust user experience when navigating through followed profiles.

#1234


cypress/e2e/contacts.cy.ts
```diff
@@ -178,4 +178,72 @@ describe('contacts', () => {
...
+    // Quick parity checks for other tabs
+    cy.get('[data-cy="profile-filter-item-followers-count"]').should('have.text', 0);
+    cy.get('[data-cy="profile-filter-item-friends-count"]').should('have.text', 0);
+
+    // Validate Following count and open Following tab
+    cy.get('[data-cy="profile-filter-item-following-count"]').should('have.text', followedProfiles.length).click();
+
+    // Trigger infinite scroll page 2 load
+    cy.scrollTo('bottom', { ensureScrollable: false });
+    cy.wait(1500);
```
catch-21: Instead of arbitrary wait, which can be fragile, we can instead have a "is expected bottom list item visible" check that decided whether to scroll down more or not.

cypress/e2e/contacts.cy.ts
```diff
@@ -178,4 +178,72 @@ describe('contacts', () => {
           });
       });
   });
+
+  it('loads complete own following list across pagination', () => {
```
catch-21: This test covers the functionality well but it takes a long time to execute due to onboarding many users and then following them all. Unless we can come up with some other test cases to utilise the same collection of users, I'd have to reject it from the CI pipeline.

What we should aim for is to get coverage of pagination scrolling at the component level in a browser test. For that we could use Cypress but we have plans for visual testing with [Vitest browser](https://vitest.dev/guide/browser/), so we can look at adding these checks there.

catch-21:
Tested with my prod profile and unable to reproduce the bug on this branch 👍

cypress/e2e/contacts.cy.ts
Nirzhuk, line 1: This is going to revert, it's only purpose of helping to test and review.

catch-21:
Tested on fbacc91ab551124846ca345ca177f25d22b3523b

There is a regression that happens to me on revisiting the lists in profile page where list item is immediately removed when unfollowed. ideally this never happens and only when page is reloaded/revisited.

Video shows I can unfollow users in every list without them being removed, then after I leave profile page and return (40s), and unfollow one they are immediately removed. I cannot reproduce this in current prod deployment.

Potentially the removal of logic commented as `// Block sync when cache has fewer users (unfollows) to preserve UI state` is the cause.

https://github.com/user-attachments/assets/a13f4be4-4fac-408a-805c-fed367a53633

catch-21:
Verified on a266bc367bc22675eaf77cb5ffc5f6b986739e33
The regression appears to be fixed and my Following list remains complete. 👍

catch-21 (approved):

PR #1750, opened 2026-04-28
fix(ui): header avatar shows blank after avatar removal (#526)

### LINKED ISSUE #526: [Bug]: Avatar thumbnail remains in header after removing avatar from profile

Taewa:
## Summary

  Closes #526.

  When a user removed their profile avatar, the header thumbnail went blank instead of falling back to the default
  placeholder. The avatar on the profile page itself updated correctly — only the header was stuck.

  ## Why it happened

  The header is a persistent layout element: it doesn't remount when you navigate or save the profile. The
  underlying Radix Avatar component caches the load state of the image internally, so once an avatar had
  successfully loaded, removing the URL didn't cause the placeholder to appear — the cached "image is loaded" state
  suppressed the fallback.

  The profile page didn't have this problem because saving navigates back to the profile route, which mounts a fresh
   Avatar tree.

  ## The fix

  Tie the Avatar's React `key` to the resolved avatar URL, so the Avatar tree remounts whenever the URL changes
  (including when it goes away). A fresh mount means a fresh internal state — and the placeholder shows up as
  expected.

  One-line change in `AvatarWithFallback.tsx`.

  ## Tests

  - Added a unit test that asserts the Avatar root is replaced when the URL changes (catches regressions of the
  `key` wiring).
  - Added a TODO in the component pointing to E2E/VRT for the user-visible flow — the actual visual bug requires
  real image loading, which jsdom can't reproduce, so unit tests can only verify the wiring.

  ## Test plan

  - [ ] Sign in with a custom avatar set
  - [ ] Go to your profile page, edit profile, remove the avatar, save
  - [ ] Confirm the header thumbnail shows the FacehashAvatar placeholder (not blank, not black)
  - [ ] Upload a new avatar and confirm the header updates immediately
  - [ ] Navigate around the app and confirm avatars on posts/lists still render correctly
  
  ## Result
  

https://github.com/user-attachments/assets/fb85d20c-f1c9-49fa-8705-ff4237a0ff3c

### DIFF
--- a/src/components/organisms/AvatarWithFallback/AvatarWithFallback.test.tsx
+++ b/src/components/organisms/AvatarWithFallback/AvatarWithFallback.test.tsx
@@ -328,6 +328,35 @@ describe('AvatarWithFallback', () => {
     });
   });
 
+  describe('Avatar root key', () => {
+    // Verifies the consumer passes `key={resolvedAvatarUrl ?? 'fallback'}` to
+    // <Atoms.Avatar>. React doesn't expose `key` to the rendered component, so we
+    // assert the observable effect: the avatar DOM node is replaced when the key
+    // changes, and preserved when it doesn't. See issue #526 for the underlying
+    // Radix state issue this workaround addresses.
+    it('replaces the avatar root node when avatarUrl changes', () => {
+      const { rerender } = render(<AvatarWithFallback {...mockProps} avatarUrl={validAvatarUrl} />);
+      const firstRoot = screen.getByTestId('avatar');
+
+      // Note: undefined avatarUrl triggers the fallback (image not shown)
+      rerender(<AvatarWithFallback {...mockProps} avatarUrl={undefined} />);
+      const secondRoot = screen.getByTestId('avatar');
+
+      expect(secondRoot).not.toBe(firstRoot);
+    });
+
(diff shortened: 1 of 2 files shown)


Taewa:
@greptile-apps
The useEffect isn't redundant — `imageError` lives in `AvatarWithFallback` (parent), but `key` is on `<Atoms.Avatar> `(child). React only resets state for the remounted subtree, so the parent's `imageError` survives the URL change. Removing the effect breaks 2 existing tests (resets error state when avatarUrl changes and resets error state when local avatar URL changes)

Taewa:
@greptile-apps
Review again plz

infin1t3 (approved):
LGTM - verified at ` a2c1a902b337417e91a71d0e74a4eaf29ce272e3` - approved.

PR #1749, opened 2026-04-28
refactor(hooks): remove hooks barrel imports

infin1t3:
## Summary
- Remove `src/hooks/index.ts` and replace `@/hooks` barrel imports with direct hook module imports.
- Update component/page/test imports to use `@/hooks/useX/useX` paths.
- Keep test mocks aligned with `component-testing.mdc`, using per-hook module mocks and `vi.hoisted` where shared mock state is needed.

Resolves: #1742

PR #1740, opened 2026-04-26
refactor(libs): remove libs barrels and use direct imports

infin1t3:
## Summary
- Removes the `src/libs` aggregate barrel and nested libs `index.ts` re-export files.
- Migrates app, tests, and docs to direct `@/libs/...` file imports.
- Syncs the branch with latest `dev` and keeps the existing `@/icons` alias convention.

## Validation
- `npm run format:check`
- `npm run lint`
- `npm run typecheck`
- `npm run test`

## Testing
Basic regression testing done

https://github.com/user-attachments/assets/a74b2fc7-bf61-4e6a-b517-e9695d039e72

Resolves #1727

PR #1756, opened 2026-04-29
refactor: remove miscellaneous barrel imports

### LINKED ISSUE #1755: Remove barrels from i18n, providers and test-utils folder

infin1t3:
## Summary
- Removed barrel files from `src/i18n`, `src/providers`, and `src/test-utils`.
- Replaced root barrel imports/mocks with direct module imports across app code and tests.
- Added Cypress TypeScript namespace support via the root `tsconfig.json` and updated docs examples.

Resolves: #1755

### DIFF
--- a/cypress/tsconfig.json
+++ b/cypress/tsconfig.json
@@ -1,4 +1,5 @@
 {
+  "extends": "../tsconfig.json",
   "compilerOptions": {
     "sourceMap": false,
     "outDir": "../dist/out-tsc",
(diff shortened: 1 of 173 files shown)


cypress/tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+  "extends": "../tsconfig.json",
```
infin1t3, line 2: So Cypress TypeScript inherits the app’s existing path aliases

Taewa (approved):
LGTM

I tested in the same way as I did previously (https://github.com/pubky/pubky-app/pull/1749, https://github.com/pubky/pubky-app/pull/1740)

Also used AI + running test locally + build the project. All passed.

PR #1766, opened 2026-05-03
chore: add automatic import sorting

### LINKED ISSUE #1748: Proposal: add an import sort ESLint rule/plugin
## Idea

Adopt a consistent import ordering rule across the monorepo to reduce noise in diffs and make import blocks easier to scan.

Two options to consider:

- [`eslint-plugin-simple-import-sort`](https://github.com/lydell/eslint-plugin-simple-import-sort) — opinionated, auto-fixable, handles both `import` and `export` ordering; minimal config.
- [`sort-imports`](https://eslint.org/docs/latest/rules/sort-imports) — built into ESLint, no extra dependency, but more limited (sorts within a single statement / by member, not across statement groups out of the box).

## Why

- Eliminates churn from differently-ordered imports across PRs.
- Auto-fixable, so it costs ~nothing in day-to-day work once configured.
- Keeps related imports grouped (builtin → external → internal → relative), which helps reviewers.

## Tradeoffs

- One-time formatting commit will touch a lot of files (noisy git blame for one line).
- Need to align with the existing import conventions in `AGENTS.md` / lint config so we don't fight other rules.

## Note

Just capturing the idea — not picking this up now.

infin1t3:
Adds the simple-import-sort ESLint rule (and related config) so imports stay consistent across the repo, and applies the resulting ordering where needed.

Resolves #1748


src/core/services/homegate/homegate.types.ts
```diff
@@ -4,6 +4,8 @@
  * Returns available false when geoblocked (403).
  * Returns available false with error true for generic errors (network failure, server error, etc.)
  */
```
Orlandohub, line 8: this comment should be under the import and close to the `TSmsInfoResult` type

src/components/atoms/Carousel/Carousel.test.tsx
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
-import { describe, it, expect, vi, beforeEach } from 'vitest';
-import { render, screen, fireEvent } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
+// Import components after mocking
```
Orlandohub: this comment is no longer valid

src/hooks/useCopyToClipboard/useCopyToClipboard.test.tsx
```diff
@@ -1,5 +1,7 @@
 import { renderHook } from '@testing-library/react';
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+// Import the hook after mocking
```
Orlandohub: this comment is no longer valid

Orlandohub (dismissed):

Orlandohub (approved):
LGTM

PR #1777, opened 2026-05-05
fix(feed): preserve custom feed scroll position on post navigation (#1561)

### LINKED ISSUE #1561: [Bug]: Navigating back from post to custom feed doesn't retain feed position
### Description

With a custom feed applied, navigating to a post and then navigating back reloads the custom feed at the top.

### Expected Behaviour

Navigating back from post should not reload the custom feed and should feed position. This is how it functions with the default feed.

### Steps to Reproduce

1. Create and apply a custom feed
2. Scroll down the feed
3. Click on some post to navigate to its post page
4. Use browser's back navigation
5. Observe custom feed reloads and position is at the top of the feed

### Screenshots

https://github.com/user-attachments/assets/7bb7d76d-e5c7-4816-a556-3f0c20631b6c

### Console Log

```shell

```

### Browser & Version

_No response_

Taewa:
## Summary
  Fixes #1561 — custom feed loses scroll position when navigating back from a post.

  ## Root cause                                                                                             
- The custom feed only knew which feed it was viewing by reading the URL.
- When the URL changed to `/post/...`, the feed lost its identity and dropped every post it had loaded
  into the timeline (the entire list of feed items in the feed page disappeared from the DOM).              
- An empty feed page is much shorter than a fully scrolled feed page, so the document height dropped from ~3000px to ~400px. When the user clicked back, the browser tried to restore the previous scroll position (e.g. 3000px), but the page was no longer that tall, so it clamped to whatever max was available — effectively snapping back near the top.

## Fix
  - **Parallel routes (mirroring `/home`)** — keep the feed component mounted while a post is open, instead of unmounting and remounting on back.
  - **Feed-id latch** — remember the last feed the user was on, so intercepted post navigation no longer makes the feed think it's "no feed at all."

## Why both
  - Parallel routes preserve the React tree.
  - The latch preserves the data inside that tree.                                                          
  - Either alone is insufficient — the bug returns.

  ## Testing                                                                                                
- Unit tests added for the latch behavior and layout slot toggling.
- Manual smoke: scroll feed → open post → back → scroll position retained, no reload, no flicker.

## Notes
  - No change to `/home` behavior.
  - Direct visits to `/post/...` and other unrelated routes unaffected.
  
## Result
https://github.com/user-attachments/assets/531849ad-f928-401c-9d98-10a2cf24e20a


infin1t3 (approved):
Tested on `7f0f1f64ae32974de28220e965adc7b37ab98087` - LGTM. Approved.

PR #1779, opened 2026-05-06
tests: fix e2e

catch-21:


catch-21:
passing e2e: https://github.com/pubky/pubky-app/actions/runs/25431160564

Issue #1721, opened 2026-04-23
[Bug]: Tags don't show selected state and have incorrect behavior


aldertnl:
### Description

The tags you supported or added yourself don’t always accurately reflect the current state. Even though my user avatar is clearly visible, the tag is not displayed in the correct selected state (colored outline). Moreover, when the tag is clicked, a toast message states that the tag has been added instead of deleted. As a result, the tag persists, and then the selected state is actually triggered, so finally, when clicking the tag once more, the tag is actually removed.

### Expected Behaviour

When a user has added a tag or supported a tag the user expects to see the selected state for a tag (colored outline) + removal behavior when tag is clicked.

### Steps to Reproduce

1. Go to a post where your avatar is shown next to tags
2. Note that the tag is not shown in selected state
3. Clicking the tag 'adds' it again in instead of removing it

### Screenshots

https://github.com/user-attachments/assets/562443c4-8b85-4abf-b583-bd2444c92252

### Console Log

```shell

```

catch-21:
Good catch. We must add a component test for this one and extend an E2E test to cover it also.

catch-21:
Verified in 000568be044b7e5225723ee831aec0632427bc9a
However, there are edge cases being tracked by https://github.com/pubky/pubky-app/issues/1781

PR #1782, opened 2026-05-06
Regression test coverage for untagging bug #1721

catch-21:
E2E regression test coverage of https://github.com/pubky/pubky-app/issues/1721
Can be merged once we have a fix


catch-21:
Passing E2E: https://github.com/pubky/pubky-app/actions/runs/25505652735

Issue #1781, opened 2026-05-06
fix(tags): Nexus refetch reverts just-toggled tag


Taewa:
> **Status update (2026-05-11)**
> Implemented as a sessionStorage-based marker mechanism that survives F5 refresh, with no IndexedDB schema change. The long-term solution (multi-device drift) is still tracked in #1800. Earlier plans considered (IndexedDB marker, in-memory Map, SSE) are kept at the bottom of this issue for reference.

## Problem

Two separate problems, easy to confuse with each other.

### 1. Race: a stale Nexus response overwrites a fresh local mutation

`usePostTags.fetchInitialTags` fires a Nexus request on every mount. If the user toggles a tag while that request is in flight, the (now stale) response comes back and `LocalPostTagService.mergeTags` writes it into IndexedDB, undoing the local change.

Repro on `/home`:

1. A post is rendered with one of the viewer's tags highlighted.
2. Viewer clicks the tag to remove it. Local commit succeeds, UI updates correctly.
3. About one second later, the in-flight Nexus response from the page mount returns. Nexus has not indexed the homeserver delete yet, so it replies with `relationship: true` and viewer in `taggers`. `mergeTags` writes that into IndexedDB.
4. The tag is highlighted again.
5. If the user clicks again to remove it, the homeserver returns 404 because the tag is already deleted there.

QA team has flagged the visible flicker as a high-priority bug.

### 2. Drift: `mergeTags` cannot accept "the data went down"

Even when the race does not happen, once IndexedDB and Nexus disagree, the local cache is stuck. Refreshing later — even after Nexus has fully caught up — does not fix it.

Two lines in `src/core/services/local/tag/post/tag.post.ts` are the structural cause:

```ts
// line 175 — union of taggers
const mergedTaggers = [...new Set([...(existingTag.taggers ?? []), ...(newTag.taggers ?? [])])];

// line 180 — max of counts
taggers_count: Math.max(existingTag.taggers_count ?? 0, newTag.taggers_count ?? mergedTaggers.length),
```

- The **union** of taggers means any later Nexus response that still lists the viewer will re-add the viewer, even after the viewer removed themselves. (If a tagger appears on either side, they stay — removals never carry over.)
- `Math.max` on the count means the count never goes down, even when Nexus says it should.
- Also, `mergeTags` never deletes a label from IndexedDB if Nexus omits it (sole-tagger delete case — deferred to a follow-up PR).

Concrete example for the count problem:

```
[...]

PR #1783, opened 2026-05-06
fix(tags): forward viewer_id when fetching post tags (#1721)

### LINKED ISSUE #1721: [Bug]: Tags don't show selected state and have incorrect behavior

Taewa:
## Summary
Fixes #1721 — tags the viewer has already supported or added now show as selected on first load, and clicking them removes them in one click as expected.

## Root cause
- The post-tags request to Nexus omitted the viewer's identity, so the server replied with "viewer hasn't tagged this" for every tag.
- That stale answer got merged into local storage and overwrote the correct state from the viewer's earlier tag actions.
- Result: a tag the viewer had personally added rendered without the selected outline, and the first click was interpreted as "add" (toast: "added"; tag was re-added) instead of "remove" — needing a second click to actually remove.

## Fix
- **Pass the viewer's id when fetching post tags** — Nexus then returns the correct selected state, matching the user-tags path that already does this.
- **Pass it through every layer** — the request, the controller, the application, and the service all carry the viewer's id now.

## Testing
- Updated existing unit tests to assert the viewer id is included in fetch calls.
- Manual smoke: on a single post page where the viewer's avatar appears next to a tag, the tag now shows the selected outline on initial load and refresh; one click removes it and shows the "removed" toast.

## Notes
- `viewerId` is in the effect's deps but the `hasFetched` guard skips a re-fetch when the viewer changes mid-view (e.g. logging in on the same page). Same as the user-tags hook (`useTagged.tsx:82–87`); #1721 only covers the initial-load case.
- Found a separate bug while testing: a slow server reply can briefly undo a tag click. Tracked in #1781.

### DIFF
--- a/src/core/application/post/post.ts
+++ b/src/core/application/post/post.ts
@@ -86,8 +86,8 @@ export class PostApplication {
    * @param limit - Maximum number of tags to return
    * @returns Array of tags from Nexus
    */
-  static async fetchTags({ compositeId, skip, limit }: TFetchMorePostTagsParams): Promise<NexusTag[]> {
-    const nexusTags = await NexusPostService.getPostTags({ compositeId, skip, limit });
+  static async fetchTags({ compositeId, skip, limit, viewerId }: TFetchMorePostTagsParams): Promise<NexusTag[]> {
+    const nexusTags = await NexusPostService.getPostTags({ compositeId, skip, limit, viewerId });
 
     // Persist new tags to local DB (merge with existing)
     if (nexusTags.length > 0) {
--- a/src/core/controllers/post/post.ts
+++ b/src/core/controllers/post/post.ts
@@ -115,8 +115,8 @@ export class PostController {
    * @param params.limit - Maximum number of tags to return
    * @returns Array of tags from Nexus
    */
-  static async fetchTags({ compositeId, skip, limit }: TFetchMorePostTagsParams): Promise<NexusTag[]> {
-    return await PostApplication.fetchTags({ compositeId, skip, limit });
+  static async fetchTags({ compositeId, skip, limit, viewerId }: TFetchMorePostTagsParams): Promise<NexusTag[]> {
+    return await PostApplication.fetchTags({ compositeId, skip, limit, viewerId });
   }
 
   /**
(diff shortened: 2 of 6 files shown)


Taewa:
@greptile-apps
Review again plz

infin1t3 (approved):
Tested at - `1c877c6563bcb8a8f4c7263f031d035612f51f58` - LGTM, given the documented follow-up items.

https://github.com/user-attachments/assets/954e6c8d-ef95-46f7-8532-95462cb52544

catch-21:
Testing with 000568be044b7e5225723ee831aec0632427bc9a

We can currently find ourselves in a situation where cannot untag, even after page reload. Reproduce steps:

1. Find a post on feed that you've tagged
2. Untag it
3. Navigate to its post page
4. Observe it thinks you've still tagged it
5. Attempt to untag again
6. Observe it prevents you untagging
7. Navigate back to feed
8. Repeat 4-6
9. Reload page
10. Repeat 4-6

https://github.com/user-attachments/assets/93d197be-3ce8-4e5d-8210-f921511c9870

catch-21:
Also, the following scenario where newly added tag is not highlighted:

1. Find a post in feed that has existing tags not by you
2. Click on an existing tag to make it highlighted and increment its counter
3. Click on the post to navigate to its page
4. Observe the tag you clicked in not highlighted (bad)
5. Click the tag again
6. Observe it becomes highlighted and toast says tag is added (bad)

https://github.com/user-attachments/assets/1c341c37-d90b-49b8-a87b-ac36d72a938b

PR #1787, opened 2026-05-06
fix(core): cross-device mute list sync via homeserver event stream

### LINKED ISSUE #1236: [Bug]: Unmuting from one instance doesn't reflect on other instances
### Description

If I have two instances of pubky-app both signed in with the same user (e.g. one desktop and one mobile), and I mute someone or unmute some users, this does not reflect on the other session. There appear to be no ttl for this settings.

### Expected Behaviour

Any changes made in pubky-app should reflect across all instances within a reasonable amount of time. This issue is specifically for muting but there could be other settings to address here or in a separate issue.

### Steps to Reproduce

1. Have two isolated instances of pubky-app signed in with the same user
2. Mute some users from one instance
3. Wait
4. Observe users are not muted on the other pubky-app instance

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
Fixed - Any changes made in pubky-app reflect across all instances almost instantly.

Add MuteListSyncCoordinator (debounced fetchMutedUsers on mute-directory SSE). Wire CoordinatorsManager; subscribe via HomeserverService eventStreamForUser. Remove Nexus client mute routes from user stream/bootstrap; document in ADRs, architecture, and environment.

Resolves #1236


infin1t3:
@greptileai review

infin1t3:
@greptileai re-review please

infin1t3:
@greptileai review please

infin1t3:
@Taewa I promise I will stop assigning PRs to you for a while after this one :)

catch-21:
Seems to work well. Tested across Chrome on MacOS and Brave on Andriod. Muting and unmuting users updates the feed and muted users list in real time.

src/core/application/mute/mute.ts
```diff
@@ -58,4 +59,16 @@ export class MuteApplication {
     });
     return stream;
   }
+
+  /**
+   * Homeserver `/events-stream` (live) for mute directory changes only.
+   * @see HomeserverService.subscribeUserEventStreamForPath
+   */
+  static subscribeMuteDirectoryEventStream(pubky: Pubky, cursor: string | null): Promise<ReadableStream> {
```
Taewa: A feedback with AI. 

[1] high | external-data-normalized-through-pipes | src/core/application/mute/mute.ts:67
What's wrong: The raw homeserver SDK ReadableStream passes through the Application/Controller layers unchanged and reaches the Coordinator. As a result, the Coordinator directly knows SDK/WASM details such as eventType, cursor, and free().

Why it matters: When external data and SDK lifecycle details leak across service boundaries, type and normalization responsibilities become unclear. If the SDK event shape changes, the Coordinator will break directly.

Fix: Wrap the stream as a domain event type at the HomeserverService or Application boundary, and handle free() internally. The Controller should return only something like ReadableStream<MuteDirectoryEvent>.

src/core/controllers/mute/mute.ts
```diff
@@ -34,4 +34,11 @@ export class MuteController {
   static async fetchMutedUsers(pubky: Pubky): Promise<Pubky[]> {
     return MuteApplication.fetchMutedUsers(pubky);
   }
+
+  /**
+   * Live SSE (via SDK) for `/pub/pubky.app/mutes/` changes on the homeserver.
+   */
+  static subscribeMuteDirectoryEventStream(pubky: Pubky, cursor: string | null): Promise<ReadableStream> {
```
Taewa: A feedback with AI.
For me the prefix `subscribe...` makes sense so updating document would be enough.

[2] medium | controller-naming | src/core/controllers/mute/mute.ts:41
What's wrong: subscribeMuteDirectoryEventStream introduces a new prefix outside the current controller naming convention (fetch*, get*, commit*).

Why it matters: This breaks the project convention where callers can infer whether a method is network-only, read-only, or a write operation from its name.

Fix: Rename it to use an existing prefix, such as fetchMuteDirectoryEventStream, or explicitly document a subscribe* exception for live streams in the architecture docs/rules.

src/core/coordinators/mute-list-sync/mute-list-sync.ts
```diff
@@ -0,0 +1,339 @@
...
+      const isAuthenticated = state.selectIsAuthenticated();
+      const wasAuthenticated = prevState.selectIsAuthenticated();
+      const profileChanged = state.hasProfile !== prevState.hasProfile;
+      const userChanged = state.currentUserPubky !== prevState.currentUserPubky;
+      if (isAuthenticated !== wasAuthenticated || profileChanged || userChanged) {
+        this.evaluateStreaming();
+      }
+    });
+
+    if (this.respectPageVisibility && typeof document !== 'undefined') {
```
Taewa, line 122: A feedback with AI
It's low-tier but I think it's worth to be covered by test.

[3] low | missing-tests | src/core/coordinators/mute-list-sync/mute-list-sync.ts:117
What's wrong: The new coordinator tests do not cover the branch that pauses/resumes the mute stream based on page visibility.

Why it matters: This path involves a long-running stream and timer cleanup, so regressions could cause stream leaks or missed resubscriptions.

Fix: Add tests that switch visibilityState between hidden and visible, then verify that the stream is canceled and resubscribed.

src/config/mute-sync.ts
```diff
@@ -0,0 +1,42 @@
...
+export const MUTE_SYNC_DEBOUNCE_MS = 500;
+
+/** Prefix for `sessionStorage` keys storing the last HS events-stream cursor per user. */
+export const MUTE_SYNC_CURSOR_STORAGE_PREFIX = 'pubky-app:mute-sync-cursor:';
+
+/**
+ * Drops homeserver event-stream cursors for mute sync. Session storage survives logout until the tab closes;
+ * clearing on sign-out avoids leaving per-user cursor state in an open tab.
+ */
+export function clearMuteSyncCursorSessionStorage(): void {
```
Taewa: Usually `config` contains configurable constants (1 exception in `theme.ts` as a helper) but `clearMuteSyncCursorSessionStorage` updates states.

src/core/coordinators/mute-list-sync/mute-list-sync.test.ts
```diff
@@ -0,0 +1,260 @@
...
+    const coordinator = MuteListSyncCoordinator.getInstance();
+    coordinator.setRoute(APP_ROUTES.HOME);
+    coordinator.start();
+
+    await flushPromises();
+    await vi.advanceTimersByTimeAsync(MUTE_SYNC_DEBOUNCE_MS);
+    expect(fetchMuted).toHaveBeenCalledTimes(1);
+  });
+
+  it('retries failed mute refreshes and stores the cursor after success', async () => {
```
Taewa, line 99: This retry logic has "max attempts" number? If it's not the case, if HS keeps returning 4xx/5xx errors the client will call infinite times which can be burden for client & server. 
How about either update code this time or create an improvement ticket and update it in the future?

src/config/mute-sync.ts
infin1t3: I was not sure where to put this honestly, I now made a new folder for this in `libs` for this. Lmk if you have better ideas.

src/core/coordinators/mute-list-sync/mute-list-sync.test.ts
infin1t3: I am not sure about this one. Question we need to answer is - what are we protecting from?

How system currently works is - if HS throws, and we are in a state where mute sync is supposed to stay active - we timeout for 1 second and then retry.

Which is not ideal for the server from a load balancing standpoint, but I also don't think that having "max attempts" is any better, because that is just going to stop at some point and if HS later recovers, we will not attempt to reconnect.

Imo neither option is good here. Maybe we do error filtration and have use cases what to do in case of 4xx and what in case of 5xx?

Taewa: Yeah both ways have pros and cons. 
I am seeing this one as "Infinite call burden for server risk" VS "Users see muted posts/users risk".
How about creating a follow-up ticket and we discuss with other members?

Taewa (approved):
LGTM

src/core/coordinators/mute-list-sync/mute-list-sync.test.ts
infin1t3: Will ping Tomos in Slack and add you to the conversation. After that, will create a follow up ticket.

PR #1797, opened 2026-05-07
feat(users): refresh who-to-follow recommendations after follow

### LINKED ISSUE #1224: [Feature]: Refresh 'Who to follow' when any listed user is now being followed
### Problem Statement

On sign in, the 'Who to follow' list appears to always show users you do not yet follow (good). If any of these are then followed, the list never updates to suggest different users.

### Proposed Solution

If any of the listed used in 'Who to follow' is now being followed, refresh the list on navigating to home/hot/bookmarks page.

### Suggested Priority

infin1t3:
## Summary
- Refreshes Who to Follow recommendations after follow actions by removing followed users from the local recommended stream and refilling from Nexus when the candidate pool gets low.
- Keeps newly followed users visible with the followed checkmark until navigation/remount, avoiding immediate list removal and UI blink.

## Video walkthrough

https://github.com/user-attachments/assets/8274cb02-a699-4853-a32b-78fbd67b720e

Resolves: #1224


src/core/controllers/stream/users/users.ts
```diff
@@ -54,7 +57,40 @@ export class StreamUserController {
...
 
-    return { nextPageIds, skip: nextSkip };
+    return { nextPageIds, skip: nextSkip, isExhausted };
+  }
+
+  /**
+   * Fetch a slice of a user stream directly from Nexus.
+   * This bypasses cache reads but still hydrates missing user details.
+   */
+  static async fetchStreamSlice({
```
Orlandohub: According to project rules at `AGENTS.md`:

```
### Controller naming

- `fetch*` — network only, no cache
...
```

`fetchStreamSlice` is a fetch* controller API whose implementation path still consults local cache in the application layer here:

```
const cachedStream = await LocalStreamUsersService.findById(streamId);
```

Therefore the name should be something like:
`refreshStreamSlice` / `fetchAndCacheStreamSlice`

src/hooks/useUserStream/useUserStream.ts
```diff
@@ -232,6 +287,29 @@ export function useUserStream({
...
+  useEffect(() => {
+    if (!excludeFollowing || isLoading || isLoadingMore || isExhausted || refillAttemptedRef.current) return;
+    if (userIds.length === 0) return;
+
+    const shouldRefill = eligibleCount < effectiveRefillThreshold || (!paginated && eligibleCount < effectiveLimit);
+
+    if (!shouldRefill) return;
+
+    refillAttemptedRef.current = true;
+    void fetchStreamSlice(false, { forceNetwork: true });
```
Orlandohub, line 299: I'm not sure but please double check if there's a possibility of unnecessary force-network refill on normal loads, because `eligibleCount` can briefly be 0 since `useLiveQuery` maps start as empty defaults while details/relationships resolve.

Potential negative effects:

- More network traffic than intended, especially on sidebar mounts.
- The recommended stream cache may get appended with extra candidate IDs earlier than needed.
- isLoadingMore can briefly become true, which may show load-more skeletons on the full Who To Follow page.
- It can advance skipRef after the extra fetch, so the next real loadMore() starts later than it would have.
- In normal cases it probably does not visibly break the list, because duplicate IDs are filtered and users are appended, but it weakens the “bounded refill only when low” behavior.

Orlandohub (dismissed):
Nothing major but please double check comments

Orlandohub (approved):
LGTM

PR #1807, opened 2026-05-11
fix(tags): protect viewer's toggled tag from stale Nexus merges (#1781)

### LINKED ISSUE #1781: fix(tags): Nexus refetch reverts just-toggled tag

Taewa:
## Summary
  Fixes a tagging bug where a user's freshly added or removed tag could "snap back" after a Nexus refresh,
  because Nexus's indexer is several minutes behind the homeserver write. Closes #1781.

  ## Root cause
  - After a user toggles a tag, the next Nexus response for that post is often still serving pre-toggle stale data
   (indexer lag, ~minutes).
  - The old merge step blindly took the union of local + Nexus and used `max(local, nexus)` for the count,
  so a stale Nexus response could resurrect a just-deleted tag or undercount a just-added one.
  - Separately, a delete that returned 404 from the homeserver was treated as a failure and rolled back
  locally, leaving the user with a "ghost" tag they couldn't remove (the homeserver kept 404-ing on every
  retry).

  ## Fix
  - **Viewer-mutation marker** — when the viewer tags or untags a post, drop a short-lived (~5 min) marker
  in sessionStorage; the merge step trusts the marker over Nexus for that one viewer slot.
  - **Per-field merge policy** — count comes from Nexus (with ±1 correction when a marker disagrees), the
  viewer's slot in the tagger list follows the marker, other taggers stay as a union (Nexus only ships a
  truncated top-N), labels not in the Nexus response are left alone (they may be on a later page).
  - **Treat homeserver 404 on delete as success** — the tag is already gone, no rollback, no ghost.
  - **Logout sweep** — clear the current user's markers before the next user signs in on the same tab.

  ## Notes
  - Scope is post tags only — user tags and other entity tags are unchanged.
  - Marker TTL (`MARKER_TTL_MS`, 5 min) is a tunable in `src/config/viewerTagMarker.ts`; pick the indexer
  lag ceiling, not a "freshness" budget.
  - Markers live in sessionStorage, so they're per-tab and naturally gone on tab close; the logout sweep
  covers same-tab user switches.
  
  ## Result
  
### 1.

https://github.com/user-attachments/assets/87215b95-d99b-4e3c-9c5a-2b8f9e1267df

### 2.

https://github.com/user-attachments/assets/53fc7af9-8819-45b2-b81d-36f43704e3f1

### DIFF
--- a/src/core/services/local/tag/post/tag.post.ts
+++ b/src/core/services/local/tag/post/tag.post.ts
@@ -2,12 +2,14 @@ import { db } from '@/database/franky/franky';
 import { DatabaseErrorCode } from '@/libs/error/error.codes';
 import { Err } from '@/libs/error/error.factories';
 import { ErrorService } from '@/libs/error/error.types';
+import { HttpMethod } from '@/libs/http/http.types';
 import { Logger } from '@/libs/logger/logger';
 import type { Pubky } from '@/models/models.types';
 import { PostCountsModel } from '@/models/post/counts/postCounts';
 import { PostTagsModel, type PostTagsModelSchema } from '@/models/post/tags/postTags';
 import { PostTtlModel } from '@/models/post/ttl/postTtl';
 import { UserCountsModel } from '@/models/user/counts/userCounts';
+import { ViewerTagMarkerStorage } from '@/services/local/tag/post/viewerTagMarkerStorage';
 import type { TLocalTagParams } from '@/services/local/tag/tag.types';
 import type { NexusTag } from '@/services/nexus/nexus.types';
 
@@ -33,8 +35,14 @@ export class LocalPostTagService {
    * @throws {DatabaseError} When database operations fail
    */
   static async create({ taggedId: postId, label, taggerId }: TLocalTagParams): Promise<boolean> {
+    // True only when the transaction actually changed IndexedDB state.
+    // We use this to decide whether to write the viewer-mutation marker after
+    // the transaction commits: if nothing changed (e.g., the user already had
+    // this tag), there's no local state to protect from stale Nexus responses,
+    // so we skip the marker.
+    let mutated = false;
     try {
-      return await db.transaction('rw', this.TAG_TABLES, async () => {
+      mutated = await db.transaction('rw', this.TAG_TABLES, async () => {
         const postTagsModel = await PostTagsModel.getOrCreate<string, PostTagsModelSchema>(postId);
         const status = postTagsModel.addTagger(label, taggerId);
         // Idempotent: user already tagged this post with this label
@@ -57,6 +65,14 @@ export class LocalPostTagService {
         cause: error,
       });
     }
+
+    if (mutated) {
+      // Record this viewer change so mergeTags ignores stale Nexus responses
+      // for the next ~5 minutes (until Nexus catches up).
+      ViewerTagMarkerStorage.set({ pubky: taggerId, postId, label, op: HttpMethod.PUT });
+    }
+
+    return mutated;
   }
 
   /**
@@ -95,7 +111,6 @@ export class LocalPostTagService {
         await UserCountsModel.updateCounts({ userId: taggerId, countChanges: { tagged: -1 } });
         await PostTtlModel.upsert({ id: postId, lastUpdatedAt: Date.now() });
       });
-      return true;
     } catch (error) {
       throw Err.database(DatabaseErrorCode.WRITE_FAILED, 'Failed to delete post tag', {
         service: ErrorService.Local,
@@ -104,6 +119,11 @@ export class LocalPostTagService {
         cause: error,
       });
     }
+
+    // Record this viewer change so mergeTags ignores stale Nexus responses
+    // for the next ~5 minutes (until Nexus catches up).
+    ViewerTagMarkerStorage.set({ pubky: taggerId, postId, label, op: HttpMethod.DELETE });
+    return true;
   }
 
   /**
@@ -147,13 +167,29 @@ export class LocalPostTagService {
   }
 
   /**
-   * Merges new tags from Nexus with existing local tags.
-   * Updates existing tags or adds new ones without removing any.
+   * Merges Nexus tags into local IndexedDB using a per-field policy:
    *
-   * @param postId - Unique identifier of the post
-   * @param tags - Array of NexusTags to merge
+   * - `taggers_count`: replaced with the Nexus value (Nexus is authoritative
+   *   for the total).
+   * - Viewer's `relationship` and the viewer's entry in `taggers`: an active
+   *   sessionStorage marker (set by `create` / `delete`) overrides; otherwise
+   *   the Nexus value is used.
+   * - Other users in `taggers`: union of existing + Nexus. The Nexus `taggers`
+   *   array is a truncated top-N sample, so replacing would drop locally-cached
+   *   taggers.
+   * - Labels in IndexedDB but not in `tags`: left alone, since Nexus paginates
+   *   by label and absent labels may exist on later pages.
+   *
+   * @param postId - Composite post ID
+   * @param tags - Tags from the Nexus response
+   * @param viewerId - Current viewer pubky for marker lookup. Null when the
+   *   user is unauthenticated, in which case Nexus values are used as-is.
    */
-  static async mergeTags({ postId, tags }: { postId: string; tags: NexusTag[] }) {
+  static async mergeTags({ postId, tags, viewerId }: { postId: string; tags: NexusTag[]; viewerId: Pubky | null }) {
+    // Side-effect GC: drop expired markers across the board so sessionStorage
+    // doesn't accumulate them.
+    ViewerTagMarkerStorage.sweepExpired();
+
     try {
       await db.transaction('rw', [PostTagsModel.table], async () => {
         const existing = await PostTagsModel.findById(postId);
@@ -170,18 +206,40 @@ export class LocalPostTagService {
           const key = newTag.label.toLowerCase();
           const existingTag = tagMap.get(key);
 
-          if (existingTag) {
-            // Merge taggers - combine unique taggers
-            const mergedTaggers = [...new Set([...(existingTag.taggers ?? []), ...(newTag.taggers ?? [])])];
-            tagMap.set(key, {
-              ...existingTag,
-              ...newTag,
-              taggers: mergedTaggers,
-              taggers_count: Math.max(existingTag.taggers_count ?? 0, newTag.taggers_count ?? mergedTaggers.length),
-            });
-          } else {
-            tagMap.set(key, newTag);
+          // If the viewer just toggled this tag locally, the marker tells us
+          // the intended viewer-state — trust it over a possibly stale Nexus
+          // value for the next ~5 minutes.
+          const marker = viewerId ? ViewerTagMarkerStorage.get({ pubky: viewerId, postId, label: newTag.label }) : null;
+          const nexusSaysViewerIsTagger = Boolean(newTag.relationship);
+          const viewerIsTagger = marker ? marker.op === HttpMethod.PUT : nexusSaysViewerIsTagger;
+
+          // Other users: union of existing + Nexus, viewer excluded
+          // (viewer's slot is decided by `viewerIsTagger`).
+          const otherTaggers = new Set(
+            [...(existingTag?.taggers ?? []), ...(newTag.taggers ?? [])].filter((tagger) => tagger !== viewerId),
+          );
+          const mergedTaggers: Pubky[] = Array.from(otherTaggers);
+          if (viewerIsTagger && viewerId) {
+            mergedTaggers.push(viewerId);
           }
+
+          // Default: trust Nexus's count.
+          // Exception: if the marker says the viewer is a tagger but Nexus
+          // doesn't know yet, Nexus's count is one short — add 1. If the marker
+          // says the viewer is NOT a tagger but Nexus still counts them, Nexus's
+          // count is one too high — subtract 1. This keeps `taggers_count`
+          // matching the final `taggers` / `relationship` we write below.
+          let taggers_count = newTag.taggers_count;
+          if (marker && viewerIsTagger !== nexusSaysViewerIsTagger) {
+            taggers_count = viewerIsTagger ? taggers_count + 1 : Math.max(0, taggers_count - 1);
+          }
+
+          tagMap.set(key, {
+            ...newTag,
+            taggers_count,
+            taggers: mergedTaggers,
+            relationship: viewerIsTagger,
+          });
         }
 
         // Convert map back to array
--- a/src/core/services/local/tag/post/tag.post.ts
+++ b/src/core/services/local/tag/post/tag.post.ts
@@ -31,6 +31,7 @@ export class LocalPostTagService {
    * @param params.label - Normalized tag label (must be pre-normalized by caller)
    * @param params.taggerId - Unique identifier of the user adding the tag
    *
+   * @returns {boolean} true if local state changed; false if the tagger already had this tag (idempotent — no writes, no viewer marker)
    * @throws {AppError} When user has already tagged this post with the same label
    * @throws {DatabaseError} When database operations fail
    */
--- a/src/core/services/local/tag/post/tag.post.ts
+++ b/src/core/services/local/tag/post/tag.post.ts
@@ -186,8 +186,7 @@ export class LocalPostTagService {
    *   user is unauthenticated, in which case Nexus values are used as-is.
    */
   static async mergeTags({ postId, tags, viewerId }: { postId: string; tags: NexusTag[]; viewerId: Pubky | null }) {
-    // Side-effect GC: drop expired markers across the board so sessionStorage
-    // doesn't accumulate them.
+    // Piggyback GC: drop expired markers so sessionStorage doesn't accumulate them.
     ViewerTagMarkerStorage.sweepExpired();
 
     try {
--- a/src/core/services/local/tag/post/viewerTagMarkerStorage.ts
+++ b/src/core/services/local/tag/post/viewerTagMarkerStorage.ts
@@ -19,18 +19,6 @@ function userPrefix(pubky: Pubky): string {
   return `${VIEWER_TAG_MARKER_STORAGE_PREFIX}${pubky}:`;
 }
 
-function isStorageAvailable(): boolean {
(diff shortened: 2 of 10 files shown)


Taewa:
@greptile-apps
Re: the load-bearing invariant — I considered a `mutateAndMark(op, fn)` wrapper but skipped it: only two call sites, and `create` / `delete` differ enough (idempotent early-return vs not) that a shared wrapper would hurt readability. 
A missing marker would also break the `mergeTags` tests immediately, so the risk is bounded. Happy to revisit if a third mutation path appears.

Taewa:
@greptile-apps
I also updated `sweepExpired` related code. Review this PR again plz

src/core/services/local/tag/post/viewerTagMarkerStorage.ts
```diff
@@ -0,0 +1,160 @@
...
+
+function buildKey(pubky: Pubky, postId: string, label: string): string {
+  return `${VIEWER_TAG_MARKER_STORAGE_PREFIX}${pubky}:${postId}:${label.toLowerCase()}`;
+}
+
+function userPrefix(pubky: Pubky): string {
+  return `${VIEWER_TAG_MARKER_STORAGE_PREFIX}${pubky}:`;
+}
+
+function isStorageAvailable(): boolean {
```
infin1t3: This is a Next.js app but we do not use any of the SSR functionalities, given this is an offline first app. So `window` and storage should always be available and I am not sure we need this check.

src/core/services/local/tag/post/viewerTagMarkerStorage.ts
```diff
@@ -0,0 +1,160 @@
...
+  return typeof window !== 'undefined' && typeof window.sessionStorage !== 'undefined';
+}
+
+function safeRemove(key: string): void {
```
infin1t3: Not sure we need this wrapper function, instead of calling `window.sessionStorage.removeItem(key);` directly.

Especially since a bit down we use other similar methods without a wrapper - `window.sessionStorage.getItem(key);`

And even if we do want to have some wrapper for storages, may be better to make that into a separate util.

src/core/services/local/tag/post/tag.post.ts
infin1t3, line 35: @returns {boolean} true if local state changed; false if the tagger already had this tag (idempotent — no writes, no viewer marker)

infin1t3 (changes requested):

src/core/services/local/tag/post/tag.post.ts
Taewa: Done

src/core/services/local/tag/post/viewerTagMarkerStorage.ts
Taewa: Done

src/core/services/local/tag/post/viewerTagMarkerStorage.ts
Taewa: Done

infin1t3 (approved):
Tested at - `8f0be7ee85183f46d8fe276a5103ca84eb2f9cab` - LGTM. Approved.
https://github.com/user-attachments/assets/e9604c87-9990-41c7-b3ec-8d0c99d0e65b

PR #1809, opened 2026-05-13
feat(profile): align other-user profile views

### LINKED ISSUE #1716: [Feature]: Align mobile and desktop views at other users profile page
### Problem Statement

At `/profile/[pubky]` we've two different templates being rendered based on device viewport. If mobile, we render a profile view, if desktop we render the profile posts view.

This is far from optimal and raises some inconsistencies:
- when sharing a profile from a mobile view, someone on desktop will experience a completely different UI
- loading different templates based on css still loads both templates no matter what which is bad for performance

### Proposed Solution

We'll use only the `ProfilePagePosts` template but in mobile we'll keep the current profile card at the top followed by the posts.

The posts link/tab on the profile page will navigate to the `/profile/[pubky]` route while in mobile it will scroll down past the profile card to the posts section.

This means we won't be making use of `/profile/[pubky]/posts` route anymore.

https://github.com/user-attachments/assets/f55b2147-a9e8-4c15-a72f-99bf6da27b96

### Others

Since we're on it, make sure the profile page UI is up to date with Figma designs, specially the mobile profile card section and navbar tabs order

**Figma mobile profile page:**
https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=27624-127175&m=dev

**Figma mobile profile posts page:**
https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=27624-127201&m=dev

**Figma desktop profile page**
https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=20307-75987&m=dev

### Suggested Priority

Orlandohub:
## Summary
- fix #1716
- Aligns mobile and desktop other-user profile pages around the canonical `/profile/[pubky]` posts view.
- Keeps the mobile profile summary above posts while avoiding separate viewport-specific templates.
- Preserves legacy `/profile/[pubky]/posts` links by redirecting them to the canonical route.

## Changes
- Render `ProfilePostsPage` directly from `/profile/[pubky]` on all viewports.
- Redirect `/profile/[pubky]/posts` to `/profile/[pubky]`.
- Update profile navigation so the Posts tab targets the canonical route and scrolls to posts on mobile.
- Adjust mobile profile tab ordering, search/profile layout behavior, tests, and snapshots for the aligned UI.

https://github.com/user-attachments/assets/9a82822b-6894-4044-a0cc-c379b47bbe58


Orlandohub:
@greptile-apps review

infin1t3:
https://github.com/user-attachments/assets/16fbbcce-4282-486e-8cd1-4ccc8c03ff8b

Not sure if this should be in scope for this PR, so just flagging it here.

After aligning the Profile and Posts mobile pages with the design, there’s now a noticeable difference in the top spacing between the first three cards on the Profile mobile view and the rest of the cards. It might be a good opportunity to adjust the spacing on the other cards as well so everything looks more uniform - but I’ll leave that up to you.

infin1t3 (changes requested):
Check [this branch](https://github.com/pubky/pubky-app/pull/1825), it was easier for me to show in code rather than to type. 

Highlights are:
* useLayoutEffect logic and ProfilePageLayout component are now simplified with scrollIntoView
* We rely on forwardRef instead on query selector
* Moved redirect logic into next.config (similar to my related PR)
* Minor cleanup - dead key removal, button refactor

I tested locally and seems like it works properly. Feel free to pick what you agree with from this branch.

Orlandohub:
@infin1t3 I've accepted your changes and I do agree my approach was way more complex, thanks for the fix!

I've added a small patch to your solution since when navigating away from profile posts on mobile to another profile page like followers it would keep some unnecessary vertical scroll.

Apart from that I think we got to a good solution

infin1t3 (approved):
LGTM - tested at `c1cd2a8e6a99d64e874a0524ad653238048392f6` - approved.

https://github.com/user-attachments/assets/c8eecb85-602d-4a5b-8d4b-82d9238bce4c

PR #1812, opened 2026-05-14
fix(ui): add discovery not-found for missing posts and invalid profiles

### LINKED ISSUE #1769: [Bug]: not found post url shows "loading post..." forever
### Description

When a post url is loaded and the post id is not available we see "loading post..." in the UI indefinitely.

### Expected Behaviour

There should be a "Post not found" page to support 404 response.

### Steps to Reproduce

1. Go to some made up post page like https://pubky.app/post/operrr8wsbpr3ue9d4qj41ge1kcc6r7fdiy6o3ugjrrhi4y77rdo/003337FCVQNEZ
2. Observe post page shows "loading post..." and never indicates the post cannot be found

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

### LINKED ISSUE #1796: [Bug]: Blank profile page shown when not found
### Description

There's no placeholder when viewing a profile page that does not exist.

### Expected Behaviour

We apply the "User Not Found" oage placeholder https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=34902-94749&t=h3ChgYCxLSxkYvRN-1

### Steps to Reproduce

1. Visit some user's profile
2. Edit url to some unknown pubky
3. Observe profile page without placeholder

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
Add discovery not-found for missing posts and invalid profiles

Resolves #1796
Resolves #1769

https://github.com/user-attachments/assets/68700d62-aff5-46d6-831b-8d8f01488da0


infin1t3:
@greptileai review again

src/components/molecules/UserNotFound/UserNotFound.tsx
```diff
@@ -1,21 +1,53 @@
 'use client';
 
-import { UserX } from 'lucide-react';
+import { ArrowLeft, Tag, UserX } from 'lucide-react';
+import { Button, ButtonVariant } from '@/atoms/Button/Button';
+import { Container } from '@/atoms/Container/Container';
 import { ProfilePageEmptyState } from '../ProfilePageEmptyState/ProfilePageEmptyState';
 
+interface UserNotFoundProps {
```
secondl1ght: Why not just do this in the component instead of passing props?

src/components/molecules/PostNotFound/PostNotFound.tsx
```diff
@@ -0,0 +1,62 @@
...
+  imageAlt,
+  backToFeedLabel,
+  viewProfileLabel,
+  exploreTagsLabel,
+  onBackToFeed,
+  onViewProfile,
+  onExploreTags,
+}: PostNotFoundProps) {
+  return (
+    <ProfilePageEmptyState
```
secondl1ght: Bit of a naming mismatch here now, could rename that comp to be more re-useable.

src/components/molecules/PostNotFound/PostNotFound.tsx
```diff
@@ -0,0 +1,62 @@
+'use client';
+
+import { ArrowLeft, FileQuestion, Tag, UserRound } from 'lucide-react';
+import { Button, ButtonVariant } from '@/atoms/Button/Button';
+import { Container } from '@/atoms/Container/Container';
+import { ProfilePageEmptyState } from '../ProfilePageEmptyState/ProfilePageEmptyState';
+
+interface PostNotFoundProps {
+  title: string;
```
secondl1ght: Why not just do this in the comp instead of all the props?

secondl1ght (commented):
Left a few minor code comments/questions. 

I noticed that a not found profile logs a warning, but a not found post throws an error. I think handling both as a warning makes sense, because these cases are expected, otherwise the new Sentry logs could be noisy with errors that we don't need to address.

Tested by running the code locally, profile and post routes appear to work as expected in the related issues, high-level design review looks to match. ☑️

secondl1ght (approved):

catch-21:
Raised a couple against this one: https://github.com/pubky/pubky-app/issues/1875 https://github.com/pubky/pubky-app/issues/1876

PR #1813, opened 2026-05-14
Update Header.tsx to fix alignment issues for smaller breakpoints

aldertnl:
This PR fixes alignment issues for the header for smaller breakpoints.

Fixed homepage header/content alignment by removing duplicate horizontal padding from the outer header wrapper. The nav keeps the shared container padding, so the Pubky logo now aligns with the landing heading across smaller and wide desktop breakpoints.

Incorrect (currently live on pubky.app):
image

image

Correct:

image

image


infin1t3 (approved):
Code looks ok. Approved given Aldert already confirmed it works visually.

PR #1815, opened 2026-05-14
fix(ui): skip clickable hashtags for invalid tag labels

### LINKED ISSUE #1764: [Bug]: `#long-tag-sequence` in post renders as clickable tag, even when too long to be tag
### Description

If the post contains hashtag followed by a word, that is rendered as clickable tag. When clicked, it searches by that tag.

However this also happens for very long "tags", like
`#test-a-very-long-possible-tag-but-who-knows-might-work`

### Expected Behaviour

One idea is to check by length: if the hashtag word is too long for a tag (max valid length for a tag label = 20 chars), then it should not be rendered as tag.

Another idea is, before rendering it as a tag, double-check in the post's data from Nexus, whether this sequence is returned as tag. Here's the Nexus data for the demo article: https://nexus.staging.pubky.app/v0/post/56o4ougdwf8wdqn5ztarg97nik77isg8kukfrfqyqjg8r9rwodbo/00351CEHH2TXG

### Steps to Reproduce

1. Go to demo post: https://staging.pubky.app/post/56o4ougdwf8wdqn5ztarg97nik77isg8kukfrfqyqjg8r9rwodbo/00351CEHH2TXG
2. Notice tag is clickable
3. Notice the search page doesn't find the post with this tag, because the tag itself is never created and associated to the post on Nexus, because it's too long

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
Gate remark hashtag links with isValidTagLabel so overlong tokens stay plain text. Extend createPatternPlugin with optional shouldConvert.

image

Resolves: #1764


catch-21:
Screenshot 2026-05-18 at 13 23 35


If not too challenging, can we allow inclusion of symbols that are allowed in tags such as `_` `@` `"` `?` `;` `*`

infin1t3:
> Screenshot 2026-05-18 at 13 23 35
> If not too challenging, can we allow inclusion of symbols that are allowed in tags such as `_` `@` `"` `?` `;` `*`

Should be good now.

src/components/molecules/PostText/PostText.test.tsx
```diff
@@ -606,6 +618,15 @@ describe('PostText', () => {
       expect(screen.getByText('bold').tagName).toBe('STRONG');
       expect(screen.getByTestId('post-hashtag')).toHaveTextContent('#hashtag');
     });
+
+    it('does not render overlong hashtag as PostHashtags link', () => {
```
Nirzhuk, line 622: Add test case for invalid inline cases.
Example of the text
Hello#Notatag

Nirzhuk (approved):

PR #1820, opened 2026-05-15
fix: delete null page in feeds route group

secondl1ght:
Fixes a Next bundle bug RE: Slack thread.


infin1t3 (approved):
https://github.com/user-attachments/assets/312677a4-eaf2-45c5-85a8-c49b13f389c2

Ephemerals work properly now.

PR #1828, opened 2026-05-18
fix(#1220): edit post can edit media

### LINKED ISSUE #1062: [Feature]: Add/Remove attachments whilst Editing post
### Problem Statement

I want to be able to drag-drop, copy paste, or use button to add new attachments to an existing post. Also be able to remove attachments that were previously attached to it.

### Proposed Solution

Add attachment support to Edit Post.

### Feature Area

Post Creation / Editing

### Feature Type

New Feature (doesn't exist yet)

### Acceptance Criteria

_No response_

### Design / Mockups

_No response_

### Alternatives Considered

_No response_

### Suggested Priority

None

### Additional Context

_No response_

### LINKED ISSUE #1220: [Bug]: Article editing does not allow banner image edit
### Description

When editing an article, i can edit everything except the header image

### Expected Behaviour

to be able to replace the header image by editing the article

### Steps to Reproduce

try to an edit an article

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

Nirzhuk:
fix #1220 

Initially the bug is only about Articles but we expanded to include normal posts too.

Demo:
https://github.com/user-attachments/assets/384a6e5f-1320-43f0-a388-55f087f0b05b

- Can replace photos from articles and posts
- You can add photos to old posts.

The first PR was closed because we need it from homeserver to be available to edit the images, now we're allowed to do it.


Nirzhuk:
@greptile-apps review it again  please

src/components/molecules/PostInputAttachments/PostInputAttachments.tsx
```diff
@@ -88,15 +121,69 @@ export const PostInputAttachments = forwardRef<HTMLInputElement, PostInputAttach
           </Card>
         ) : null}
 
-        {attachmentsWithPreviews.length ? (
+        {hasExisting || hasNew ? (
           <Container className="gap-4">
+            {/* Existing attachments (from server) */}
+            {existingAttachments?.map((a, i) => {
```
Orlandohub: Two ~90-line render blocks for existing vs new attachments are near-identical — same delete-button shape, same type-conditional positioning classes, same image/video/audio/pdf branches. Only data-cy prefix, onClick source, and disabled source differ.

I believe we can extract an `AttachmentPreviewItem` sub-component

src/hooks/usePostInput/usePostInput.ts
```diff
@@ -469,6 +626,11 @@ export function usePostInput({
     setTags,
     attachments,
     setAttachments,
+    existingAttachments,
+    setExistingAttachments,
+    isLoadingExistingAttachments,
+    editHadAttachments: (editAttachments?.length ?? 0) > 0,
```
Orlandohub: This does not seem to be used anywhere in the code base

src/components/organisms/PostInput/PostInput.tsx
```diff
@@ -121,9 +158,24 @@ export function PostInput({
   };
 
   const isEdit = variant === POST_INPUT_VARIANT.EDIT;
+  const isArticleEdit = isEdit && Boolean(editIsArticle);
 
   const { toast } = useToast();
 
+  const hasRequiredMediaForEdit = (
```
Orlandohub: `hasRequiredMediaForEdit`, `handleOnRemoveExisting`, `isRemoveExistingDisabled`, `handleOnRemoveAttachment`, `isRemoveAttachmentDisabled` all live in the component but operate purely on hook state (existingAttachments, attachments, editHadMediaAttachments, setExistingAttachments, setAttachments). 

I would move the five handlers into `usePostInput.ts` return them as a single `editAttachmentControls` object when `variant === EDIT`. `PostInput` then passes one prop instead of five, and the predicate lives next to the `editHadMediaAttachments` it depends on.

src/hooks/usePostArticle/usePostArticle.ts
```diff
@@ -72,26 +75,95 @@ export function usePostArticle({
...
+        }
+
+        if (attachmentMetadata.length > 0) {
+          Logger.warn('[usePostArticle] Cover image metadata is present but not an image', {
+            attachments,
+          });
+          return;
+        }
+
+        const fallbackCoverImage = resolveCoverImageFromUri(attachments[0]);
```
Orlandohub: AI review:

The cover-image URI-fallback flow is implemented twice — inside the try block (when attachmentMetadata.length === 0) and inside the catch block. The two blocks differ only in log messages.

Why it matters: Any change to fallback semantics (e.g., trying more than attachments[0], or toast copy) has to happen in two places. The function bloated from ~25 to ~95 lines.

Fix: Extract a local `tryUriFallbackOrToast(attachmentUri, reason, error?)` helper that returns boolean. Both try and catch branches collapse to ~3 lines.


I had a quick glance and it seems right, let me know if you agree

Orlandohub (changes requested):

src/hooks/usePostArticle/usePostArticle.ts
Nirzhuk: Yeah lets fix it.

Orlandohub (changes requested):
When changing article banner image, the old one jumps back creating an unexpected double banner image on edit mode:

https://github.com/user-attachments/assets/33c42bee-e9db-4eb0-a5c3-d9ab05670273

PR #1817, opened 2026-05-15
perf: create shared layout for feed routes

secondl1ght:
## 🎯 Summary

Hoist the `ContentLayout` shell (sidebars, mobile header, drawers, right rail) into a shared layout for the feed route cluster — `/home`, `/feed/[id]`, `/bookmarks`, `/search` — so it stays mounted across intra-cluster navigation. This eliminates per-route shell remounts greatly improving performance, removes layout shift and loading flash for shared components resulting in a smooth UX, and centralizes the parallel `@post` intercepted-modal slot in one place.

## 🗂️ Route Group: `app/(feeds)/`

Moved all four feed routes under a new Next.js route group `(feeds)` (no URL impact):

- `app/home/page.tsx` → `app/(feeds)/home/page.tsx`
- `app/feed/[id]/page.tsx` → `app/(feeds)/feed/[id]/page.tsx`
- `app/bookmarks/page.tsx` → `app/(feeds)/bookmarks/page.tsx`
- `app/search/page.tsx` → `app/(feeds)/search/page.tsx`

The previous per-route `layout.tsx` files (`home/layout.tsx`, `feed/[id]/layout.tsx`) and their `@post` slot scaffolding (`(..)post`, `(...)post`, `default.tsx`, `[...catchAll]/page.tsx`) have been consolidated into a single hoisted layout + slot at the `(feeds)` level using the `(.)post` interception marker.

### `app/(feeds)/layout.tsx`

A single persistent `<ContentLayout>` for the whole feed cluster:

- Resolves per-route shell props from the pathname via `tryResolveFeedsShellConfig`
- Hoists the `@post` parallel slot so the intercepted post modal renders as a sibling of `ContentLayout` (full-width, not nested in the main column)
- When `(.)post` is active, children are hidden (`class="hidden"`) but stay mounted — preserves scroll position on back-nav
- **Intercepted-modal pathname fallback**: while `(.)post` is active, `usePathname()` reports `/post/<user>/<post>` which doesn't resolve to a feeds config. The layout caches the last successful config via a `useState` "store-during-render" pattern and reuses it so the previous route's chrome stays mounted under the modal
- **Config-drift safety net**: throws if a non-feeds pathname mounts this layout without `(.)post` being active (e.g. adding `(feeds)/notifications/page.tsx` without registering a config)

### `app/(feeds)/_shell/configs.tsx`

Single source of truth for per-route shell props:

[...]

PR #1831, opened 2026-05-18
refactor: feeds layout cleanup

secondl1ght:
A few cleanup follow-up items from #1817, there should be no functional changes.

- silence console warning regarding unique key from Next.js internal code
- graceful fallback for routes within the layout group that have not had shell config yet


infin1t3 (approved):
Code checked and basic regression testing on ephemeral done - LGTM.

PR #1856, opened 2026-05-20
refactor: disable Sentry on testnet and E2E builds

Orlandohub:
## Summary
- Stop testnet and Cypress E2E runs from reporting errors to the production Sentry project, which inflated issue counts and made production triage unreliable
- Gate runtime Sentry initialization on `NEXT_PUBLIC_TESTNET` alongside existing test/dev guards
- Remove unnecessary Sentry DSN and sample-rate build args from the franky-testnet Docker workflow (source map upload args retained)

## Changes
- Add `NEXT_PUBLIC_TESTNET` check to `shouldEnableSentry()` in `src/libs/observability/sentry.ts`
- Drop `NEXT_PUBLIC_SENTRY_*` runtime build args from the testnet job in `.github/workflows/build-docker-image.yml`
- Add unit test coverage for the testnet guard in `sentry.test.ts`
- Document testnet disable behavior in `docs/sentry.md`

## Test plan
- [x] `npm run test -- src/libs/observability/sentry.test.ts` (17 tests pass)
- [ ] After merge: confirm franky-testnet deploy no longer sends HeadlessChrome events to Sentry
- [ ] After merge: confirm production/staging Sentry init unchanged when `NEXT_PUBLIC_TESTNET` is unset

## Notes
- No tracking issue linked — follow-up from production Sentry triage (CI/E2E events were ~99% of top issue volume).
- `SENTRY_AUTH_TOKEN`, `SENTRY_ORG`, and `SENTRY_PROJECT` remain in the testnet Docker build for source map upload only.


.github/workflows/build-docker-image.yml
```diff
@@ -173,11 +173,6 @@ jobs:
...
             SUPPORT_ACCOUNT_ID=123
             SUPPORT_FEEDBACK_INBOX_ID=26
-            NEXT_PUBLIC_SENTRY_DSN=${{ vars.SENTRY_DSN }}
-            NEXT_PUBLIC_SENTRY_ENVIRONMENT=testnet
-            NEXT_PUBLIC_SENTRY_TRACES_SAMPLE_RATE=0.1
-            NEXT_PUBLIC_SENTRY_REPLAYS_SESSION_SAMPLE_RATE=0.0
-            NEXT_PUBLIC_SENTRY_REPLAYS_ON_ERROR_SAMPLE_RATE=1.0
             SENTRY_AUTH_TOKEN=${{ secrets.SENTRY_AUTH_TOKEN }}
             SENTRY_ORG=${{ vars.SENTRY_ORG }}
             SENTRY_PROJECT=${{ vars.SENTRY_PROJECT }}
```
catch-21, line 178: remove these too?

catch-21:
Instead of using `NEXT_PUBLIC_TESTNET` you could do like we already have here to detect is using Cypress https://github.com/pubky/pubky-app/blob/9c718ebee3ceaa18de6a2b13e0776ad25abdba35/src/components/organisms/HumanSelection/HumanSelection.tsx#L17-L18

.github/workflows/build-docker-image.yml
Orlandohub: No, we keep this ones for the source map on build

Orlandohub:
@catch-21 on https://github.com/pubky/pubky-app/pull/1856#issuecomment-4497830939:

I think the Cypress window check works for client-only UI paths, but I’d keep NEXT_PUBLIC_TESTNET here because Sentry enablement is shared across browser/server/edge runtimes and testnet should be disabled independent of whether Cypress is attached. The E2E workflow builds the testnet image, so NEXT_PUBLIC_TESTNET=true covers those runs too. A browser-only Cypress guard could be additive if needed, but it should not replace the environment gate.

Orlandohub:
@greptile-apps review

catch-21 (approved):

PR #1860, opened 2026-05-20
chore: add .dockerignore so a local build doesn't interfere with image build

catch-21:
I had build errors in the docker stack due to dependency incompatibility within my checkout of pubky-app. This keeps them independent.


Orlandohub (approved):
LGTM

PR #1888, opened 2026-05-25
fix: paddings and size in MobileHeader and related components

### LINKED ISSUE #1757: [Bug]:  Regression Mobile Header - Incorrect margin bottom
### Description

There's a few pages with an extra pb-6 in Mobile Header.

we need to add a possibility to add variations for those p-6

Image

Pages which are affected:
- Logout
- Settings

### Expected Behaviour

Image

### Steps to Reproduce

Go to settings in mobile
Go to Logout in mobile

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

Nirzhuk:
- Refactor mobile header and tab bar to support compact header height.
- Update related tests and snapshots to reflect new header height adjustments.
- Ensure consistent styling across mobile menus and headers.

fix #1757

Demo:

https://github.com/user-attachments/assets/b5762a36-8910-448e-88e2-2ef4ab383e00


Nirzhuk:
@greptile-apps review it again

src/components/molecules/HotMobileMenu/HotMobileMenu.tsx
```diff
@@ -40,7 +40,8 @@ export function HotMobileMenu({ activeSection, onSectionChange }: HotMobileMenuP
       items={items}
       showLabels
       position="sticky"
-      className="-mx-6 -mt-6 mb-6"
+      headerTop="compact"
```
Orlandohub, line 41: HotMobileMenu now pins at `headerTop="compact"` (top-(--header-height-settings) = 72px), but on the Hot page the header is rendered by `ContentLayout` with no `containerClassName`, i.e. full chrome = p-6 (24+24) + min-h-12 (48) = 96px.

The issue lists Hot as a page that should get the compact (72px) header, but `ContentLayout`/`HotDiscoveryContentLayout` were not given any way to render a compact header, so Hot's header is still 96px.

The bar only lines up because the leftover 24px lands in the header's empty bottom padding and is masked by the -mx-6 -mt-6 override — a coincidental, brittle alignment, not the intended fix. Settings/Profile work correctly because their headers are compact

Fix: Thread `containerClassName` (or a compact flag) through `HotDiscoveryContentLayout` → `ContentLayout` → `MobileHeader` and pass `pb-0` for Hot, then drop the `-mt-6` compensation. Verify on a mobile viewport.

src/app/globals.css
```diff
@@ -126,14 +126,36 @@ body {
...
-   * Mobile header height
-   * - Used for sticky positioning of mobile menu below header
-   * - Matches MobileHeader component height
-   * - Applied: ProfilePageMobileMenu sticky top offset
+   * Mobile header height (full chrome)
+   * - Matches MobileHeader with default padding (p-6, side buttons)
+   * - Applied: MobileTabBar `headerTop="mobile"` (e.g. Hot feed via ContentLayout)
    * - Responsive: Mobile only (hidden on desktop via lg:hidden)
    */
   --header-height-mobile: 80px;
```
Orlandohub, line 137: 80px no longer matches reality. The rewritten comment claims it "Matches MobileHeader with default padding (p-6, side buttons)", but that header is 96px.

The "Applied: MobileTabBar headerTop="mobile"" note is also stale — all three consumers (Hot, Profile, Settings) now use headerTop="compact".

The value is also consumed by `ProfilePageLayout.tsx` (min-h-[calc(100dvh-var(--header-height-mobile))]) where that page's header is actually 72px, so the calc is off there too.

Fix: Set --header-height-mobile to the true full-chrome height (96px) and correct both comments, or remove the variable + the headerTop="mobile" branch if full-chrome tab-bar positioning is genuinely unused.

src/components/molecules/Logo/Logo.tsx
```diff
@@ -11,12 +11,37 @@ interface LogoProps {
   noLink?: boolean;
 }
 
+const DEFAULT_LOGO_WIDTH = 109;
+/** Rendered logo height in px (matches `h-[36px]` / SVG aspect ratio). */
+const DEFAULT_LOGO_HEIGHT = 36;
+
 const FORCE_HOME_SCROLL_TOP_KEY = 'pubky:force-home-scroll-top';
 
+const logoLinkClassName = (width: number, height: number, className?: string) =>
```
Orlandohub, line 23: Four module-level helpers (logoLinkClassName, logoImageClassName, logoSizeStyle, logoImageStyle) are each used once and each recompute the identical `width === DEFAULT_LOGO_WIDTH && height === DEFAULT_LOGO_HEIGHT` guard. 

four single-use helpers with the same boolean repeated four times add indirection without reuse for a small component. 

Compute const isDefaultSize = ... once and inline the cn()/style expressions (pass isDefaultSize to LogoImage).

Orlandohub (changes requested):

src/app/globals.css
Nirzhuk: That's some technical debt I'm okay with taking, but changing this value means refactoring a big chunk of the pull request, which honestly is pretty involved and adds at least an hour of extra testing. Not really worth the time right now.

Also, this wasn't part of someone request, it's something I noticed while testing on mobile.

The root cause is actually deeper though, the layout structures are kind of all over the place, designs don't really match up and end up needing exceptions for almost every component.

So if it's possible I will decide to not spend time on this nit

Orlandohub: fair enough, just make sure to leave a comment so we know this value is not correct for future iterations.

Orlandohub (approved):

PR #1906, opened 2026-05-26
fix(1499): show auth expired dialog as soon as relay polling fails

### LINKED ISSUE #1499: [Bug]: auth relay link renewal occurs too long after the final http relay request times out
### Description

When sitting on the sign in page (QR or 'Authorize with Pubky Ring' button), if the three http relay requests all fail (timeout) then we inform the user and they can recreate the link. This was implemented as a dialog in https://github.com/pubky/pubky-app/pull/1392. The problem is the dialog doesn't show for a little while after the third http relay request times out.

### Expected Behaviour

This might be because `AUTH_POLL_MAX_ATTEMPTS` is 5 minutes and the requests themselves are timeing out after 1 minute each, for my local instance and in staging at least. Although, it would be more robust to react to the third request failing rather than have a separate poll, so let's do that?

### Steps to Reproduce

1. Go to sign in page
2. wait 3 minutes for all 3 http relay requests to time out
3. observe no 'Authorization session expired' dialog shows
4. (meanwhile the user could use the qr/button to sign in with guaranteed failure
5. wait a few minutes more
6. observe dialog shows

### Screenshots

Shows all three http relay requests timed out but no dialog

Image

After waiting a few minutes more the dialog shows accompanied by console errors

Image

### Console Log

```shell

```

### Browser & Version

_No response_

Taewa:
## Summary
Fixes #1499. The "QR code expired" dialog on the sign-in / sign-up QR screen now appears   as soon as the relay polling actually fails, instead of after an extra silent wait of several minutes.

## Root cause
- The SDK's relay polling already retries internally and only throws once it's actually failed.
- On top of that, our polling loop had its own retry layer that kept polling after the SDK gave up.       
- Result: the dialog was effectively gated on a separate ~5-minute timer, not the real failure, so users
sat on a dead flow for minutes before seeing anything.

## Fix
- **Fail fast on relay error** — when the SDK reports a polling failure, surface it as a session-expired
error immediately instead of retrying.
- **Drop duplicate retry layer** — remove the status-code retry helper that's no longer reachable, since
the SDK is the single source of truth for retry.
- **Comment fix** — clarify in code that we're intentionally choosing the fast "expired + refresh" UX
over a silent wait on a possibly-dead flow.
## Testing
- Updated unit tests to verify that one relay throw is terminal (no extra polling attempts) and produces   the session-expired error.
- Manual smoke: opened the sign-in QR page, let relay requests time out — dialog now appears right after
the SDK gives up rather than minutes later.

## Notes
- No change to other session-expired paths (regular API calls / `requiresLogin` checks).
- Behavior change is scoped to the QR auth flow polling loop only.

## Reulst
Screenshot 2026-05-27 at 08 52 28

## Test
From ephemeral deployment:
https://github.com/user-attachments/assets/24c0fe1b-7596-4924-a712-6e37e0e619d9

From local test: Change the value of `NEXT_PUBLIC_DEFAULT_HTTP_RELAY`.

## Follow-up
- #1910 — revisit "QR code expired" wording vs. relay-failure cause.

### DIFF
--- a/src/core/services/homeserver/error.utils.ts
+++ b/src/core/services/homeserver/error.utils.ts
@@ -20,18 +20,6 @@ const PUBKY_ERROR_NAMES = {
   AUTHENTICATION_ERROR: 'AuthenticationError',
 } as const;
 
-const RETRYABLE_STATUS_CODES = [
-  HttpStatusCode.NOT_FOUND,
-  HttpStatusCode.REQUEST_TIMEOUT,
-  HttpStatusCode.TOO_MANY_REQUESTS,
-  HttpStatusCode.INTERNAL_SERVER_ERROR,
-  HttpStatusCode.BAD_GATEWAY,
-  HttpStatusCode.SERVICE_UNAVAILABLE,
-  HttpStatusCode.GATEWAY_TIMEOUT,
-] as const;
-
-type RetryableStatusCode = (typeof RETRYABLE_STATUS_CODES)[number];
-
 /**
  * Extracts status code from error objects (checks error directly first, then nested data)
  * @param error - The error to extract status code from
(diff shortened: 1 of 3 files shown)


Taewa:
@greptile-apps
  Updated related comments. Checked the SDK docs — tryPollOnce is one-shot, no retry inside.
  A transient hiccup could trigger the dialog, but there's no self-healing on our side either.
  When a poll fails, informing the user and letting them retry is more intuitive than waiting silently.

Taewa:
@greptile-apps
Review again plz

Taewa:
Closed due to the plan changed. Check this https://github.com/pubky/pubky-app/issues/1925

PR #1912, opened 2026-05-27
fix(ui): scope custom feed layout to feed routes

### LINKED ISSUE #1911: [Bug]: Layout doesn't apply when switching from custom feed to default feed
### Description

When switching from an custom feed to the default feed, the default feed's Layout is not applied.

### Expected Behaviour

Layout is correctly applied every time any type of feed to selected.

### Steps to Reproduce

1. Have a custom feed that changed Layout to Wide
2. Be on default feed that has Column Layout applied
3. Switch to custom feed
4. Observe Wide Layout is applied (good)
5. Switch back to default feed
6. Observe Column Layout is not applied to the page (bad)

### Screenshots

https://github.com/user-attachments/assets/ba3ca71a-970e-4119-88c7-c70517be5d54

### Console Log

```shell

```

### Browser & Version

_No response_

Orlandohub:
## Summary
- fix #1911
- Prevents wide/visual custom feed layout state from leaking into default feed routes after navigation.
- Restores the default feed shell immediately when returning to `/home`, `/bookmarks`, or other non-custom feed routes.

## Changes
- Scope `ContentLayout` custom feed layout overrides to `TIMELINE_FEED_VARIANT.CUSTOM`.
- Keep default feed routes driven by the persisted home layout even when `useCustomFeed` has a latched custom feed.
- Add regression coverage for home route layout behavior with a latched wide custom feed.

https://github.com/user-attachments/assets/756500a9-a2da-4ecf-b2d4-69de7be3b69b

## Notes
- `useCustomFeed` remains latched for intercepted post navigation; only the layout consumer is scoped by route variant.

### DIFF
--- a/src/components/organisms/ContentLayout/ContentLayout.test.tsx
+++ b/src/components/organisms/ContentLayout/ContentLayout.test.tsx
@@ -326,13 +326,15 @@ describe('ContentLayout', () => {
 describe('ContentLayout - Custom Feed Layout Override', () => {
   beforeEach(() => {
     mockUseCustomFeed.mockReturnValue(undefined);
+    mockHomeLayout = 'columns';
   });
 
-  it('hides sidebars when custom feed layout is wide', () => {
+  it('hides sidebars when custom feed route layout is wide', () => {
     mockUseCustomFeed.mockReturnValue({ layout: 'wide' });
 
     render(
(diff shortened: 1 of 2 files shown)


infin1t3 (approved):
LGTM

PR #1916, opened 2026-05-27
fix: only update `last_read` when user has unread notifications

ok300:
I noticed that each time I go to my profile, `last_read` is updated on the HS. Even though it is used to determine read / unread notifications, it is saved regardless of whether or not I have any unread notifications.

This means, each time the user goes to their profile and they have no new notifications, the `last_read` update causes unnecessary overhead:
- on the FE (to push the new `last_read` to the HS)
- on the HS (network traffic + new event in DB)
- on Nexus (as it will fetch the new `last_read` event as part of the user's event stream)
- on future user backups of their HS data (for each new `last_read` event)

When user has no unread notifications, saving the file serves no purpose.

This PR updates the `last_read` logic, so it is only saved when indeed the user has unread notifications.


catch-21:
Tested by receiving notification whilst:
- viewing list on profile page
- on feed
- signed out
All functions the same as before.
Notification E2E tests pass too.

Taewa (approved):
LGTM

The `last_read` only fires when `unread` is greater than 0.
Also I did smoke test and everything works fine.

PR #1918, opened 2026-05-27
fix: remove 'why invite codes' from FAQ, and remove Ring seed import

### LINKED ISSUE #1905: [UI]: Remove FAQ for Invite Code
We now have phone and payment options to sign up so we can remove "Why does Pubky require invite codes?" from the FAQ.

Image

catch-21:
Fixes #1905 

Removes the following from FAQ:
- "Why does Pubky require invite codes?"
- "(Seed phrase support will be added in future versions of Pubky Ring.)"
- "Pubky Ring does not yet support seed import."

Also fix Scan template snapshot test.


Taewa (commented):
I tested updated FAQ section and it's all well updated.

@catch-21 
Can you address what greptile's feedback? It seem like `FAQ_SECTIONS` is a dead code.

catch-21:
@Taewa Done, thanks.

Taewa:
@greptile-apps

Review again plz

Taewa (approved):
LGTM

PR #1926, opened 2026-05-28
fix(ui): preserve wide layout across route changes

### LINKED ISSUE #1913: [Bug]: Wide Layout gets reset to Column when navigating back from Settings or Hot page
### Description

When Wide Layout is applied to the default feed, if the Settings or Hot page is navigated to and then returned from the Layout is set to Column.

### Expected Behaviour

Layout setting for default feed should persist regardless of where in the app you navigate.

### Steps to Reproduce

1. On default feed, change Layout to Wide
2. Navigate to Settings or Hot page
3. Navigate back to feed (home button)
4. Observe Column Layout has been applied

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

Orlandohub:
## Summary
- fix #1913
- Preserve the user's saved wide layout preference when navigating through pages that do not support the wide shell.
- Keep Settings, Hot, Who To Follow, and discovery pages rendering their column-style chrome without mutating the home feed store.

## Changes
- Removed the `useLayoutReset` hook and route-level calls that wrote `LAYOUT.COLUMNS` into persisted home state.
- Added render-time wide layout fallback for unsupported timeline variants while preserving `requestedLayout`.
- Added `disableWideShellLayout` to `ContentLayout` for page-local column chrome.
- Updated component tests and snapshots for the new non-mutating layout behavior.

https://github.com/user-attachments/assets/6f3bdbd0-50f4-4346-a53c-06f7dc3105e7


Orlandohub:
@greptile-apps review

infin1t3 (approved):
https://github.com/user-attachments/assets/958b3ee6-2857-4f9b-9319-6f1c92035270

Approved - LGTM.

Issue #1927, opened 2026-05-28
Remove 5-minute auth poll timeout logic (follow-up to #1925)


Taewa:
### Description

Follow-up to #1925.

The auth QR polling loop still enforces a hard 5-minute cap via `AUTH_POLL_MAX_ATTEMPTS` (3000 × 100ms). Once exceeded, it throws a `REQUEST_TIMEOUT` error and tears down the flow.

Per #1925, the pubky client now reestablishes the http-relay connection indefinitely, so **we no longer expect a timeout and should not handle one**. This timeout cap is the remaining piece of timeout-handling that should be removed.

### ⚠️ Precondition (must do first)

Confirm with the **pubky-core team** that the 5-minute rule is genuinely no longer needed — i.e. that the pubky client reestablishes the http-relay connection indefinitely and there is no underlying timeout we'd be re-exposing by removing the cap. Do not remove the cap until this is confirmed.

### Location

- `src/core/services/homeserver/homeserver.utils.ts`
  - `AUTH_POLL_MAX_ATTEMPTS` constant (line ~32) — "Maximum auth poll attempts (3000 × 100ms = 5 minutes max wait)"
  - `awaitApproval` loop (line ~182-188) — throws `Err.timeout(TimeoutErrorCode.REQUEST_TIMEOUT, ...)` after `maxPollAttempts`

### Expected Behaviour

- No artificial attempt cap / timeout on the auth poll loop.
- The loop only terminates on: success (session approved), user cancellation, or a genuine relay/poll error (handled per #1925 with blurred QR + retry).
- Remove now-dead timeout code (`REQUEST_TIMEOUT` throw, `maxPollAttempts` cap, related test cases) if nothing else depends on them.

### Naming cleanup (moved here from #1925)

Once the timeout cap is gone, the remaining trigger is purely a genuine relay/poll error — at which point the "expired" framing is misleading and should be renamed. This was intentionally **left out of #1925** to keep that PR scoped to the dialog→blurred-QR UX, and because the rename only makes sense after the timeout case is removed.

Rename (current name → proposed):
- `isExpired` → `isAuthUnavailable` — `src/hooks/useAuthUrl/useAuthUrl.tsx`, `useAuthUrl.types.ts` (`UseAuthUrlReturn`), `useMobileAuth.tsx` (`UseMobileAuthReturn`), and consumers `SignIn.tsx` / `Scan.tsx`
[...]

Issue #325, opened 2025-10-22
[Bug]: Auth request rapid fire when http relay is unavailable


catch-21:
### Describe the bug

The sign in page fires requests to http relay in rapid succession when it fails to connect. Recall using an AbortController to achieve this in pubky-app, see https://github.com/pubky/pubky-app/blob/dev/src/app/sign-in/Card/_SignIn.tsx

### Reproduce

1. Configure pubky-app to use non-existent http replay for auth
2. Go to sign in page
3. Observe rapid fire

### Screenshots / Recording

https://github.com/user-attachments/assets/5c12ad49-0ac4-4f5c-932d-d9a18c259997

### Pubky App Version

6e02be58bb8702b55574fae1c118996817dc6da6

### Console log output

```shell

```

catch-21:
In 3ae860e the rapid fire still occurs but only three times. This is due to work done in https://github.com/pubky/pubky-app/pull/1392 that causes UI prompt after third request timeout. It probably still is a good idea to slow down failed requests in case of some temporary issue with the relay. Although I'd be happy if [pubky-app/issues/1499](https://github.com/pubky/pubky-app/issues/1499) were addressed and then the user can reliably retry on final request failure/timeout.

catch-21:
Closing in favour of #1925 because the rapid-fire is now prevented by the pubky client after third failed attempt.

PR #1930, opened 2026-05-29
Fix/1925 auth blurred qr reload

### LINKED ISSUE #1925: [Bug]: auth http-relay timeout unnecessarily initiated after 5 minutes and shows blocking dialog
### Description

We are closing both https://github.com/pubky/pubky-app/issues/325 and https://github.com/pubky/pubky-app/issues/1499 in favour of this issue because the pubky client and http-relay behaviour has since changed.

https://github.com/pubky/pubky-app/pull/1392 implemented a dialog that shows once the auth QR link is shown for at least 5 minutes. This was an attempt to solve the http-relay connection timeout, although this no longer occurs since the latest pubky client update that appears to reestablish the http-relay connection indefinitely.

The problem is that now the dialog no longer adds value. It unnecessarily closes the http-relay connection and dialog blocks the UI causing the user to dismiss it.

### Expected Behaviour

The pubky client handles the auth connection entirely. We don't expect any timeout, so don't handle for one.
We only need to handle an error scenario, such as when the http-relay is not contactable (e.g. local network issue).
We don't want a dialog that blocks the rest of the UI. The user must be able to sign in by other means (phrase/file) when http-relay is not an option. Ideally, in an error scenario the QR is blurred and the user can click a button to restart a connection attempt.

### Steps to Reproduce

1. Block requests to http relay (dev tools), or change env var url not to a working http-relay
2. Go to sign in
3. Observe QR code on page shows for up to over 5 minutes even though the connection with the http-relay is broken

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

Taewa:
## Summary
On the sign-in and sign-up QR screens, a blocking "session expired" dialog interrupted users when the relay connection failed. This replaces it with a non-blocking blurred QR plus a "Click to reload" control, so users can retry or sign in another way (phrase / file). Closes #1925.

## Root cause
- The QR auth flow showed a full-screen modal on failure, blocking the rest of the page — including the phrase/file sign-in options.
- When the pubky SDK reported a relay poll failure, our loop retried instead of surfacing it — so the QR stayed on a broken flow for up to 5 minutes.

## Fix
- **Non-blocking error state** — on relay failure the QR blurs and a "Click to reload" control appears in place of the dialog; the rest of the screen stays usable.
- **Fail fast on relay error** — the failure is surfaced as soon as polling fails, instead of silently retrying.
- **Mobile** — the deeplink "Authorize" button turns into an enabled "Click to reload" retry instead of a dead/disabled button.

## Testing
- Updated unit tests for the auth hook and both QR screens: reload control appears when the flow fails, and clicking it restarts the connection.

### Manual Testing
1. **Relay failure** — point `NEXT_PUBLIC_DEFAULT_HTTP_RELAY` at a dead URL. On desktop sign-in the QR blurs with a "Click to reload" banner; clicking it restarts the flow.
2. **5-minute cap** — leave the QR untouched for over 5 minutes; it should reach the expired state (blurred QR + reload) as the backstop.
3. **Mobile** — there is no QR, so the blurred QR + banner never appears; instead the "Authorize" button becomes an enabled "Click to reload" retry.
4. **Non-blocking** — while expired, the rest of the page stays usable (e.g. phrase / file sign-in remains clickable).

## Notes
- The 5-minute poll cap is intentionally **kept** here; its removal and the related "expired" → "unavailable" naming cleanup are tracked in #1927 (pending pubky-core confirmation).
- Replaces the old dialog approach from #325 / #1499; no change to other session-expired paths (regular API calls).

### Desktop sign-in
Screenshot 2026-05-29 at 11 24 51

### Desktop onboard
Screenshot 2026-05-29 at 11 27 13

### Mobile sign-in
Screenshot 2026-05-29 at 11 27 43

### Mobile onboard
Screenshot 2026-05-29 at 11 47 46

https://github.com/user-attachments/assets/ae6f63a3-3331-4910-b859-81c6b98921ee


src/core/services/homeserver/homeserver.utils.ts
Orlandohub, line 182: does this `maxPollAttempts` still make sense?

I see we've on line 32:
```
/** Maximum auth poll attempts (3000 × 100ms = 5 minutes max wait) */
const AUTH_POLL_MAX_ATTEMPTS = 3_000;
```

But the idea is go away from this approach, correct?

src/core/services/homeserver/homeserver.utils.ts
```diff
@@ -192,11 +192,19 @@ export const createCancelableAuthApproval = (
...
-        if (isRetryableRelayPollError(error)) {
-          await sleep(pollIntervalMs);
-          continue;
-        }
-        throw error;
+        // From the caller's view, tryPollOnce is one-shot: one call, one outcome
+        // (pubky SDK 0.8 — it doesn't loop or retry on our behalf). If it throws,
+        // we treat the flow as dead and fail fast — showing "session expired" now
+        // is better UX than letting the user wait minutes on a flow that may already be dead.
+        throw Err.auth(AuthErrorCode.SESSION_EXPIRED, 'Auth flow polling failed', {
```
Orlandohub, line 199: nit:
The trigger per the issue is "relay not contactable / local network issue," not a session expiry. The isExpired / SESSION_EXPIRED naming is now a slight misnomer for "connection error, click to reconnect." Cosmetic — the UX is correct regardless.

src/components/organisms/SignIn/SignIn.tsx
```diff
@@ -165,21 +164,31 @@ export const SignInContent = () => {
...
                 </Container>
               ) : isExpired ? (
-                <Container className="items-center gap-2">
-                  <QrCode className="size-8 text-muted-foreground" />
-                  <Typography as="small" size="sm" className="text-muted-foreground">
-                    {t('expired')}
-                  </Typography>
-                </Container>
+                <>
+                  <Image
```
Orlandohub: identical `<Image>` + clip-path `<span>` overlay in both organisms, `Scan` and `SignIn`  is still worth extracting into a shared molecule.

Orlandohub (changes requested):

src/core/services/homeserver/homeserver.utils.ts
Taewa: It doesn't make sense and there is plan to remove 5 min related logics (https://github.com/pubky/pubky-app/issues/1927) once @tipogi comes back and confirms it.

src/core/services/homeserver/homeserver.utils.ts
Taewa: I agree with you and my original implementation was updating `isExpired` to `isAuthUnavailable ` but I reverted it since it could be too early to confirm it. 
The follow-up ticket (https://github.com/pubky/pubky-app/issues/1927) already has plan to update it.

src/components/organisms/SignIn/SignIn.tsx
Taewa: Refactored.

Orlandohub (approved):

catch-21:
Verified on a9f88a11b64bc98840879ab52a3a246d817a2b47

Connection with http relay now reestablishes indefinitely, no more cancel after 3rd attempt.
If there is a connection error then QR is blurred with interactive 'Click to reload'. The button changes to 'Click to reload' on mobile.
No more blocking dialog 🙌 

Screenshot 2026-06-05 at 09 50 21

Screenshot 2026-06-05 at 09 52 34

PR #1931, opened 2026-05-29
feat: validate invite code before applying when manually inputting or using `/invite/<code>`

### LINKED ISSUE #1738: [Bug]: Invite code not validated on input
### Description

During sign up, when an invite code is entered or the `invite/<code>` url is used and the code is valid format but not registered with the homeserver then no error is displayed and the user can proceed. They will encounter an error after key generation when they try to use the invite code. The error isn't happening soon enough in the flow.

### Expected Behaviour

The homeserver now has an endpoint to validate an invite code without consuming it. We should adapt the onboarding flow to perform validation so that it no longer allows user to proceed to maybe fail at key generation in pubky-app or pubky-ring. Apply to both:
- on the sign up page where you enter an invite code
- on `/invite/<code>` endpoint

If an invite code with valid format but that is not recognised by homeserver is used then a toast error occurs at that point and prevents the user from proceeding.

### Steps to Reproduce

1. Follow onboarding with invite code
2. Input invalid code of valid format e.g. AAAA-AAAA-AAAA
3. Observe no error, you can proceed to key generation

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

catch-21:
Fixes https://github.com/pubky/pubky-app/issues/1738

1. validate invite code on `/invite/abc...`

https://github.com/user-attachments/assets/38725ad8-52ed-4201-a6b4-24622d75078b

2. validate invite code on manual input

https://github.com/user-attachments/assets/cacbbc9e-b6cd-49ca-b46d-4a99f1cab725

Need to refactor to use sdk once this is addressed: https://github.com/pubky/pubky-core/issues/410. Raised https://github.com/pubky/pubky-app/issues/1933 to track that work.


src/core/services/homeserver/homeserver.ts
```diff
@@ -143,6 +143,28 @@ export class HomeserverService {
...
+    try {
+      const url = `${HOMESERVER_URL}/signup_tokens/${encodeURIComponent(signupToken)}`;
+      const response = await this.getPubkySdk().client.fetch(url, { method: HttpMethod.GET });
+      Logger.debug('Signup token verification response', { status: response.status });
+      return response.ok;
+    } catch (error) {
+      Logger.warn('Signup token verification failed', { error });
+      return false;
+    }
+  }
```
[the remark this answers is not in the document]
catch-21: Addressed in f505d8bd239cd60ce997e653191981fef741e313


Screenshot 2026-05-29 at 17 04 52

tipogi (commented):
Approved the `core` related flow but in the UI side, I would not do the verification when the user hits the _continue_ button. I would do automatically when we fill the 16 chars the verification. When that verification is ok, we activate the _continue_ button

catch-21:
@tipogi I've changed UI to apply check on invite code entry rather than on Continue button press.  Also, improved logic to handle when invite code is not found by homeserver and when it is 'used'.
I'll push better E2E tests shortly.

catch-21:
This video shows that now the invite code is checked on entry (not Continue button) and a different toast message is shown when code is unrecognised by homeserver or already used.

https://github.com/user-attachments/assets/1861005e-dca8-4a36-bc8d-ba7c983b4579

src/components/templates/Onboarding/Human/Human.tsx
```diff
@@ -37,6 +42,44 @@ export function Human() {
...
+  // entered. On failure we surface a toast and keep the user on this step so they can try again.
+  // A homeserver that reports the code as invalid shows a warning toast; a homeserver we couldn't
+  // reach shows an error toast so the two are not conflated.
+  async function onInviteCodeVerify(inviteCode: string): Promise<boolean> {
+    try {
+      const status = await AuthController.verifySignupToken(inviteCode);
+
+      if (status === 'valid') {
+        toast({
+          title: t('inviteCodeApplied'),
```
tipogi, line 57: what about the message something like, "Invite code verified"

src/components/organisms/HumanInviteCode/HumanInviteCode.types.ts
```diff
@@ -1,5 +1,7 @@
 export interface HumanInviteCodeProps {
   onBack: () => void;
-  /** Called when user submits invite code. May be async; throws on validation failure. */
-  onSuccess: (inviteCode: string) => void | Promise<void>;
+  /** Verifies the invite code with the homeserver when the full code is entered. Returns true when valid. */
+  onVerify: (inviteCode: string) => Promise<boolean>;
```
tipogi: **UI: network errors should not use the same destructive styling as invalid/used codes**

The service-layer change in `homeserver.ts` (throwing on unreachable/unexpected HTTP responses) correctly separates **toasts** in `Human.tsx` — invalid/used get warning toasts, connectivity failures get `showErrorToast`.
That doesn’t fully address input styling in `HumanInviteCode`.
`onVerify` currently returns `Promise<boolean>`. In `HumanInviteCode.tsx`, any `false` result sets `isVerificationFailed = true`, which applies `border-destructive`:
```tsx
const isValid = await onVerifyRef.current(trimmedInviteCode);
// ...
} else {
  setIsVerificationFailed(true);
}
```
So when `Human.tsx` catches a network error, shows the error toast, and returns `false`, the input still gets the red “invalid code” border — even though the code may be fine and the failure was transient.

_Suggested fix_: return a richer result from onVerify (e.g. `'valid' | 'used' | 'invalid' | 'error'`) and only apply destructive styling for 'invalid' / 'used'. Keep 'error' on neutral border-input styling; the error toast is enough for connectivity issues.

```tsx
// HumanInviteCode.types.ts
/** Outcome of homeserver invite-code verification for manual entry UI state. */
export type THumanInviteCodeVerifyResult = 'valid' | 'used' | 'invalid' | 'error';

export interface HumanInviteCodeProps {
  onBack: () => void;
  /** Verifies the invite code with the homeserver when the full code is entered. */
  onVerify: (inviteCode: string) => Promise<THumanInviteCodeVerifyResult>;
  /** Called when the user clicks Continue after successful verification. */
  onSuccess: (inviteCode: string) => void;
}

```

tipogi (dismissed):
looks ready to merge for me, apart of the NITs

src/components/templates/Onboarding/Human/Human.tsx
catch-21: My thoughts are that "applied' goes a step further than "verified". The user doesn't need to know about verification happening, what's important is that it is applied.  If we say "verified" the user may assume they still need to apply it somehow.

src/components/organisms/HumanInviteCode/HumanInviteCode.types.ts
catch-21: Addressed in a7496d91be5a9b0eeac61749a19f3c19d1417047. The border is now only red when invite code is unrecognised (invalid) or already used. It is grey initially and on network error. It is green only when valid to proceed.

tipogi (approved):
LGTM, merge!

PR #1938, opened 2026-06-01
fix(database): recover from transient IndexedDB init failures on iOS

### LINKED ISSUE #1852: [Bug]: IndexedDB initialization failure blocks local-first app on Mobile Safari
### Description

Production Sentry reports **IndexedDB initialization failures** that prevent the local-first Dexie database from starting. Franky depends on IndexedDB for all cached data, optimistic writes, and offline behavior — if init fails, the app cannot function as designed.

**Primary Sentry issue:** [PUBKY-APP-2Q](https://synonym-52.sentry.io/issues/PUBKY-APP-2Q)  
**Events:** 2 (production, last 30 days) | **Status:** unresolved (new) | **Release:** 1.3.0  
**Route:** `/home` (`https://pubky.app/home`)  
**First seen:** 2026-05-17 | **Last seen:** 2026-05-17

**Error surfaced to Sentry:**

```
AppError: Failed to initialize database, indexedDB
  error.category: database
  error.code: INIT_FAILED
  error.service: local
  error.operation: initialize
```

**Underlying browser error (cause):**

```
UnknownError: Connection to Indexed Database server lost. Refresh the page to try again
```

**Stack trace (first-party):**

```
src/core/database/franky/franky.ts:269 (AppDatabase.initialize)
  throw Err.database(DatabaseErrorCode.INIT_FAILED, 'Failed to initialize database, indexedDB', …)
```

**Initialization flow:** `DatabaseProvider` (`src/providers/DatabaseProvider/DatabaseProvider.tsx`) calls `db.initialize()` on mount, which:
1. Checks if IndexedDB is available
2. Opens or recreates the Dexie database (`AppDatabase` in `franky.ts`) based on schema version
3. On failure, throws `INIT_FAILED` and sets error state on `DatabaseContext`

**Current behavior on failure:**
- `DatabaseProvider` sets `isReady: false` and exposes `error` + `retry()` via context
- The provider **still renders child components** when an error is present (spinner is only shown when `!isReady && !error`)
- No dedicated user-facing error UI appears to consume `DatabaseContext.error` — users may see a broken/partial app rather than a clear recovery prompt

**Affected environment (Sentry tags):**
- **Browser:** Mobile Safari
- **OS:** iOS 26.4.2
- **Device:** iPhone
- **Locale:** pt-BR (Curitiba, Brazil)
- **Environment:** production

**Session replay:** https://synonym-52.sentry.io/explore/replays/520f3435c72648d2a323d78adfc669d2/

**Sentry discover query (30d, production):**  
https://synonym-52.sentry.io/explore/discover/homepage/?dataset=errors&queryDataset=error-events&query=environment%3Aproduction+indexedDB+%21browser.name%3AHeadlessChrome&project=4510935265640528&field=issue&field=title&field=count%28%29&sort=-count%28%29&statsPeriod=30d&mode=aggregate&yAxis=count%28%29

**Potentially related issue (same user profile, 2 days later):**

| Issue | Error | Events | Notes |
|---|---|---:|---|
| [PUBKY-APP-34](https://synonym-52.sentry.io/issues/PUBKY-APP-34) | `Failed to bulk save records in user_details` | 1 | Same browser/OS/geo/locale; cause: `AbortError: Transaction aborted` on `/home` |

This may indicate broader IndexedDB instability on iOS Safari for the same user/session class, not only init-time failures.

**Why this deserves attention despite low event count:**
- **Severity per event is critical** — local-first architecture cannot operate without IndexedDB
- Error message explicitly tells the user to refresh, indicating a browser-level IDB process crash
- No retry/recovery UX is wired up despite `DatabaseContext.retry` existing
- iOS Safari is a primary mobile platform for the app

**Suggested fix directions:**
1. Add user-facing error UI when `DatabaseContext.error` is set (block render + “Refresh to retry” action wired to `retry()`).
2. Implement automatic retry with backoff when the cause is transient (`UnknownError: Connection to Indexed Database server lost`).
3. Handle iOS Safari IDB lifecycle edge cases (tab backgrounding, memory pressure, multi-tab `deleteDatabase` blocking — `recreateDatabase` already logs blocked deletion).
4. Investigate whether init should attempt `recreateDatabase` when connection is lost vs. failing immediately.
5. Monitor [PUBKY-APP-34](https://synonym-52.sentry.io/issues/PUBKY-APP-34) for post-init write/transaction failures on the same platform.

### Expected Behaviour

1. IndexedDB should initialize reliably on supported mobile browsers (including Mobile Safari on iOS).
2. If initialization fails due to a transient browser error, the app should retry automatically and/or show a clear recovery UI.
3. The app should not render main content in a broken state when the database is unavailable.
4. Users should be guided to refresh or retry without losing context where possible.

### Steps to Reproduce

Exact reproduction is intermittent (browser-level IndexedDB server crash), but Sentry captured this on production:

1. Sign in and navigate to `https://pubky.app/home` on **Mobile Safari** (iPhone, iOS 26.4.2 reported).
2. App mounts `DatabaseProvider`, which calls `AppDatabase.initialize()`.
3. IndexedDB connection fails with: `Connection to Indexed Database server lost. Refresh the page to try again`.
4. App throws `AppError: Failed to initialize database, indexedDB` (`INIT_FAILED`).

**Known iOS Safari conditions that can trigger similar failures (for manual testing):**
- Low memory / background tab eviction
- Rapid tab switching during DB open or schema migration
- Multiple tabs open against the same origin during `recreateDatabase()` (deletion blocked)
- Private browsing mode or strict storage settings (though this event was standard Mobile Safari)

### Screenshots

_Session replay available in Sentry:_  
https://synonym-52.sentry.io/explore/replays/520f3435c72648d2a323d78adfc669d2/

_Related write failure replay (PUBKY-APP-34):_  
https://synonym-52.sentry.io/explore/replays/e96f37ce22f24ebfb7e7103dbb2d4d3b/

### Console Log

```shell
# Surfaced AppError
AppError: Failed to initialize database, indexedDB
  at AppDatabase.initialize (src/core/database/franky/franky.ts:269)
  error.code: INIT_FAILED
  error.service: local
  error.operation: initialize

# Underlying browser cause
UnknownError: Connection to Indexed Database server lost. Refresh the page to try again

# Related write failure (PUBKY-APP-34, same platform)
AppError: Failed to bulk save records in user_details
  at RecordModelBase.bulkSave (src/core/models/shared/base/record/baseRecord.ts:19)
  cause: AbortError: Transaction aborted
```

### Browser & Version

Confirmed in production Sentry (release 1.3.0):
- Mobile Safari / iOS 26.4.2 (iPhone)
- Locale: pt-BR
- Geo: Curitiba, Brazil

Orlandohub:
## Summary
- fix #1852
- iOS Safari and in-app browsers (e.g. the Twitter webview) intermittently drop the IndexedDB connection during init, surfacing in Sentry as `INIT_FAILED` (PUBKY-APP-2Q). We now retry transient failures instead of hard-failing.
- The provider previously rendered the broken app on failure and never consumed the existing `retry()`. It now blocks the app and shows a recovery screen wired to retry.

## Changes
- Add `isTransientIndexedDbError()` (walks the `cause` chain for known WebKit/iOS IDB abort signatures) and a bounded retry-with-backoff loop in `AppDatabase.initialize()`; deterministic errors (schema, etc.) still fail fast.
- `DatabaseProvider` no longer renders children on error — it gates on state (error → recovery screen, not ready → spinner, ready → app).
- New `DatabaseErrorScreen` molecule with "Try again" (wired to `retry()`) and "Reload page".
- Add `errors.database.*` i18n keys across all 9 locales.
- Add/refresh unit tests for the retry loop, the detector, the recovery screen, and provider behavior.

Screenshot 2026-06-01 at 16 05 08

## Test plan
- [ ] Manual: throttle/interrupt IndexedDB on iOS Safari (or simulate a transient `UnknownError`) and confirm the app retries, then shows the recovery screen with working Try again / Reload

## Notes
- Root cause is environmental (WebKit/iOS IndexedDB instability), so this is resilience + recovery UX rather than a deterministic logic fix. The issue text described one cherry-picked event; the fix targets the whole class of transient init failures across any route/webview.
- `isTransientIndexedDbError` is exported and reusable — a sensible follow-up is to harden the post-init write path (related `PUBKY-APP-34` bulk-save aborts) with the same helper.


Orlandohub:
@greptile-apps review

src/core/database/franky/franky.ts
```diff
@@ -215,65 +258,85 @@ export class AppDatabase extends Dexie {
...
-        return { wasDbReset };
-      }
+        // iOS Safari / in-app browsers can drop the IndexedDB connection mid-open.
+        // Reset Dexie's internal state and retry before surfacing a hard failure.
+        if (!isLastAttempt && isTransientIndexedDbError(error)) {
+          Logger.warn('Transient IndexedDB error during initialization. Retrying...', {
+            attempt,
+            maxAttempts: DB_INIT_MAX_ATTEMPTS,
+          });
+          this.close();
```
Taewa, line 290: I found an issue during the manual test.
Basically 2nd, 3rd try might cause unexpected infinite call.
Here is an AI generated lengthy explanation but informative.

https://github.com/user-attachments/assets/e0b24e8a-d27c-4d92-8b08-6ecfcff21515

## TL;DR

Existing-user retry path leaves Dexie closed → all reads/writes throw `DatabaseClosedError`.
`runInitialize()` version-match branch (`franky.ts:354`) returns without `this.open()`,
but the retry loop's `this.close()` disables Dexie's auto-open.

---

## Intended vs actual behavior

**Intended** (per PR design):

- attempt 1: throws transient
- attempt 2: succeeds → `initialize()` returns success
- App loads normally

**Actual**:

- attempt 1: throws → catch calls `this.close()` → Dexie enters explicitly-closed state
- attempt 2: `Dexie.exists()` returns `true`, version matches → **returns without `this.open()`**
- `initialize()` resolves successfully (`wasDbReset: false`)
- `DatabaseProvider` calls `setIsReady(true)` → app starts rendering
- Dexie instance still in closed state
- Every component mount triggers a table call → **`DatabaseClosedError` floods the console**

The discrepancy: `initialize()` reports success, but the underlying Dexie instance was never re-opened.

## Reproduction (local dev)

Add a temporary fault injection inside `runInitialize()` (`franky.ts:295`, right
after `await Dexie.exists(this.name)`):

```ts
private async runInitialize(): Promise<{ wasDbReset: boolean }> {
const dbExists = await Dexie.exists(this.name);

// DEV ONLY — remove before commit.
// Toggle from DevTools Console BEFORE reload (sessionStorage survives reload,
// window globals do not):
//   sessionStorage.setItem('__idbFaultMode', 'once');    location.reload();  // 1st attempt throws,
retry succeeds
//   sessionStorage.setItem('__idbFaultMode', 'persist'); location.reload();  // every attempt throws
→ recovery screen
//   sessionStorage.removeItem('__idbFaultMode');         location.reload();  // disable
if (process.env.NODE_ENV !== 'production' && typeof window !== 'undefined') {
  const mode = window.sessionStorage.getItem('__idbFaultMode');
  if (mode === 'persist' || mode === 'once') {
    if (mode === 'once') window.sessionStorage.removeItem('__idbFaultMode');
    const e = new Error('Connection to Indexed Database server lost');
    e.name = 'UnknownError';
    throw e;
  }
}

if (!dbExists) { /* ... */ }
// ... rest unchanged
}
```

Steps:

1. Sign in once with the fault injection disabled so a local Dexie DB is created
  (this simulates an existing user)
2. In DevTools Console: `sessionStorage.setItem('__idbFaultMode', 'once'); location.reload();`
3. Expected (per PR intent): one retry, then app loads normally
4. Actual: console shows `[WARN] Transient IndexedDB error... attempt: 1`, retry
  reports success, app proceeds — but every subsequent Dexie read/write throws
  `DexieError: Database has been closed`. App eventually surfaces an
  `AppError: Failed to find stream in user_streams` and falls into the error
  boundary.

Note: `'persist'` mode hits `DatabaseErrorScreen` correctly (3 attempts all
throw before any open). The bug is specific to the **partial recovery** path.

## Root cause

`initialize()` retry loop calls `this.close()` on transient errors
(`franky.ts:268`), putting Dexie into an explicitly-closed state. After explicit
`close()`, Dexie does NOT auto-open on table calls — `open()` must be called
explicitly.

`runInitialize()` has three exit paths:

- `!dbExists` → `await this.open()` ✅
- version mismatch → `await this.recreateDatabase(...)` (opens internally) ✅
- **version match → returns without `open()`** ❌

The version-match path (`franky.ts:354`):
The version-match path (`franky.ts:354`):

```ts
Logger.debug('Database version is current');
return { wasDbReset: false };   // db left in closed state
```

In normal flow (no prior `close()` call) Dexie auto-opens on the first table
operation, so this is invisible. After the retry loop's explicit `close()`,
the explicitly-closed flag sticks → `DatabaseClosedError` on every read/write.

## Production scenarios that hit this

- Existing user (local Dexie DB on disk from prior session)
- iOS WebKit drops the connection at any IDB-touching call before
`runInitialize()` reaches its open: `Dexie.exists()`, `getExistingDbVersion()`,
or `this.open()` itself
- Retry succeeds (WebKit recovers), `Dexie.exists()` now returns true,
`getExistingDbVersion()` returns a matching version → version-match return
path → Dexie instance never opened → every query fails

## Why existing tests don't catch this

`franky.test.ts:307`:

```ts
existsSpy.mockRejectedValueOnce(transientError(...));
existsSpy.mockResolvedValue(false);  // ← always false on retry → !dbExists path
```

Tests only cover the new-DB recovery path (`!dbExists` → `this.open()`).
No coverage for existing-DB + version-match + transient.

## Impact

`DatabaseClosedError` is not in `isTransientIndexedDbError`'s signature list,
so the app can't self-recover from this state. For affected users it surfaces
as broken table reads (worse UX than the original `INIT_FAILED` — at least
that one was caught by the recovery screen). The PR's recovery UX is bypassed
for this class of failure.

Taewa (changes requested):

src/components/molecules/DatabaseErrorScreen/DatabaseErrorScreen.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    render(<DatabaseErrorScreen onRetry={onRetry} />);
+
+    await act(async () => {
+      fireEvent.click(screen.getByRole('button', { name: 'errors.database.tryAgain' }));
+    });
+
+    expect(onRetry).toHaveBeenCalledTimes(1);
+  });
+
+  it('shows the retrying label while the retry is in flight', async () => {
```
Taewa, line 35: Nit,
This is unreachable test because when the retry button is hit, the full screen spinner UI pops up therefore the button is invisible.

Taewa (approved):
LGTM

There is a nit comment which doesn't impact on the production code. Maybe we can refactor later.

src/components/molecules/DatabaseErrorScreen/DatabaseErrorScreen.test.tsx
Orlandohub: This test passes against the component in isolation — `DatabaseErrorScreen` keeps the button mounted and disabled with a retrying label, no full-screen spinner. The full-screen spinner only comes from the parent DatabaseProvider, which is out of scope here. Keeping the test as a valid component-level check.

PR #1943, opened 2026-06-02
feat(config): implement runtime config via server-injected window config

### LINKED ISSUE #1939: [Feature]: Implement runtime config
### Problem Statement

based on: https://github.com/pubky/pubky-app/pull/1644 and https://synonymworkspace.slack.com/archives/C09HQTLAARF/p1775468666923319

the problem: currently the variables like `NEXT_PUBLIC_NEXUS_URL` , `NEXT_PUBLIC_HOMESERVER_ID` provided during build, so we have to build separate docker images for each of env. 

### Proposed Solution

We need to support dynamic variables using runtime config.

### Suggested Priority

None

Orlandohub:
## Summary
- fix #1939
- Makes the environment-specific network values runtime-configurable so a single Docker image can be promoted across staging/prod/testnet, unblocking the Ansible migration.
- Values are read from non-inlined `PUBKY_RUNTIME_*` env on the server, injected synchronously into the HTML (`window.__PUBKY_CONFIG__`), and read via lazy isomorphic getters — no async fetch, no build-artifact rewriting.

## Changes
- Add `src/libs/runtime-config`: a shared zod schema (split `runtimeEnvInputSchema` for string env vs `runtimeConfigValueSchema` for the parsed window object) plus a memoized isomorphic resolver, lazy getters, and an HTML-safe serializer.
- Inject `window.__PUBKY_CONFIG__` from the already `force-dynamic` root layout (`ContainerRoot`).
- Route `config/nexus`, `config/network`, `homeserver`, `homegate`, and avatar utils through getters; remove module-load `IS_TESTNET` and the direct `Env.NEXT_PUBLIC_HOMESERVER` read.
- `env.ts` reuses the shared field validators so `Env` stays behavior/type-compatible.
- Fail loud in deployed envs (`NODE_ENV=production` or `PUBKY_RUNTIME_CONFIG_REQUIRED=true`); fall back to `NEXT_PUBLIC_*` in dev/test. ESLint guard blocks direct `NEXT_PUBLIC_` network reads (Sentry testnet gate kept build-time as a documented limitation).
- Drop the 7 runtime build ARGs from the `Dockerfile`; update `.env.example`, `docs/environment.md`, and add ADR 0017.

## Test plan
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes (guard ignores `config/test.ts` assignments)
- [x] New unit tests pass: schema split, defaulted-schema parsed defaults, resolver (server present/partial-throw/dev-fallback/required-throw/build-guard, client read/invalid/fallback/required-throw), serializer escaping (19 tests)
- [x] Updated existing nexus/homegate/avatar tests to getter form
- [ ] Deploy verification with real `PUBKY_RUNTIME_*` (staging/prod/testnet) — to be done in deployment pipeline

## Notes
- Contract shift: staging runs `NODE_ENV=production`, so staging must now set `PUBKY_RUNTIME_*`; `NEXT_PUBLIC_*` network values become local dev/test defaults only.
- Local development needs no changes — the resolver falls back to `NEXT_PUBLIC_*`.
- Known limitation: `shouldEnableSentry()` keeps the build-time `NEXT_PUBLIC_TESTNET` read because Sentry initializes before the injected config exists (documented in ADR 0017).
- Pre-existing, environment-only test failure unrelated to this PR: `env.test.ts` "Sentry secrets undefined" fails locally when a `SENTRY_AUTH_TOKEN` is present in the environment/`.env`.


Orlandohub:
@greptile-apps review

Orlandohub:
@greptile-apps review

src/libs/runtime-config/runtime-config.schema.ts
```diff
@@ -0,0 +1,151 @@
...
+const pkarrRelaysFromString = z.string().transform((val, ctx) => {
+  try {
+    return parsePkarrRelaysString(val);
+  } catch (error) {
+    ctx.addIssue({ code: 'custom', message: `Invalid PKARR_RELAYS value: ${(error as Error).message}` });
+    return z.NEVER;
+  }
+});
+
+const testnetFromString = z.string().transform((val) => val === 'true');
```
infin1t3: If we misconfigure production runtime config, this will not fail loudly as stated in this PR, instead it will just default to `false`.

Consider making this string parser stricter:
```
const testnetFromString = z.enum(['true', 'false']).transform((val) => val === 'true');
```
and adding test, like
```
it('throws on invalid TESTNET value', () => {
  expect(() =>
    runtimeEnvInputSchema.parse({
      ...VALID_ENV_INPUT,
      testnet: 'tru',
    }),
  ).toThrow();
});
```

Dockerfile
infin1t3, line 33: Im not sure this is in the scope of this ticket, but may deserve a separate ticket on its own.

infin1t3 (changes requested):

Dockerfile
infin1t3: Edit: Go to diff screen and you will see warnings for this line

docs/adr/0017-runtime-config-injection.md
```diff
@@ -0,0 +1,76 @@
...
+- Resolver/getters/serializer: `src/libs/runtime-config/runtime-config.ts`
+- Injection: `src/components/molecules/ContainerRoot/ContainerRoot.tsx`
+- Consumers: `src/config/nexus.ts`, `src/config/network.ts`, `src/core/services/nexus/nexus.utils.ts`, `src/core/services/homegate/homegate.api.ts`, `src/core/services/homeserver/homeserver.ts`, `src/components/organisms/AvatarWithFallback/AvatarWithFallback.utils.ts`
+- ESLint guard: `eslint.config.mjs` (`no-restricted-syntax`), with a single inline exception in `src/libs/observability/sentry.ts`
+- Optional ops hardening (not implemented in v1): validate + warm the server config at startup via `register()` in `src/instrumentation.ts` so a misconfigured deploy fails fast on boot/health check.
+
+## References
+
+- [Next.js — Runtime environment variables](https://nextjs.org/docs/app/guides/environment-variables#runtime-environment-variables)
+- GitHub issue #1939; draft PR #1644
```
infin1t3: Remove draft prefix

src/libs/runtime-config/runtime-config.ts
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * - Server: reads non-inlined `PUBKY_RUNTIME_*` env at request time, validates, and memoizes.
+ *   The same memoized object is serialized into the HTML (see `serializeRuntimeConfig`).
+ * - Client: reads the injected `window.__PUBKY_CONFIG__`, validates, and memoizes.
+ * - dev/test: falls back to `NEXT_PUBLIC_*` (honoring `.env.local` / `src/config/test.ts`).
+ *
+ * This module must NOT import `Env` (keeps the import graph a leaf, avoids the env<->logger cycle).
+ */
+
+export const RUNTIME_CONFIG_WINDOW_KEY = '__PUBKY_CONFIG__' as const;
```
infin1t3: Why as const?

Orlandohub:
@greptile-apps review

Dockerfile
Orlandohub: resolved as part of ADR 0019

Orlandohub:
@greptile-apps review

infin1t3 (approved):
Tested builds locally and runtime success paths.

```
npm run build, and it passed.
npm run start with full PUBKY_RUNTIME_*, and the injected config matched the runtime values.
npm run start with no runtime env, and it failed fast as expected.
npm run start with partial runtime env, and it failed fast as expected.
docker build -t pubky-app-runtime-config-local ., and it passed.
Docker run with full PUBKY_RUNTIME_*, both on 3001 first and then on 3000, and injected config matched.
Docker run with no runtime env, failed fast as expected.
Docker run with partial runtime env, failed fast as expected.
```

Looks good to me overall, and I’m okay with merging this as-is. I think the runtime-config direction is solid. A few follow-up cleanup items I’d suggest tracking after merge:

- `src/libs/env/env.ts`: this still duplicates quite a few defaults/parsing rules that now also live in `src/libs/runtime-config/runtime-config.schema.ts` (`APP_RUNTIME_DEFAULTS`, boolean/number parsing, moderation tags, metadata/link defaults, etc.). It would be good to derive more from the runtime-config defaults/helpers so local fallback validation can’t drift from deployed runtime validation. In particular, `NEXT_PUBLIC_MODERATED_TAGS` still uses raw `JSON.parse` while runtime config uses the safer shared parser path.

- `src/libs/env/env.test.ts`: small stale wording. The test says “These are now REQUIRED variables,” but after this PR they’re local/dev fallback values; deployed required vars are `PUBKY_RUNTIME_*`.

- `src/libs/observability/sentry-test-harness.ts`: `NEXT_PUBLIC_DEBUG_MODE=true` bypasses the production-environment gate for `/sentry-test`. The comment says this force-enables local builds, but the code does not restrict it to local/dev. Safer follow-up:
  ```ts
  if (Env.NODE_ENV !== 'production' && Env.NEXT_PUBLIC_DEBUG_MODE) return true;
  return resolveEnvironment() !== PRODUCTION_ENVIRONMENT;
  ```
  or remove the debug override if runtime `sentryEnvironment` is now the source of truth.

- `src/config/nexus.ts`: tiny cleanup only. The exports can be a single line:
  ```ts
  export { getCdnUrl, getNexusUrl, getStreamCacheMaxAgeMs };
  ```

- `src/core/application/moderation/moderation.utils.ts`: not a blocker, but `detectModerationFromTags()` now rebuilds `new Set(getModeratedTags())` on every call. Runtime config itself is memoized, and the list is tiny, so it’s fine. If this path becomes hot during stream sync, we could memoize the set by tag-array identity.

- `src/hooks/useUserProfile/useUserProfile.test.tsx`: stale comment still says `DEFAULT_URL`:
  ```ts
  // Uses DEFAULT_URL from config (SSR-safe)
  ```
  Should be updated to runtime/default URL wording.

- Dockerfile comment thread `r3372267892`: the thread is marked resolved with “resolved as part of ADR 0019,” but I don’t see ADR 0019 in this PR. Current Dockerfile still has `ARG SUPPORT_API_ACCESS_TOKEN`, which was the original warning. If ADR 0019 is meant to accept that risk, can you link it or add it in a follow-up? Otherwise we should move support credentials out of Docker build args later.

Main thing I’d want fixed sooner rather than later is the E2E workflow compatibility: the reusable Docker workflow removed the `env` input, but `e2e-test.yml` still passes it, and the external stack compose still appears to expect `pubky-app-${NETWORK}` rather than the new unified `pubky-app` image plus `PUBKY_RUNTIME_*` env.

Also after this PR, `npm run dev` still falls back to `NEXT_PUBLIC_*`, while `npm run start` / Docker require `PUBKY_RUNTIME_*`. That is technically fine, but it can confuse local setup because dev may keep working while production-style runs fail.

Can we make sure the follow-up PR updates `.env.example` and `docs/environment.md` clearly, especially showing that `NEXT_PUBLIC_*` network values are now local/dev fallbacks and deployed/runtime values should use `PUBKY_RUNTIME_*`?

PR #1945, opened 2026-06-03
fix(ui): correct feed scroll position on post return and feed navigation (#1348)

### LINKED ISSUE #1348: [Bug]: Feed is scrolled down when navigating back via Pubky logo or home button
### Description

When I navigate back from a post by clicking on Pubky logo or home button, I get scrolled down to the bottom of the page.

### Expected Behaviour

Feed position should be retained when navigating back from a post.
Feed position is not retained when navigating back from other pages such as profile.
Should not scroll down beyond when I left off.

### Steps to Reproduce

See vid

### Screenshots

https://github.com/user-attachments/assets/41ea20e3-64f4-40f7-b8ec-a4f66332010a

### Console Log

```shell

```

### Browser & Version

_No response_

Orlandohub:
## Summary
- fix #1348
- Feed scroll position is now correct in two related situations across the persistent `(feeds)` cluster (`/home`, `/feed/[id]`, `/bookmarks`, `/search`):
  1. Returning to a feed from an intercepted post (Pubky logo, home button, feed nav, browser back) retains the prior scroll position instead of jumping to the bottom.
  2. Forward-navigating between feeds (e.g. home → bookmarks) lands at the top instead of inheriting/clamping the previous feed's offset (it used to load "in the middle").
- All scroll positioning is owned centrally by the feeds layout; nav components only signal intent, so the fix is uniform across logo, header, mobile footer, feed nav, and browser back.

### 1. Retain feed scroll when returning from a post
- `src/app/(feeds)/layout.tsx`: track the visible feed's `window` scroll offset and its route; on an intercepted-post close, restore the offset only when returning to the SAME feed route. Cross-route closes never splash one feed's offset onto another.
- Scroll listener gated on both post activity and feed-route visibility (via render-safe guard refs updated in the layout effect) so the post overlay and non-feed pathnames can't contaminate the saved offset.

### 2. Reset to top on intra-cluster feed navigation
- `src/config/feed.ts`: rename `FORCE_HOME_SCROLL_TOP_KEY` → `FORCE_FEED_SCROLL_TOP_KEY` (now feed-wide).
- `src/libs/utils/feedScrollTop.ts` (new): shared `handleFeedNavClick` util — sets a one-shot top-scroll flag when navigating to a feed the user isn't already on; smooth-scrolls to top when already there; leaves modified clicks alone.
- `src/app/(feeds)/layout.tsx`: consume the flag centrally — scroll to top on a flagged feed route change, then clear it. Browser back never sets the flag, so native history scroll restoration is preserved.
- Wire the flag into `Logo`, `MobileFooter` (Home/Search/Bookmarks), desktop `Header` (Home/Bookmarks), and `FeedNavigation` (Home + custom feeds).
- `Home.tsx`: remove the now-redundant per-page consume effect — the layout is the single owner of feed scroll positioning.

https://github.com/user-attachments/assets/2f9e5bae-4ebf-4dbf-b41f-cf6704678667

### Tests
- `src/app/(feeds)/layout.test.tsx`: same-route restore, cross-route no-restore, non-feed pathname not tracked/contaminating, no restore on mount/while opening, scroll-while-post-active doesn't overwrite, flag cleared on same-route restore, forward-nav top-scroll (flag set) + clear, and no force-scroll without the flag (browser back).
- `MobileFooter.test.tsx` adds coverage for setting the flag when navigating to a non-home feed; `Logo`/`Home` tests updated for the rename / centralized ownership.

## Test plan
- [ ] Manual: same-route return via logo / home button / feed nav / browser back retains position; entering `/home` from a profile lands at top
- [ ] Manual: cross-feed (post from `/bookmarks` → logo → `/home`) lands at top, not the bookmarks offset; browser back returns to `/bookmarks` at its prior position

## Notes
- The feed scrolls the window (no inner scroll container) and the `(feeds)` layout is persistent across intra-cluster navigation, so `window.scrollY` is neither reset on feed switches nor preserved when a post overlay clobbers it — hence the explicit save/restore + intent-flag, which together mirror native browser-back behavior while still allowing deliberate top-scroll on forward navigation.
- Guard refs are updated at the top of the layout `useLayoutEffect` (not during render) to satisfy the project's `react-hooks/refs` rule; layout effects run after the `display:none` commit and before the clamp scroll event, so the timing hole is still closed.
- Out of scope: search-box-driven navigation to `/search` (typing a query) does not set the flag and relies on default behavior.


Orlandohub:
@greptile-apps review

src/app/(feeds)/layout.tsx
Nirzhuk, line 1: My concern in general on this file is related to having 'use client' in the first level file (layout.tsx) I woud like to have all these code in deeper levels but I'm being Next.JS purist more than anything.

src/app/(feeds)/layout.tsx
```diff
@@ -76,6 +80,79 @@ export default function FeedsLayout({ children, post }: { children: React.ReactN
   //      optional, so spreading `undefined`s is safe.
   const shellConfig = resolved ?? (isPostActive ? lastFeedsConfig : null);
 
+  // Feed scroll preservation across the intercepted post overlay (#1348).
```
Nirzhuk, line 83: Everything is extremely verbose commented by AI, I will appreciate it if we can make it more human.

for example:

```
   // While the post overlay is active (pathname is `/post/...`) tracking is
    // frozen: `savedScrollRef` + `lastFeedPathnameRef` hold the source feed.
    if (isPostActive) return;
```
I feel doesn't need a comment.

Related to this code I feel it's a very interesting code which can be reused as hook pretty easily.

src/app/(feeds)/layout.tsx
Orlandohub: Agreed, this is mostly a boundary hygiene concern. Now this is something which we're doing at almost all our first level files and if we move in this direction I suggest we create an issue where we should update all first level files and likely also update docs so it becomes a rule. I would not block this PR on that though. wdyt?

Nirzhuk (changes requested):
If you change from the feed using the Sidebar is going to create this crazy interaction scrolling by his own until decides it's good, so we need to prevent this.

https://github.com/user-attachments/assets/f83b0c90-bc00-4bf7-a2b9-99fd65466793


This is happening with any interaction with the sidebar which changes the feed content.

src/app/(feeds)/layout.tsx
Nirzhuk: I feel the proper path it's create a tech debt task related to these layouts with states to refactor them

src/app/(feeds)/layout.tsx
Orlandohub: I believe this is the same behavior from prod and staging which is not a crazy scroll interaction but the feed loading at the same position as before

Nirzhuk (approved):

PR #1948, opened 2026-06-04
tests: fix e2e - remove toast check for tag add/remove and check for add new tag sort order

catch-21:
Remove check for toast on add/remove tag.
Add check for tag ordering (newest shown first)


catch-21:
Passing E2E tests: https://github.com/pubky/pubky-app/actions/runs/26952442415

PR #1955, opened 2026-06-05
feat(notification): sync last_read across devices via homeserver SSE

### LINKED ISSUE #1595: [Bug]: Read notifications on devices still shows unread on another
### Description

I use PubkyApp mostly on desktop. I usually read all notifications. When I then occasionally use it on my mobile phone, then I still see a lot of unread notifications that I already read on desktop.

### Expected Behaviour

Looks like the notification read status is not synced over multiple devices which it should IMO

### Steps to Reproduce

1. Get a new notification. 
2. Read it on device 1.
3. Go to device 2 and see the same notification again.

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

Taewa:
## Summary

Fixes #1595 — read notifications on one device now clear on other active sessions. Sibling to the mute-list sync (#1787).

## Root cause

- Read state lives on the homeserver (`last_read`), but a device only **read** it once, at startup.       
- Opening the notifications page **writes** a new `last_read`; other devices never re-read it, so their
unread badge stayed stale.
- The cross-device re-read also returned a **cached** value: single-file reads went through the browser
HTTP cache, so a value written elsewhere stayed invisible until the local cache was invalidated. (Mute
didn't hit this — it reads a directory listing, which isn't cached.)

## Fix

- **Live sync coordinator** — subscribes to the homeserver event stream for `last_read`; on change,
debounces then refreshes the local read state and recomputes the unread badge.
- **Race guard** — never moves `last_read` backward (ignores stale/echo values).
- **Cache bypass** — the cross-device refresh reads the live value instead of the cached copy. Normal
startup read keeps the cached path.
- **Bounded reconnect/refresh** — exponential backoff with a max-attempt cap so a persistently
unreachable homeserver can't loop forever.

## Testing

- Unit: application, controller, coordinator (events, debounce, race guard, route/visibility gating,
cursor persistence, retry cap), homeserver cache-bypass, bootstrap regression.
- Manual: marked read on Chrome → Firefox badge clears, and vice versa; verified the fresh (non-cached)
`last_read` GET in the network tab.

## Manual test

https://github.com/user-attachments/assets/49f09fab-88f2-4695-9a28-7c327bdb0442

Tested on URL: https://pubky-app-preview-pr-1955-fzxmjul7ya-oa.a.run.app/
pubky-app-preview-pr-1955 • 549d379


Taewa:
@greptile-apps
Fixed your feedback "Upward layer dependency: Application imports from Controller".

Review again plz.

infin1t3:
@greptileai review

src/core/coordinators/notification-last-read-sync/notification-last-read-sync.ts
```diff
@@ -0,0 +1,410 @@
...
+    }
+  }
+
+  public destroy(): void {
+    this.stop();
+    this.removeListeners();
+  }
+
+  private setupListeners(): void {
+    this.authStoreUnsubscribe = useAuthStore.subscribe((state, prevState) => {
```
infin1t3, line 120:
```
    this.authStoreUnsubscribe = useAuthStore.subscribe((state, prevState) => {
      const isAuthenticated = state.selectIsAuthenticated();
      const wasAuthenticated = prevState.selectIsAuthenticated();
      const profileChanged = state.hasProfile !== prevState.hasProfile;
      const userChanged = state.currentUserPubky !== prevState.currentUserPubky;
      if (isAuthenticated !== wasAuthenticated || profileChanged || userChanged) {
        this.evaluateStreaming();
      }
    });
```

Exists verbatim in `mute-list-sync.ts` - you could consider extracting to an utility or a helper.


More importantly - This should compare the snapshot fields directly instead of calling `selectIsAuthenticated()` on `state` / `prevState`.

`selectIsAuthenticated()` is defined as a function that uses Zustand’s `get()` internally. That means the function does not read from the object it is attached to. It always asks the store for the latest current state.

So inside this subscription:

```ts
useAuthStore.subscribe((state, prevState) => {
  const isAuthenticated = state.selectIsAuthenticated();
  const wasAuthenticated = prevState.selectIsAuthenticated();
});
```

both calls can read the same latest store value. If `session` just changed from `null` to restored, we expect `isAuthenticated = true` and `wasAuthenticated = false`, but both calls may return `true`.

Suggested fix:

```ts
const isAuthenticated = state.session !== null;
const wasAuthenticated = prevState.session !== null;
```

Same duplicated pattern exists in `MuteListSyncCoordinator`, so we should update both coordinators for consistency.

src/core/controllers/notification/notification.ts
```diff
@@ -117,4 +118,42 @@ export class NotificationController {
...
+  /**
+   * Pulls the homeserver `last_read` value and reconciles it with the local store.
+   * Called by `NotificationLastReadSyncCoordinator` after a homeserver PUT event.
+   *
+   * Race guard: skip if remote is not newer than local. Otherwise updates `lastRead`
+   * and recomputes the preference-filtered unread count.
+   * Example: device A marks read at t=1000 → PUT → SSE echo back to A → refresh()
+   * fetches remote=1000, local=1000 → no-op.
+   */
+  static async refreshLastReadFromHomeserver(pubky: Pubky): Promise<void> {
```
infin1t3, line 139: This one is from Cursor - `refreshLastReadFromHomeserver` introduces a refresh* controller method. 

Why it matters: Controller prefixes are standardized (fetch*, get*, getOrFetch*, commit*, subscribe*) so call sites communicate cache/network/write behavior. 

Fix: Rename to a documented prefix, e.g. fetchLastReadFromHomeserverAndReconcile or similar, then update coordinator/tests/docs.

- `fetch*` — network only, no cache
- `get*` — local only
- `getMany*` — bulk local reads, returns `Map<Pubky, T>`
- `getOrFetch*` — local first, network fallback
- `getMany*OrFetch` — bulk local first, fetch missing
- `commitCreate*` / `commitUpdate*` / `commitDelete*` — optimistic local write + network sync
- `subscribe*` — long-lived live stream subscription

infin1t3 (changes requested):

Issue #1869, opened 2026-05-21
[Task]: Collections content filter visibility


infin1t3:
## Context

Collections are supported as a content type and have been implemented in the home content filter.

## Updated requirement

The **Collections** content filter should exist in the home content filter UI.

However, collection posts should **not** appear in:

- **All**
- **Short**
- **Long**
- **Images**
- Any other non-Collections content filter

Collection posts should only be visible when the user explicitly selects the **Collections** content filter.

## Acceptance criteria

- [x] Sidebar/mobile content filter includes **Collections**
- [x] Selecting **Collections** shows collection posts only
- [x] Collection posts are hidden from **All**
- [x] Collection posts are hidden from every other non-Collections content filter
- [x] Existing filters continue to work without showing collections

PR #1964, opened 2026-06-09
feat(feeds): add collections content filter

infin1t3:
Wire collection posts into the home content filter and stream mappings so users can filter the main timeline to collection items.

https://github.com/user-attachments/assets/31fd361d-3b14-4b41-9d8e-9b63d57a2f7f

Resolves: #1869


infin1t3:
@greptileai review

Issue #1466, opened 2026-03-06
[Epic] Collections: Feature Suite


SHAcollision:
# Context

From [Northstar Roadmap #56](https://northstar-eta.vercel.app/?task=56) | Due: Mar 24 | [Figma prototype](https://www.figma.com/proto/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?page-id=18345%3A2204&node-id=22990-172646)

**REVIEW LATEST FIGMA DESIGN** Take this screenshots as an example only.

https://www.figma.com/proto/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?page-id=18345%3A2204&node-id=22990-172646&p=f&viewport=-16945%2C2901%2C1&t=X9szReaz6YVgax6l-8&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=22990%3A172646&show-proto-sidebar=1&hide-ui=1

https://www.figma.com/proto/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?page-id=18345%3A2204&node-id=23375-140294&p=f&viewport=-16945%2C2901%2C1&t=X9szReaz6YVgax6l-8&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=23375%3A140294&show-proto-sidebar=1&hide-ui=1

## Background

Pubky has a flat bookmarks system. You bookmark posts, they go into one list, done. Collections add a grouping layer on top: named lists of bookmarked posts, with social features (follow other people's collections, discover public ones), visibility controls (public/private), and eventually payment gating (locked collections via Paykit).

The Figma designs are thorough. 13 screenshots in `figma/` covering desktop, mobile, dialogs, locked/unlocked states, empty states, and the collection picker component. This is well-designed but large in scope.

Nothing exists in the codebase. No models, no hooks, no routes, no Dexie tables, no Nexus API types. The bookmark system (`BookmarkModel`, `BookmarkApplication`, `LocalBookmarkService`, `useBookmark`) is the direct reference pattern for how to build this, following ADR-0004 (layering) and ADR-0009 (orchestration).

### Collections page (Desktop / Mobile)

<img width="2560" height="2512" alt="Image" src="https://github.com/user-attachments/assets/59c908e4-7fd3-4710-a2c4-39cd6ae21fee" />

Mobile (figma):
[...]

PR #1967, opened 2026-06-10
feat: edit and delete collections

secondl1ght:
# feat: edit collection dialog + instant cover preview via local-files store

  ## Summary

  Wires up the Edit button in `CollectionHero` to a new **Edit Collection** dialog that lets owners change a collection's
  **title, description, and cover image**. The same dialog also opens from the post menu when editing a collection post in
   the timeline / post route — mirroring how `DialogEditPost` already swaps in an "Edit Article" UI for `kind === 'long'`
  posts.

  Along the way, fixes a pre-existing gap in the local-first cover-image flow: covers on freshly created/edited
  collections used to render broken until the CDN caught up. They now preview instantly via the `useLocalFilesStore`,
  matching the pattern already used for post attachments and avatars.

  ## Why a separate dialog (not `DialogEditPost`)?

  `DialogEditPost` is built around `PostInput` — a textarea + attachment grid. Collections aren't text content; they're a
  structured JSON envelope (`name` / `description` / `cover_image` / `items`). The create flow already lives in its own
  `NewCollectionDialog` (not in `PostInput`), so the edit flow is the symmetric mirror: a dedicated `EditCollectionDialog`
   rather than a `DialogEditPost` that conditionally renders an entirely different form.

  ## What changed

  ### Core / controllers
  - **`PostController.commitEditCollection`** (`src/core/controllers/post/post.ts`) — uploads new cover Files via
  `FileApplication`, parses the current envelope, merges name/description/items/cover, and round-trips through
  `PostNormalizer.toEdit` → `PostApplication.commitEdit`. Mirrors the shape of `commitCreateCollection`.
  - **`TEditCollectionParams`** type added next to existing collection param types.

  ### Why cover editing works for collections (but not articles/posts)

  Articles store their cover as `attachments[0]` and `PostNormalizer.toEdit` treats `attachments` as immutable (the
  `pubky-app-specs` `builder.editPost` only accepts new `content`). Collections store `cover_image` as a **top-level field
   inside the JSON content envelope**, so changing it is mechanically identical to `commitUpdateCollectionItem`'s existing
   `items` updates — no spec change needed.

  ### Hooks
  - **`useEditCollection`** — mirrors `useCreateCollection`'s shape (`{ form, cover, submit, reset }`) but loads via
  `usePostDetails`, prefills name/description, seeds the cover picker, returns `Promise<boolean>` from `submit()`.
  - **`useCreateCollection`** — now seeds the local-files store with a blob URL after a successful commit (same pattern as
   `usePostInput.handleSuccess`).

  ### Components
  - **`CollectionFormDialog`** (new presentational shell) — title/submitLabel/isSaving as props; `disableOpenAutoFocus`
  prop for opting into Radix's `onOpenAutoFocus` suppression.
  - **`NewCollectionDialog`** — refactored into a thin wrapper around the shared shell. Snapshot DOM preserved.
  - **`EditCollectionDialog`** — new wrapper, controlled `open`, mounts `useEditCollection`. Suppresses autofocus so the
  prefilled title field isn't highlighted on open.
  - **`CollectionHero`** — Edit button now opens `EditCollectionDialog` (was a stub). Reads local-files store for the
  cover with envelope fallback.
  - **`CollectionCard`** — reads local-files store for the cover with envelope fallback. Added `isolate` to the outer Card
   so the `-z-10` cover background doesn't escape into a nested `PostMain` stacking context (was rendering nothing in the
  timeline feed).
  - **`PostMenuActions`** — routes to `EditCollectionDialog` instead of `DialogEditPost` when `postDetails.kind ===
  'collection'`.

  ### Local-first cover previews
  - **`useLocalFilesStore`** — added a `collections: Record<string, string | undefined>` map and `setCollectionCover(id,
  blobUrl)` action. `reset()` revokes collection blob URLs on logout.
  - Hook layer (`useCreateCollection`, `useEditCollection`) writes blob URLs on success; component layer
  (`CollectionHero`, `CollectionCard`) reads with envelope fallback.
  - The edit hook also **seeds the picker** from the local store on reopen — so reopening the dialog right after a cover
  change shows the freshly-uploaded cover instead of a broken CDN URL.

  ### Translations
  Added `collections.edit` namespace (title, save, updated, updateFailed) across all 9 locales: en, es, fr, de, it, pt-BR,
   ja, zh, ar. Shared form labels reuse `collections.new` to keep duplication minimal.

  ## Bug fixes captured during QA

  ### 🖼️  Cover image missing in timeline `CollectionCard` even on brand-new collections
  **Root cause:** `-z-10` escaped `CollectionCard`'s relative parent (which doesn't create a stacking context without an
  explicit z-index) and landed behind `PostMain`'s opaque `bg-card`.
  **Fix:** Added `isolate` to `CollectionCard`'s outer Card.

  ### ⏳ Saving state appeared late / not at all when uploading a cover
  **Root cause:** RHF's `formState.isSubmitting` flip was batched into a microtask, and heavy `stripImageMetadata` canvas
  re-encoding could block the main thread before React painted.
  **Fix:** Local `isSavingLocal` state per wrapper, flipped synchronously via `flushSync`. Shell reads `isSaving` from a
  prop.

  ### ✨ Prefilled title field shown highlighted on edit-dialog open
  **Root cause:** Radix Dialog autofocuses the first focusable element.
  **Fix:** `CollectionFormDialog` accepts `disableOpenAutoFocus`; `EditCollectionDialog` opts in via Radix's
  `onOpenAutoFocus` with `preventDefault()`.

  ### ❌ X close button required two clicks on an empty title
  **Root cause:** First click triggered blur → `mode: 'all'` validation → required-error rendered → dialog grew → centered
   dialog shifted up → X moved out from under the mouseup.
  **Fix:** Switched RHF mode from `'all'` to `'onChange'`. User still gets required feedback as soon as they type and
  erase.

  ### 🖼️  Cover empty when reopening the edit dialog right after a cover change
  **Root cause:** `useEditCollection` only seeded `initialPreviewUrl` from
  `resolveCollectionCoverImage(envelope.cover_image)`; CDN hadn't indexed the new URL yet.
  **Fix:** Hook now reads `useLocalFilesStore.collections[compositeId]` and prefers it over the resolved CDN URL. Envelope
   value preserved for submit-without-changes round-trips.

  ## Tests

  **+50 new tests** across the controller, store, hooks, and component layers. Final suite: **639 files, 10,282 passing, 4
   skipped**, no regressions. Typecheck and lint clean.

  - `PostController.commitEditCollection` — 7 tests (File upload, string passthrough, null clearing, items preserved,
  missing/invalid/non-author/upload-failure errors)
  - `useLocalFilesStore.setCollectionCover` — 8 tests (set/clear/revoke/non-blob/independence/no-op + reset coverage)
  - `useCreateCollection` — 2 tests covering local-store blob URL seeding
  - `useEditCollection` — 9 tests (hook prefill, submit paths, cover-state branches, local-store preview seeding,
  regression: submit-without-changes preserves the *envelope* URL even when the local store holds a blob)
  - `CollectionFormDialog` — 14 tests + 3 snapshot scenarios (default, with cover preview, saving state) (new file)
  - `EditCollectionDialog` — 6 tests + 1 snapshot (prefilled state) (new file)
  - `CollectionCard` — 2 tests for local-cover fallback + snapshot updated for the `isolate` class
  - `CollectionHero` — 4 tests (Edit button opens dialog with composite id, non-owners don't mount it, local cover
  preferred, local cover renders even with no envelope cover)
  - `PostMenuActions` — 5 tests for the post-kind routing (short/long/collection paths, click flow, loading-state
  fallback)

  ## Related

#1466 

  ## Notes

  - Replacing a cover orphans the previous file on the homeserver (no cleanup). This mirrors how all current file uploads
  behave today — out of scope here.
  - The architecture rule "Application never accesses Stores" is upheld: all `useLocalFilesStore.setCollectionCover`
  writes happen at the hook layer, matching the pattern used by `usePostInput.handleSuccess` and `useProfileForm`.


secondl1ght:
@greptile

secondl1ght:
@greptile

secondl1ght:
@greptile stop making random comments every time you are pinged, catch all the real issues in the initial review and on re-review confirm they were addressed or not.

infin1t3 (approved):
Fine with minor comments communicated in Slack.

src/components/organisms/CollectionHero/CollectionHero.tsx
```diff
@@ -99,10 +111,34 @@ function CollectionHeroContent({ authorPubky, compositeId, postDetails, classNam
...
+    toastMessages: {
+      deleted: tCollectionToast('collectionDeleted'),
+      deletedDesc: tCollectionToast('collectionDeletedDesc'),
+      deleteFailed: tCollectionToast('deleteFailed'),
+    },
+  });
+  const [deleteConfirmOpen, setDeleteConfirmOpen] = useState(false);
+  const handleDelete = () => setDeleteConfirmOpen(true);
+  const handleDeleteConfirm = async () => {
+    await deletePost(compositeId);
```
infin1t3, line 139: `deletePost` never throws on error, instead it shows a toaster, which means that, on error, you will get a toast and `router.replace(APP_ROUTES.COLLECTIONS)` will still be executed.

One of the ways to fix this is to have `deletePost` returning `Promise<boolean>` and then doing something like
```
const handleDeleteConfirm = async () => {
  const deleted = await deletePost(compositeId);
  if (deleted) {
    router.replace(APP_ROUTES.COLLECTIONS);
  }
};
```

Or any other strategy that prevents execution if a method throws.

infin1t3:
@greptile review

src/components/organisms/CollectionHero/CollectionHero.tsx
```diff
@@ -99,10 +111,34 @@ function CollectionHeroContent({ authorPubky, compositeId, postDetails, classNam
...
+  const { deletePost, isDeleting } = useDeletePost({
+    toastMessages: {
+      deleted: tCollectionToast('collectionDeleted'),
+      deletedDesc: tCollectionToast('collectionDeletedDesc'),
+      deleteFailed: tCollectionToast('deleteFailed'),
+    },
+  });
+  const [deleteConfirmOpen, setDeleteConfirmOpen] = useState(false);
+  const handleDelete = () => setDeleteConfirmOpen(true);
+  const handleDeleteConfirm = async () => {
```
infin1t3, line 138: With the tombstone approach, you have a small visual race condition here.
Tombstone is written, but the component flips to `CollectionNotFound ` before `router.replace` runs, and you can see a short flash. Check my video at mark 01:53

https://github.com/user-attachments/assets/8edfe44c-f74b-4de1-88db-b77d43e63cd9

src/hooks/useDeletePost/useDeletePost.tsx
infin1t3, line 98: Looking at the code and tests, but did not try to reproduce this — with the tombstone fix, if homeserver DELETE fails after local delete already succeeded, getDetails won't return null anymore (you get a [DELETED] row). The hook still treats that as "post exists" and can call prependPosts, which would put a deleted post back in the feed.

Consider something like:
```
const stillLive =
  postStillExists === 'unknown' ||
  (postStillExists !== null && !isPostDeleted(postStillExists.content));

if (stillLive) {
  timelineFeed?.prependPosts(postId);
}
```

src/components/organisms/PostMenuActions/PostMenuActions.tsx
```diff
@@ -24,6 +26,11 @@ export function PostMenuActions({ postId, trigger }: PostMenuActionsProps) {
   const [deleteConfirmOpen, setDeleteConfirmOpen] = useState(false);
   const { deletePost, isDeleting } = useDeletePost();
```
infin1t3, line 27: Collection delete from the post menu uses:
* Generic DialogConfirmDelete (post copy, not dialogs.deleteCollection)
* Generic useDeletePost() toast ("Post deleted" not "Collection deleted")

CollectionHero / CollectionCard pass collection-specific overrides. Timeline/menu path doesn't.

src/hooks/useEditCollection/useEditCollection.ts
```diff
@@ -0,0 +1,172 @@
...
+    // `originalName` / `originalDescription` above can still be stale. Clearing
+    // the ref lets the effect catch up once the fresh envelope arrives — so a
+    // reopened dialog shows the committed values rather than the pre-save ones.
+    hasPrefilledRef.current = false;
+  };
+
+  return {
+    form,
+    cover,
+    isLoaded: !!collection,
```
infin1t3, line 168: This isn't used anywhere.

src/hooks/useEditCollection/useEditCollection.types.ts
```diff
@@ -0,0 +1,4 @@
+export type UseEditCollectionParams = {
```
infin1t3, line 1: Can be inlined cause it has only one caller.

src/components/molecules/CollectionDeleted/CollectionDeleted.tsx
```diff
@@ -0,0 +1,39 @@
+'use client';
+
+import { useTranslations } from 'next-intl';
+import { Card, CardContent } from '@/atoms/Card/Card';
+import { Container } from '@/atoms/Container/Container';
+import { Typography } from '@/atoms/Typography/Typography';
+import { cn } from '@/libs/utils/utils';
+
+interface CollectionDeletedProps {
+  className?: string;
```
infin1t3, line 10: Not sure you ever use it. I see you pass it through the chain but it looks `undefined` so maybe you don't need it.

src/components/organisms/Collections/DiscoverCollections/DiscoverCollections.tsx
```diff
@@ -179,15 +181,28 @@ export function DiscoverCollections() {
         let reachedPageCap = false;
         if (result.nextPageIds.length > 0) {
           // Read the local bookmark set after persist has committed.
-          // Filter own + already-bookmarked.
+          // Filter own + already-bookmarked + deleted collections.
           const bookmarkedIds = new Set(await BookmarkController.getAll());
           if (token?.cancelled) return;
+          // Slice post details to identify already-deleted collections in this
```
infin1t3, line 187: This logic can probably go into the hook

infin1t3 (changes requested):

src/components/organisms/CollectionHero/CollectionHero.tsx
secondl1ght: The issue is actually bigger than that and it is pre-existing, if delete fails the post appears deleted locally and is not rolled back even though there was no update to the homeserver or Nexus. I think this is out of scope for the Collections feature but worth looking into in the future.

src/components/organisms/CollectionHero/CollectionHero.tsx
secondl1ght: This was an intentional decision and I think it is a fine UX, the component flipping to not found is expected because the collection was deleted. I tried routing first before flipping the state, but that unmounts the components and causes a potential race condition where the delete flow has not fully executed yet. So it's better to let it finish to be sure and only redirect after.

src/hooks/useDeletePost/useDeletePost.tsx
secondl1ght: Pushing fix now. 👍🏼

src/components/organisms/PostMenuActions/PostMenuActions.tsx
secondl1ght: This is also one I left one purpose, IMO it's fine because when you view the collection on the timeline or single post page, in that context it is a post. So post views get post copy and collection views get collection copy.

src/components/molecules/CollectionDeleted/CollectionDeleted.tsx
secondl1ght: This has been a problem since I first started working on the codebase, there was lots of unused `className` props and I brought it up to Miguel. But since we never addressed it, now it is a loop because the AI uses existing components as examples for new ones. I think we need to add an explicit AI rule in our docs to avoid this pattern, and also do a cleanup pass to remove any unused ones. But out of scope here.

src/hooks/useEditCollection/useEditCollection.ts
secondl1ght: Pushing fix. 👍🏼

src/components/organisms/Collections/DiscoverCollections/DiscoverCollections.tsx
secondl1ght: Which hook?

src/hooks/useDeletePost/useDeletePost.tsx
secondl1ght: This was also a pre-existing issue for the soft-delete path, so now both will be handled the same here.

infin1t3 (approved):

PR #1973, opened 2026-06-10
fix(#1218): perform account deletion when confirming in settings

### LINKED ISSUE #1218: [Bug]: Cannot delete account
### Description

I want to delete my account ( `pubky98t9k4uisdzzz8y9bujpa9zz5uwjeq8fsotc1izdme85b7cjwzey` ), because I moved to a new one ( `pubkygabri5dk51q1c1kzi53hjpqw5yhmw1nqsryzsfx5kubd7jpm63jy` ).

When I click Delete Account in my settings, a popup shows, with the buttons Cancel and Delete Account. I press 'Delete Account' again, but nothing happens.

Browser Console shows no error, Network tab doesn't show any requests ...

All that happens is that the Delete Account popup disappears.

### Expected Behaviour

Account gets deleted

### Steps to Reproduce

Settings gear icon > Delete Account > Delete Account

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

Chromium 144.0.7559.109 (Official Build) built on Ubuntu 24.04.3 LTS (64-bit)

Orlandohub:
## Summary
- fix #1218
- Confirming "Delete Account" in settings did nothing — the dialog just closed and no deletion ever ran (no network requests, no errors)
- The confirm button is now wired to the existing `ProfileController.commitDelete` flow: all account data is deleted with progress feedback, then the user is signed out and redirected to `/logout`
- **Root cause found while verifying the fix**: even once wired, `commitDelete` only deleted `profile.json`. `HomeserverService.list` was called with `limit: Infinity`, which coerces to `0` at the SDK's WASM boundary and silently returns an empty page — so every other file (posts, follows, tags, …) stayed on the homeserver and reappeared after signing back in with the same pubky. Listing is now cursor-paginated so deletion removes everything.

## Changes
- Add `useDeleteAccount` hook (modeled on `useSignOut`): calls `ProfileController.commitDelete({ pubky, setProgress })`, then `AuthController.logout()` + redirect; on deletion failure shows a destructive toast and resets state for retry
- Split error handling in `useDeleteAccount` (review feedback): a logout failure *after* successful deletion no longer surfaces a misleading "Failed to delete account" toast inviting a retry on an already-deleted account — it logs a warning and redirects anyway
- Add `HomeserverService.listAll`: paginates `list` with the SDK cursor (pages of 500, "start after last URL") until exhausted; verified against staging (712 files → 300/300/112, no overlap)
- `ProfileApplication.commitDelete` and `downloadData` now use `listAll` instead of `limit: Infinity`; this also fixes the data-export ZIP, which would have been generated empty (export is not yet reachable from the UI — noted in a code comment)
- Retry transient per-file delete failures ([PUBKY-APP-6Y](https://synonym-52.sentry.io/issues/PUBKY-APP-6Y)): deleting an active account issues hundreds of sequential DELETEs, and a single transport blip aborted the whole flow mid-way with a "failed" toast. Each file now gets up to 3 attempts with backoff, and a 404 counts as already deleted (covers a lost response after a server-side applied delete)
- Wire `DialogDeleteAccount` confirm button; while deleting it shows `Deleting... {progress}%`, disables both buttons, and blocks dialog dismissal (Cancel, X, overlay, Escape)
- Localize the dialog (existing `dialogs.deleteAccount.*` keys) and add `button`, `buttonLoading`, and `errors.deleteAccountFailed` keys to all 9 locales
- Rewrite dialog unit tests (previous tests asserted the broken close-only behavior), add `useDeleteAccount` and `listAll` tests, update delete/download suites; regenerate snapshots
- Re-enable the previously commented delete-account block in `cypress/e2e/settings.cy.ts`

https://github.com/user-attachments/assets/521b93dc-8709-4378-a8b1-dc3a6126a497

## Test plan
- [x] Unit suites: `homeserver.test.ts`, `profile.delete.test.ts`, `profile.download.test.ts`, `useDeleteAccount.test.tsx`, `DialogDeleteAccount.test.tsx`
- [x] SDK cursor pagination semantics verified empirically against staging
- [x] Manually verified on the PR preview deploy with a ~200-file account (first run surfaced the transport-blip issue, now retried automatically)
- [ ] Cypress `settings.cy.ts` in CI — not run locally since it requires staging infra (spec onboards a throwaway user and deletes it)

## Notes
- Deletion cannot be interrupted once started, so the dialog intentionally blocks dismissal while in progress
- Do not manually verify with the shared staging account — deletion permanently wipes it; use a throwaway account
- `ProfileApplication.commitDelete` local-first ordering and per-file homeserver deletion are unchanged; only file enumeration (paginated `listAll`) and per-file resilience (retries + 404 tolerance) changed
- After deletion, Nexus may still report non-zero post/reply counts for the pubky (tombstoned posts with interactions from others) and followers remain since those are other users' files — server-side indexing behavior, out of scope here


Orlandohub:
@greptile-apps review

infin1t3 (changes requested):
https://github.com/user-attachments/assets/9df41fb7-fdca-4f82-80b7-6569536de356

Orlandohub:
@greptile-apps review

infin1t3 (approved):
I am OK with the implementation, team confirmed on Slack that they are ok with the functionality.

PR #1988, opened 2026-06-12
feat: delete collections

secondl1ght:
# Wire up collection delete + fix latent post-delete reanimation race 🗑️

  ## Summary

  This PR wires up the Delete button on `CollectionCard` (in `/collections`) and `CollectionHero` (on the
  single-collection page), reusing the existing post delete pipeline. While building it we surfaced and fixed a
  preexisting race in the post delete code path that affected any hard-deleted post — `LocalPostService.delete` would
  fully remove `PostDetails` locally while the homeserver DELETE was still in flight, and any `usePostDetails` consumer on
   the page would race in, fire `fetchFn`, get the original content back from Nexus's eventually-consistent by-ids
  endpoint, and reanimate the post via `PostDetailsModel.bulkSave`.

  ## Collection delete UX

  ### `CollectionHero` (single-collection page)
  - 🆕 Delete button now opens a `DialogConfirmDelete` confirmation.
  - On confirm: **awaits** `deletePost(compositeId)` then `router.replace(APP_ROUTES.COLLECTIONS)`. Awaiting before the
  redirect avoids a race where the landing page's `MyCollections` would re-fetch a still-present id (the original
  optimistic-redirect approach hit this bug live during QA).
  - While `isDeleting`, **all owner action buttons** (Share, Edit, Delete) are disabled — signals progress and prevents
  racing Share / Edit against an imminent route replace.
  - Owners only — non-owners get no dialog and no delete button.

  ### `CollectionCard` (everywhere it renders)
  - 🆕 Owner's Delete button opens `DialogConfirmDelete` and calls `deletePost(compositeId)` on confirm. Optimistic
  removal works via the new `TimelineFeedContext.Provider` wrap on `MyCollections` (see below).
  - 🆕 `isPostDeleted` safety-net: if `usePostDetails` resolves with `content === '[DELETED]'`, the card renders the new
  `CollectionDeleted` molecule instead of an empty husk. Short-circuits before `parseCollectionContent` is ever called
  against the `[DELETED]` sentinel.

  ### Toast copy
  - Added optional `toastMessages` override to `useDeletePost` (mirrors `useBookmark`'s pattern). Per-field fallback to
  the generic `toast.post.*` copy.
  - Collection components pass collection-specific copy so the user sees "Collection deleted" / "Failed to delete
  collection" instead of "Post deleted".

  ### Dialog copy
  - `DialogConfirmDelete` got an optional `i18nNamespace` prop (`'dialogs.deletePost' | 'dialogs.deleteCollection'`).
  Collection components pass `'dialogs.deleteCollection'` so the dialog reads as "Delete collection?" rather than "Delete
  post?".

  ## New `CollectionDeleted` molecule

  Fully self-contained drop-in replacement for `CollectionCard` when content is the `[DELETED]` tombstone. Owns its own
  `Container` + `Card` + `CardContent` shell, mirroring `CollectionCard`'s outer footprint (`block h-full w-full
  lg:max-w-187`) so the deleted slot has the same dimensions in any grid it lands in. Renders a muted-foreground centered
  "This collection has been deleted by its author." à la `PostDeleted`. Accepts an optional `className` for caller
  composition.

  ## Section filtering (`CollectionsSections`)

  Mirrors the timeline's existing `isPostDeleted` filter pattern (`useVisualFeedTiles.ts:237`) so deleted collections
  disappear from the three landing sections:

  - **`MyCollections`**: wraps the grid in `<TimelineFeedContext.Provider>` exposing the hook's `removePosts` /
  `prependPosts`, so `useDeletePost`'s optimistic removal fires for cards deleted inline. Also adds a `useLiveQuery`
  filter that drops ids whose local `PostDetails.content === '[DELETED]'` — catches the soft-delete-leaves-stream-entry
  case after a redirect-from-hero delete, and any mid-session delete from another tab/device.
  - **`FollowedCollections`**: widens the existing live query to also require `!isPostDeleted(detail.content)`.
  - **`DiscoverCollections`**: adds the `isPostDeleted` guard at fetch-time (drops slice ids whose local details are
  tombstoned) and via a new live overlay that subtracts mid-session deletions from `displayIds`.

  ## 🔥 The reanimation fix (root cause)

  This is the part that turned out to be bigger than scoped. During QA we saw:
  - Delete a collection → redirect to `/collections` → card still showed up with original content.
  - Navigate back to the single page → page renders normally even though Nexus 404s `/tags`.
  - Hard refresh → same.

  Diagnostic logging confirmed the actual sequence:

  1. `useDeletePost` → `commitDelete` runs.
  2. `LocalPostService.delete` (hard-delete branch, because the collection had no connections) removes `PostDetails`
  locally.
  3. **Four `usePostDetails` consumers** on the page (`CollectionHero` + `CollectionItems` + tag panel etc.) see `data ===
   null`, fire `fetchFn` → `PostController.fetch` → `fetchMissingPostsFromNexus`.
  4. Nexus's by-ids endpoint **still returns the post** (eventual consistency between its delete index and the by-ids read
   path — `/tags` 404s separately).
  5. `LocalStreamPostsService.persistPosts` → `PostDetailsModel.bulkSave` writes the original content back **four times**.
  6. Homeserver DELETE finally returns; the await resolves. By now local DB has been reanimated.

  Posts seemed to work because most test posts have connections, taking the **soft-delete** branch (`content =
  '[DELETED]'` tombstone). `useLocalFirstQuery` sees non-null local data and never fires `fetchFn`. The bug existed for
  any post with no tags/replies/reposts — we just hadn't QA'd that path.

  ### Fix

  Two narrow surgical changes:

  - **`LocalPostService.delete` hard-delete branch**: `PostDetailsModel.deleteById` → `PostDetailsModel.update(id, {
  content: DELETED })`. The row stays as a tombstone; relationships / counts / tags / stream entries / parent count
  decrements still get cleaned up exactly as before. `useLocalFirstQuery` now sees a non-null row, so `fetchFn` doesn't
  fire and Nexus can't reanimate.
  - **`LocalStreamPostsService.persistPosts` tombstone guard**: before `bulkSave`, looks up existing `PostDetails` for the
   incoming ids; any id whose existing `content === '[DELETED]'` is dropped from every per-table batch (details / counts /
   relationships / tags / ttl / bookmarks / moderations). Defense-in-depth in case a stream slice fetch races a delete.

  ### Idempotency guard

  Added an explicit short-circuit in `LocalPostService.delete`: if the post is already tombstoned, return early. Prevents
  a stray re-delete (e.g. stale page state during a race) from re-running the hard-delete transaction and
  double-decrementing the author's `UserCountsModel.posts` count.

  ## i18n

  Added across all 9 locale files (`ar`, `de`, `en`, `es`, `fr`, `it`, `ja`, `pt-BR`, `zh`):

  - `collections.deleted` — used by the new `CollectionDeleted` molecule.
  - `dialogs.deleteCollection.{title,description,confirmButton,cancelButton}` — used by `DialogConfirmDelete` when called
  with the collection namespace.
  - `toast.collection.{collectionDeleted,collectionDeletedDesc,deleteFailed}` — used by `useDeletePost` when called with
  collection-specific toast overrides.

  ## New files

  - `src/components/molecules/CollectionDeleted/CollectionDeleted.tsx` (+ `.test.tsx` + snapshot) — collection-flavored
  sibling of `PostDeleted`.

  ## Test coverage 🧪

  - ✅ 18 new tests covering every behavior change in the PR.
  - ✅ Updated 4 existing `LocalPostService` tests to reflect the new tombstone behavior (hard-delete now leaves a
  `[DELETED]` row instead of fully removing).
  - ✅ Updated 1 `DiscoverCollections` mock to branch on `useLiveQuery` deps now that the component fires two distinct
  live queries.

  Highlights:
  - `LocalPostService.delete`: idempotency guard test (no user-count drift on re-delete).
  - `LocalStreamPostsService.persistPosts`: 2 tombstone-guard tests (skips overwriting `[DELETED]` rows; still persists
  non-tombstoned posts in the same batch).
  - `useDeletePost`: 3 `toastMessages` override tests (success / failure / per-field fallback).
  - `DialogConfirmDelete`: `i18nNamespace` swap test.
  - `CollectionDeleted`: full card-shell shape, footprint classes matching `CollectionCard`, centered alignment, className
   passthrough.
  - `CollectionCard`: delete-flow opens dialog with collection namespace + invokes `deletePost(compositeId)`;
  `CollectionDeleted` molecule renders for tombstoned content.
  - `CollectionHero`: delete-flow opens dialog with collection namespace + awaits `deletePost` then calls
  `router.replace(APP_ROUTES.COLLECTIONS)`; non-owner gets no dialog.
  - `MyCollections`: live-query filter drops tombstoned ids.
  - `FollowedCollections`: live-query excludes soft-deleted collections.
  - `DiscoverCollections`: fetch-time filter drops tombstoned slice ids; live-overlay subtracts ids that flipped to
  `[DELETED]` mid-session.


secondl1ght:
@greptile

infin1t3 (approved):
I have a couple of comments, will post in edit PR.

PR #1992, opened 2026-06-12
style: add collections are public message

secondl1ght:
# Add public-collections disclaimer to the My Collections header

  ## Summary

  Adds the missing design element on the `/collections` route: a muted right-aligned note next to the "My Collections"
  header reading **"Please note: all collections are public and discoverable."**

  The note is opt-in via a prop so it only appears on the dedicated `/collections` landing — not on the other surfaces
  (`Collection`, `BookmarksCollection`) that reuse the same `MyCollections` organism through `CollectionsSections`.

  ## Changes

  ### UI

  - 🆕 `MyCollections` now accepts an optional `showPublicNote?: boolean` prop (default `false`). When `true`, a
  right-aligned `Typography` (`as="span"`, `size="md"`, `text-input`) renders the note.
  - 🧱 Wrapped the existing header (title + avatar + "New Collection" button) in an inner `Container` and switched the
  outer header to `justify-between` so the note pushes to the right, while still wrapping cleanly on narrow viewports via
  `flex-wrap`.
  - ⚛️  Used the `Typography` atom from `@/atoms/Typography/Typography` rather than a raw `<span>` so the styling stays
  consistent with the rest of the codebase.

  ### Prop plumbing

  - `CollectionsSections` accepts a new `showMyCollectionsPublicNote?: boolean` prop (default `false`) and forwards it to
  `MyCollections` as `showPublicNote`.
  - The `Collections` template (used by `src/app/collections/page.tsx`) passes `showMyCollectionsPublicNote`. The
  `Collection` and `BookmarksCollection` templates leave it off so the note doesn't appear there.

  ### i18n

  - Added `collections.my.publicNote` translation key to all 9 locale files: `en`, `ar`, `de`, `es`, `fr`, `it`, `ja`,
  `pt-BR`, `zh`.

  ### Tests

  - ✅ `MyCollections.test.tsx` — new `public note` describe block:
    - Note is hidden by default.
    - Note renders when `showPublicNote` is passed.
  - ✅ `CollectionsSections.test.tsx` — `MyCollections` mock updated to capture the forwarded prop; new tests assert
  `showMyCollectionsPublicNote` defaults to `false` and forwards `true` when set.
  - ✅ `Collections.test.tsx` — `CollectionsSections` mock updated similarly; new test confirms the `/collections`
  template enables `showMyCollectionsPublicNote`.
  - 🔄 4 snapshots refreshed to account for the inner-container wrapper around the header's left-side group.
  - 👍 Sibling template tests (`Collection`, `BookmarksCollection`) untouched — they don't pass the prop, so
  `showPublicNote` stays `false` and their snapshots/assertions are unaffected.


secondl1ght:
image

infin1t3:
Tested on web and mobile, lgtm.

PR #2017, opened 2026-06-16
fix(notifications): defer mark-all-read until session is restored

### LINKED ISSUE #1895: [Bug]: markAllAsRead fails on profile notifications with homeserver validation error
### Description

Visiting the profile notifications page triggers `markAllAsRead()`, which attempts a homeserver `PUT` for the `lastRead` timestamp. The request fails validation with:

```
AppError: Authenticated writes must target an owned /pub/* path for the current session.
```

**Sentry:** [PUBKY-APP-3R](https://synonym-52.sentry.io/issues/PUBKY-APP-3R)
**Environment:** staging (also seen on production URL once)
**Release:** 1.4.0
**Occurrences:** 9 (first seen 2026-05-24, last seen 2026-05-25)
**Route / transaction:** `/profile`

**Stack trace (application frames):**
- `NotificationsContainer.tsx:33` — `markAllAsRead()` on mount
- `useNotifications.tsx:192` — `NotificationController.markAllAsRead()`
- `notification.ts:61` — `NotificationApplication.markAllAsRead(lastRead)`
- `homeserver.ts:305` — `Err.validation` when PUT URL is not resolved as an owned `/pub/*` session path

**Error metadata:**
- Category: `validation`
- Code: `INVALID_INPUT`
- Service: `homeserver`
- Operation: `request`
- HTTP method: `PUT`
- Status code: `400`

**Session replays available:**
- https://synonym-52.sentry.io/explore/replays/39c138d44941462187be59f0028b14f8/
- https://synonym-52.sentry.io/explore/replays/0a65777fac864eabb52c71234165946c/
- https://synonym-52.sentry.io/explore/replays/db830186659e49d2b8e0931f0ee3b4a8/

### Expected Behaviour

Opening profile notifications should mark notifications as read without throwing a validation error. The `lastRead` homeserver write should target a valid owned `/pub/*` path for the authenticated session, or the app should skip/defer the write when the path cannot be resolved.

### Steps to Reproduce

1. Sign in on staging (https://staging.pubky.app)
2. Navigate to the profile notifications page (`/profile` notifications tab — uses `NotificationsContainer`)
3. Observe the error when the page mounts and `markAllAsRead()` runs

**Affected URLs (from Sentry tag distribution):**
- https://staging.pubky.app/profile (8 events)
- https://pubky.app/profile (1 event)

**Note:** Reproduction may depend on session/path state — the failure occurs when `HomeserverService.resolveOwnedSessionPath()` does not recognize the `lastRead` PUT URL as an owned pub path.

### Screenshots

_No screenshots attached. See Sentry session replays linked above._

### Console Log

```shell
AppError: Authenticated writes must target an owned /pub/* path for the current session.
  at new AppError (./src/libs/error/error.ts:74)
  at createAppError (./src/libs/error/error.factories.ts:46)
  at HomeserverService.request (./src/core/services/homeserver/homeserver.ts:305)
  at NotificationApplication.markAllAsRead (./src/core/application/notification/notification.ts:59)
  at NotificationController.markAllAsRead (./src/core/controllers/notification/notification.ts:61)
  at markAllAsRead (./src/hooks/useNotifications/useNotifications.tsx:192)
  at NotificationsContainer useEffect (./src/components/organisms/NotificationsContainer/NotificationsContainer.tsx:33)
```

### Browser & Version

Primarily **Mobile Safari 16.6.1** on **iPhone / iOS 16.7.14** (8/9 events). Also seen once on **Safari 26.3**.

Orlandohub:
## Summary
- fix #1895
- `markAllAsRead` fired an authenticated homeserver write before the session finished restoring (on reload / mobile tab-resume), throwing `Authenticated writes must target an owned /pub/* path for the current session.` and leaving the unread badge desynced from the server.
- It was guarded only on `currentUserPubky` (restored synchronously from storage) while the write actually needs the `session` (restored asynchronously).

## Changes
- `NotificationController.markAllAsRead`: require a restored `session` (not just `currentUserPubky`), and commit `lastRead` / `unread = 0` **only after** the homeserver write succeeds.
- `NotificationApplication.markAllAsRead`: now `async` — awaits the PUT and propagates failures instead of swallowing them with `.catch`.
- `NotificationsContainer`: re-run the mark-as-read effect once the session becomes ready (depends on auth state), so the badge still clears after restore.
- `useNotifications`: call the now-async controller via `void`.
- Tests: add "skip when session not yet restored" and "don't update local store when the write fails" cases; update the application test to assert propagated failures.

## Test plan
- [x] Sign in, open Profile → Notifications with unread items, then reload directly on the page (and reopen the app on mobile Safari) — no validation error; badge clears to 0 and stays 0 after refresh.
- [ ] Simulate a failed homeserver write — the unread badge is not cleared locally.

## Notes
- Behavior change: the unread badge now clears after the homeserver write confirms (a quick round-trip) instead of optimistically, keeping local state honest when the write fails.


src/core/controllers/notification/notification.ts
```diff
@@ -62,11 +67,13 @@ export class NotificationController {
     // Create new lastRead with current timestamp using normalizer
     const lastRead = LastReadNormalizer.to(pubky);
```
infin1t3: This operation can also throw error, and if that happens it will become an unhandled rejection because we are calling this outside of the try block.

It's a small possibility for this, for example corrupted/invalid Pubky - but still a simple fix would be to just put this line in your try block

```
try {
  const lastRead = LastReadNormalizer.to(pubky);
  await NotificationApplication.markAllAsRead(lastRead);
  notificationStore.setLastRead(Number(lastRead.last_read.timestamp));
  notificationStore.setUnread(0);
} catch (error) {
  Logger.warn('Failed to mark notifications as read on homeserver', { error });
}
```

src/components/organisms/NotificationsContainer/NotificationsContainer.tsx
```diff
@@ -27,11 +28,14 @@ export function NotificationsContainer() {
...
   });
 
-  // Mark all notifications as read when entering the notifications page
-  // This allows the tab counter to show 0 while viewing notifications
+  // Re-run once the session is restored, since the write needs an authenticated session.
+  const isAuthenticated = useAuthStore((state) => state.session !== null);
+
+  // Mark all notifications as read when entering the page (once authenticated), so the
+  // tab counter shows 0 while viewing.
   useEffect(() => {
```
infin1t3, line 36: The effect depends on isAuthenticated but still calls markAllAsRead() while unauthenticated, relying on the controller to no-op. 

The comment says the action should run once authenticated; the condition is clearer at the call site. 

Fix: Add if (!isAuthenticated) return; before markAllAsRead() in the effect.

infin1t3 (changes requested):

src/core/controllers/notification/notification.ts
Orlandohub: Done in de5f3ac9. I moved `LastReadNormalizer.to(pubky)` inside the guarded `try` so synchronous normalizer failures are swallowed instead of becoming unhandled rejections. I also kept `AppError` logging deduped because `Err.*` factories already log automatically.

src/components/organisms/NotificationsContainer/NotificationsContainer.tsx
Orlandohub: Done in de5f3ac9. Added the `isAuthenticated` guard at the call site and covered the unauthenticated path in `NotificationsContainer` tests.

Orlandohub:
@greptile-apps review

infin1t3 (approved):
Fix is simple and well targeted - LGTM.

Issue #1966, opened 2026-06-09
[Task]: Add content to collection


infin1t3:
# Background

This is separate from the collection picker/save flow: sub-issue #2 covers adding a post to collections from the post action itself. This issue covers the reverse entry point: starting from a collection and adding content into that collection.

## Figma reference

**Web Add Content dialog**:
https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=34188-139658&m=dev

**Mobile Add Content dialog**:
https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=34503-276796&m=dev

Dialog content:
- Title: "Add Content"
- Description: "Add a post or article from your feed or paste a url directly."
- Option 1: "Add from feed"
  - Instruction: "Find a post you want to add and click the collections button."
  - Shows the post actions row with the collections/library button highlighted
- Option 2: "Paste post url"
  - Instruction: "You can also paste a post URL to quickly add it to your collection."
  - URL input with `https://` placeholder and paste icon
- Desktop: two cards side by side
- Mobile: cards stacked vertically

## What needs to happen

1. **Open Add Content dialog from collection detail.** The floating "+" button on a collection detail view opens this dialog for the active collection.

2. **Add from feed path.** The "Add from feed" card should guide the user back to the feed/post list flow. It can close the dialog and leave the user in the normal post action flow from sub-issue #2. If there is an obvious existing feed route, link there.

3. **Paste post URL path.** User can paste a Pubky post URL into the input and add that post directly to the current collection.

4. **URL validation and resolution.** Validate that the pasted URL points to a supported Pubky post/article. Resolve it to the internal post id used by collection items.

5. **Local-first add.** Adding by URL should use the same collection item creation path as the collection picker: Dexie first, homeserver sync in background.

6. **Duplicate handling.** If the post is already in the collection, do not create a duplicate. Show useful feedback instead.

7. **Feedback states.** Show loading while resolving/saving, success toast on add, and validation/error messaging for invalid URLs or failed resolution.

## Acceptance criteria

- [x] Floating "+" on collection detail opens the Add Content dialog
- [x] Dialog matches Figma on desktop and mobile
- [x] "Add from feed" explains the feed/post-action flow and routes/closes appropriately
- [x] "Paste post url" accepts a Pubky post URL
- [x] Invalid URLs show inline validation
[...]

PR #2034, opened 2026-06-18
feat(collections): add content paste flow

infin1t3:
Allow collection owners to paste a post URL via the Add Content dialog on collection pages.

Resolves: #1966 

Can't paste walkthrough video its too big. Will send in Slack.


src/core/services/local/stream/posts/posts.ts
```diff
@@ -382,12 +386,21 @@ export class LocalStreamPostsService {
     // Combine existing and new posts
     const combinedStream = [...postStream.stream, ...newPostsToAdd];
 
+    if (this.isBookmarkStream(streamId)) {
+      await PostStreamModel.upsert(streamId, combinedStream);
+      return;
+    }
```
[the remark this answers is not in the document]
infin1t3: @greptileai

Single-collection item streams are skip-paginated and intentionally bypass local stream persistence, so `persistNewStreamChunk()` is not called for that path. There is already a regression test for this in `src/core/application/stream/posts/post.test.ts`: `should skip cache check AND local persistence for collection item streams (skip-paginated)`.
Because of that, the bookmark guard is needed for bookmark streams, but an equivalent collection guard would currently be a dead path for single-collection items.

Re-run review to upgrade score after this please.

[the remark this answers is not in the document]
infin1t3: @greptileai Can you upgrade score to 5/5?

src/hooks/useStreamPagination/useStreamPagination.types.ts
```diff
@@ -62,6 +62,16 @@ export interface UseStreamPaginationResult {
...
+  /**
+   * Function to show membership-ordered posts at the top without timestamp sorting.
+   * Used by bookmarks and single-collection optimistic inserts where membership
+   * order differs from post creation time.
+   *
+   * This does not affect pagination cursors.
+   *
+   * @param postIds - A single post ID or array of post IDs to add
+   */
+  prependOptimisticPosts: (postIds: string | string[]) => void;
```
infin1t3, line 74: `prependPosts` is timestamp-based: it inserts the IDs and then re-sorts the visible list by `post.indexed_at`.

That works for Home/New Posts because timeline order is based on post creation/indexing time.

It does not work for membership feeds like Bookmarks or single Collection pages.

In Bookmarks, the meaningful order is “when the post was bookmarked”, not when the post was originally created.

In Collections, the meaningful order is the explicit order in the collection items array.

If we use `prependPosts` for an old post added to Bookmarks/Collections, it can be pushed lower after timestamp sorting.

`prependOptimisticPosts` adds the newly added membership item at the top immediately, without timestamp sorting.

It only affects the displayed list, not the underlying pagination cursor.

That matters for collection streams because they are offset-paginated; counting optimistic items as loaded items would break skip.

src/components/organisms/Collections/CollectionItems/CollectionItems.tsx
```diff
@@ -45,48 +42,41 @@ export function CollectionItems({ authorPubky, postId }: CollectionItemsProps) {
   // here stays as a defensive fall-through to the feed's own empty/error state.
   const collection = postDetails ? parseCollectionContent(postDetails.content) : null;
   const isConfirmedEmpty = postDetails != null && (collection?.items?.length ?? 0) === 0;
+  const emptyState = <CollectionItemsEmpty />;
 
-  if (isConfirmedEmpty) {
-    return <CollectionItemsEmpty isOwner={isOwn} />;
+  if (!isOwn && isConfirmedEmpty) {
```
infin1t3, line 46: If it's not ours and if we know it's empty - don't mount the timeline component at all.

src/components/organisms/AddContentDialog/AddContentDialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+    onSuccess,
+  });
+
+  const handleSubmit = async (event: SyntheticEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    await addContentForm.submit();
+  };
+
+  return (
+    <form onSubmit={handleSubmit} className="flex w-full flex-col gap-6">
```
infin1t3, line 170: We use plain html form component in other components too - not our own Form component.

infin1t3:
@greptileai review

infin1t3:
@greptileai review

PR #2041, opened 2026-06-19
fix(search): treat compact pubky prefix as ambiguous in autocomplete

### LINKED ISSUE #1909: [Bug]: Top search-bar autocomplete for `pubky*` searches is broken
### Description

I tried to see what tags autocomplete for `pubky-` prefix.

I typed it in the top search bar, but got nothing.

Typing it in the input field for tagging a specific post does show the correct tag autocomplete suggestions.

### Expected Behaviour

When I type `pubky-` in the top search bar, I see autosuggested tags (`pubky-feedback`, etc) and users (`Pubky`, etc).

### Steps to Reproduce

1. Go to top search bar
2. Slowly type `pubky-`
3. Notice that
    - for `pub` you get suggestions
    - for `pubk` you get suggestions
    - for `pubky` you get _no_suggestions

Also worth noting the network requests made by the browser:
- for `pub` there are 3 REST calls searching for `pub`
    - at Nexus search endpoints `tags/by_prefix`, `users/by_name`, `users/by_id`
- for `pubk` there are also 3 for `pubk`
- for `pubky` there are *none*
- for `pubky-feedback` there are 3 calls searching for `-feedback`

I suspect the `pubky` prefix may be removed from the search term too aggressively. Intention was probably to remove it from user PK searches (`users/by_id`), but if that's true, it should only be removed for this query and not for the other two.

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

Same behavior in Chrome, Safari

Orlandohub:
## Summary
- fix #1909
- Search autocomplete no longer treats partial `pubky` typing or tag-like text (e.g. `pubky-feedback`) as an explicit user ID search
- Tags and usernames continue to resolve for compact `pubky` input while user ID lookup only runs when the stripped suffix is a valid z32 prefix

## Changes
- Extract `resolveSearchAutocompletePlan` to decide tag, username, and user ID queries independently
- Split legacy delimited prefixes (`pubky:`, `pk:`) from the ambiguous compact `pubky` prefix
- Skip user ID search while the query is still a progressive prefix of `pubky` (`p`, `pu`, `pub`, `pubk`)
- Preserve raw query casing for tag/username endpoints; normalize only for compact prefix detection

## Test plan
- [x] `npm run test -- src/hooks/useSearchAutocomplete/` (48 tests passed)
- [ ] Type `pubky-feedback` in search autocomplete — expect tag/username results, no ID-only search
- [ ] Type `pubkyabc` — expect tag, username, and user ID searches (ID prefix `abc`)
- [ ] Type `pk:abc` or `pubky:abc123` — expect ID-only search with prefix stripped

## Notes
- Legacy delimited prefixes remain unambiguous ID-only searches; behavior for `pk:` / `pubky:` is unchanged

### DIFF
--- a/src/hooks/useSearchAutocomplete/useSearchAutocomplete.constants.ts
+++ b/src/hooks/useSearchAutocomplete/useSearchAutocomplete.constants.ts
@@ -17,11 +17,8 @@ export const AUTOCOMPLETE_USER_LIMIT = 10;
 /** Minimum character length for user ID searches after "pubky" prefix */
 export const MIN_USER_ID_SEARCH_LENGTH = 3;
 
-/**
- * Prefixes for user ID searches
- * Order matters - longer prefixes should come first to match correctly
- * - pubky: legacy format with colon
- * - pk: legacy format
- * - pubky: new format without colon (must be last to not match 'pubky:' first)
- */
-export const USER_ID_PREFIXES = ['pubky:', 'pk:', 'pubky'] as const;
+/** New compact user ID prefix. It is also valid text in tags and usernames. */
+export const COMPACT_USER_ID_PREFIX = 'pubky';
+
+/** Legacy, unambiguous user ID prefixes that remain ID-only searches. */
+export const DELIMITED_USER_ID_PREFIXES = ['pubky:', 'pk:'] as const;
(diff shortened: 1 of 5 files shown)


SHAcollision (approved):
ACK. Code review and tested functionality,

PR #2048, opened 2026-06-22
fix(notifications): improve tagged notification readability

### LINKED ISSUE #1793: [Bug]: Mobile notifications unreadable with long tags
### Description

Mobile notifications on https://pubky.app/profile that include a long tag are functionally useless.

If I click on the tag, it searches for the tag. If I click on the username/picture, it shows the profile. There is not way to navigate to the post that was tagged.

At the same time, informationally I do not see what happens. I just see a truncated username and a tag. Did he remove the tag? Did he add it? The whole notification screen would benefit from improvements. 

Could well be that this problem is not exclusive to long tags but one can create different notifications that cause this.

Image

FYI @aldertnl 

### Expected Behaviour

I need a way to navigate to the tagged post. Probably needs a redesign, wider layout, maybe multiple instead of squeezing everything in one line.

### Steps to Reproduce

Get tagged with a long tag.

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

iPhone 13

Orlandohub:
## Summary
- fix #1793
- Tagged notifications rendered their interactive tag badge at every viewport size. On mobile, long labels consumed the available row width and truncated the actor/action copy, making the notification difficult to understand and the post target difficult to discover.
- Mobile tagged notifications now match the design: the tag badge is hidden, the actor/action copy can wrap naturally, and the existing post/profile navigation remains intact.

## Changes
- Allow notification actor/action copy to wrap with normal line height below the `lg` breakpoint while retaining single-line truncation on desktop.
- Hide `PostTag` below `lg`; preserve the desktop tag-search interaction for post and profile tag notifications.
- Add focused unit coverage for responsive classes, tagged-post routing, actor-profile routing, and tagged-profile routing; refresh snapshots.
- Update the Cypress notification scenario to use the near-limit `first-world-problem` tag and assert mobile/desktop visibility plus their respective navigation targets.

## Test plan
- [ ] Desktop/mobile Cypress notification flow — not run locally per request

## Notes
- Notification payloads and route resolution are unchanged: the actor avatar/name still opens the actor profile; the action, timestamp/icon, and non-interactive row area still open the tagged post; tagged-profile actions still open `/profile/tagged`.
- The existing `lg` breakpoint remains the mobile/desktop boundary.

### DIFF
--- a/cypress/e2e/notifications.cy.ts
+++ b/cypress/e2e/notifications.cy.ts
@@ -118,6 +118,21 @@ describe('notifications', () => {
   });
 
   it('can be notified for tagged post and profile', () => {
+    // helper function to verify the tag is visible on desktop and not on mobile
+    const expectLatestNotificationTag = (tagLabel: string) => {
+      const isMobile = Cypress.expose('isMobile');
+      cy.get('[data-cy="notifications-list"]')
+        .children()
+        .first()
+        .find('[data-cy="post-tag"]')
+        .should(isMobile ? 'not.be.visible' : 'be.visible')
+        .then(($tag) => {
+          if (!isMobile) {
+            cy.wrap($tag).should('contain', tagLabel);
+          }
+        });
+    };
+
     // * profile 1 creates a post
     const postContent = `I will be notified when this post is tagged! ${Date.now()}`;
     createQuickPost(postContent);
@@ -140,7 +155,9 @@ describe('notifications', () => {
     goToProfilePageFromHeader();
     verifyNotificationCounter(0);
     // check latest notification on profile page
-    checkLatestNotification([profile1.username, 'tagged your profile', profileTag], LatestNotificationReadState.Unread);
+    checkLatestNotification([profile1.username, 'tagged your profile'], LatestNotificationReadState.Unread);
+
+    expectLatestNotificationTag(profileTag);
 
     // * profile 2 tags profile 1's post (from their profile page)
     cy.get(`@${profile1.pubkyAlias}`).then((pubky) => {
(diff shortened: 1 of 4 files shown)


catch-21:
@Orlandohub you'll want to truncate long usernames

Screenshot 2026-06-24 at 19 46 20

cypress/e2e/notifications.cy.ts
catch-21: I guess test would fail here on mobile because the `postTag` string will not be found when calling `checkLatestNotification`. We should probably do the check for correct tag label on desktop and absence of it in here.

cypress/e2e/notifications.cy.ts
```diff
@@ -168,6 +168,27 @@ describe('notifications', () => {
...
     checkLatestNotification([profile2.username, 'tagged your post', postTag], LatestNotificationReadState.Read);
 
+    cy.get('[data-cy="notifications-list"]')
+      .children()
+      .first()
+      .within(() => {
+        cy.contains('a', 'tagged your post')
+          .should('be.visible')
+          .parent()
+          .should('have.css', 'white-space', Cypress.expose('isMobile') ? 'normal' : 'nowrap');
```
catch-21: no need to check for css in e2e test, do this check in the unit/snapshot tests

cypress/e2e/notifications.cy.ts
catch-21: removed check for tag value here in 154c0bcfc776ea416bd6fbeda3fcc8493e0059b2

cypress/e2e/notifications.cy.ts
catch-21: removed this block in 154c0bcfc776ea416bd6fbeda3fcc8493e0059b2

catch-21:
LGTM now 👍 

https://github.com/user-attachments/assets/e8f1e735-d528-4163-95cc-b63319081271

Made changes to the cypress test in 154c0bcfc776ea416bd6fbeda3fcc8493e0059b2.

catch-21 (approved):

PR #2053, opened 2026-06-23
feat: collection moderation

secondl1ght:
## 🙈 Add moderation (blur) to Collections

  ### Context

  Collections are just posts with `kind === 'collection'`, and the moderation/blur plumbing already exists end-to-end —
  `usePostDetails` returns an `EnrichedPostDetails` with an `is_blurred` flag, and
  `ModerationController.unBlur(compositeId)` clears it. `PostContentBase` already intercepts blur and renders
  `PostContentBlurred`.

  The gap: `CollectionCard` (landing sections) and `CollectionHero` (single-collection page) are rendered **directly**,
  bypassing that interception — so a moderated collection rendered in full. This PR adds the render path so those surfaces
  blur correctly, with a click-to-unblur placeholder that matches the real card/hero footprint.

  No moderation plumbing changes were needed — this is purely a render path plus presentational components.

  ### ✨ What changed

  **New blurred placeholders** (one per surface, since the card and hero have different dimensions):
  - `CollectionCardBlurred` — mirrors the card shell (`p-6`, header/description/tag-action rows, `lg:max-w-187`).
  - `CollectionHeroBlurred` — mirrors the hero shell (`p-8 lg:p-12`, large title → owner → description → tags → actions
  stack).
  - Each is a single click-to-unblur `Button` (no navigation) wrapping a `Card` whose mock content is rendered under
  `blur-2xl select-none`, with an `EyeOff` + "content moderated" overlay — matching how `PostContentBlurred` handles
  moderated posts.

  **Render-path wiring:**
  - `CollectionCard`: `skeleton → deleted → blurred → content` (deleted wins, matching `PostContentBase`'s ordering).
  - `CollectionHero`: `skeleton → blurred → content`.
  - Both pass the **composite id** to `unBlur`, consistent with `PostContentBlurred`.

  **🎨 White placeholders:** The mock placeholders (title, description, icon, bars) are intentionally **white**
  (`text-white` / `bg-white/30`) rather than reusing the real colors — colored elements bled through the blur as invisible
  color blobs, whereas white blurs into a clean light redaction.

  **📐 Bottom-aligned action row:** The card's tag/action row uses `mt-auto` so it pins to the bottom when a card
  stretches to match a taller sibling in the same row (matching the alignment behavior the real cards will adopt in an
  upcoming refactor).

  **♻️  Shared overlay extraction:** The "content moderated" overlay was identical across the two new collection components
  and near-identical in `PostContentBlurred`, so it's extracted into a presentational molecule, `ModerationBlurOverlay`,
  that takes a `label` prop. All three components now consume it. The `PostContentBlurred` DOM output is byte-for-byte
  unchanged by this extraction.

  **🌍 Internationalization:** The previously-hardcoded English strings are now translated. Added a new `moderation`
  namespace with `postContentModerated` and `collectionContentModerated` to **all 9 locale files** (`en`, `de`, `es`,
  `fr`, `it`, `ja`, `pt-BR`, `zh`, `ar`), with wording aligned to each locale's existing `blurCensored` / collection
  vocabulary. Each component now resolves its label via `useTranslations('moderation')`.

  ### ✅ Tests

  - Added `moderation — blurred state` cases to the `CollectionCard` and `CollectionHero` suites: the blurred placeholder
  renders instead of the real content, clicking it calls `unBlur` with the composite id (and stops propagation), and a
  deleted-and-moderated collection still shows the deleted fallback.
  - Added a dedicated suite for the new `ModerationBlurOverlay` molecule (label, icon, overlay positioning, `className`
  merge, snapshot).
  - Updated `PostContentBlurred` tests for the i18n change (added `next-intl` mock; the one snapshot change is the label
  text node now resolving to the translation key — expected, structure otherwise unchanged).

  ### 🗂️  Notable files

  - **New:** `ModerationBlurOverlay.tsx`, `CollectionCardBlurred.tsx`, `CollectionHeroBlurred.tsx` (+ tests/snapshots)
  - **Modified:** `CollectionCard.tsx`, `CollectionHero.tsx`, `PostContentBlurred.tsx`, all 9 `messages/*.json`

  ### 👀 Reviewer notes

  - Blurred collection posts that flow through `PostContentBase` (feeds/timelines) were already handled by its existing
  blur intercept — this PR covers only the direct-render surfaces that bypass it.
  - The non-English moderation strings are aligned to existing in-app vocabulary but may warrant a later native review.

### DIFF
--- a/src/components/molecules/ModerationBlurOverlay/ModerationBlurOverlay.tsx
+++ b/src/components/molecules/ModerationBlurOverlay/ModerationBlurOverlay.tsx
@@ -1,0 +1,36 @@
+import { EyeOff } from 'lucide-react';
+import { Container } from '@/atoms/Container/Container';
+import { Typography } from '@/atoms/Typography/Typography';
+import { cn } from '@/libs/utils/utils';
+
+interface ModerationBlurOverlayProps {
+  /** Message shown under the icon, e.g. "Collection content moderated." */
+  label: string;
+  className?: string;
+}
+
+/**
+ * Centered "content moderated" overlay (eye-off icon + label) layered over a
+ * blurred placeholder. Absolutely fills its positioned parent. Shared by the
+ * blurred collection card / hero (and any other moderated-content placeholder)
+ * so the unblur affordance stays visually consistent.
+ *
+ * Expects a `group` ancestor for the hover color transition, and a `relative`
+ * (positioned) parent so the `inset-0` fill lands correctly.
+ */
+export function ModerationBlurOverlay({ label, className }: ModerationBlurOverlayProps) {
+  return (
+    <Container
+      overrideDefaults
+      className={cn(
+        'absolute inset-0 flex flex-col items-center justify-center gap-2 text-muted-foreground transition-colors group-hover:text-secondary-foreground',
+        className,
+      )}
+    >
+      <EyeOff className="size-6" />
+      <Typography overrideDefaults as="p" className="text-sm">
+        {label}
+      </Typography>
+    </Container>
+  );
+}
--- a/src/components/organisms/Collections/CollectionCard/CollectionCardBlurred.tsx
+++ b/src/components/organisms/Collections/CollectionCard/CollectionCardBlurred.tsx
@@ -1,0 +1,89 @@
+'use client';
+
+import { Library } from 'lucide-react';
+import { useTranslations } from 'next-intl';
+import { Button } from '@/atoms/Button/Button';
+import { Card, CardContent } from '@/atoms/Card/Card';
+import { Container } from '@/atoms/Container/Container';
+import { Typography } from '@/atoms/Typography/Typography';
+import { ModerationController } from '@/controllers/moderation/moderation';
+import { cn } from '@/libs/utils/utils';
+import { ModerationBlurOverlay } from '@/molecules/ModerationBlurOverlay/ModerationBlurOverlay';
+
+interface CollectionCardBlurredProps {
+  /** `author:postId` composite id — the moderation record key passed to `unBlur`. */
+  compositeId: string;
+  className?: string;
+}
+
+/**
+ * Blurred placeholder for a moderated `CollectionCard`.
+ *
+ * Mirrors `CollectionCard`'s outer shape (rounded card, `p-6` body, header +
+ * description + tag/action rows) so the moderated state occupies the same
+ * footprint as the real card — only blurred, with an unblur overlay. The whole
+ * card is a single click-to-unblur button (no navigation), matching how
+ * `PostContentBlurred` handles moderated posts.
+ */
+export function CollectionCardBlurred({ compositeId, className }: CollectionCardBlurredProps) {
+  const t = useTranslations('moderation');
+  return (
+    <Button
+      overrideDefaults
+      onClick={(e) => {
+        e.stopPropagation();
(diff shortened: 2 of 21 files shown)


src/components/molecules/ModerationBlurOverlay/ModerationBlurOverlay.tsx
```diff
@@ -0,0 +1,36 @@
+import { EyeOff } from 'lucide-react';
+import { Container } from '@/atoms/Container/Container';
+import { Typography } from '@/atoms/Typography/Typography';
+import { cn } from '@/libs/utils/utils';
+
+interface ModerationBlurOverlayProps {
+  /** Message shown under the icon, e.g. "Collection content moderated." */
+  label: string;
+  className?: string;
```
infin1t3, line 9: You never pass `className` prop to this component

src/components/organisms/Collections/CollectionHero/CollectionHeroBlurred.tsx
```diff
@@ -0,0 +1,86 @@
+'use client';
```
infin1t3, line 1: Missing test and snapshot for this one.
I realize also that CollectionHero.test.tsx has a moderation - blurred state block that checks the placeholder renders so I am fine with not doing this atm if we're in a rush - feel free to ignore.

src/components/organisms/Collections/CollectionCard/CollectionCardBlurred.tsx
```diff
@@ -0,0 +1,89 @@
...
+      overrideDefaults
+      onClick={(e) => {
+        e.stopPropagation();
+        ModerationController.unBlur(compositeId);
+      }}
+      className={cn('group relative block h-full w-full cursor-pointer text-left lg:max-w-187', className)}
+    >
+      <Card className="relative isolate h-full gap-0 overflow-hidden rounded-md py-0">
+        {/* Blurred mock of the real card layout */}
+        <CardContent aria-hidden="true" className="flex h-full flex-col gap-3 p-6 blur-lg select-none">
```
infin1t3, line 41: Collection blurred placeholders use blur-lg; PostContentBlurred uses blur-2xl.

infin1t3 (changes requested):
Only thing that should be addressed IMO is blur-lg > blur-xl or vice versa. Others are minor.

src/components/organisms/Collections/CollectionCard/CollectionCardBlurred.tsx
secondl1ght: Yes I did this on purpose, these ones have a different blurred background content and it looked better with less blur.

src/components/organisms/Collections/CollectionHero/CollectionHeroBlurred.tsx
secondl1ght: I left tests excluded here because I treated these the same as skeleton components which we don't test.

src/components/molecules/ModerationBlurOverlay/ModerationBlurOverlay.tsx
secondl1ght: Yeah we need to add an agent rule for this and cleanup all comps with unused class name props. Miguel's approach was to just add them everywhere just in case but I am pro not doing that.

infin1t3 (approved):
LGTM - approved

PR #2055, opened 2026-06-23
fix(collections): pin collection action bar to bottom and unify expandable tag actions

infin1t3:
* Pin action bar to bottom
* Add expandable tags to collection cards and single collection view

https://github.com/user-attachments/assets/bcb5ebd2-257b-49fb-af6c-e810c6735c5b


infin1t3:
@greptileai review

PR #2061, opened 2026-06-23
fix: edge navigation clicks on collection card

secondl1ght:
# Fix collection card tag interactions: dead zone & accidental navigation 🎯

  ## Summary

  The collection card is wrapped in a `Link`, with the tags row and action button suppressing navigation via handlers on
  the **parent containers**. This caused two QA issues:

  1. **Dead zone** — clicks in the spacing *around* the tags/button (not directly on them) neither navigated nor triggered
  an action. On narrow/mobile cards where tags wrap to multiple rows, this dead area was large.
  2. **Accidental navigation** — opening the emoji picker from the tag input and selecting an emoji (or clicking the
  dialog overlay) navigated away from the card unexpectedly.

  This PR moves navigation suppression onto the interactive elements themselves and fixes the portal-related navigation
  leak.

  ## Changes

  ### 🧩 Dead zone — suppress at the leaf, not the wrapper

  Moved `preventDefault` + `stopPropagation` directly onto the interactive controls so clicks in the gaps between them
  fall through to the wrapping `Link` and navigate as expected:

  - `PostTag` — added `preventDefault` alongside the existing `stopPropagation` on both the chip click and the close
  button.
  - `PostTagAddButton` — its click now suppresses navigation before invoking the callback.
  - `TagInput` — same on the container click (focus still happens on `mousedown`, so typing/focus behaviour is unaffected;
  `cursor-pointer` styling stays keyed on the caller's `onClick`).
  - `CollectionCard` — removed the wide `onClick`/`onAuxClick` handlers from the full-width bottom row; the action button
  keeps its own suppression plus `onAuxClick` on its hugging wrapper.

  The result is **zero dead zone**: only the chips/button/input themselves swallow the click; everything between them
  navigates.

  ### 🪟 Accidental navigation from portaled overlays

  React synthetic events bubble through the **React tree**, not the DOM tree — so even though the emoji picker and
  tag-suggestions dropdown are portaled to `<body>`, a click inside them still bubbled up to the card's `Link` and
  triggered navigation.

  - `EmojiPickerDialog` — `stopPropagation` on `DialogContent` so emoji clicks stay contained (this also covers the
  overlay, which dismisses via the in-content close button).
  - `TagInput` — `stopPropagation` on the suggestions `PopoverContent`, which had the identical latent bug (clicking a
  suggestion would have navigated too).

  `preventDefault` is intentionally **not** added on these overlays: because they're portaled outside the anchor's DOM
  subtree, there is no native `<a>` navigation to prevent — the only path is the `Link`'s React `onClick`, which
  `stopPropagation` fully blocks.

  ## Why keep the `Link` (vs `router.push`)?

  Switching the card to an `onClick` + `router.push` would **not** remove the portal bubbling issue (a `router.push`
  handler is still a React ancestor that portaled content bubbles to), and it would sacrifice real anchor semantics —
  open-in-new-tab via cmd/middle-click, right-click "copy link", `role=link`, and keyboard activation. The `Link` is
  retained and suppression is handled at the interactive elements instead.

  ## Safety / audit 🔍

  Audited every usage of `PostTag`, `PostTagAddButton`, and `TagInput` across the app (search, notifications, article
  detail, composer, feed dialog, etc.). No usage relies on a tag chip / add button / input navigating via a native anchor
  — all navigation is done through `router.push` in handlers or callbacks, neither of which `preventDefault` affects. The
  change is therefore a defensive improvement, not a regression, and is safe to apply universally.

  ## Tests ✅

  - Added `preventDefault`/`stopPropagation` regression tests for `PostTag`, `PostTagAddButton`, and `TagInput`.
  - Added a regression test for `EmojiPickerDialog` verifying a click inside the dialog does **not** reach a clickable
  ancestor (while emoji selection still fires).
  - All affected suites pass; lint and typecheck clean.

### DIFF
--- a/src/components/molecules/EmojiPickerDialog/EmojiPickerDialog.test.tsx
+++ b/src/components/molecules/EmojiPickerDialog/EmojiPickerDialog.test.tsx
@@ -16,8 +16,16 @@ vi.mock('@/atoms/Dialog/Dialog', () => {
         {open && children}
       </div>
     ),
-    DialogContent: ({ children, className }: { children: React.ReactNode; className?: string }) => (
-      <div data-testid="dialog-content" className={className}>
+    DialogContent: ({
+      children,
+      className,
+      onClick,
+    }: {
+      children: React.ReactNode;
+      className?: string;
+      onClick?: (e: React.MouseEvent) => void;
+    }) => (
+      <div data-testid="dialog-content" className={className} onClick={onClick}>
         {children}
       </div>
     ),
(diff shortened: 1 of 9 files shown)


infin1t3 (approved):
https://github.com/user-attachments/assets/8e2ab6fc-82de-4ca4-bd75-f6c6bbe9c07c

PR #2062, opened 2026-06-23
refactor: show deleted posts in collections and bookmarks

secondl1ght:
# Show deleted posts in bookmark and collection feeds 🗑️

  ## Overview

  The **bookmarks feed** (`/collections/bookmarks`) and the **single-collection items feed**
  (`/collections/[userId]/[postId]`) were filtering deleted posts out on the client, but the bookmark and collection item
  **counts include deleted posts**. That left a mismatch between the count shown and what actually rendered in the feed
  (e.g. a count of 60 against 5 visible cards). The count cannot currently be made accurate on the backend, so rather than
  fight it, this PR aligns the feeds to the count: deleted items are kept in these two feeds and rendered with the
  existing deleted-post placeholder, so the number and the feed line up. It also tidies up how that placeholder sits
  inside a stretched grid cell.

  ## What changed

  ### 🔹 Retain deleted posts in bookmark & collection-item streams

  All post streams previously ran every page through a single deletion filter in
  `PostStreamApplication.filterStreamPosts`, which dropped any post whose content is `[DELETED]`. We now skip that filter
  for the two target stream shapes while every other feed (home, hot, search, profile, replies, discover, etc.) continues
  to filter deleted posts out as before.

  A new `isDeletedRetainingStream(streamId)` predicate centralizes this decision:

  - ✅ Single-collection item feeds — `collection:<author>:<postId>`
  - ✅ Bookmark post feeds — `<sorting>:bookmarks:<kind>` (all / short / long / image / video / link / file)
  - ❌ The collection-kind bookmark stream — `<sorting>:bookmarks:collection` — is **excluded**

  The `:bookmarks:collection` exclusion is important: that stream is not a bookmarks post feed, it is the seed for the
  **Followed Collections** section, which deliberately hides deleted collections via its own Dexie live query. Excluding
  it keeps stream-level behavior consistent with that section rather than surfacing deleted collections there.

  ### 🔹 Scope & safety notes

  - The visual (image/video) tile layout still filters deleted posts — a deleted post has no attachments, so it can't form
  a media tile; the deleted-state card only makes sense in the standard list/grid layout.
  - `FollowedCollections`, `MyCollections`, and `DiscoverCollections` are unaffected — they use different stream shapes
  and/or re-filter deleted items in their own live queries.
  - Write-side bookmark cache updates don't flow through the changed filter, so they're untouched.

  ### 🔹 Center the "deleted" placeholder vertically in grid cells

  In the collection/bookmark **grid**, each card stretches to the tallest card in its row, which left the short "This post
  has been deleted by its author." message pinned to the top with empty space below. `PostDeleted` now fills its parent
  card and centers the message vertically.

  Because the placeholder is a shared component, this was verified to be a no-op everywhere else it's used (single-post
  page, timeline list, nested post previews) — in those contexts the card is content-sized, so the fill-and-center has no
  extra space to act on and nothing visibly changes. The effect is naturally scoped to the stretched grid case, so no new
  prop or component was needed.

  ## Testing ✅

  - Added unit tests covering `filterStreamPosts`: deleted posts are **removed** from timeline streams and the
  collection-kind bookmark stream, and **retained** in bookmark post feeds and single-collection item streams.
  - Updated the `PostDeleted` snapshot for the new centering classes.
  - Full suite, typecheck, and lint all pass.

### DIFF
--- a/src/components/molecules/PostDeleted/PostDeleted.test.tsx.snap
+++ b/src/components/molecules/PostDeleted/PostDeleted.test.tsx.snap
@@ -2,7 +2,7 @@
 
 exports[`PostDeleted - Snapshots > matches snapshot with default props 1`] = `
 <div
-  class="py-2"
+  class="flex flex-1 items-center justify-center py-2"
   data-testid="card-content"
 >
   <p
--- a/src/components/molecules/PostDeleted/PostDeleted.tsx
+++ b/src/components/molecules/PostDeleted/PostDeleted.tsx
@@ -8,7 +8,7 @@ export const PostDeleted = () => {
   const t = useTranslations('post');
 
   return (
-    <CardContent className="py-2">
+    <CardContent className="flex flex-1 items-center justify-center py-2">
       <Typography size="sm" className="text-center font-normal text-muted-foreground">
         {t('deleted')}
       </Typography>
--- a/src/core/application/stream/posts/post.test.ts
+++ b/src/core/application/stream/posts/post.test.ts
@@ -293,6 +293,74 @@ describe('PostStreamApplication', () => {
 
       expect(result).toEqual([missingPostId]);
     });
+
+    const createDeletedPostDetail = async (postId: string) => {
+      await PostDetailsModel.create({
+        id: postId,
+        content: DELETED,
+        kind: 'short',
+        indexed_at: BASE_TIMESTAMP,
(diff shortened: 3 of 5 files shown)


secondl1ght:
@greptile regarding your concern, it is a good point except that `NewPostsSection` is not enabled for the bookmarks or single collection stream. We don't want it for bookmarks, and for collections it is a bigger refactor that we will do later. So the issue you describe is not reachable right now and will be addressed when that path is hooked up

infin1t3 (approved):

PR #2077, opened 2026-06-25
feat: ingest external key during the sign-in

### LINKED ISSUE #2054: [Feature]: Ingest external-homeserver users on sign-in when `profile.json` is missing
## Problem

When a user _signs in_ without a `profile.json` (they registred in the hs with their pkarr records but haven't finished pubky.app onboarding), the app currently does nothing to register them with Nexus.

This is fine for users on the **Synonym homeserver** — Nexus already watches that server.

For users on an **external homeserver**, Nexus doesn't know about them yet. If we don't tell Nexus at sign-in, it may not pick up their data when they later create their profile during onboarding. That can lead to a broken or empty experience after onboarding (feed, profile, notifications).

## Proposal

During app sign-in, after authentication succeeds:

1. Check if the user has `profile.json` (we already do).
2. Check which homeserver they are on - `await pubkySdk.getHomeserverOf(publicKey);`
3. If they **don't** have `profile.json` **and** they are **not** on the Synonym homeserver → call:
   `PUT {NEXUS_URL}/v0/ingest/{user_id}`
4. Otherwise, skip the call.
5. Continue the normal sign-in flow (onboarding if unsigned, bootstrap if signed up).

## Work required

- Add ingest call in the sign-in flow (`AuthController.initializeAuthenticatedSession` or equivalent).
- Resolve the user's homeserver and compare it to `NEXT_PUBLIC_HOMESERVER`.
- Only call ingest when: no `profile.json` + external homeserver.
- Handle ingest failures gracefully (log and continue, or show a clear retry message — TBD).
- Cover all sign-in paths: Pubky Ring auth URL, mnemonic, recovery file.
- Add tests for the three cases: external HS + no profile, Synonym HS + no profile, has profile.

tipogi:
## ⚠️ Deployment coordination — merge with Nexus `main` or after

**This PR must be deployed together with the Nexus release that ingests external keys.**

The app will call `PUT {NEXUS_URL}/v0/ingest/{user_id}` during sign-in for external-homeserver users without a profile.


SHAcollision:
hey @tipogi! I was working on the same thing in parallel without noticing your PR, sorry 😅. I pushed a commit here that merges both efforts, hope you dont mind. Basically I moved the ingest call to two places:

- right after profile.json is saved (new users)
- when bootstrap returns `indexed: false` (users that already have a profile but nexus dont know them yet, and it also works as retry if the first call failed)

and I removed the client side homeserver check, I think we dont need it because the endpoint is idempotent and nexus already skips known users and handles the default homeserver fine, so we save the pkarr lookup during sign-in. Also the calls are fire and forget now, so nexus being down never blocks sign-in.

details are in the commit message. maybe we should also update the PR title since is not "during the sign-in" anymore, wdyt?

SHAcollision:
@greptile review

Orlandohub (dismissed):
LGTM

Small feedback from AI review:

**[1] medium | simplification | src/core/application/bootstrap/bootstrap.ts:62**
**What's wrong:** The same fire-and-forget pattern (`void NexusBootstrapService.ingest(pubky).catch(() => {})` + nearly identical comment) is duplicated in `BootstrapApplication` and `ProfileApplication`.
**Why it matters:** The swallow policy and rationale live in two places; both application tests re-prove the same non-blocking contract.
**Fix:** Move `.catch(() => {})` into `NexusBootstrapService.ingest` (or a dedicated fire-and-forget method), document once in the service JSDoc, and call a single non-throwing entry point from both applications. Keep reject-does-not-fail coverage on the service.

**[2] low | simplification | src/core/services/nexus/nexus.utils.ts:95**
**What's wrong:** `fetchNexusNoContent` re-copies the `safeFetch + !response.ok` block from `fetchNexus`.
**Why it matters:** A no-content path is required, but the request/ok-check logic does not need a second full copy.
**Fix:** Extract a shared helper (e.g. `fetchNexusResponse`) for `fetch + ok-check`; have `fetchNexus` parse and `fetchNexusNoContent` return void.

secondl1ght (dismissed):
LGTM

There are no tests for the new `fetchNexusNoContent` but there is no existing tests for the similar `fetchNexus`, and it is a lower value test anyway.

SHAcollision:
thanks for the review @Orlandohub! applied your first point in c2e76b2f, moved the catch into `NexusBootstrapService.ingest` so the never-block policy lives in one place, and the app call sites are just `void ingest()` now. Also the non-throwing contract is tested once at service level instead of per caller, PR got ~20 lines shorter 👍

for the second one I think I would skip it, I did the math and extracting the shared helper ends with basically the same line count but one more indirection level, not sure is worth it. wdyt?

secondl1ght (approved):

PR #2081, opened 2026-06-25
feat: render path for not found embedded posts

secondl1ght:
## 🩹 Fix: Render a "not found" state for posts/collections that 404 instead of skeletoning forever

  ### Problem

  When Nexus returns a **404** for a post, `usePostDetails` correctly resolves to `postDetails === null` once the fetch
  settles. However, the **embedded** post/collection components only branched on _loading_ vs _deleted_ vs _content_ —
  they never handled the terminal "missing" state. As a result, a 404'd post would sit in its **skeleton loading state
  forever**.

  The page-level routes were already covered:
  - `/post/[userId]/[postId]` correctly bails out to `PostNotFoundDiscoveryView`.
  - `/collections/[userId]/[postId]` correctly bails out to `CollectionNotFound`.

  The gap was entirely in the **embedded** render paths (feed posts, replies, repost/quote previews, embedded collection
  cards).

  ### Solution

  Added a new inline "not found" render path that mirrors the existing soft-deleted (`[DELETED]`) handling, using the
  shared rule: a post is **missing** when `postDetails === null && !isLoading` (`undefined` still means loading → keep the
  skeleton).

  **New inline molecules** (small, centered card content — parallel to `PostDeleted` / `CollectionDeleted`):
  - 🆕 `PostMissing` → "Post not found."
  - 🆕 `CollectionMissing` → "Collection not found."

  > ℹ️  Note: the names `PostNotFound` / `CollectionNotFound` were already taken by the large illustrated route-level
  empty-states, so new distinct inline molecules were introduced.

  **Wired the missing branch into all affected embedded paths:**
  - `PostMain` — the actually-observed feed/reply bug; symmetric with its existing deleted branch.
  - `PostContentBase` — returns the skeleton while loading, `PostMissing` once settled.
  - `PostPreviewCard` — repost/quote previews. ⚠️  This is the one surface that didn't already handle the deleted state, so
  it's **not a 1-for-1 swap**: `PostMissing` _is_ a `CardContent`, so it's rendered as a **direct child of the preview
  `<Card>`** (replacing the inner `<CardContent>`) to avoid double-nesting and to stop the `PostHeader` from skeletoning
  forever.
  - `CollectionCard` — defensive coverage for the (so-far theoretical) collection 404 case.

  **Translations:**
  - Added `post.missing` and `collections.missing` keys to all **9** locale files (`en`, `es`, `de`, `fr`, `it`, `ja`,
  `ar`, `pt-BR`, `zh`), localized per language. 🌍

  ### Testing

  - ✅ New unit tests + snapshots for `PostMissing` and `CollectionMissing`.
  - ✅ Added explicit **missing-state** and **still-loading** cases to `PostMain`, `PostContentBase`, `PostPreviewCard`,
  and `CollectionCard`.
  - ✅ Hardened `PostPreviewCard.test.tsx` — it previously passed only because the unmocked real hook returned
  `undefined`; now `usePostDetails` is mocked for determinism and the new branch is covered.
  - ✅ Full suite: **10,660 passed / 4 skipped / 0 failed** (662 files); only the 2 intended new snapshots were generated.
  - ✅ `tsc --noEmit` clean, `eslint` clean, Prettier clean on message files.

  ### Notes / Follow-ups

  - 👀 Worth a quick **visual check** on `PostPreviewCard`: render a repost whose original post 404s and confirm
  `PostMissing` sits correctly inside the preview card (padding/centering, no clipping) — this can't be verified from
  tests alone.
  - 🌐 The 8 non-English translations were authored as plausible equivalents and may benefit from a native-speaker review.


infin1t3 (approved):
LGTM overall.

PR #2082, opened 2026-06-25
refactor: fix and update toasts after changes from dev

secondl1ght:
## 🧹 Collections Toast Review & Cleanup

  Audits and standardizes every toast notification in the Collections feature. The toasts had grown inconsistent: some
  error toasts rendered with success styling, several piped dynamic collection names into the copy, and failure messages
  hid the actual reason from the user.

  ### 🎯 What & Why

  **1. Removed dynamic values from toast copy**
  Collection toasts that interpolated the collection `{name}` (follow / unfollow / share, add / remove post) are now
  generic. This avoids having to truncate long names and keeps the messaging consistent.

  - `"You've followed the {name} collection"` → `"You've followed this collection"`
  - `"Post added to {name}."` → `"Post added to collection."`

  **2. Explicit `variant` on error toasts**
  There is no `success` variant — `variant` defaults to `'default'`, which *is* the success/checkmark styling. As a
  result, error toasts that omitted `variant` were rendering as success. Every collection error toast now explicitly sets
  `variant: 'error'`. Success toasts intentionally stay on the default variant (the established app-wide convention).

  **3. Collapsed redundant title + description**
  Aligned with the app convention of a single field per toast:

  - ✅ Success toasts → `title` only (dropped the generic `"Success"` title in favor of the descriptive message)
  - ❌ Error toasts → `variant: 'error'` + `description` only (the Toaster auto-fills the `"Error"` title for the error
  variant)

  **4. Forward backend error messages on write failures** 🔎
  Collection write failures (create, edit, add/remove item) previously showed generic copy even when the backend threw a
  readable reason. They now forward the backend message, falling back to generic copy when the error isn't an `AppError`:

  ```ts
  description: isAppError(error) ? error.message : t('updateFailed')
  ```

  This means a user who hits the 100-item cap now sees **"Collection has too many items"** instead of a generic failure —
  and the toast is correctly styled as an error rather than success.

  ### 🌍 Localization

  Removing the `{name}` placeholder from the call sites means the value is no longer passed to `next-intl`. The `{name}`
  placeholder was therefore removed from the 5 affected collection-toast keys across **all 9 locale files** (`en`, `ar`,
  `de`, `es`, `fr`, `it`, `ja`, `pt-BR`, `zh`) — otherwise non-English languages would have rendered a literal `{name}`.
  Each translation was reworded to a grammatically-correct generic equivalent, for example:

  - 🇫🇷 `"Vous suivez désormais la collection {name}"` → `"Vous suivez désormais cette collection"`
  - 🇩🇪 `"Beitrag zu {name} hinzugefügt."` → `"Beitrag zur Sammlung hinzugefügt."`
  - 🇯🇵 `"投稿を「{name}」に追加しました。"` → `"投稿をコレクションに追加しました。"`
  - 🇨🇳 `"你已分享 {name} 收藏"` → `"你已分享此收藏"`

  All 9 locale files remain valid JSON with a consistent placeholder count.

  ### 📦 Changes

  - `messages/*.json` (all 9 locales) — removed `{name}` from 5 collection toast keys
  (`card.toast.followed/unfollowed/shared`, `postSave.addedToCollection/removedFromCollection`)
  - `useCreateCollection`, `useEditCollection`, `usePostSaveTargets` — added `variant: 'error'`, collapsed to single-field
  toasts, and forward `AppError.message` on failures
  - `FollowedCollections`, `MyCollections`, `DiscoverCollections` — load-failure toasts switched to `variant: 'error'` +
  single field; removed now-unused `tToast` declarations
  - `CollectionHero`, `CollectionCard` — dropped the `{ name: toastName }` args and removed the unused `toastName` locals
  - Updated affected test suites (hooks, list sections, and the `NewCollectionDialog` / `EditCollectionDialog` integration
  tests) and added a new test asserting the backend max-items error message is forwarded

  ### 🚫 Out of scope

  - Shared hooks (`useBookmark`, `useDeletePost`, `usePostReplyRepostDialogs`) were left untouched. Their success toasts
  already use the correct default variant and the delete-failure toast already sets `variant: 'error'` — only the
  collection-specific override strings/args at the call sites were adjusted.
  - The delete-collection **dialog** body still uses `{name}` — it's a dialog (not a toast) and its call site still passes
  the value, so it's outside this toast review.

  ### ✅ Verification

  - `tsc --noEmit` clean
  - `eslint` clean
  - Full test suite green: **660 test files, 10,640 tests passing** (4 skipped, 0 failures) — including the new
  backend-error-forwarding test
  - All 9 locale files validated as parseable JSON with consistent `{name}` counts

  ### 📝 Note

  Forwarded `AppError.message` strings (e.g. `"Collection has too many items"`) are English literals, not localized —
  consistent with the app's existing copy/feedback error toasts.

### DIFF
--- a/messages/ar.json
+++ b/messages/ar.json
@@ -1368,9 +1368,9 @@
       "unfollow": "إلغاء المتابعة",
       "delete": "حذف",
       "toast": {
-        "followed": "لقد تابعت مجموعة {name}",
-        "unfollowed": "لقد ألغيت متابعة مجموعة {name}",
-        "shared": "لقد شاركت مجموعة {name}"
+        "followed": "لقد تابعت هذه المجموعة",
+        "unfollowed": "لقد ألغيت متابعة هذه المجموعة",
+        "shared": "لقد شاركت هذه المجموعة"
       }
     },
     "single": {
@@ -1420,8 +1420,8 @@
     "collectionNamePlaceholder": "اسم المجموعة",
     "createCollection": "إنشاء مجموعة",
     "loadingCollections": "جارٍ تحميل المجموعات...",
-    "addedToCollection": "تمت إضافة المنشور إلى {name}.",
-    "removedFromCollection": "تمت إزالة المنشور من {name}.",
+    "addedToCollection": "تمت إضافة المنشور إلى المجموعة.",
+    "removedFromCollection": "تمت إزالة المنشور من المجموعة.",
     "updateCollectionFailed": "فشل تحديث المجموعة.",
     "collectionCreated": "تم إنشاء المجموعة وحفظ المنشور.",
     "createCollectionFailed": "فشل إنشاء المجموعة."
--- a/messages/de.json
+++ b/messages/de.json
@@ -1374,9 +1374,9 @@
       "unfollow": "Entfolgen",
       "delete": "Löschen",
       "toast": {
-        "followed": "Du folgst jetzt der Sammlung {name}",
-        "unfollowed": "Du folgst der Sammlung {name} nicht mehr",
-        "shared": "Du hast die Sammlung {name} geteilt"
+        "followed": "Du folgst jetzt dieser Sammlung",
+        "unfollowed": "Du folgst dieser Sammlung nicht mehr",
+        "shared": "Du hast diese Sammlung geteilt"
       }
     },
     "single": {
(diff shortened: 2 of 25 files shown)


infin1t3 (approved):

infin1t3:
Tested. LGTM.

PR #2100, opened 2026-06-30
fix: bookmarks stream

secondl1ght:
# 🔖 Fix bookmark stream pagination & cache staleness to use bookmark time

  ## Context

  Bookmark post streams (`timeline:bookmarks:all` and `timeline:bookmarks:collection`) are ordered by **bookmark time** on
  Nexus — when a post was bookmarked — not by the post's own creation/`indexed_at` time. Backend confirmed the stream's
  timestamp cursor (both the `start` it accepts and the `last_post_score` it returns) is the bookmarked-at time.

  The frontend, however, synthesized timestamps from the **post's** `indexed_at` whenever it worked from the local cache
  (no Nexus response to read a cursor from). That created two related bugs:

  1. 📄➡️  **Pagination seam.** When "load more" crossed the local-cache → Nexus boundary, the `start` cursor was seeded
  from the last cached post's `indexed_at` instead of its bookmark time. A post bookmarked recently but created long ago
  would send an old `start`, skipping or repeating bookmarks at the page seam.
  2. 🗑️  **Cache staleness.** `prepareStreamForInitialLoad` runs on every bookmarks load and judged staleness from the head
  post's `indexed_at`. With cache max-age at 5 minutes, bookmarking any post older than 5 minutes (i.e. almost always)
  marked the bookmark cache stale and cleared it on every load — defeating local-first caching.

  These interact: the staleness bug kept wiping the cache, so loads usually started cold from Nexus (whose own cursor is
  correct), which is why the seam bug was rarely visible. Both predate the recent bookmarks refactor and also affect the
  `FollowedCollections` section, which rides the same `source=bookmarks` stream.

  ## ✅ What changed

  **`postStream.types.ts`**
  - Added an exported `isBookmarkStream(streamId)` predicate, alongside the existing `isSkipPaginatedStream` /
  `isDeletedRetainingStream`.

  **`post.ts` (`PostStreamApplication`)**
  - New `getBookmarkTimestamp(postId)` helper that reads the bookmark's `created_at` from the local `bookmarks` table.
  - New `getStreamCursorTimestamp(streamId, postId)`: for `:bookmarks:` streams it returns the bookmark time, falling open
  to the post's `indexed_at` if the bookmark row is missing; every other stream still paginates by `indexed_at`.
  - Wired the bookmark-aware cursor into the three cache-served sites: `getCachedLastPostTimestamp` (initial seed), the
  full-cache-hit branch, and `partialCacheHit`.
  - Made staleness bookmark-aware: `getMainStreamHeadTimestamp` and `getUnreadStreamHeadTimestamp` now key off bookmark
  time for bookmark streams. Non-bookmark streams are unchanged.
  - Refactored an existing inline `:bookmarks:` check to use `isBookmarkStream`.

  `FollowedCollections` benefits automatically since it shares the `timeline:bookmarks:collection` stream.

  ## 🧪 Tests

  - `getCachedLastPostTimestamp` bookmark cases: uses bookmark time not post `indexed_at`; falls back to post time when
  the bookmark row is missing; iterates backwards to the most recent resolvable cursor.
  - `getOrFetchStreamSlice`: a full cache hit on a bookmark stream returns the bookmark-time cursor (the actual cache →
  Nexus seam).
  - `prepareStreamForInitialLoad`: a bookmark cache stays fresh by bookmark time even when the bookmarked post is old; and
  is cleared when the most recent bookmark is older than max-age (proving bookmark time governs staleness, not post
  recency).

  ## 🔍 Verified safe (no change needed)

  - The post-time `getStreamHead` is only used by the stream coordinator, which polls `/home`, `/post`, `/feed` — never
  bookmark streams (and bookmarks disable pull-to-refresh), so the new-posts/`end` path is unreachable for bookmarks.
  - Bookmark optimistic-add uses the membership-order `prependOptimisticPosts`, not the timestamp-sorting `prependPosts`,
  so ordering stays correct.

  ## ✔️  Verification

  - `tsc --noEmit` clean
  - `eslint` clean on changed files
  - Full unit suite green: **671 files / 10,767 tests** passing (4 pre-existing skips)
  - Suggested manual check: bookmark a very old post, then "load more" past the first cached page — older bookmarks should
  neither disappear nor duplicate at the seam.

  ---


infin1t3 (approved):
LGTM.

PR #2115, opened 2026-07-01
fix: route shared collection previews to collection pages

### LINKED ISSUE #2090: [Bug]: Collections that are shared will open as a post on first click
### Description

When sharing a collection is shows up in feed as a repost. When clicking on the collection to open then it opens in post page instead of a collection.

### Expected Behaviour

Open collection in the collection page instead of post page on first click.

### Steps to Reproduce

1. Create or find a collection to share in the collections page
2. Click share button
3. Share the collection
4. Navigate to feed
5. Find the shared collection at top of feed and click on its embedded area to open
6. Observe it opens in post page rather than collection page (bad)
7. Click it again
8. Observe it opens as collection page (as it should have at step 6)

### Screenshots

https://github.com/user-attachments/assets/f8f75f29-f3e9-4539-8f87-9c450629ca43

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
fix(posts): route shared collection previews to collection pages

Clicking a collection embedded in a repost or reply preview now opens /collections/:pubky/:id instead of the post detail page. Detect postDetails.kind === 'collection' in PostPreviewCard and branch navigation accordingly, with an updated aria-label for collections.

Centralize collection routing in usePostNavigation via navigateToCollection and getCollectionHref so other components can reuse it without duplicating parseCompositeId and getCollectionRoute. Remove useCallback from the hook now that React Compiler handles memoization.

https://github.com/user-attachments/assets/895620f1-7503-499c-893e-47f8df0160da

Closes #2090

### DIFF
--- a/src/components/molecules/PostPreviewCard/PostPreviewCard.test.tsx
+++ b/src/components/molecules/PostPreviewCard/PostPreviewCard.test.tsx
@@ -5,17 +5,12 @@ import { PostPreviewCard } from './PostPreviewCard';
 
 // Mock hooks
 const mockNavigateToPost = vi.fn();
-const mockRouterPush = vi.fn();
+const mockNavigateToCollection = vi.fn();
 const mockTtlRef = vi.fn();
-vi.mock('next/navigation', () => ({
-  useRouter: () => ({
-    push: mockRouterPush,
-  }),
-}));
-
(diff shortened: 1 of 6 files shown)


infin1t3:
@greptileai review

Nirzhuk (approved):

PR #2136, opened 2026-07-04
refactor: resolves long posts due new lines

### LINKED ISSUE #1652: [Bug]: Article preview in feed can be too tall
For articles with fewer characters but many lines, such as a bullet list, it has caused them to become taller than before. Perhaps we need a height limit too. This is an example of such an article:

**Before**

Screenshot 2026-04-06 at 16 58 14

**After**

Screenshot 2026-04-06 at 16 58 03

_Originally posted by @catch-21 in https://github.com/pubky/pubky-app/issues/1594#issuecomment-4193359920_

### LINKED ISSUE #2109: [Bug]: Posts with many shorts lines can be quite tall
### Description

Similar to https://github.com/pubky/pubky-app/issues/1652 but for normal `short` kind posts instead of `long` articles.

Posts that have many short lines can appear quite tall in the feed. This is likely because the content is truncated with 'Show more" link/button based on char count.

### Expected Behaviour

Come up with an additional rule to truncate based on number of lines as well as char length.
Must also consider `long` kind that is not parsed as an article.

### Steps to Reproduce

1. Create a post with many short lines
2. Observe in feed

### Screenshots

https://github.com/user-attachments/assets/3a3596fb-61c4-407b-a1eb-fe0848e4e3b6

### Console Log

```shell

```

### Browser & Version

_No response_

Nirzhuk:
resolves #1652 
resolves #2109

## Description
Resolves to prevent long posts when using new lines in the natural formatting of the text

Additionally I've desactivated the fake formatting of tables, last images show the actual behavior in staging.

### Users with multiple type of Long posts:
https://pubky-app-pr-2136-fzxmjul7ya-oa.a.run.app/profile/3sy373cm3ghzg5fjaihie7r5t61466pggdekk4dwzq4j8gjtj5ro
https://pubky-app-pr-2136-fzxmjul7ya-oa.a.run.app/profile/asxhj5t5yqtsoh55zjuq5e8dyg9dtoqg6dcirzciazk3aczpsn4o

## Images

Screenshot 2026-07-06 at 08 41 54
Screenshot 2026-07-06 at 08 41 51
Screenshot 2026-07-06 at 08 41 47
Screenshot 2026-07-06 at 08 41 28

Screenshot 2026-07-06 at 09 05 29


Nirzhuk:

aldertnl:
LGTM

src/libs/post/articleContent.ts
```diff
@@ -0,0 +1,28 @@
+export interface ArticleContent {
```
Taewa: AI feedback but worth to consider.

medium | no-unused-exports | src/libs/post/articleContent.ts
What’s wrong: ArticleContent is exported, but it is only used inside articleContent.ts and is not imported anywhere else.

Why it matters: Unused exports widen the module’s public API surface and make the type look externally supported when it is not.

Fix: Remove export from ArticleContent unless another module is meant to import it.

src/components/molecules/PostText/PostText.utils.ts
```diff
@@ -226,20 +563,171 @@ const truncateAstParagraph = (paragraph: Paragraph, limit: number): void => {
...
+const buildArticlePreviewChildren = (children: RootContent[], state: ArticlePreviewState): RootContent[] => {
+  const previewChildren: RootContent[] = [];
+
+  for (const child of children) {
+    if (state.remainingLines <= 0 || state.remainingChars <= 0) {
+      state.hiddenContent = true;
+      break;
+    }
+
+    if (child.type === 'heading') {
```
Taewa, line 628: AI feedback. 

medium | simplification | src/components/molecules/PostText/PostText.utils.ts
What’s wrong: The heading and paragraph branches in buildArticlePreviewChildren are not identical, but they repeat the same state update and character-limit truncation flow: extracting text, pushing the node, checking remainingChars, truncating on overflow, setting previewHasEllipsis / hiddenContent, and updating remainingChars.

Why it matters: The branches have real differences, so keeping them separate is fine. But the shared truncation rules now live in two places, which makes future changes easier to apply inconsistently.

Fix: Keep the separate heading and paragraph branches if that reads better, but extract the shared “text preview node” handling into a small helper. Pass the differences, such as consumed line count, max line limit, and the truncate function, as parameters.

Taewa (commented):
Left 2 medium AI feedback

Taewa (commented):
@Nirzhuk 
For table, I understand it as a temporary solution. Do we have a follow-up ticket for a proper table rendering with markdown? 
Do we need to support markdown table rendering? 
I see that it's not also properly being rendered in article mode.

Nirzhuk:
> For table, I understand it as a temporary solution. Do we have a follow-up ticket for a proper table rendering with markdown?

No, we don't. I believe it's a good thing, before this, we were doing something very weird with tables that wasn't real usage, which was a bug. If we want to add tables, that's a new thought more than anything.
Here's how they look in staging:
https://staging.pubky.app/post/3sy373cm3ghzg5fjaihie7r5t61466pggdekk4dwzq4j8gjtj5ro/0035BVPZBZ4Q0
The formatting looks like a bug, but no one reported it, so I believe it was never intended to be compatible.

> Do we need to support markdown table rendering? 

I believe what I really did was fix a bug, we don't support them because they look like this, and removing them prevents buggy behavior.


For giving you the perspective we didn't even had a component in our markdown parser.

@Taewa

Taewa:
> I believe what I really did was fix a bug, we don't support them because they look like this, and removing them prevents buggy behavior.

We use `remarkGfm` which helps to understand markdown and I noticed that some markdown works while some are not. For example, list works while link or title don't.
I guess table rendering is one of them. But it's strange that markdown table is not rendered in article more which supports markdown.

I know that @aldertnl already signed as LGTM but can you please confirm about the table rendering please?

Nirzhuk:
Screenshot 2026-07-06 at 14 07 12

@Taewa 

In Article we dont have a button to help render the table

Taewa:
@Nirzhuk 
I noticed a little tick during my test.
I am not 100% sure but this may caused by
( `src/hooks/usePostNavigation/usePostNavigation.ts` )
```js
// Post detail pages are scrolled to whatever position the click happened at (e.g. clicking
// "Show more" near the bottom of a card). Scroll to top before push so the new page doesn't
// briefly render at the old scroll position.
function scrollToTop() {
  window.scrollTo({ top: 0, behavior: 'auto' });
}
```
My screen recording compares with ephemeral and staging env.
Do we need `scrollToTop()`?

https://github.com/user-attachments/assets/4e993161-6f33-4a25-97c0-8562037055a8

Nirzhuk:
> @Nirzhuk I noticed a little tick during my test. I am not 100% sure but this may caused by ( `src/hooks/usePostNavigation/usePostNavigation.ts` )
> 
> ```js
> // Post detail pages are scrolled to whatever position the click happened at (e.g. clicking
> // "Show more" near the bottom of a card). Scroll to top before push so the new page doesn't
> // briefly render at the old scroll position.
> function scrollToTop() {
>   window.scrollTo({ top: 0, behavior: 'auto' });
> }
> ```
> 
> My screen recording compares with ephemeral and staging env. Do we need `scrollToTop()`?
> 
>  2136-auto-scroll.mp4

Removed, yeah like we said in slack, I dont know why I was getting a wrong behavior and added this, sorry

catch-21:
Different length (height/width) of posts look better in feed and in collections:

https://github.com/user-attachments/assets/72ce120e-c4f6-4807-94e5-dad0933f16b1

Long and wide articles appear well in feed and collections too:

https://github.com/user-attachments/assets/bdf57a9d-05ce-4837-9d1c-6a6ab23fd181

Taewa (approved):
LGTM

PR #2144, opened 2026-07-07
chore: add blank vrt-update-baselines.yml for testing in branch

catch-21:
Add placeholder `.github/workflows/vrt-update-baselines.yml` so the one in https://github.com/pubky/pubky-app/pull/1995 can be triggered for testing before merging into `dev`.


Taewa (approved):
LGTM

PR #2155, opened 2026-07-07
fix(posts): skip optimistic repost prepend on collection and bookmarks feeds + minor cleanup

infin1t3:
## Summary
- Stop optimistically prepending reposts into the active feed unless it is a timeline feed (`home` / `custom` / `hot`), so sharing a collection from collection no longer inserts the repost into that membership feed.
- Keep stream-kind gating for prepends, and catch/log `getDetails` failures so a failed lookup still collapses the composer instead of leaving an unhandled rejection.
- Small cleanups: reuse `useEffectiveTagsLayout` in `PostInput`, derive `isEmbed` inside `CollectionCardContent`, and remove dead collection navigation from `PostPreviewCard` (collections already render embed `CollectionCard`).


Taewa (approved):
LGTM

I tested with @infin1t3 and comparing between staging and ephemeral.

Issue #2149, opened 2026-07-07
[Merged into #2148] Wire Nexus stream requests for Network and domain feeds


Orlandohub:
Parent: #2111

This issue has been merged into #2148 so the cache-safe stream identity work and Nexus request wiring are implemented together in one end-to-end slice.

Use #2148 as the source of truth for:

- `Network` without profile tags -> `source=wot&depth=2`.
- `Following/Friends + profile tags` -> `source=wot_domain&depth=1&domain_tags=...`.
- `Network + profile tags` -> `source=wot_domain&depth=2&domain_tags=...`.
- Keeping `domain_tags` distinct from existing post/content `tags`.
- Query builder and stream-id coverage for all supported V1 mappings.

Closing this standalone tracker to avoid duplicate implementation issues.

Orlandohub:
Merged into #2148 so stream identity and Nexus request wiring are tracked and implemented together.

Issue #2147, opened 2026-07-07
[WoT V1] Reach sidebar UI and profile tag selector


Orlandohub:
Parent: #2111

Implementation: #2154 (merged into `feat/wot-v1` on 2026-07-10)

Follow-up enablement: #2231 (merged into `feat/wot-v1` on 2026-07-21)

## Status

Complete on the `feat/wot-v1` integration branch. The Reach sidebar, Network and Me options, and profile-tag selector are implemented. The temporary Me gate was removed by #2231 after Nexus depth-zero support was verified on staging and production.

## Implemented Scope

- Reach shows All, Network, Following, Friends, and Me.
- Plain Me keeps the existing author stream.
- The profile-tag selector appears under Reach and uses the existing validation, suggestions, colors, and emoji selector.
- All clears selected profile tags and disables the selector.
- Network, Following, Friends, and Me enable profile tags.
- Me with profile tags maps through the depth-zero domain-feed implementation in #2231.
- Selected profile tags are stacked vertically with an independent maximum of 5, matching the Nexus `domain_tags` API limit.
- Multiple profile tags use OR semantics; AND semantics remains out of scope.
- Desktop sidebar and drawer/mobile variants share the same state policy.

Figma reference: https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=19902-14534&m=dev

## Acceptance Criteria

- [x] Reach shows All, Network, Following, Friends, and Me.
- [x] Profile tags match the intended Reach-sidebar design.
- [x] All clears and disables profile tags.
- [x] Network, Following, Friends, and Me enable profile tags.
- [x] Plain Me remains usable through the existing author stream.
- [x] Up to 5 selected tags are displayed and enforced.
- [x] Existing suggestions and emoji selection are available.
- [x] Desktop and mobile/drawer variants behave consistently.
- [x] Component and store coverage verifies reach/profile-tag transitions.

PR #2156, opened 2026-07-07
[WoT V1]  Wire cache-safe wot_domain streams and Nexus params

### LINKED ISSUE #2148: [WoT V1] Add cache-safe domain-feed streams and Nexus wiring
Parent: #2111

Merged scope: #2149

Implementation: #2156 (merged into `feat/wot-v1` on 2026-07-15)

Follow-up depth-zero enablement: #2231 (merged into `feat/wot-v1` on 2026-07-21)

## Status

Complete on the `feat/wot-v1` integration branch. Cache-safe domain-feed stream identity, Nexus request wiring, strict depth parsing, and Me depth-zero support are implemented.

Beyond the original scope, #2156 also fixed two review findings: `matchesFilters` returns `false` for Me rather than throwing, and post-kind prepend gating uses the canonical `getPostStreamKind` helper so `wot_domain`, tag, and author-kind streams cannot bypass the kind gate.

## Confirmed Nexus Mappings

- Network without profile tags: `source=wot&observer_id={viewerPubky}&depth=2`.
- Network with profile tags: `source=wot_domain&observer_id={viewerPubky}&depth=2&domain_tags=...`.
- Following/Friends with profile tags: `source=wot_domain&observer_id={viewerPubky}&depth=1&domain_tags=...`.
- Me without profile tags: existing author stream.
- Me with profile tags: `source=wot_domain&observer_id={viewerPubky}&depth=0&domain_tags=...`.
- All with profile tags remains unsupported for V1.
- Multiple profile tags use OR semantics, up to 5 tags.

## Implemented Behavior

- Profile/domain tags are encoded deterministically in stream/cache identity.
- Profile-tag order is canonicalized so equivalent sets share identity.
- Profile/domain tags (`domain_tags`) remain separate from post tags (`tags`).
- Existing plain reach/content streams remain unchanged.
- Depth is fixed app-side: Network 2, Following/Friends 1, Me 0.
- Depth parsing is strict and accepts only app-supported values `0 | 1 | 2`.
- Unsupported combinations are gated before network requests.
- WoT depth was not added to tag-visibility endpoints.
- Combined `tags` and `domain_tags` remain supported for custom feeds through #2190/#2191.

## Acceptance Criteria

- [x] Different profile-tag combinations produce distinct cache identities.
- [x] Equivalent tag sets produce canonical identities.
- [x] Existing post-tag streams remain separate and unchanged.
- [x] All clears profile tags and uses the normal All stream.
- [x] Plain Network uses `source=wot` with depth 2.
- [x] Following/Friends profile-tag feeds use `wot_domain` depth 1.
- [x] Network profile-tag feeds use `wot_domain` depth 2.
- [x] Me profile-tag feeds use `wot_domain` depth 0.
- [x] Query-builder and stream-ID tests cover supported mappings.
- [x] Existing stream-ID and consumer tests continue to pass.

Orlandohub:
## Summary
- fix #2148
- Profile tag selections on Network, Following, and Friends now produce distinct, cache-safe `wot_domain` stream IDs instead of reusing plain reach streams.
- Nexus requests for those feeds send `source=wot_domain` with the correct `depth` and `domain_tags`, keeping post content tags separate from author profile tags (includes merged #2149 scope).

## Changes
- Add `WotDomainStreamCompositeId`, `buildWotDomainStreamId`, and widen `PostStreamId` / home stream builders to emit canonical domain IDs from profile tags and reach depth.
- Refactor `breakDownStreamId` to return named fields and parse `wot_domain` segments with strict `parseWotDomainDepth` validation.
- Wire `profileTags` through `useStreamIdFromFilters`, `StreamCoordinator`, and `UserController.getActiveStreamId`; restart polling when profile tags change.
- Map Following/Friends → depth 1 and Network → depth 2 in `home.utils.ts`, with `isProfileTagGatedReach` narrowing instead of runtime casts.
- Add unit coverage for ID canonicalization, `domain_tags` vs `tags` separation, collections-kind regression, and coordinator re-evaluation on tag changes.

https://github.com/user-attachments/assets/a1d8292f-80ab-455b-810c-e60a55703fbe

## Test plan
- [ ] add/remove profile tags on Network reach and confirm the feed refetches with distinct cache keys per tag set

## Notes
- Builds on the Reach sidebar / profile tag selector UI from #2147 on this branch.
- `Me` and `All` profile-tag feeds remain gated until Nexus support lands (pubky/pubky-nexus#981).
- Follow/unfollow still invalidates only fixed timeline enum streams; cached `wot_domain` streams refresh via coordinator polling rather than explicit invalidation.


Orlandohub:
@greptile-apps review

secondl1ght (approved):
Approved with one currently low severity bug finding that is worth fixing either here or in the next PR. But it is a symptom of a larger concern where our stream ID system has become very messy. 

The problem compounds when we continue to add more segments to the magic string and parse it in many places across the app. 

I think the system started out fairly well defined, but it has now grown to the point where any part of the string segment can mean anything. Even though we try to type it. 

When the string is later split and read there are many latent bugs waiting to happen. 

I think this is some tech debt that is worth discussing and addressing as a team in the future. May be worth even having an entire tech debt sprint sometime to tackle issues like this.

Here is the actual bug report though:

#### 🐛 `wot_domain` streams bypass the post-kind prepend gate

**`src/core/stores/home/home.utils.ts:174`** — `parseStreamId` returns `null` for any id where `split(':').length !== 3`. The new `wot_domain` ids have **5 segments**, so it always returns `null`, and `postKindBelongsToStream` then hits its `!parsed → return true` ("unparseable → accept all") fallback, dropping the kind gate.

**Repro:** authed user, reach = Network, content = Images, profile tag `bitcoin` → active stream `timeline:wot_domain:2:image:bitcoin`. Post a **short** post → both prepend gates (`usePostInput.ts:232`, `NewPostsSection.tsx:80`) call `postKindBelongsToStream('short', 'timeline:wot_domain:2:image:bitcoin')` → `true`, so the short post is prepended into an images-only feed. Same breaks `:collection` wot_domain feeds — exactly what the `NewPostsSection` comment says to prevent.

Orlandohub:
Addressed the bug from [review 4698856902](https://github.com/pubky/pubky-app/pull/2156#pullrequestreview-4698856902) in 2cb6f918.

Rather than teaching `parseStreamId` about 5-segment ids (depth 1 cannot reconstruct Following vs Friends), the kind is now extracted by a canonical model-layer parser, `getPostStreamKind()` in `postStream.types.ts`, and `postKindBelongsToStream` delegates to it. This closes the prepend-gate bypass for:

- `wot_domain` streams (the reported repro: short post into `timeline:wot_domain:2:image:bitcoin`)
- 4-segment tag streams (`timeline:all:image:bitcoin` from search/custom feeds) — same class, pre-existing
- author-kind streams (`<pubky>:author:<kind>` from reach=Me + content filter) — same class, pre-existing

The helper also already understands the sorted-author shape (`{sorting}:author:<pubky>:<kind>[:tags]`) coming in #2190, so that branch only needs to reuse it. Coverage includes unit tests for every stream id shape plus mismatched-`wot_domain` regressions on both reported prepend paths (`usePostInput`, `NewPostsSection`).

PR #2163, opened 2026-07-08
fix(images): guarantee uploaded images stay under 5MB

infin1t3:


infin1t3:
Closing this PR - decided to take a different more minimal route.

Issue #2181, opened 2026-07-10
Locks orphan case 2: content lock left behind when the announcement post fails


Taewa:
Sub-issue of #1998. Sibling of #2039, which covers the *previous* gap in the same sequence.

## Problem

Publishing a locked post is three writes, in order:

1. `PUT /creator/priv-resources/content/<id>` — the guarded resources (one per file).
2. `POST /creator/content-locks` — the content lock; the Lock Server publishes the public lock file at `/pub/locks.app/<lock_id>.json`.
3. The **announcement** — a normal `PubkyAppPost` written to the creator's homeserver with a `lock` field pointing at that lock file.

#2039 handles an interruption between **1 and 2**. This issue is the interruption between **2 and 3**.

When step 3 fails — network error, homeserver rejection, or the creator simply closing the tab — the outcome is:

- The guarded content **exists** in the private homeserver.
- The content lock **exists** on the Locks server, and its lock file is **publicly readable** at `/pub/locks.app/<lock_id>.json`.
- **No post references it.** Nobody can discover the lock: it is not in any feed, and Nexus never indexes it.

The lock is not dangerous — the content stays guarded — but it is dead weight the creator cannot see or remove, and it silently consumes their private storage.

## Why it is not #2039

#2039's failure leaves *resources without a lock*. This one leaves *a lock without a post*. The recovery is different: here the FE holds a valid `lock_id` and could simply retry the post, whereas #2039 has lost the resource descriptors.

## Options

- **Retry the announcement.** The lock is already valid, so the retry body is unchanged. Cheapest fix for the transient case (network / 5xx), and it covers most of the real failures.
- **Delete the content lock** on a deterministic failure or an explicit cancel, so the creator's `/pub` and `/priv` are left clean. Needs a Lock Server delete route for content locks (only `deleteGuardedResource` exists today).
- **Reconcile on next launch** — list the creator's lock files and drop any that no post points at. Most complete, most expensive.

## Current behaviour

Nothing. `useCreateLockContent` publishes the lock and the announcement in sequence; a failure in the second step surfaces as a toast, reports to Sentry via the `Err.*` factory, and leaves the lock behind.

Marked in code as `TODO:[Locks] #2181`.

## Definition of Done

A failed announcement no longer leaves an unreferenced content lock — either it is retried, cleaned up, or reconciled — and the creator is told what happened.

Issue #2040, opened 2026-06-19
Locks: release gate — audit & remove all TODO:[Locks] #1998 markers before ship


Taewa:
Sub-issue of #1998. **Release gate — audit & resolve every `TODO:[Locks] #1998` marker before Phase 1 ships.**

## Goal
Phase 1 left dev/mock shortcuts tagged with `TODO:[Locks] #1998` in the code. Before release, find every marker, confirm the real implementation landed (or is consciously deferred), and remove the comment. This ticket is the single checklist that guarantees no dev-only shortcut leaks to production.

## Why
Each marker is a known production hazard: a hardcoded dev value, a trusted-input shortcut, or a silent localhost fallback. They are individually small but easy to forget; centralizing them here prevents a partial cleanup.

## Markers to resolve (as of branch `feat/2025-locks-create-lock-post-ui-mock`)
- [ ] **`src/app/api/locks/content/route.ts:14`** — `creator` is trusted from the request body (dev/mock). Production must derive the creator from the Locks-local session/auth (depends on #2001 — the Locks login establishes that session; real wiring under #2026). **Security-relevant.**
- [ ] **`src/libs/env/env.ts:177`** — `NEXT_PUBLIC_LOCKS_SERVER_URL` defaults to `http://localhost:3100`; an unset prod env silently falls back to localhost. Point the default at real/staging Locks server, or drop the default so it is required.
- [ ] **`src/core/application/locks/locks.ts:7`** — `VERIFIER_TYPE = 'dev-static'` is a dev-only verifier; must be replaced by real `password`/`payment` verifiers before release.
- [ ] **`src/core/application/locks/locks.ts:4`** — `CREDENTIAL_TTL_SECONDS = 900` is a fixed placeholder; creator-configurable TTL is deferred/under discussion (#2025). Confirm fixed value is acceptable for release or wire the input.

> **Future markers (not in code yet):** #2001 (Locks login modal + lock-service config) stubs the lock-sdk auth flow behind `TODO:[Locks]` until the SDK lands. Those markers must resolve here too once that work merges — the grep gate below will surface them.

## Definition of Done
- `grep -rn "TODO:\[Locks\]" src/` returns **zero** results.
- For each marker: either the real implementation landed and the comment is removed, or the deferral is explicitly approved and re-tracked under a non-1998 ticket (not left as a bare TODO).
- No dev-only shortcut (trusted creator, localhost fallback, `dev-static` verifier, fixed TTL) is reachable in a production build.

[...]

Issue #2039, opened 2026-06-18
Locks orphan case 1: uploaded guarded resources left behind when publish fails before the content lock


Taewa:
Sub-issue of #1998. Sibling of #2181 (orphan case 2), which covers the *next* gap in the same sequence.

## Problem

Publishing lock content is a sequence of uploads followed by one bundling call (`ApplicationLocks.createLockContent`):

1. `PUT /creator/priv-resources/content/<id>` — once **per attachment**, sequentially, then once more for the locked-post JSON. Each upload returns a descriptor (path, hash, size) that lives **in memory only**.
2. `POST /creator/content-locks` — bundles all the uploaded resources into one content lock.

Every resource that has been uploaded is orphaned the moment the flow dies **anywhere after that upload and before step 2 succeeds**. The danger window is not the single gap "between step 1 and step 2" — it spans from the first upload until `createContentLock` returns, and it grows with attachment count and file size.

## How it happens

Mostly **not** rare:

- **Step-2 failure while the app is still running** — network error, Locks-server outage. The FE shows a toast and does nothing else; the uploaded resources stay behind.
- **Deterministic repro**: the Lock Server enforces size limits only at `createContentLock` (server config; defaults 10 MB/file, 10 files, 100 MB total). An oversized file uploads fine and then fails the lock — a guaranteed orphan, every time.
- **Retry multiplies the damage**: each attempt mints fresh UUID paths and re-uploads everything, so every failed publish leaves one more full set of orphans.
- **Rare but unrecoverable**: the tab/app is closed mid-sequence (e.g. losing connection on the metro). The descriptors were memory-only, so after restart nothing anywhere knows the orphans' paths.

## Why it matters

These orphans are invisible. Their paths are random UUIDs under `/priv`, referenced by nothing — unlike #2181, where the public `lock.json` still records every resource path, here **no record of the paths exists at all** once the flow dies. The creator cannot see them, cannot delete them, and they silently consume private storage.

## Directions

- **Cheap FE fix, covers most real cases**: in the failure path of `createLockContent`, the descriptors are still in memory — call `deleteGuardedResource` for each uploaded resource before surfacing the error. The route already exists.
- **UX pre-check**: validate sizes against the known limits before uploading anything. Server stays the authority; this only avoids the guaranteed-orphan path.
- **Server-side reaping**: the Locks server tracks/reaps blobs never referenced by a lock — the only thing that covers the tab-close remainder.

## Context

[...]

Issue #2025, opened 2026-06-16
Locks: publishing — create lock post — UI with mock


Taewa:
Sub-issue of #2002. **Phase 1 — UI only.**

## Goal
Build the creator lock-content **UI and interaction**, fully decoupled from how the FE reaches the Locks backend. Applying a lock shows a success toast only — no real submission yet (the transport is being reworked, see below).

## Scope (UI only)
- [x] Lock toggle in the New Post composer (POST variant)
- [x] Lock Content dialog with verifier tabs — **Password** (functional) / **Payment** ("To be developed")
- [x] Password UI: password + repeat, match check, policy rules (≥8 chars / ≥1 number / ≥1 special) shown only when unmet
- [x] Apply Lock → success toast + close the lock dialog and the compose modal
- [x] Modal styling per Figma (node 35688-348348)
- [ ] Lock-content **TITLE input** in the Locked Post Preview modal — text field above the preview. Empty → placeholder + locked-content card title default to **"Locked post"**. On input, the card's "Locked post" title is replaced live by the input value. Figma: empty state node 40174-294862, filled state node 40234-103134

## Out of scope (→ Wiring sub-issue #2026)
- **Any Locks backend communication.** Decision: the FE must NOT call the Locks server directly (nor via a BFF proxy) — it will talk to the Locks BE **through the Pubky SDK, like the homeserver**. That transport is being built and is **blocked** until the SDK supports Locks. The password-submit handler is a UI stub tagged `TODO:[Locks]`.
- Preview post referencing the lock file + showing the locked post in the feed — #2026 (depends on lock-reference post model #2029)
- `password` / `payment` verifiers (server-side, not implemented)

## Deferred — under discussion, may be added later
- Creator-configurable **TTL input** (900s/15min default is fixed server-side for now)
- **Lock logic `all`/`any`** UI (N/A for the single-condition MVP)
- **Guarded-resource drag & drop upload** (current flow guards the composed post text)

## Definition of Done
The lock-content UI is complete and demoable: toggle → dialog → password validation → Apply shows a toast and closes the dialogs. No backend wiring (that is #2026).

Taewa:
The connected PR has been dropped ( https://github.com/pubky/pubky-app/pull/2045 )
Instead, there is a new PR ( https://github.com/pubky/pubky-app/pull/2186 ). Any update will happen on #2002.
So closing this issue.

PR #2186, opened 2026-07-12
feat(locks): creator lock-content — auth + composer + publish (#2026, #2025, #2001)

### LINKED ISSUE #2001: Locks: creator auth + lock-service config
Part of #1998.

## Goal
A lock-content creator authenticates to the **Locks server** (a session separate from the Pubky/homeserver one) and then configures the lock service on the homeserver. The Locks login is triggered the moment the creator turns on locking in the composer.

## Context — trigger
A creator is **already signed into Pubky** (homeserver session) but is **not necessarily signed into the Locks server** — the two are separate sessions. The moment the creator flips the **Lock switch ON** in the New Post composer (POST variant, see #2025), if there is **no active Locks session**, open the **Locks login modal** before the lock flow can continue.

- If `isLocksAuthenticated` → proceed with the existing lock flow (#2025).
- If **not** authenticated → open the 3-step Locks login modal. On success, continue with the switch ON. On cancel/close, **revert the switch to OFF** (no half-locked state).

## Same identity (pubky) for both

> **⚠️ UPDATE (2026-07-03 — supersedes this section):** Same-pubky enforcement is **no longer required**. We now **allow** the Locks pubky to differ from the pubky.app/HS pubky and simply **surface which identity is being used** after auth (e.g. show profile / "signed in as X; lock content stored in X's homeserver") instead of rejecting. The text below is kept for history but is **superseded**. Rolling as-is for end of July.

**The pubky used to log into the Locks server MUST be the same pubky used to log into pubky.app (i.e. the one backing the active homeserver session).** Locks is a second session for the *same* identity, not a different account. The Locks login (step 2's Pubky Ring authorize) must be performed with — or validated against — the currently signed-in pubky (`useAuthStore.selectCurrentUserPubky()`); if the authorized pubky differs from the HS-session pubky, reject the Locks login and surface a mismatch error rather than storing it.

## Login and session are separate per service; logout is unified
- **Login — separate.** HS login (Pubky Ring → pubky-sdk) and Locks login (Pubky Ring → lock-sdk, name TBD) are distinct steps for the **same identity** (see above). Being signed into Pubky does NOT sign you into Locks.
- **Session — separate.** Each session is its own object in its own persisted store (`auth-store` vs `locks-auth-store`), restored independently on app load.
- **Logout — unified.** Logging out of pubky.app logs out of **both**. App logout already tears down the HS session (`AuthController.logout()` → `HomeserverService.logout()` → `session.signout()`, then `cleanupLocalState()`), and the Locks teardown is added to that same path. There is no standalone Locks logout.

## 3-step login modal
1. **Intro** — explains the creator needs to sign into the Locks service (what it is / why). CTA → start.
2. **QR + Pubky Ring** — show a QR code and instruct the creator to authorize with Pubky Ring (desktop: scan QR; mobile: deeplink button). Poll for approval. Mirrors the homeserver `SignIn` QR phase.
3. **Success** — confirms the Locks session was established; closes the modal and resumes the lock flow.

## Session management — mirror the homeserver pattern
Reuse the **same shape** as the existing pubky-sdk/homeserver auth, backed by the **lock-sdk** and a separate persisted store. Reference flow in code today:

- **Service** — `HomeserverService.generateAuthUrl()` → `pubkySdk.startAuthFlow(capabilities, signin, relay)` returns an `authorizationUrl` (the QR) and an approval poller (`createCancelableAuthApproval` → `flow.tryPollOnce()`). `src/core/services/homeserver/homeserver.ts:235`, `homeserver.utils.ts:155`.
- **Application / Controller** — `AuthApplication.generateAuthUrl()` / `AuthController.getAuthUrl()`; QR lifecycle + approval polling driven by `useAuthUrl` / `useMobileAuth`. `src/core/application/auth/auth.ts:139`, `src/core/controllers/auth/auth.ts:310`, `src/hooks/useAuthUrl/useAuthUrl.tsx`.
- **Session persistence** — on approval the SDK returns a `Session`; serialized via `session.export()` and persisted in the Zustand `auth-store` (`partialize`: `sessionExport`, `currentUserPubky`, `hasProfile`). `src/core/stores/auth/auth.store.ts:9`, `auth.actions.ts:6`, `persistedKeys.ts:1`.
- **Restore on load** — `pubkySdk.restoreSession(sessionExport)` on rehydrate, with retry. `src/core/services/homeserver/homeserver.ts:521`, `src/core/application/auth/auth.ts:41`.
- **Logged-in check** — `selectIsAuthenticated` = `session !== null`. `src/core/stores/auth/auth.selectors.ts:18`.

### Proposed Locks equivalent
- **`LocksAuthService`** wrapping the lock-sdk: `startAuthFlow` (→ QR url + poller), `restoreSession`, `signout`. Singleton like `getPubkySdk()`.
- **`LocksAuthApplication` / `LocksAuthController`** mirroring the auth layers; expose `getLocksAuthUrl()`, `restorePersistedLocksSession()`, `logoutLocks()`.
- **New persisted store** `locks-auth-store` (add to `persistedKeys.ts` + `PERSISTED_STORE_KEYS`), partialized to `locksSessionExport` (+ creator pubky if the lock-sdk surfaces it). Separate store from `auth-store` only because it holds a separate session object — NOT an independent lifecycle (see logout above).
- **Selector** `selectIsLocksAuthenticated` and a `useLocksAuthStatus` hook for the composer toggle gate.
- **Restore** the Locks session on app load via `lockSdk.restoreSession(locksSessionExport)`, same retry pattern.
- **No manually-attached token.** Like the homeserver — the FE never builds an `Authorization: Bearer` header itself. Authenticated homeserver writes go through the SDK `Session` object (`session.storage.putJson/putBytes/list`, `src/core/services/homeserver/homeserver.ts:340/394/439`) and public reads through `pubkySdk.client.fetch` with `credentials: 'include'` (`homeserver.ts:368/504`); the SDK session carries the auth. The Locks equivalent is the lock-sdk **`LocksSession`**: creator API calls are made **through that session object** (mirroring `session.storage` / `client.fetch`) and the lock-sdk carries the auth — the FE attaches nothing by hand. (Server-side, deriving the creator from that session instead of trusting the request body is the dev shortcut tracked in #2040 / `src/app/api/locks/content/route.ts`.)

### Logout wiring
- Add `locks-auth-store` to the stores reset by `cleanupLocalState()` and to `PERSISTED_STORE_KEYS`.
- Call the lock-sdk session signout/teardown inside `logout()` next to the homeserver `session.signout()`.
- Result: one logout drops **both** sessions — the user can never be logged out of Pubky while still holding a live Locks session.

## ⚠️ Auth mechanism — reconcile before finalizing
This issue now describes the auth as **Pubky Ring QR → lock-sdk session, mirroring the homeserver** (the working assumption per the wider Locks design). An earlier draft of this ticket described a different, **redirect-based** flow: a "Connect Locks" URL with `return_to` + `state`, a redirect back with a single-use `code`, then `code → session token` exchange. These are two different mechanisms — **confirm which one is correct with Denys + Pubky core** (the auth flow is still WIP under their review) and drop the other. The rest of this ticket assumes the Pubky-Ring/lock-sdk path.

## Open questions
**Working assumption: the Locks session behaves identically to the homeserver session in every respect below — same SDK auth-flow shape, same Pubky Ring relay/keypair, same long-lived no-refresh lifecycle, same persistence location.** Listed only to confirm/deny with the lock-sdk owners; absent any signal otherwise, mirror the homeserver exactly.
- **lock-sdk readiness** — the lock-sdk does not exist yet (name TBD); this issue defines the FE shape so it can be wired the moment the SDK lands. Until then the auth flow is a UI stub tagged `TODO:[Locks]` (tracked at the release gate #2040).
- Pubky Ring relay/keypair — **assumed same** as pubky-sdk (so step 2's QR is the same authorize). Confirm.
- Session expiry/refresh — **assumed same** as homeserver (long-lived, no refresh today). Confirm.
- Long-term storage of the session export — **assumed same** as homeserver (localStorage now → homeserver private folder `/priv/...`, pubky/pubky-core#426, for device independence, per the #1998 storage direction). Confirm.
- Dual-auth is **creator-only** and separate from the normal pubky.app session — confirmed for Phase 1; single-auth is future work.

## Scope
- [ ] Composer Lock-switch gate: open Locks login modal when toggled ON without a Locks session; revert toggle on cancel.
- [ ] 3-step modal (Intro → QR/Pubky Ring → Success), reusing `QrCodeSlot` + the SignIn QR/poll pattern.
- [ ] `LocksAuthService` (lock-sdk wrapper) + Application/Controller layers, mirroring the homeserver auth layering.
- [ ] `locks-auth-store` persisted store + `selectIsLocksAuthenticated` + `useLocksAuthStatus`.
- [ ] Restore Locks session on app load; **clear it as part of Pubky logout** (hook `locks-auth-store` reset + lock-sdk signout into `cleanupLocalState()` / `logout()`).
- [ ] ~~**Enforce same-pubky:** the Locks login must use / match the active HS-session pubky; reject + surface an error on mismatch.~~ **CHANGED (2026-07-03):** do NOT enforce — surface the authenticated identity instead (see "Same identity" update).
- [ ] Stub the actual lock-sdk calls behind `TODO:[Locks]` until the SDK exists.

## Out of scope
- Real lock-sdk transport / creator API wiring → #2026 (and blocked on lock-sdk).

## Definition of Done
Toggling the Lock switch ON with no Locks session opens the 3-step login modal; completing it (with the **same pubky** as the active pubky.app/HS session) establishes a persisted Locks session (managed like the homeserver session, via lock-sdk) and resumes the lock flow; the session restores on reload and **is cleared together with the Pubky session on logout**. Real lock-sdk calls remain stubbed until the SDK is available.

_(NOTE 2026-07-03: same-pubky no longer enforced — auth may use a different pubky; we only surface which identity is used. See "Same identity" update.)_

---

## 🔄 UPDATE (2026-07-07) — Auth mechanism LOCKED: **iframe + postMessage, NOT redirect**

**Supersedes the "⚠️ Auth mechanism — reconcile before finalizing" section.** Decision: the Locks login runs **inside an iframe** in the 3-step modal. The creator never leaves pubky.app via a top-level redirect. There is also no in-app Pubky-Ring QR rendered by pubky.app — the Lock Server owns and renders the authorize UI inside the iframe (its `authorization_url` is secret-bearing and must not be handled by pubky.app).

Flow (`useLocksAuthFlow`):
1. Modal embeds the Lock Server `/connect` page in an iframe (`iframeRef`), passing a CSRF `state`.
2. Creator approves in Ring / signer inside the iframe.
3. One-time `code` + `state` come back to the parent via **`postMessage`**; parent validates (origin / source / schema + `state` match) and exchanges `code` → session in the service layer (SDK access stays there).

Two sub-variants (differ only in how the `code` returns to the parent):
- **Target — B-style (no redirect):** the Lock Server `/connect` page posts the `code` **directly** to the parent (cross-origin postMessage). No same-origin callback page.
- **Interim — A-style (current local code):** iframe redirects to our own same-origin callback page `src/app/locks/auth-callback/page.tsx`, which relays the URL up via postMessage. Kept only to prove the flow locally. **Delete on B-style landing:** `src/app/locks/auth-callback/page.tsx` + the same-origin branch in `src/libs/locks/locksAuthBridge.ts`.

**Dependency on Lock Server team (Denys):** `locks/docs/_front_end_integration.md` currently specifies a full-page **redirect** flow (`pubky.app/locks/callback?code&state`). For the iframe target the Lock Server must (a) allow `/connect` to be embedded in an iframe from the pubky.app origin (`frame-ancestors` / drop `X-Frame-Options: DENY`), and (b) post the `code` back cross-origin instead of only redirecting. Confirm both.

## Scope addition — `lock-service-config` write (Configuration step)

After the Locks session is established, register the creator's default Lock Server on their homeserver so consumers can discover it.

- FE calls **`POST {lockServer}/creator/lock-service-config`** with `Authorization: Bearer <frontend_session_token>` and body `{ "default_lock_server": "<lock server pubky>" }`.
- The **Lock Server** writes the pointer JSON to the creator homeserver at `/pub/locks.app/config.json` — **FE does NOT write to the homeserver directly** (corrects the earlier "session.storage.putJson" assumption).
- Response: `{ creator, path: "/pub/locks.app/config.json", lock_service_pointer: { version, default_lock_server, created_at } }`.
- Runs in the **background** on auth success — no extra UI step.
- **Why it matters:** SDK `Locks.forCreator` fetches `/pub/locks.app/config.json` to resolve the creator's default Lock Server. Without this write, locked-content consumption can't find the server.
- Refs: `locks/docs/_front_end_integration.md:32-62`, `locks/docs/API.md:486-512`, `SDK.md:184`.

- [ ] `LocksAuthController.setLockServiceConfig({ defaultLockServer })` → service `POST /creator/lock-service-config` with the frontend-session bearer.
- [ ] Call it in the background after auth success (session persisted first).
- [ ] Surface a config-write failure without dropping the established session.

---

## 🔄 UPDATE (2026-07-08) — config write goes through the SDK session, not a hand-built HTTP call

**Refines the `lock-service-config` section above** after reviewing the locks-sdk (`locks/docs/SDK.md:322`, `API.md:486`).

- The FE does **NOT** hand-build `POST {lockServer}/creator/lock-service-config` nor attach the `Authorization: Bearer` header itself. It calls the SDK method on the established session:
  ```ts
  await session.creator.setLockServicePointer(defaultLockServer)
  ```
  The SDK issues the `POST /creator/lock-service-config` and carries the frontend-session bearer internally — consistent with this issue's **"No manually-attached token"** principle (the FE attaches nothing by hand; the session object carries auth).
- The **Lock Server** (not the FE) then writes the pointer JSON to the creator homeserver at `/pub/locks.app/config.json`. Confirms the 2026-07-07 correction of the earlier `session.storage.putJson` assumption.
- Not automatic on auth: `exchangeFrontendSessionCode` only establishes the session. `setLockServicePointer` is a **separate explicit call** the FE fires in the background after the session is persisted.

Updated task shape:
- [ ] `LocksAuthService.setLockServiceConfig(session, defaultLockServer)` → wraps `session.creator.setLockServicePointer(defaultLockServer)`.
- [ ] `LocksAuthApplication` / `LocksAuthController.setLockServiceConfig({ defaultLockServer })` layers on top.
- [ ] Fire in the background after auth success (session persisted first); tolerate failure without dropping the session.

### LINKED ISSUE #2002: Locks: publishing — create lock post
Part of #1998

## Goal
Creator publishes guarded content + a public lock file + a preview.

## Tasks
- [ ] Upload guarded resource (drag & drop) → `POST /creator/guarded-resources`
- [ ] Build criteria (password / `dev-static` for Phase 1) + lock logic (`all`/`any`) + TTL
- [ ] Create content lock → `POST /creator/content-locks` (returns `lock_id` + public lock file)
- [ ] Create a preview post that references the lock file

## Open / blocked
- ⚠️ pubky-app-spec post-model extension (how exactly the post references the lock file) — **unassigned**, shape TBD.
- Preview ↔ lock-file assembly responsibility (FE vs Nexus) — under discussion.

### LINKED ISSUE #2026: Locks: publishing — create lock post — Wiring
Sub-issue of #2002. Wire the create-lock-post flow to the real Locks backend. #2025 ships UI only (Apply = toast stub); this builds the transport + persistence behind it.

## Transport
FE talks to the Locks BE **through the Pubky SDK, like the homeserver** — not direct HTTP, not a BFF proxy. Blocked until the SDK supports Locks.

## Upload (decided 2026-06-25)
Guarded content = single opaque blob. Multiple files → FE zips them, base64-encodes, and registers the guarded resource:
```json
{
  "creator": "pubky<z32>",
  "path": "/priv/locks.app/content/<name>.zip",
  "content_type": "application/octet-stream",
  "content_base64": "<base64(zip bytes)>"
}
```
- `path` is FE-chosen (must start `/priv/locks.app/content/`, unique to avoid overwriting prior content). Server stores raw bytes by creator/path; it does not zip, validate, or transform.
- `content_type` = **`application/octet-stream`**, not `application/zip`: the server validates the MIME via the `mime` crate, which does not accept `application/zip`. FE always uploads a zip; the reader-side FE assumes zip and decompresses after proxy-read.

Then create the content lock (criteria, lock_logic, …) → public lock file `/pub/locks.app/<lock_id>.json`.

## Preview post
Published as a normal `PubkyAppPost`: optional `lock` field = flat `pubky://` URL to the lock file, creator-supplied preview metadata (`name` / `description`) embedded in `content`. The public lock file has no title/description (that's behind the lock); FE copies metadata into the post on publish, Nexus does not generate previews. **Shape owned by #2029.**

## Scope
- [ ] FE↔Locks transport via the Pubky SDK (register guarded resource → create content lock), once the SDK path exists
- [ ] Zip multi-file on upload + decompress on read
- [ ] Preview post: `lock` URL field + metadata in `content` (#2029)
- [ ] Creator auth/session on the calls (#2001)
- [ ] Real verifiers beyond `dev-static` (`password`, …) once server-side
- [ ] Local-first feed render of the creator's just-published post (write to IndexedDB + prepend, like a normal post)

## Blocked by
- Pubky SDK Locks support
- Lock-reference post model — #2029
- Auth — #2001

## Done
Create-lock-post works end-to-end against the real Locks BE via the SDK: authenticated creator, published preview post, just-published locked post rendered in the creator's feed from IndexedDB.

Taewa:
## What

Creator-side "lock content" in the post composer: a creator writes something, locks it behind a password, and publishes a public teaser that points at the lock.

This branch bundles three stacked tickets that grew together while the Lock Server stayed local-only (no staging to split them across):

- **#2001** — creator authenticates to the Lock Server + writes the lock-service config; session restored on app load.
- **#2025** — composer lock switch + Lock Content dialog (password/payment tabs) UI.
- **#2026** — wiring: publish the guarded content and the announcement, two-phase composer, teaser.

Part of **#2002** (publishing). Supersedes **#2045** (its UI commits are included here).

## Flow

Switch on → capture the written body as the content to lock → sign in to the Lock Server if needed → set a password (Apply only configures; Post publishes). Post uploads the guarded files, creates the content lock, then publishes the public announcement carrying the lock URL. The announcement is authored by the pubky.app account while the guarded content lives on the Lock-Server-authenticated account — the two may differ and every reference points at the right one. Cancelling at any step turns the draft back into a normal post.

## Not for `dev` yet

Base is **feat/2002**, not `dev`. The Lock Server isn't deployed to staging, so this can only be tested locally and must not reach `dev` until the server and homeserver private-data support land (see #1998). This PR feeds the integration branch (#2046).

## Locks TODOs (temporary, tracked)

The diff carries `TODO:[Locks] #NNNN` markers — all deliberate placeholders that get removed with their ticket. `grep -rn "TODO:\[Locks\]"` audits them before release:

- **#2040** — the Lock Server has no password verifier yet, so a `dev-static` placeholder is sent; also the staging manual-test harness and untyped-SDK casts. All removed when the verifier lands.
- **#2039** — orphan cleanup + size pre-check when an upload partly fails.
- **#2181** — the lock survives if the announcement fails (unreferenced); recovery deferred.
- **#2003** — locked-post card / attachment-URI rendering finishes on the reader side.
- **#2001** — one iframe-sandbox note to verify against the live /connect flow.

## Tests

Unit tests across every layer, plus composer wiring + integration tests for the safety rule that matters most: the to-be-locked body is never published as a public post.

## Note

`package.json` / `package-lock.json` are intentionally left out — `@pubky/locks-sdk` currently resolves to a local tarball. They land when the SDK is published to npm.


Taewa:
@greptile-apps
Updated code to fix the "Locked Draft Bypasses Discard Prompt". 
Review again plz

.env.example
```diff
@@ -116,3 +116,11 @@ SUPPORT_ACCOUNT_ID=123
...
 # used only for source-map upload. They are not runtime app config and are not
 # needed to run the public image.
+
+# =============================================================================
+# Locks (dev/staging only)
+# =============================================================================
+# Lock Server pubky — used only by the dev/staging Lock-auth test trigger (#2001; TODO:[Locks] #2026).
+# Resolved via pkarr (no URL needed). No shared staging Lock Server yet; set it in your local .env.
+# Locally it is the `lock_server_public_key` from ~/.pubky-lock/config.toml (per-machine).
+# NEXT_PUBLIC_LOCK_SERVER=
```
infin1t3: Stale wording, we don't use `NEXT_PUBLIC_` anymore. I guess you meant `PUBKY_RUNTIME_LOCK_SERVER`?

src/core/services/locks/locks.utils.ts
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * The runtime-configured Lock Server pubky. Locks is a feature that can simply be off (no config),
+ * and every Locks entry point is gated on the config, so this never fires in a healthy build —
+ * a warning is enough; no typed error, no Sentry noise for a disabled feature.
+ */
+export function getLockServerPubky(): string {
+  const lockServerPubky = getLockServer();
+  if (!lockServerPubky) {
+    Logger.warn('[LocksService] No Lock Server configured (Locks disabled); call should be unreachable');
+    throw new Error('No Lock Server configured');
```
infin1t3: This still throws raw error - why are you even throwing it if you say a warning is enough?

infin1t3 (changes requested):
Given the current state of this feature I reviewed only the code and not the functionality.

A lot of things I found are already listed in various todos and as future parts of #2040/#2181 etc so when you put all that aside - there is not much things I can comment other than these two :)

.env.example
Taewa: Right, code updated.

src/core/services/locks/locks.utils.ts
Taewa: Yeah right, the throw needs to stay, the warning was the wrong part. Fixed.
Thanks for finding it out.

src/hooks/useCreateLockContent/useCreateLockContent.ts
```diff
@@ -0,0 +1,114 @@
...
+      return { status: 'published', postId: announcementPostId };
+    } catch (caught) {
+      // No Logger here: the service Err factory already logged and reported this.
+      const appError = isAppError(caught)
+        ? caught
+        : toAppError(caught, ErrorService.Locks, 'useCreateLockContent.publish');
+      setError(appError);
+      // The Lock Server rejected the session. Restore is offline, so a stale secret would keep the UI
+      // looking authenticated — drop it here and tell the caller to reopen the sign-in modal.
+      if (appError.category === ErrorCategory.Auth) {
```
infin1t3: **[1/10 · correctness]** This catch treats *any* `ErrorCategory.Auth` error as a Lock Server session rejection, but the announcement publish (`PostController.commitCreate`, line 86) sits inside the same `try` and throws `Err.auth(SESSION_EXPIRED, { service: ErrorService.Homeserver })` when the pubky.app homeserver session expires (`checkSessionExpiration`, homeserver.utils.ts:270).

Scenario: homeserver session expires while the Locks session is healthy. The lock is created (line 77), then the homeserver write 401s → this catch **clears the valid Locks session** and returns `'auth-expired'` → PostInput reopens the **Lock Server** sign-in dialog. Re-authing there can never fix the homeserver 401, and every retry mints another orphaned lock (compounding #2181).

Fix: also require `appError.service === ErrorService.Locks` here (or move the announcement step out of this `try`), and let homeserver auth errors flow to the normal homeserver re-auth path.

src/components/organisms/PostInput/PostInput.tsx
```diff
@@ -140,11 +147,102 @@ export function PostInput({
...
+      setArticleTitle(draft.articleTitle);
+    },
+    clearComposer: clearComposerForLock,
+  });
+
+  // Once the switch is on the composer holds the announcement; the locked post is the captured draft.
+  // The locked post may be any kind, so infer it exactly as a normal post would (link / image / video / …).
+  const { publish: publishLock, isPublishing: isPublishingLock } = useCreateLockContent({
+    lockedPost: {
+      content: lockDraft?.content ?? '',
```
infin1t3, line 203: Locking an **article** silently drops its title. Nothing gates the lock switch in article mode (`PostInputActionBar` renders `lockSwitch` without the `!isArticle` wrap the emoji/image buttons have; `canEnable` only checks content/attachments), and the locked payload here is the bare body with `kind` inferred `Long` — while the normal article path publishes `JSON.stringify({ title: articleTitle.trim(), body: content.trim() })` (usePost.tsx:101). `lockDraft.articleTitle` is only used for restore-on-cancel.

Result: the guarded post is a `Long` post whose content has no `{title, body}` envelope — the title is silently lost and the content won't parse as an article on the reader side.

Fix: either serialize the article envelope into the locked content when `lockDraft.isArticle`, or exclude article mode from `canEnable` until supported.

src/components/organisms/PostInput/PostInput.tsx
```diff
@@ -140,11 +147,102 @@ export function PostInput({
...
+      kind: inferPostKindForCreate({
+        content: lockDraft?.content ?? '',
+        attachments: lockDraft?.attachments,
+        isArticle: lockDraft?.isArticle,
+      }),
+      attachments: lockDraft?.attachments ?? [],
+    },
+    announcement: {
+      teaser: { lock_title: lockTitle, teaser_description: content },
```
infin1t3, line 212: The announcement's published content is `JSON.stringify({ lock_title, teaser_description })`, but the 2000-char limit is enforced only on the raw teaser (`maxLength={POST_MAX_CHARACTER_LENGTH}` on the textarea; counter counts `content` only) and the lock-title input has **no maxLength** (`isValid()` never checks it). The ~41-char JSON envelope plus escaping pushes a near-limit teaser — or any long title — past the spec limit.

Verified against the installed specs wasm: `builder.createPost` **throws** (`Post content exceeds maximum length for Short kind`), it does not truncate. Because the lock is created strictly before the announcement (useCreateLockContent.ts:77 vs 86), the failure is deterministic: generic `lockError` toast, and each retry mints another orphaned lock while the counter tells the user they're within limits.

Fix: validate the *serialized envelope* length before `createLockContent`, and cap the title input.

src/core/pipes/post/post.normalizer.ts
```diff
@@ -87,7 +87,14 @@ export class PostNormalizer {
...
       }
 
-      return builder.createPost(post.content, post.kind, post.parentUri ?? null, embedObject, attachments);
+      return builder.createPost(
+        post.content,
+        post.kind,
+        post.parentUri ?? null,
+        embedObject,
+        attachments,
+        post.lock ?? null,
```
infin1t3, line 96: `createPost` now carries `post.lock ?? null`, but `toEdit` (line ~160, unchanged) still rebuilds with the 5-arg `new PubkyAppPost(content, kind, replied, embed, attachments)` — the specs d.ts shows `lock` is only settable via the static `new_with_lock`. So editing a published lock announcement (menu → Edit → Save) rewrites its post file **without the lock reference**, permanently unlinking the locked content — and the announcement is the only public pointer to it.

The local post model has no `lock` field to round-trip either, and announcements render as plain posts, so the edit path is fully reachable (no menu gate for locked posts).

Fix: carry `lock` through the local model and `toEdit`, or hide/block Edit for announcement posts until #2003 lands.

src/hooks/usePostInput/usePostInput.ts
```diff
@@ -150,6 +150,11 @@ export function usePostInput({
       const dialogContent = document.querySelector('[data-slot="dialog-content"]');
       if (dialogContent?.contains(target)) return;
 
+      // The lock-title input renders just above the composer container but belongs to it: focusing it
+      // must not collapse the composer. `closest` on the target (not a document query) so multiple
+      // mounted composers cannot shadow each other.
+      if (target instanceof Element && target.closest('[data-lock-title-input]')) return;
```
infin1t3, line 157: The collapse condition a few lines below still treats an empty composer as "nothing in progress", but enabling the lock deliberately *empties* the composer (`captureComposer()` + `clearComposer()`) while holding the captured draft. On the inline Home/Custom-feed composer (mounted without `expanded`, so this handler is active), any click elsewhere after flipping the switch fires `setIsExpanded(false)` — unmounting the lock card, lock switch, and Post button mid-flow, and stranding the lock-title input (rendered outside the container, gated only on `isLockEnabled`) above a collapsed composer. The draft survives but looks lost.

Fix: include lock state (`isLockEnabled` / `lockDraft`) in the collapse guard — the same fix this PR already applied to `useConfirmableDialog`.

src/hooks/useLocksAuthFlow/useLocksAuthFlow.ts
```diff
@@ -0,0 +1,118 @@
...
+    setStatus(LocksAuthFlowStatus.IDLE);
+  }, []);
+
+  // Run when the modal opens: probe readiness before the user can start, so a dead / not-ready server
+  // disables "Continue" (with a message) instead of loading a broken iframe on click.
+  const prepare = useCallback(async () => {
+    setError(null);
+    setStatus(LocksAuthFlowStatus.CHECKING_SERVER);
+    try {
+      const reachable = await LocksController.isServerReachable();
```
infin1t3, line 52: `prepare()` sets status unconditionally after the `await`, with no generation counter or abort. The hook lives in `DialogLocksAuth`, which stays mounted across open/close (PostInput renders it unconditionally), and the effect runs `prepare()` on every open — so close+reopen while a slow probe is in flight (`isServerReachable` has no timeout) produces two concurrent probes where **last-resolved wins**.

Race: probe A (slow) starts → user closes (reset → IDLE) → reopens → probe B resolves `reachable=false` → `SERVER_UNAVAILABLE` → stale probe A resolves `true` and sets `IDLE`, re-enabling Continue against a server the latest probe said is down. The inverse ordering leaves the modal stuck on `SERVER_UNAVAILABLE`.

Fix: a generation counter (or `open`-keyed AbortController) checked before each `setStatus` after the `await`.

src/core/controllers/auth/auth.ts
```diff
@@ -295,6 +296,10 @@ export class AuthController {
     useNotificationStore.getState().reset();
     useSettingsStore.getState().reset();
 
+    // Unified logout: tear down the Locks session (Lock Server signout + local store) alongside the
+    // homeserver session, so the user can never stay logged into Locks after leaving pubky.app.
+    await LocksController.logout();
```
infin1t3, line 301: This `await` runs a Lock Server network signout (`LocksController.logout` → `LocksApplication.signout` → `session.signout()`) with no timeout or AbortSignal **before** `clearCookies`/`clearDatabase` — the try/catch in `LocksController.logout` only fires after the promise settles, not while it hangs.

Scenario: user with a live Locks session logs out while the Lock Server is blackholed (connection accepted, no response). Local cleanup is blocked until the OS-level TCP timeout and `Logout.tsx` keeps the user on a spinner the whole time. It mirrors the existing unbounded homeserver signout, but adds a second, independent hang path gated only on a Locks session existing.

Fix: race the signout against a short timeout (or fire-and-forget it) so local teardown always proceeds.

src/core/application/locks/locks.ts
```diff
@@ -0,0 +1,117 @@
...
+  static async createLockContent({
+    attachments = [],
+    buildPost,
+  }: TCreateLockContentParams): Promise<TCreateContentLockResult> {
+    // The guarded bytes land on the Lock-Server-authenticated account, which may differ from the
+    // pubky.app user. Capture that owner from the upload response so `buildPost` can reference the
+    // attachments by their real host. With no attachments the owner stays undefined and no URIs are built.
+    let owner: string | undefined;
+    const attachmentResources: TGuardedResource[] = [];
+    for (const file of attachments) {
```
infin1t3, line 86: Attachment uploads run strictly sequentially, but each upload is independent: a fresh `crypto.randomUUID()` path per file, no shared mutable state, and `creator` is identical for every response under the same session (it's overwritten each iteration anyway). N attachments = N serial round-trips carrying raw bytes before the post upload and `createContentLock` can start, with the Post button blocked throughout.

Fix: `Promise.all(attachments.map((f) => this.upload(f)))` (or a bounded-concurrency map) — results preserve input order for `attachmentResources`, and `creator` can come from any result. Only the post upload and `createContentLock` need to stay sequential. (Partial-failure orphaning already exists sequentially and is tracked in #2039.)

src/core/controllers/locks/locks.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * True when the Lock Server is reachable + ready.
+   *
+   * `/readyz` needs the server's HTTP origin, which the SDK only exposes by building a connect URL —
+   * so build a throwaway one just to read its origin; the
+   * real auth builds its own URL at "Continue".
+   */
+  static async isServerReachable(): Promise<boolean> {
+    const url = await LocksApplication.generateConnectUrl({
```
infin1t3, line 40: `isServerReachable` mints a CSRF `state` and builds a full `/connect` URL — a network pkarr resolution — solely to read `new URL(url).origin` for the `/readyz` probe (the comment above admits it), and `start()` then resolves a *second* connect URL on Continue. The Lock Server is a runtime-configured constant, so its origin never changes within a session.

Side effect on UX: `prepare()`'s catch maps any throw here to `SERVER_UNAVAILABLE` — and since `isServerReady` swallows fetch errors, the only thing that can throw is `createConnectUrl`/client init. A transient pkarr relay hiccup therefore shows "Lock Server unavailable" even when the server is fine.

Fix: cache the resolved origin alongside the cached client singleton in `LocksService` (resolve once per session) and probe `/readyz` against that; then a probe failure genuinely means the server, not pkarr.

src/hooks/useLocksAuthFlow/useLocksAuthFlow.ts
```diff
@@ -0,0 +1,118 @@
+'use client';
+
+import { useCallback, useEffect, useRef, useState } from 'react';
```
infin1t3: No need for useCallback we use react-compiler

infin1t3 (changes requested):

src/hooks/useCreateLockContent/useCreateLockContent.ts
Taewa: Code updated. But the orphan issue will be addressed later in #2181.

src/components/organisms/PostInput/PostInput.tsx
Taewa: This is fixed in 54bb9860a (`feat/2005-locks-unlocked-profile`)
But it's in the following PR #2301.

src/core/pipes/post/post.normalizer.ts
Taewa: Oh that's a good finding! I actually found a related issue. 
But it'd be much easier to handle with a separated PR since it requires `lock` field. 
I have create an issue https://github.com/pubky/pubky-app/issues/2312 and it will be implemented as a high priority.

1e74867aa59ea2e9cba6a97162b388ea7136366f

src/components/organisms/PostInput/PostInput.tsx
Taewa: That's good finding, thanks.
Resolved by following the existing article content pattern.

src/core/application/locks/locks.ts
Taewa: I'd like it to be handled in a separated PR: https://github.com/pubky/pubky-app/issues/2316

src/hooks/usePostInput/usePostInput.ts
Taewa: Code updated.

src/hooks/useLocksAuthFlow/useLocksAuthFlow.ts
Taewa: Code updated.

src/core/controllers/auth/auth.ts
Taewa: Code updated.

src/core/controllers/locks/locks.ts
Taewa: - Two resolutions because the SDK never exposes an origin: `lockServer()` returns the pubky, the resolved endpoint stays in the `wasm`, and `createConnectUrl` is the only call that leaks it.
- Right fix is in the SDK: filed pubky/locks#12 for `isReady()`. It already resolves the endpoint, 
`/readyz` is its protocol.
- Tracked as #2317, blocked on that. A cache would make this happen once instead of on every modal          
open, but if that one lookup fails we still blame the Lock Server when the real problem is pkarr.
Waiting for the SDK to add `isReady()`.

src/hooks/useLocksAuthFlow/useLocksAuthFlow.ts
Taewa: Code updated.

infin1t3 (approved):
Checked just the code part - didn't went through the usual verification, a lot of parts missing.

PR #2228, opened 2026-07-16
fix(ui): align collection counts with avatars

aldertnl:
## Summary

- Move collection post counts directly to the left of owner avatars
- Apply the same layout to the pinned Bookmarks card
- Update affected component snapshots

## Validation

- 49 focused component tests passed
- ESLint passed
- `git diff --check` passed

image


infin1t3 (approved):
Verified, LGTM.

PR #2230, opened 2026-07-16
fix(ui): show collection-aware notification copy

### LINKED ISSUE #2125: [Bug]: Notification for updated collection shows as "edited a post" (v25)
### Description

Having interacted with a collection (e.g. reposted it), when that collection is edited by its owner a notification is received that reads "Testarossa edited a post you have interacted with". This doesn't match the design.

### Expected Behaviour

v25 design shows notification text as "[username] updated a Collection"

Also consider that it should link to the collection and not a post page.

### Steps to Reproduce

1. Create a collection from profile A
2. Repost that collection from profile B
3. Edit the collection with profile A
4. Observe notification as profile B doesn't match design

### Screenshots

Actual

Image

Design

Image

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
## Summary

- consume the Nexus `post_kind` field for supported post-bearing notifications
- show collection- and article-specific notification copy across all nine locales
- route collection subjects to `/collections/<owner>/<postId>`
- preserve generic copy for missing, `unknown`, and future post kinds
- preserve existing live-post targets for repost and deleted-post notifications

## Root cause

Notification bodies previously did not expose the subject post kind to the app, so collection edits fell through to the generic “edited a post you have interacted with” copy and the generic post route. pubky/pubky-nexus#994 now supplies a backward-compatible `post_kind` field.

## User impact

Collection edit notifications now display the v25 copy “updated a Collection” and navigate to the collection detail page. Existing and unknown notification kinds continue to use the current generic behavior.

Closes #2125

## Validation

- `npm test` — 679 files passed; 10,969 tests passed; 2 skipped
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `git diff --check`

https://github.com/user-attachments/assets/5fab0198-e6d5-4c90-a533-01652b3dea7d


aldertnl:
Some small requests:
- Can we list the collection name (in muted color) like we do with posts? `Anna updated collection 'x'`
- Can we show the correct category icon at the end of the row?
- Let’s use lowercase `collection` to be consistent with `post`.

image

aldertnl:
Thanks, LGTM

src/components/organisms/NotificationItem/NotificationItem.utils.test.ts
```diff
@@ -0,0 +1,138 @@
...
+describe('NotificationItem utilities', () => {
+  it('shows the subject preview for edited-collection notifications', () => {
+    expect(hasPostPreview(NotificationType.PostEdited, 'collection')).toBe(true);
+  });
+
+  it.each([undefined, 'long', 'unknown'])('does not show an edited-post preview for post kind %s', (postKind) => {
+    expect(hasPostPreview(NotificationType.PostEdited, postKind)).toBe(false);
+  });
+
+  it.each([
```
catch-21, line 14: Could be worth covering `short` uses correct copy also, in this or other test.

catch-21: ah perhaps a similar data-driven approach for the test below "keeps generic copy for unsupported post kind "?

messages/en.json
```diff
@@ -844,12 +844,24 @@
...
+      "repostedCollection": "reposted your collection",
+      "repostedArticle": "reposted your article",
       "mentionedYou": "mentioned you in post",
+      "mentionedYouInCollection": "mentioned you in a collection",
+      "mentionedYouInArticle": "mentioned you in an article",
       "deletedPost": "deleted a post",
+      "deletedCollection": "deleted a collection",
+      "deletedArticle": "deleted an article",
       "editedPost": "edited a post you have interacted with",
+      "updatedCollection": "updated collection",
```
catch-21, line 866: "updated a collection"

src/components/organisms/NotificationItem/NotificationItem.utils.test.ts
```diff
@@ -0,0 +1,138 @@
...
+      edited_by: 'owner',
+      edited_uri: 'pubky://owner/pub/pubky.app/posts/post-id',
+      linked_uri: 'pubky://viewer/pub/pubky.app/posts/repost-id',
+      post_kind: postKind,
+    } satisfies FlatNotification;
+
+    expect(getNotificationActionKey(notification)).toBe('editedPost');
+  });
+
+  it('links an updated collection to its collection detail page', () => {
```
catch-21: We can also add a test to ensure the routing to post page is still functioining, e.g.:

```
  it('keeps the post route for an updated article', () => {
    const notification = {
      id: 'post_edited:123:owner',
      type: NotificationType.PostEdited,
      timestamp: 123,
      edit_source: PostChangedSource.Repost,
      edited_by: 'owner',
      edited_uri: 'pubky://owner/pub/pubky.app/posts/collection-id',
      linked_uri: 'pubky://viewer/pub/pubky.app/posts/repost-id',
      post_kind: 'long',
    } satisfies FlatNotification;

    expect(getNotificationLink(notification).notificationLink).toBe('/post/owner/collection-id');
  });
```

messages/en.json
aldertnl: I think `updated collection` may be correct because the collection name follows after this copy

`Anna updated collection 'Cats'`

catch-21: If that's how it should be, then we need to fix that because it currently shows like:
Screenshot 2026-07-16 at 22 56 48
Screenshot 2026-07-16 at 22 56 44

catch-21: ... turns out this is because my browser is not full size. We should keep collection name for all responsive sizes where possible. That also applies to others like "tagged your post" and "mentioned you in post", possibly others aswell.

Screenshot 2026-07-16 at 23 06 02

catch-21:
Can look a bit odd on narrow display and long username. Perhaps it should be hidden when truncation doesn't show enough?

Screenshot 2026-07-17 at 17 43 20

aldertnl:
Figma mobile design does not show those details on the mobile breakpoint, so you can just hide the muted details on that mobile breakpoint @infin1t3 

image

Taewa (dismissed):
LGTM
I tested manually on ephemeral link and it works (link also redirect to the collection)

Image

Taewa (approved):

PR #2231, opened 2026-07-16
[WoT V1] Enable Me profile-tag domain feeds at depth 0

### LINKED ISSUE #2150: [WoT V1] Enable Me + profile tag feeds
Parent: #2111

Implementation: #2231 (merged into `feat/wot-v1` on 2026-07-21)

Completes Me + profile tags for the Home work in #2147/#2148 and custom feeds in #2190. Plain Me remains unchanged.

## Status

Complete. The Nexus contract, staging and production deployment checks, functional verification, and app implementation are finished.

- pubky/pubky-nexus#982 provides the required depth-zero trust set.
- Confirmed request: `source=wot_domain&observer_id={viewerPubky}&depth=0&domain_tags=...`.
- Semantics: posts by authors whom the viewer directly tagged with any selected profile labels. The viewer's own posts remain excluded by WoT-source semantics.
- Staging was verified on 2026-07-15 against Nexus 0.4.1.
- Functional staging verification returned real posts for observers with direct profile tags; a nonexistent tag returned an empty result.
- Production Nexus was verified on 2026-07-21 against Nexus 0.4.1: OpenAPI documents depth 0 and a valid request returns HTTP 200.
- #2231 is merged into `feat/wot-v1`.

OpenAPI:
- Staging: https://nexus.staging.pubky.app/api-docs/v0/openapi.json
- Production: https://nexus.pubky.app/api-docs/v0/openapi.json

## Implemented Behavior

- Me without profile tags keeps the existing author stream.
- Me with profile tags uses `wot_domain` with explicit depth 0.
- Selected profile tags are preserved when switching to Me.
- All continues to clear and disable profile tags.
- Network, Following, and Friends mappings remain unchanged.
- Home and custom feeds share the supported-reach capability policy.
- Cache identity, strict parsing, and request serialization preserve depth 0.
- Remote Me profile-tag custom feeds are admitted during bootstrap without rewriting.

## Work Checklist

- [x] Deploy pubky/pubky-nexus#982 to staging.
- [x] Verify staging request acceptance and OpenAPI documentation.
- [x] Extend app depth types, builders, and strict parsing to `0 | 1 | 2`.
- [x] Enable Me through the shared profile-tag reach policy.
- [x] Wire and test Home and custom-feed depth-zero paths.
- [x] Functionally verify non-empty depth-zero results on staging.
- [x] Verify production Nexus deployment.
- [x] Merge #2231 into `feat/wot-v1`.

## Acceptance Criteria

- [x] Nexus accepts explicit depth-zero domain requests without default-depth fallback.
- [x] Functional verification returns tag-driven Me-domain results.
- [x] Cache identity and request parsing represent depth 0 explicitly.
- [x] URL-builder and stream-ID tests cover Me + profile tags.
- [x] All other reach mappings remain unchanged.
- [x] #2231 is merged into the integration branch.

Orlandohub:
Fixes #2150

## Summary

Nexus depth-0 support (pubky/pubky-nexus#982) is live on staging, so `Me + profile tags` now maps to `source=wot_domain&observer_id={viewer}&depth=0&domain_tags=...` — a feed of posts by authors the viewer directly tagged with the selected labels. Plain `Me` (no profile tags) keeps the existing author stream.

- **Depth plumbing**: `WotDomainDepth` widened to `0 | 1 | 2` (model), `TStreamIdBreakdown.wotDepth` now reuses that type (service), and the stream-id parser accepts `'0'` while still rejecting anything else.
- **Truthiness bug fix**: `createPostStreamParams` used `if (wotDepth)`, which silently dropped depth `0` and would have let Nexus apply its default depth — quietly widening a Me feed into a network feed. Now checks `!== undefined`.
- **Single gate flip**: `'me'` added to `PROFILE_TAG_SUPPORTED_REACHES` in `src/config/feed.ts`. That one constant drives the Home sidebar gating, store tag-clearing on reach switch, Home stream-id building, custom-feed dialog gating, save validation, and bootstrap admission — no per-surface logic changes were needed.
- **Depth mappings**: `REACH.ME → 0` in the Home depth map; the custom-feed depth ternary (`Wot ? 2 : 1`) replaced with an exhaustive typed map (`satisfies Record<ProfileTagSupportedReach, WotDomainDepth>`) so a future supported reach fails compilation instead of silently defaulting to depth 1.
- **Cleanup**: removed the now-dead `isProfileTagGatedReach` helper (zero consumers; encoded the stale All/Me gate).

https://github.com/user-attachments/assets/a770b868-3462-4de3-b182-1bce53eb94c8

## Tests

- Model/service: depth-0 stream-id build + parse round-trip, `params.depth === 0` regression for the truthiness bug, and a URL-serialization table case proving the literal `depth=0` survives into the request URL.
- Store/hooks: `Me + tags` produces `timeline:wot_domain:0:...` in Home and custom-feed stream-id hooks; `setReach(ME)` no longer clears profile tags; All still clears.
- Custom feeds: Me profile-tag feeds build/save/hydrate; dialog keeps tags enabled on switch to Me and still clears for All; controller accepts Me + domain tags and still rejects All/Followers.
- Bootstrap: replaced the gated-Me skip test with an admission test proving a remote Me + `domain_tags` feed keeps its original HashId, persists locally, and triggers no PUT/DELETE/migration; the skip path is still covered via an All fixture.

## Functional staging verification (issue checkbox)

Verified depth-0 returns real results (not accepting-but-empty) against staging with observers that have direct profile tags:

- `observer_id=6mfxozzq…` + `domain_tags=test-moderation` → HTTP 200, posts by `pjsm9sr7…` (the author that observer tagged).
- `observer_id=npyph6zg…` + `domain_tags=ta` → HTTP 200 with posts; same observer with a nonexistent tag returns an empty array, confirming results are tag-driven.

## Verification

- Targeted suites: 13 files / 505 tests + 13 adjacent consumer files / 342 tests, all passing
- `tsc --noEmit`, ESLint on changed files, `git diff --check` clean


secondl1ght (approved):
🕸️

secondl1ght:
LGTM

Claude has a couple comments, so I will drop them here but they are non-blocking. And the specs vs. local strings issue is pre-existing before WoT, I've commented on it before but we've not done any cleanup on this tech debt so far to date.

```
  One substantive nit

  The PR body claims the satisfies map makes "a future supported reach fail compilation instead of silently defaulting to
  depth 1." That's only half true, and only on the Home side:

  - Home side (home.utils.ts:62) — genuinely type-safe. The type guard narrows ReachType with no cast, so adding a reach
  to the config without a depth entry is a compile error.
  - Model side (feed.helpers.ts:134) — the as ProfileTagFeedReach cast severs that link. If someone adds 'followers' to
  PROFILE_TAG_SUPPORTED_REACHES without touching ProfileTagFeedReach, it compiles fine; at runtime the map lookup returns
  undefined and you get timeline:wot_domain:undefined:..., which only fails later at parse time (Invalid wot_domain depth:
  undefined). It fails loudly rather than silently, but at request time, not compile time.

  Suggestion (non-blocking): add a small unit test iterating every supported reach through buildFeedStreamId and asserting
  a valid depth segment, or derive ProfileTagFeedReach so the config list and the union can't drift.

  Minor observations

  - The depth policy now lives in two maps (WOT_DOMAIN_DEPTH_BY_REACH in home.utils and wotDomainDepthForReach in
  feed.helpers). They key on different types (REACH strings vs the specs enum), so unifying them isn't free, but a comment
  cross-referencing them would reduce drift risk when depths are ever retuned.
  ```

Orlandohub:
Addressed the substantive nit from https://github.com/pubky/pubky-app/pull/2231#issuecomment-5028086759 in 4719833d: the custom-feed depth map is now keyed by the exported `ProfileTagSupportedReach` union from `config/feed.ts` (including the `network`/`wot` vocabulary aliases), indexed with the guard-narrowed reach string. The `as ProfileTagFeedReach` cast and the hand-maintained enum union are gone, so adding a reach to `PROFILE_TAG_SUPPORTED_REACHES` without a depth decision now genuinely fails compilation, matching the PR description. String keys also remove the module-scope enum access concern for partially mocked suites.

Both depth maps now carry cross-referencing comments per the minor observation. The specs-vs-local-strings vocabulary debt stays out of scope as noted.

PR #2239, opened 2026-07-17
fix(posts): redirect collections from post routes

### LINKED ISSUE #2168: [Bug]: Can open Collection in post page by editing url
### Description

Navigating to a collection using the post url opens the collection in post page. There's no use for this in the app because it cannot be navigated to via the UI. (Although it does allow you to have a secret reply thread)

### Expected Behaviour

Redirect from `/post/*` to `/collection/*` when kind is collection.

### Steps to Reproduce

1. Find a collection
2. Change it's url from `/collection/*` to `/post/*`
3. Observe the collection loads in post page

### Screenshots

https://github.com/user-attachments/assets/78521fd7-cbf1-45a9-bb8c-87fa4684cdd2

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
fix(posts): redirect collection posts to canonical collection routes

Permanently redirect collection-kind posts opened through /post URLs
to their canonical /collections URLs. Preserve normal post rendering
when the post is not a collection or the metadata lookup fails.

Resolves #2168


Nirzhuk:
I dont like quite a lot the flash we do in post page then we redirect (look in my video), I would like to have a better solution like making in the middleware so this way even before we see anything we will get redirect by the server and we never have this loading post page.

I know this effort could be much harder because it's more a architectural and structural effort but  gives a much nicer idea and we're making the same type of request anyway so.
https://github.com/user-attachments/assets/27a20190-b1f5-4906-bcdf-f3ccc7d8a08e

infin1t3:
> I dont like quite a lot the flash we do in post page then we redirect (look in my video), I would like to have a better solution like making in the middleware so this way even before we see anything we will get redirect by the server and we never have this loading post page.
> 
> I know this effort could be much harder because it's more a architectural and structural effort but gives a much nicer idea and we're making the same type of request anyway so. https://github.com/user-attachments/assets/27a20190-b1f5-4906-bcdf-f3ccc7d8a08e

@aldertnl and @catch-21 are ok with this.

src/app/post/[userId]/[postId]/page.tsx
```diff
@@ -73,6 +79,19 @@ export async function generateMetadata({ params }: PostPageProps): Promise<NextM
...
+  try {
+    const result = await fetchUserAndPostForMetadata(userId, postId);
+    isCollection = result?.post.kind === 'collection';
+  } catch {
+    // Preserve the existing post page when the kind lookup fails.
+  }
+
+  if (isCollection) {
+    permanentRedirect(`/collections/${userId}/${postId}`);
+  }
```
Comment (Nirzhuk) [inline src/app/post/[userId]/[postId]/page.tsx:93]: To make that explicit and remove any doubt, we can wrap `fetchUserAndPostForMetadata` in React.cache() so the helper itself runs once per request so this way we know it's cache and we're not making ALL posts pages slowers due being duplicated from generateMetadata?

I have another concern: `result` brings `user` from `fetchUserAndPostForMetadata` which we're not using it at all in this case, why not separate them? The cache is going to be there anyway.

Nirzhuk (approved):

src/app/post/[userId]/[postId]/page.tsx
Comment (infin1t3) [reply src/app/post/[userId]/[postId]/page.tsx:93]: React cache is redundant. Next.js automatically memoizes identical fetch calls across generateMetadata and the Page during the same request. React’s cache is primarily recommended when the underlying operation is not fetch. Next.js documentation
https://nextjs.org/docs/app/api-reference/functions/generate-metadata

PR #2240, opened 2026-07-17
feat: add years support to useRelativeTime, fix tests and weird cases with weeks

### LINKED ISSUE #2238: [Feature]:  Support years in useRelativeTime
Follow up from PR #2226 & issue #2184

### Problem Statement

- Add support for years
- Fix for 
`P1 Eight Weeks Becomes One Month
At 56 days, diffWeeks reaches 8 and skips the weeks branch, while diffMonths is still Math.floor(56 / 30) === 1. A timestamp therefore changes from 7w at 55 days to 1M at 56 days, under-reporting posts that are nearly two months old.`

This issue and PR is going to fix the date labels strings in notificatio nlist.

### Proposed Solution

More tests & support for years

### Suggested Priority

None

Nirzhuk:
resolves #2238 

- We add years support to useRelativeTime (Quite hard to test in staging due not having super old posts)
- usage of useRelativeTime in notifications profile list
- Fix weird cases for weeks

## Demo Profile Notification List
Tooltip included
Screenshot 2026-07-21 at 11 04 04
General
Screenshot 2026-07-21 at 11 04 00


Nirzhuk:

aldertnl:
LGTM

src/components/organisms/NotificationItem/NotificationItem.tsx
```diff
@@ -116,9 +119,15 @@ export function NotificationItem({ notification, isUnread }: NotificationItemPro
...
-  // Format timestamps (short for mobile, long for desktop)
-  const timestampShort = formatNotificationTime(notification.timestamp, false);
-  const timestampLong = formatNotificationTime(notification.timestamp, true);
+  const timestampDate = new Date(notification.timestamp);
+  const timestampElement = (
+    <RelativeTimestamp
+      timeAgo={formatRelativeTime(timestampDate)}
+      date={timestampDate}
+      as="p"
+      className="text-xs font-medium tracking-[1.2px] text-muted-foreground uppercase"
```
infin1t3: After your changes this `uppercase` is problematic.

It can make `2m` (2 minutes) look like `2M` (2 months).

Before this PR, notifications did not display months as units they used raw day counts.

src/components/molecules/RelativeTimestamp/RelativeTimestamp.tsx
```diff
@@ -0,0 +1,60 @@
...
+      </Typography>
+    </>
+  );
+
+  const showTooltip = !isMobile && !!date;
+  if (!showTooltip) {
+    return label;
+  }
+
+  const exactTimeLabel = date.toLocaleString(undefined, {
```
infin1t3, line 44: Why are you passing `undefined` instead of using real application locale? You could also just hardcode `en`, given we wanna remove other languages soon.

src/components/organisms/NotificationItem/NotificationItem.tsx
```diff
@@ -116,9 +119,15 @@ export function NotificationItem({ notification, isUnread }: NotificationItemPro
   // Get post preview text
   const previewText = hasPostPreview(notification.type, postKind) ? formatPreviewText(postContent) : null;
 
-  // Format timestamps (short for mobile, long for desktop)
-  const timestampShort = formatNotificationTime(notification.timestamp, false);
-  const timestampLong = formatNotificationTime(notification.timestamp, true);
+  const timestampDate = new Date(notification.timestamp);
+  const timestampElement = (
+    <RelativeTimestamp
```
infin1t3, line 124: Every notification renders `RelativeTimestamp`, which calls `useIsMobile`. That hook attaches a separate window resize listener for every row.

So if someone loads 90 notifications, the page has roughly 90 listeners all reacting when the window is resized or the device rotates. It probably won’t break anything immediately, but it’s unnecessary work and could make a large notification list feel sluggish.

It would be cleaner to detect the viewport once in the notification list and pass the result down, or use CSS/media queries for the tooltip behavior.

src/components/organisms/NotificationItem/NotificationItem.tsx
```diff
@@ -116,9 +119,15 @@ export function NotificationItem({ notification, isUnread }: NotificationItemPro
...
+      timeAgo={formatRelativeTime(timestampDate)}
+      date={timestampDate}
+      as="p"
```
infin1t3: Minor polish - not blocking.

Notifications pass `as="p"` to `RelativeTimestamp`, but the tooltip wraps that paragraph in a `<span>` The result is:
```
<span>
  <p>2m</p>
</span>
```

Which is invalid HTML. You can just render the notification timestamp as a `<span>` instead of a `<p>`

infin1t3 (changes requested):

src/components/molecules/RelativeTimestamp/RelativeTimestamp.tsx
Nirzhuk: Because when we use undefined we're going to use the one default from the system, so the user will see the one they use in they're computer, it's a good thing actually.

This is related to the formatting for example US is `MONTH/DAY/YEAR` but in mostly europe is `DAY/MONTH/YEAR`.
Dates needs to be formatted for the user, not the language they use, because my formatting is different than english but I dont share the same format.

src/components/organisms/NotificationItem/NotificationItem.tsx
Nirzhuk: removed as from `RelativeTimestamp` no need to be as anymore

src/components/molecules/PostHeaderTimestamp/PostHeaderTimestamp.tsx
```diff
@@ -1,51 +1,30 @@
...
-        </TooltipTrigger>
-        <TooltipPortal>
-          <TooltipContent>{exactTimeLabel}</TooltipContent>
-        </TooltipPortal>
-      </Tooltip>
+      <RelativeTimestamp
+        timeAgo={timeAgo}
+        date={indexedAt}
+        isMobile={isMobile}
+        className="text-xs leading-4 font-medium tracking-[0.075rem] whitespace-nowrap text-muted-foreground"
```
infin1t3: Minor - I think you can use `tracking-widest` here instead of `tracking-[0.075rem]`.

src/hooks/useRelativeTime/useRelativeTime.ts
```diff
@@ -24,14 +24,20 @@ export function useRelativeTime() {
...
     if (diffMins < 60) return t('minutesShort', { count: diffMins });
     if (diffHours < 24) return t('hoursShort', { count: diffHours });
     if (diffDays < 7) return t('daysShort', { count: diffDays });
     if (diffWeeks < 8) return t('weeksShort', { count: diffWeeks });
-    return t('monthsShort', { count: diffMonths });
+    // ponytail: switch to real-day years here (not diffMonths/12) so a year stays 365 days, not 336 (12 * 28-day "weeks/4" months)
+    if (diffDays < 365) return t('monthsShort', { count: Math.min(diffMonths, 11) });
+
+    const diffYears = Math.floor(diffDays / 365);
+    const remainderMonths = Math.min(Math.floor((diffDays % 365) / 30), 11);
```
infin1t3: Small consistency question: before one year, months are calculated from `diffWeeks / 4`, so one month represents 28 days. Here, the remaining months are calculated using 30 days. This means 56 days displays `2M`, but `1Y 28d` still displays only `1Y`. 

Let's we use the same month approximation in both branches, or document and test this difference if it’s intentional?

src/components/organisms/NotificationItem/NotificationItem.tsx
```diff
@@ -116,9 +119,15 @@ export function NotificationItem({ notification, isUnread }: NotificationItemPro
...
-  // Format timestamps (short for mobile, long for desktop)
-  const timestampShort = formatNotificationTime(notification.timestamp, false);
-  const timestampLong = formatNotificationTime(notification.timestamp, true);
+  const timestampDate = new Date(notification.timestamp);
+  const timestampElement = (
+    <RelativeTimestamp
+      timeAgo={formatRelativeTime(timestampDate)}
+      date={timestampDate}
+      isMobile={isMobile}
+      className="text-xs font-medium tracking-[1.2px] text-muted-foreground"
```
infin1t3: You can also use `tracking-widest` here as well.

infin1t3 (changes requested):

infin1t3 (approved):
https://github.com/user-attachments/assets/2550ef88-6449-44c7-9c2a-811515700c93

LGTM

catch-21:
Screenshot 2026-08-06 at 10 04 50

I was expecting just `1Y` for up to `2Y`, not combination of year and month like `1Y 2M`, looks weird to me. None of the evidence in this PR's description or comments shows this so I want to check, @aldertnl You approved this?

aldertnl:
No, that is indeed not correct and I missed this edge case in review. It should not show a combination of year and month, just `1Y` for up to `2Y` in that particular case.

PR #2246, opened 2026-07-21
fix(onboarding): accept Argentina mobiles missing trunk 9

### LINKED ISSUE #2079: [Bug]: `Send Code` button remains disabled in Proof Of Phone page
### Description

While trying to setup my pubky account through the phone verification wizard, I could not finish the first step due to `Send Code` button disabled, even after typing my phone, neither I see any error message. I reveal my area code (+54) just in case it works for debugging

### Expected Behaviour

When I type a valid phone number in the input I should either see:
- `Send Code` button disabled
- An error indicating that there is something wrong with my phone number

### Steps to Reproduce

1. Go to pubky app onboarding (https://pubky.app/onboarding/human)
2. Pick "Receive SMS"
3. Type a valid phone number (+54 prefixed) -> "Send Code" button is disabled
4. If I change the +54 by +1 -> "Send Code" button is enabled.
5. If I delete the last digit keeping the +54 prefix -> "Send Code" button remains disabled 

Then the issue seems to be related with the prefix of the phone (area code)

### Screenshots

Image

### Console Log

```shell
No relevant logs
```

### Browser & Version

Brave 1.91.168

Orlandohub:
## Summary
- fix #2079
- Argentina mobiles entered without the international trunk `9` (e.g. `+54341…`) left Send Code silently disabled on Proof of Phone
- Normalize those numbers client-side so users can proceed, and show clear invalid feedback when a send is attempted with an invalid number

## Changes
- Insert Argentina mobile trunk `9` in `parsePhoneNumber` when `+54` input fails mobile validation but becomes valid with the trunk digit
- Send normalized E.164 from HumanPhoneInput to Homegate instead of the raw field value
- Keep Send Code enabled; show inline `invalidPhone` only after the user attempts to send with an invalid number (clears on edit)
- Associate the error with the input via `aria-describedby` for screen readers
- Add dedicated phone unit tests and organism coverage for AR-without-9 send path and submit-time validation UX

https://github.com/user-attachments/assets/ced3e2e5-412c-4112-9688-1009c6e47a03

https://github.com/user-attachments/assets/30cff5d9-9a65-436f-94f8-ec44ce6e9113

## Test plan
- [x] Go to `/onboarding/human` → Receive SMS
- [x] Enter an Argentina number without trunk 9 (e.g. `+543416620881`) → checkmark appears and Send Code works
- [x] Send code → request uses `+5493416620881` and OTP step shows the normalized number
- [x] Enter an invalid number → no error yet; Send Code stays enabled
- [x] Click Send Code with that invalid number → inline error appears; no Homegate request
- [x] Edit the number → error clears
- [x] Enter a valid `+1` / already-correct `+549…` number → still works as before
- [x] Unit tests: `src/libs/phone/phone.test.ts`, HumanPhoneInput / HumanPhoneInputField tests

## Notes
- Country-code selector is out of scope (Slack follow-up); this only handles the AR trunk-9 exception
- Landlines remain rejected (SMS requires mobile metadata validation)


Orlandohub:
@greptile-apps review

catch-21:
@MCarlomagno has kindly confirmed this works end-to-end

catch-21:
LGTM. A couple of notes to highlight:
- this fix applies to all `+54`, which includes both mobile AND landline. The insertion of `9` is only valid for mobile. Therefore, if someone wanted to use a landline for this it would likely fail at the homegate/prelude stage because adding `9` is unexpected for the carrier. Probably not likely to be a problem because most use mobile for SMS/whatsapp.
- the 'Send code' button is always enabled instead of only becoming enabled when a valid phone number has been inputted. The new error is clearer.

catch-21 (approved):

PR #2248, opened 2026-07-22
feat(collections): add creator-selected list layout

### LINKED ISSUE #2208: [UI]: List Layout for Collections page (v25)
v25 design shows List Layout on Collections page.

https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=35688-285472&t=uzj6rQ1I7WHCTbfB-1

Image

We also need to add the Layout option when creating/editing a Collection if it doesn't exist already: https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=40341-335749&t=uzj6rQ1I7WHCTbfB-1

infin1t3:
## Summary

- Add creator-selected Grid/List defaults to collection create and edit flows, including a clearer **Default layout** label in the edit dialog and Grid fallback for legacy collections without a stored layout.
- Persist the typed collection layout end to end through forms, controllers, pipes, models, save-target updates, and collection membership changes without dropping existing collection fields.
- Add a temporary Grid/List override for non-owner desktop viewers while keeping the creator default unchanged; hide that override for owners and on mobile.
- Render collection List mode through the shared timeline/feed layout pipeline, with collection-scoped layout resolution and consistent post/tag presentation.
- Keep collection views focused on the curated posts by hiding inline replies, while still showing replies on the dedicated post page.
- Stabilize loading so a List collection does not flash Grid skeletons before switching layouts; the collection envelope resolves the requested layout before the feed mounts.
- Restore collection-specific empty states and Add Post affordances in both layouts, including the full-width final Add Post row in List mode and the regular hero action.
- Open collection posts with clean URLs and the existing global Home layout. Post content, shell, and sidebar share that preference; changing layout updates the Home preference, while Back returns directly to the collection.
- Keep post composer dialogs on their intended inline layout instead of inheriting collection/List presentation.
- Upgrade `pubky-app-specs` to `0.6.2` for typed collection layout and collection-post support, while keeping unsupported WoT/Me feed reaches explicitly rejected.
- Expand translations plus unit, snapshot, and Cypress regression coverage for persisted defaults, viewer overrides, stable loading, hidden collection replies, Add Post behavior, Home-layout post navigation, and single-step Back navigation.

Visual collection layout remains out of scope and will be handled separately.

## Verification

- `npm run lint`
- `npm run typecheck`
- `npx tsc --noEmit -p cypress/tsconfig.json`
- `npm test` — 697 test files passed; 11,316 tests passed and 2 skipped
- Added Cypress coverage for creator defaults, temporary collection overrides, Home-layout post navigation from List collections, layout preference changes, and single-step Back navigation

Closes #2208

https://github.com/user-attachments/assets/0f368619-352e-4ec8-9241-c5681b05fa2b


aldertnl (dismissed):
**Looking good. A few small remarks:**

1. Perhaps we should hide the layout override dropdown for the creator in the collection header, since it may be confusing. As it stands, the creator may think the layout type can also be changed and saved there (instead, it's set in the edit dialog).

image

2. Perhaps in the edit dialog we should change the label to "DEFAULT LAYOUT" to imply ability to override elsewhere.
image

3. Clicking a post to view it in full, on its dedicated page, should open in List layout. Then the user stays in list view when navigating collections. Currently, it switches to Column layout.

image

4. Layout dropdown override should be hidden on mobile 
image

aldertnl:
One more request, can we move the layout option above the background option in the dialog? 

image

Nirzhuk:
Screenshot 2026-07-23 at 11 56 29
Seems this could be a bug, we want to show the replies of posts in the list layout in collections? I feel Collections we want to save the main post but we dont care that much about the replies, We could confirm this desired behavior?  




https://github.com/user-attachments/assets/d129807b-32cd-40d2-9085-d0350130ea86
This video shows me doing a very fasts reload to show we have 2 skeletons modes
This collection is on list layout mode but we see first the grid layout then we have the list layout.
Grid layout skeleton -> List Layout skeleton -> Loaded content.

We might need to choose one or another and not change the skeletons in middle of the loading state.

src/libs/post/collectionContent.ts
```diff
@@ -17,6 +18,7 @@ import { FileVariant } from '@/services/nexus/file/file.types';
  *   - `cover_image: string | undefined`
  */
 export type { PubkyAppCollectionContent } from 'pubky-app-specs';
```
Nirzhuk:
```suggestion
export { type PubkyAppCollectionContent };
```
to prevent 2 froms from app-specs from the same thing

src/hooks/useCreateCollection/useCreateCollection.types.ts
```diff
@@ -1,9 +1,17 @@
...
+  NAME: 'name';
+  DESCRIPTION: 'description';
+  LAYOUT: 'layout';
+};
+
+export const CREATE_COLLECTION_FORM_FIELDS: CreateCollectionFormFields = {
   NAME: 'name',
   DESCRIPTION: 'description',
+  LAYOUT: 'layout',
 };
```
Nirzhuk:
```suggestion
export const CREATE_COLLECTION_FORM_FIELDS = {
  NAME: 'name',
  DESCRIPTION: 'description',
  LAYOUT: 'layout',
} as const;
```

src/core/models/post/collection/collectionPost.types.ts
```diff
@@ -1,18 +1,23 @@
...
+export type CollectionContent = PubkyAppCollectionContent & { layout: CollectionLayout };
+
 export type CollectionContentInput = {
   name: string;
   description?: string | null;
   items?: string[] | null;
   /** Optional cover image URL. Validated against the spec attachment URL rules. */
   coverImage?: string | null;
+  /** Creator-selected default layout. Missing values retain the legacy Grid behavior. */
+  layout?: CollectionLayout | null;
```
Nirzhuk: Why we add `null` and not just play around `undefined`? What brings us here null above `undefined`

package.json
```diff
@@ -62,7 +62,7 @@
     "next": "16.2.6",
     "next-intl": "4.12.0",
     "next-themes": "0.4.6",
-    "pubky-app-specs": "0.5.3",
+    "pubky-app-specs": "0.6.2",
```
Nirzhuk, line 65: I feel updating this brings things from WoT but at the same time we dont need them, theres no other version to avoid things from WoT?

src/config/collections.ts
```diff
@@ -1,10 +1,28 @@
...
 
+export type CollectionLayout = Extract<PubkyAppCollectionLayout, 'grid' | 'list'>;
+
+export const COLLECTION_LAYOUT: {
+  GRID: CollectionLayout;
+  LIST: CollectionLayout;
+} = {
+  GRID: 'grid',
+  LIST: 'list',
+};
```
Nirzhuk:
```suggestion
export const COLLECTION_LAYOUT: Record<'GRID' | 'LIST', CollectionLayout> = {
  GRID: 'grid',
  LIST: 'list',
};
```

src/config/collections.ts
```diff
@@ -1,10 +1,28 @@
+import type { PubkyAppCollectionLayout } from 'pubky-app-specs';
+
 /**
  * Collections feature configuration.
  *
  * Section page sizes / avatar caps for the `/collections` landing and the
  * `/collections/[id]` bottom sections.
  */
 
+export type CollectionLayout = Extract<PubkyAppCollectionLayout, 'grid' | 'list'>;
```
Nirzhuk, line 12: I might add a comment adding the reason of extracting.


nit I prefer exclude

src/components/organisms/SinglePostContent/SinglePostContent.tsx
```diff
@@ -26,7 +27,9 @@ import type { SinglePostContentProps } from './SinglePostContent.types';
  */
 export function SinglePostContent({ postId, postDetails }: SinglePostContentProps) {
   const t = useTranslations('post');
-  const layout = useHomeStore((state) => state.layout);
+  const homeLayout = useHomeStore((state) => state.layout);
+  const routeLayout = usePostRouteLayout();
+  const layout = routeLayout ?? homeLayout;
```
Nirzhuk: Image

This forces and we remove any interaction, My recommendation we need a helper to play with searchParams & internal state so add something like `nuqs` so we have a default value (searchParam) and we're available to be free to change it after that.

Video:
https://github.com/user-attachments/assets/a33b452b-f72f-4a22-ba49-0daa9c219f85

Nirzhuk: My main concern with this right now is we have 2 source of truths and they're not friendly between eachother, we need to unify them if not we will have bugs (no regressions) related to the URLs,

src/components/organisms/DialogReply/DialogReply.tsx
```diff
@@ -18,45 +19,47 @@ export function DialogReply({ postId, open, onOpenChangeAction }: DialogReplyPro
...
-        hiddenTitle={t('hiddenTitle')}
-      >
-        <DialogHeader>
-          <DialogTitle>{t('title')}</DialogTitle>
-          <DialogDescription className="sr-only">{t('description')}</DialogDescription>
-        </DialogHeader>
-        <Container className="min-h-0 flex-1 gap-3 overflow-x-hidden overscroll-contain pr-1">
-          {/* Post being replied to */}
-          <PostPreviewCard postId={postId} />
+    <PostMainLayoutProvider tagsLayout="inline">
```
Nirzhuk: P0 - Now we need to use `PostMainLayoutContext` everywhere and it's maybe were not going to use it properly just to prevent errors, this is the perfect case for: use zustand as general store, we dont need to create 1 provider per dialog and most important and there's no set inside this components, we only use it to not prop drill.

src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
```diff
@@ -315,6 +332,7 @@ function CollectionHeroContent({ authorPubky, compositeId, postDetails, classNam
               {tagToggle}
             </>
           )}
+          {!isOwn && <CollectionLayoutPicker layout={layout} onLayoutChange={onLayoutChange} />}
```
Nirzhuk, line 336: - The layout picker is unreachable for owners and on mobile
  - CollectionHero.tsx:332: {!isOwn && <CollectionLayoutPicker .../>} — the creator can never preview their own collection
  in the other layout without opening the edit dialog. The e2e asserts this (collection-layout-menu should not exist), so it
  looks deliberate; confirm with design, because it reads as an oversight.

src/components/organisms/Collections/CollectionItems/CollectionItems.tsx
Nirzhuk, line 1: ### List layout silently drops the empty state and the "Add Post" tile


How to Reproduce 
  1. npm run dev, sign in, go to /collections
  2. New collection → set Layout = List → save
  3. You land on the empty /collections/<you>/<id>

  Expected: "This collection is empty." (collections.single.empty, data-cy="collection-items-empty")
  Actual: "No posts found" (empty.noPosts, generic timeline placeholder), and no dashed Add-Post tile.

Image

Image

Nirzhuk (changes requested):

package.json
infin1t3: Unfortunately no.

I checked the published versions: 0.6.0 has neither WoT/Me nor collection layout, 0.6.1 introduces WoT/Me but still has no collection layout, and 0.6.2 is the first published version containing collection layout support. So there is currently no published version with collection layout support but without the WoT types.

The app does not enable WoT or Me feeds. The related changes only keep our enum handling exhaustive and explicitly reject those reaches in buildFeedStreamId until the Nexus stream implementation supports them. We need 0.6.2 specifically for the collection layout field and createCollectionPost support.

src/libs/post/collectionContent.ts
infin1t3: I ended up deleting this export. The original purpose is to have this file as a facade but on second look we don't really need it.

src/hooks/useCreateCollection/useCreateCollection.types.ts
infin1t3: Prefer a more explicit types, but I agree this is less code and does the same thing - accepted.

src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
infin1t3: This is deliberate and already confirmed with the design.
Owners already have a way to change the collection layout - this is for non-owners and is supposed to be a temporary switch not permanent.

On mobile we have only one layout so this is not needed.

src/components/organisms/SinglePostContent/SinglePostContent.tsx
infin1t3: I implemented something in-between that seemed like the best option at the time. Let me know what you think.

Also I think we will be returning to this part in a follow-up PRs after product feedback we got in Slack.

infin1t3:
@greptileai review

src/hooks/usePostRouteLayout/usePostRouteLayout.ts
```diff
@@ -1,15 +1,46 @@
...
+}
+
 /**
- * Returns the supported temporary layout override carried by a post URL.
- * Other values deliberately fall back to the persisted home preference.
+ * Resolves the post layout through one interface. A valid route layout keeps
+ * changes temporary in the current URL; otherwise changes use the home store.
  */
-export function usePostRouteLayout(): LayoutType | undefined {
+export function usePostRouteLayout(): PostRouteLayout {
```
Nirzhuk: 👍

Nirzhuk (changes requested):
https://github.com/user-attachments/assets/18c64de8-27b2-40ed-9e37-6e1051ac6307

Heres a video with a huge bug related layouts
How to reproduce it: home feed → List → open any post → change layout in the left sidebar → go back → preference unchanged. 

Lets fix it first before I keep my code review because I believe this will change a lot the code again

src/libs/post/collectionContent.ts
Nirzhuk: This was not deleted at the end

infin1t3: Merge conflict resolution removed it. Fixed.

src/components/organisms/Collections/CollectionLayoutPicker/CollectionLayoutPicker.tsx
```diff
@@ -0,0 +1,125 @@
...
+  );
+}
+
+function CollectionLayoutPickerContent({
+  layout,
+  onSelect,
+}: {
+  layout: CollectionLayout;
+  onSelect: (layout: CollectionLayout) => void;
+}) {
```
Nirzhuk: Create a interface for this please, the others have one.

Consider create a type file

src/components/organisms/Collections/DialogAddContent/DialogAddContent.tsx
```diff
@@ -79,6 +79,44 @@ const DialogAddContentGridTrigger = forwardRef<ComponentRef<typeof Button>, Dial
...
+        {...props}
+      >
+        <Plus className="size-3 shrink-0" />
+        <Typography as="span" overrideDefaults className="text-sm font-bold">
+          {t('addContent')}
+        </Typography>
+      </Button>
+    );
+  },
+);
```
Nirzhuk, line 103: Why you're using `forwardRef` I dont see where we use that ref anywhere

src/components/organisms/Collections/DialogCollectionForm/DialogCollectionForm.tsx
```diff
@@ -152,6 +157,51 @@ export function DialogCollectionForm({
...
+                <Container
+                  overrideDefaults
+                  role="radiogroup"
+                  aria-labelledby="collection-layout-label"
+                  className="flex w-full border-b border-border"
+                >
+                  {[
+                    { value: COLLECTION_LAYOUT.GRID, label: t('layoutGrid'), icon: Grip },
+                    { value: COLLECTION_LAYOUT.LIST, label: t('layoutList'), icon: Rows4 },
+                  ].map(({ value, label, icon: Icon }) => {
```
Nirzhuk: Why no create a proper constant inside the component, if this file is react compile friendly will be auto memotized

Nirzhuk (changes requested):

src/components/organisms/Collections/DialogAddContent/DialogAddContent.tsx
infin1t3: Shadcn/Radix requirement.

These trigger components are rendered through <DialogTrigger asChild>.

Radix attaches its own ref for focus management, so custom child components must spread the incoming props and forward the ref to the underlying button. The ref is used indirectly by Radix rather than directly inside our component.

src/components/organisms/Collections/CollectionLayoutPicker/CollectionLayoutPicker.tsx
infin1t3: I will make an interface in this component. 

Prefer creating separate `.types.ts` files when an interface has to be imported in multiple places.

src/components/organisms/Collections/DialogCollectionForm/DialogCollectionForm.tsx
infin1t3: Makes sense. Done.

Nirzhuk (approved):

PR #2249, opened 2026-07-22
feat(posts): remove deleted and missing posts from saved feeds

### LINKED ISSUE #2102: [UI]: Option to remove deleted post from Collection
We need a way for the creator of a Collection to be able to delete a post from their Collection that has been deleted by its author.

The suggestion is to show a "❌ Remove" button in the card that says "This post has been deleted by its author." The button can be positioned centered below the label.

Image

Image

infin1t3:
## Summary

- Add a centered **Remove** action on unavailable post cards (author-deleted *and* not-found) in **owned collections** and **bookmarks** — available in every layout: grid, list, and visual mosaic
- Gate the action on collection ownership / authenticated bookmark access (`useRemoveDeletedPost`)
- Remove cards optimistically with explicit commit/rollback that keeps ordering stable across refreshes, concurrent removals, and stream changes
- Keep collection pagination offsets correct while removals are pending **and after they commit** (the consumed skip offset shrinks with the backend collection, so load-more never skips a live post)
- Make bookmark writes local-first (IndexedDB before homeserver sync); when the local write commits but sync fails, the card stays removed and a warning toast reports the failed sync instead of a contradictory error
- Move keyboard focus to the adjacent feed card before unmount (`focusAdjacentGridItem`) — scans past non-focusable siblings, never targets disabled/hidden elements, and works in the visual mosaic via `data-grid-item` cells
- Unify `PostDeleted` / `PostMissing` into one `PostUnavailable` molecule (shared UI, distinct copy + `data-cy`)

## Test plan

- [x] Own a collection with a **deleted** item → Remove clears it and toast confirms (grid, list, and visual layouts)
- [ ] Own a collection with a **Post not found** item → same Remove flow
- [ ] Bookmarks with deleted / missing items → Remove unbookmarks
- [ ] Non-owner viewing someone else's collection → no Remove button
- [ ] Failed network remove → card restores; if the local write already committed, card stays gone with a warning toast
- [ ] Keyboard: after Remove, focus lands on the next/previous card
- [ ] Pagination/load-more still correct after several removes — including after they commit and the collection shrinks

## Verification

- `npm run test` — 702 files; 11,484 passed, 2 skipped
- `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run build` — all green

https://github.com/user-attachments/assets/a57bc7c7-8454-4659-95e6-f487b0e28375

Closes #2102


src/hooks/useStreamPagination/useStreamPagination.ts
```diff
@@ -243,17 +326,77 @@ export function useStreamPagination({
...
+          // removal deletes one of those rows server-side, shifting every
+          // later index down — so drop the removed stream rows from the
+          // offset too, or the next loadMore skips one live row per removal
+          // once the backend reindexes the shorter list. Membership is
+          // checked at commit time: rows refetched after a refresh are
+          // already recounted in the new offset, and optimistic prepends
+          // never counted toward it.
+          const removedStreamRowCount = postIdsRef.current.filter((id) => idsToRemoveSet.has(id)).length;
+          if (removedStreamRowCount > 0) {
+            setStreamTail((tail) => Math.max(0, tail - removedStreamRowCount));
```
Orlandohub, line 406: **the committed-removal offset can be silently dropped**

`finalize` applies the correction as a functional update (`setStreamTail(tail => tail - n)`), but an in-flight `fetchStreamSlice` later overwrites it with the absolute `setStreamTail(result.nextCursor)` at line 135. `advanceCursor` derives `nextCursor` as `prevCursor + rawPage.ids.length` from the offset captured when that request started, so any page fetch resolving after a commit throws the decrement away and because the write is absolute, it discards *every* decrement accumulated during its flight, not just one.

Scope is narrower than it first looks: only `collection:…` streams are skip-paginated, so bookmark removals never reach this. It also needs `hasMore` plus a Nexus reindex before the next page.

With 10-post pages against a 3000px `rootMargin`, the collection feed chain-prefetches, so a fetch is in flight a good fraction of the time while scrolling, which is exactly the bulk-cleanup flow this PR targets. A stationary click is usually safe (the homeserver write outlasts the fetch); scrolling through the write is not. It isn't durable either — a refresh resets the offset to 0.

Still worth fixing, because it inverts the failure mode: an offset that's too low refetches a row that gets deduped away (invisible), while one that's too high skips a live row with no signal at all.

```ts
// finalize() increments committedRemovalOffsetRef instead of calling setStreamTail
if (result.nextCursor != null) {
  setStreamTail(Math.max(0, result.nextCursor - committedRemovalOffsetRef.current));
  committedRemovalOffsetRef.current = 0;
}
```

Regression test: hold a `loadMore` on an unresolved promise, commit a removal of a consumed row, resolve with `nextCursor: 5`, assert the next request uses `streamTail: 4`.

src/hooks/useRemoveDeletedPost/useRemoveDeletedPost.ts
```diff
@@ -0,0 +1,144 @@
...
+    const { pubky } = parseCompositeId(feed.collectionId);
+    return pubky === currentUserPubky ? { type: 'collection', collectionId: feed.collectionId } : null;
+  } catch {
+    return null;
+  }
+}
+
+async function isPostStillInCollection(collectionId: string, postId: string): Promise<boolean | null> {
+  const collection = await PostController.getDetails({ compositeId: collectionId });
+  if (!collection || isPostDeleted(collection.content)) return false;
```
Orlandohub: **What's wrong:** The error-recovery check draws the opposite conclusion from the error that triggered it. `PostController.commitUpdateCollectionItem` throws `Err.client(NOT_FOUND)` when `!collection || isPostDeleted(collection.content)` `(src/core/controllers/post/post.ts:434-440)`.

The catch handler then calls isPostStillInCollection, which returns false for those same two conditions — and false means "already gone", so `shouldRestore` becomes false, `optimisticRemoval.commit()` runs, and the user is told `postSave.removedFromCollectionSyncFailed` ("Removed on this device, but syncing failed"). Nothing was written locally or remotely.

**Why it matters:** The card and the toast both assert a removal that never happened, and commit() additionally decrements the pagination offset for a row the backend still has. The bookmarks branch doesn't have this problem `BookmarkController.exists` genuinely distinguishes "gone" from "unverifiable".

**Fix**: Return `null` (unverifiable) instead of `false` when the collection row is missing or tombstoned, so the existing `if (isStillInCollection !== null)` guard leaves `shouldRestore = true`. Alternatively check `error.code === ClientErrorCode.NOT_FOUND` before consulting local state, which also matches the project's "check code, not message" convention.

src/core/application/bookmark/bookmark.ts
```diff
@@ -49,10 +49,7 @@ export class BookmarkApplication {
 
     const { postId, bookmarkUrl, bookmarkJson } = params;
 
-    // Execute local and homeserver operations in parallel
-    await Promise.all([
-      LocalBookmarkService.persist(action, { userId, postId }),
-      HomeserverService.request({ method: action, url: bookmarkUrl, bodyJson: bookmarkJson }),
-    ]);
+    await LocalBookmarkService.persist(action, { userId, postId });
+    await HomeserverService.request({ method: action, url: bookmarkUrl, bodyJson: bookmarkJson });
```
Orlandohub, line 56: **What's wrong**: The new doc comment promises "Homeserver failures leave the optimistic local state available for caller reconciliation", but `persist` does no compensating write and only one of its four callers reconciles. `useRemoveDeletedPost` re-checks `BookmarkController.exists`; `useBookmark.toggle` (`src/hooks/useBookmark/useBookmark.ts:105-122`), `useAddContentForm`, and `useSaveCreatedPostToTarget` don't. In `useBookmark`, a homeserver failure skips `setIsBookmarked` and toasts "Could not remove bookmark" while IndexedDB has already committed the delete and dropped the post from `TIMELINE_BOOKMARKS_ALL`.

**Why it matters**: The user sees a failure toast and an unchanged bookmark icon while the bookmarks page already shows the item gone. This was reachable before (the `Promise.all` local write still landed when the network leg rejected), but sequencing makes it deterministic, and the comment now presents it as an intentional contract most callers don't honor. Note the sibling collection path *does* compensate — `PostApplication.commitEdit` (`src/core/application/post/post.ts:317-329`) restores the previous local content on PUT failure.

**Fix**: Either mirror `commitEdit` and roll the local write back on homeserver failure, or narrow the comment to name which callers implement reconciliation, and follow up in `useBookmark.toggle` by re-reading `BookmarkController.exists(postId)` in its catch block.

Orlandohub (changes requested):

Orlandohub (approved):
https://github.com/user-attachments/assets/e853a035-fa4b-4662-b07b-bc4e4a4a1b73

PR #2250, opened 2026-07-22
feat(ui): update auth brand illustrations to Pubky v2 (#1902)

### LINKED ISSUE #1902: [UI]: Update brand images in sign up flow (v23)
Brand visuals: Replaced all visuals with new Pubky brand illustrative graphics (please check all frames)

Image

1 of 6 frames: https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=29153-80489&t=0s74QURDY5s8upgv-1

Orlandohub:
## Summary
- fix #1902
- Swap signup/auth/onboarding brand plates to the new Illustrations Pubky v2 graphics from Handoff v25
- Give enter-code its own padlock plate instead of reusing the phone graphic
- Migrate Join Pubky modal illustrations from SVG to WebP to match the raster handoff

## Changes
- Overwrite in-scope `public/images/*.webp` brand assets (SMS, Bitcoin, invite, key, shield, backup methods, check, signed out, keyring)
- Add `sms-verification-code.webp`, `new-here.webp`, and `sign-in.webp`; remove `new-here.svg` / `sign-in.svg`
- Point `HumanPhoneCode` at `/images/sms-verification-code.webp`
- Update `DialogSignIn` image paths and related tests/snapshot

## Test plan
- [x] Focused Vitest for Human*, PublicKeyCard, BackupMethodCard, DialogBackup, DialogSignIn, Logout, Install (105 passed)
- [ ] Manual smoke: `/onboarding/human` (verify + phone + code + invite), `/onboarding/pubky`, `/onboarding/backup`
- [ ] Manual smoke: `/sign-in`, `/logout`, Join dialog, backup method + complete dialogs
- [ ] Spot-check each mapped surface against Figma Handoff v25 (including SMS verify padlock plate)

## Notes
- Sign-in page keeps the live QR; no static phone+QR mock
- Most plates exported from Figma 4k fills; `note.webp` / `folder.webp` were cropped from the issue backup-methods screenshot when standalone fill nodes were not available in Handoff v25


Orlandohub:
@greptile-apps review

src/components/organisms/SignIn/SignIn.tsx
```diff
@@ -149,14 +152,28 @@ export const SignInContent = () => {
   }
   return (
     <>
-      {/** Desktop view */}
+      {/** Desktop view — Figma: scan plate + live QR + spacer, centered in card */}
```
Nirzhuk: We could not modify the comment, seems silly but I feel this output is too much AI, doesn't provide anything to us

src/components/molecules/QrCodeSlot/QrCodeSlot.tsx
```diff
@@ -6,8 +6,8 @@ import { Typography } from '@/atoms/Typography/Typography';
 import { cn } from '@/libs/utils/utils';
 import type { QrCodeSlotProps } from './QrCodeSlot.types';
 
-const QR_SIZE = 220;
-const RING_LOGO_SIZE = 48;
+const DEFAULT_QR_SIZE = 220;
```
Nirzhuk: We only use this in `Scan` & `Signin`, why we dont simplify and move the default QR SIZE to 176 and remove the constant in the those files.

src/components/organisms/SignIn/SignIn.tsx
Nirzhuk, line 1: Substantially changes the desktop card, but SignIn.test.tsx has no snapshot coverage.

src/components/molecules/Logout/Logout.tsx
Nirzhuk, line 1: introduces a new layout without a snapshot

src/components/molecules/HeaderHome/HeaderHome.test.tsx
Nirzhuk, line 1: Does not snapshot the new logout-page state.

src/components/molecules/QrCodeSlot/QrCodeSlot.tsx
```diff
@@ -16,9 +16,12 @@ export function QrCodeSlot({
   url,
   generatingLabel,
   clickToReloadLabel,
+  size = DEFAULT_QR_SIZE,
```
Nirzhuk, line 19: adds a meaningful size variant without snapshotting a non-default size & adding tests

Nirzhuk (changes requested):
[P2] Centralize these layout overrides. This PR adds mx-0 43 times across 12 files. Twenty-one are no-ops because Container already supplies w-full, while the remaining 22 only compensate for its mx-auto default on fixed/max-width children. Please remove the redundant utilities and extract the repeated illustrated-card and QR-card layouts—or use overrideDefaults/plain wrappers internally—so alignment is defined once rather than duplicated across live and skeleton components.

I feel we're adding too much mx-0, you could take a look and take a review  please

aldertnl (changes requested):
Looks OK to me, except there seem to be two incorrect exports that use the wrong cropping and don’t include the transparent background.

Image

Image

src/components/organisms/SignIn/SignIn.tsx
Orlandohub: imo this is the wrong narrative, the more our code is read and written by AI, I don't see a valid argument on "this output is too much AI".

The fact that "doesn't provide anything to us" is **harmless in this context** but it provides precious context for whatever other models come in the future and need to apply changes.

I could agree with some contexts where AI slop with the wrong context can be harmful, but when the AI model leaves a pointer on where to find the source of truth for the content at hand, I find it valid!

Orlandohub: For extra context, here you have the author of Clean Code, a dev which became famous for advocating high quality code, now explaining that he barely looks at code and how code should be mainly for AI to read/write:

https://x.com/unclebobmartin/status/2080257779395154409?s=20

All this to say I don't think small comment lines which "feel like to much AI" should be taking our time and effort.

P.S.: I've seen this type of comments before and now it's the first time I've made an effort to push back, that's why it might feel a bit over the top, but it's just so you can understand my reasoning on this topic.

aldertnl (dismissed):

src/components/organisms/SignIn/SignIn.tsx
Orlandohub: Addressed in d8c5ef73. `SignIn.test.tsx` now snapshots the QR sign-in layout in `SignIn.test.tsx.snap`.

src/components/molecules/QrCodeSlot/QrCodeSlot.tsx
Orlandohub: Addressed in d8c5ef73. `QrCodeSlot` now owns the `176` default, and `Scan`/`SignIn` no longer define or pass duplicate QR-size constants.

src/components/molecules/Logout/Logout.tsx
Orlandohub: Addressed in d8c5ef73. `Logout.test.tsx` now snapshots the signed-out illustration layout in `Logout.test.tsx.snap`.

src/components/molecules/HeaderHome/HeaderHome.test.tsx
Orlandohub: Addressed in d8c5ef73. `HeaderHome.test.tsx` now includes a dedicated snapshot for `AUTH_ROUTES.LOGOUT`, covering the Learn, Explore, and Sign in state.

src/components/molecules/QrCodeSlot/QrCodeSlot.tsx
Orlandohub: Addressed in d8c5ef73. `QrCodeSlot.test.tsx` now has both behavior coverage and a snapshot for `size={220}`, including proportional ring-logo scaling.

Orlandohub:
Addressed @Nirzhuk’s review in d8c5ef73:

- Extracted the repeated illustrated-card layout into `IllustratedCard`, shared by the live onboarding cards and their skeletons.
- Extracted the centered desktop QR layout into `BalancedQrCard`, shared by `Scan` and `SignIn`.
- Kept `ContentCard` as the compatible public API while delegating its layout internally.
- Removed the PR’s redundant layout overrides; relative to `dev`, the PR now adds no `mx-0` utilities.
- Centralized the `176` QR default in `QrCodeSlot` and added the requested SignIn, Logout, logout-header, and non-default QR-size snapshot coverage.

src/components/molecules/BalancedQrCard/BalancedQrCard.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface BalancedQrCardProps extends Omit<ComponentProps<typeof Card>, 'children'> {
+  children: ReactNode;
+  illustration: ReactNode;
+}
+
+/**
+ * Keeps a QR code centered in the card by balancing the illustration with an
+ * equal-width spacer at desktop sizes.
+ */
+export function BalancedQrCard({ children, illustration, className, ...props }: BalancedQrCardProps) {
```
Nirzhuk, line 14: I believe balanced is not a very accurate name here, maybe CenteredQRCard to play around words we normally use in css properties?

src/components/molecules/HumanPhoneCodeInput/HumanPhoneCodeInput.tsx
```diff
@@ -92,7 +92,7 @@ export const HumanPhoneCodeInput = ({ value, onChange, onEnter = () => {} }: Hum
   );
 
   return (
-    <Container className={cn('mx-0 flex-row flex-wrap justify-start gap-3')} data-testid="code-input-container">
+    <div className="flex w-full flex-row flex-wrap justify-start gap-3" data-testid="code-input-container">
```
Nirzhuk, line 95: I dont know about moving away from Container, theres no a huge reason for it and I believe the thing we need it to do it's the of overrideDefaults, take a deeper look please I understand the reason of removing mx-0

Theres a lot more in this commit but I will not comment every Container changed

Use the display="flex"

Nirzhuk (changes requested):

src/components/molecules/HumanPhoneCodeInput/HumanPhoneCodeInput.tsx
Orlandohub: Thanks but I’d prefer to keep the plain wrappers here:

> I dont know about moving away from Container
- On your [earlier review](https://github.com/pubky/pubky-app/pull/2250#pullrequestreview-4771418578)  you (or your AI) explicitly suggested `overrideDefaults/plain wrappers internally`, so the current approach follows one of those options.
- “Shadcn First” doesn’t prohibit native `<div>` wrappers, and Shadcn has no generic flex-layout primitive.

> I believe the thing we need it to do it's the of overrideDefaults
- Although `Container` is widely used in this repository, with `overrideDefaults` it contributes no layout behavior, it effectively becomes another `<div>`. @infin1t3 has brought this up in a previous discussion on how we should reconsider our approach here since we're consistently overriding values which begs the question, are we on the right path?
- This isn’t a general move away from `Container`, we should use it where its defaults add value. For structural wrappers where every default must be disabled, the native element is clearer and avoids an unnecessary abstraction.

> Use the display="flex"
- `display="flex"` is ignored when `overrideDefaults` is enabled, so `flex` would still need to remain explicit in `className`.

Nirzhuk (approved):

Issue #2134, opened 2026-07-04
[Bug]: Feed breaks as many consecutive Collections are being created


Nirzhuk:
### Description

Infinite loading in feed glitches and dont get new posts.

### Expected Behaviour

Get normally the feed

### Steps to Reproduce

I believe is due slow internet so get a throtled internet with 3g and start slow scroll down the feed.

### Screenshots

https://github.com/user-attachments/assets/c2014a3a-5244-47e6-b259-7bb0a29cf021

### Console Log

```shell
forward-logs-shared.ts:95 [WARN] StreamId not found in cache
forward-logs-shared.ts:95 [WARN] StreamId not found in cache
```

catch-21:
This is most likely caused by too many collections being created at a similar time because pubky-app is filtering those out from the 'All' content feed.

catch-21:
I think from 30 consecutive collections a degradation begins. "No posts found" seen in feed on sign in or feed reload after time spent away, but it can sort itself out after some navigating. I believe this is because querying is done in groups of 10 up to 20 iterations of batches. The greater number of filtered out posts, the higher latency and chances of seeing no older posts and flickering skeletons due to more 'wasted' fetches. As approaching 200 the degradation gets extreme with total failure of the global feed at 200.

Issue #2200, opened 2026-07-15
[Bug]: 'Show more' button for Discover section of Collections can take multiple clicks to show more


catch-21:
### Description

Depending on how many collections are being filtered out by the Discover section, the 'Show more' button can take multiple clicks to show more collections.

It currently filters out: your own collections + followed collections + collections with 0 posts + deleted collections
This means that a large combination of these could cause this bug, although it isn't happening in production at the time of writing.

### Expected Behaviour

Discover section always loads more collections when 'See more' button is clicked regardless of filtering on the FE.

### Steps to Reproduce

1. Create over 100 collections (I have a script)
2. Go to Collections page
3. Click 'See more' button at the bottom of Discover section
4. Observe no more collections are displayed (nothing happens)
5. Click again (at least once)
6. Observe collections load

### Screenshots

https://github.com/user-attachments/assets/481d2375-b6c5-4885-9fab-bb2e97d17bd7

### Console Log

```shell

```

catch-21:
Duplicate of the more general issue: https://github.com/pubky/pubky-app/issues/2170

catch-21:
<img width="635" height="136" alt="Image" src="https://github.com/user-attachments/assets/201c2ebf-28c4-49dc-b749-c13f28439820" />

This toast now shows when it cannot find more Collections. Although in truth, trying again could yeild new results. No need to 'wait'

aldertnl:
Toast appears to use the body text instead of the title.. pls use title for consistency (see other toasts)

We could shorten it to 'No new collections found right now.'

aldertnl:
Alternatively we could:

- Autoload more when user is at end of the page. 
- Repeat every X seconds (when at end of page) to account for the flakyness.
- No button, no toast.

PR #2251, opened 2026-07-22
fix: skip stream cursor pagination and refactor

### LINKED ISSUE #2170: [Bug]: Pagination stalls / dead "Show more" under client-side filtering (skip-paginated feeds aren't filter-resilient)
Feeds that paginate by a `skip` offset, the engagement / "Hot" feed and Discover Collections, aren't resilient to client-side filtering. When the app filters out most of a page (muted users, your own or already-bookmarked collections, deleted or empty ones), the next fetch re-requests posts you've already seen. The feed stalls, or "Show more" appears to do nothing, the reported "works on the 3rd/4th click".

This isn't "too few collections". It's a cursor bug that heavy filtering makes visible, and it gets worse as content and usage grow.

## Where it shows up

- **Hot / engagement feed** (and single-collection item streams): stalls when scrolling through a heavily-filtered region.
- **Discover Collections → "Show more"**: a click can surface nothing until you happen to reach an unfiltered region.

## Root cause

Client-side filtering is here to stay. These filters are frontend policy, and several are permanently FE-only, no backend change can move them:

- **muted**: mute lists are homeserver-only, never Nexus-indexed. Nexus physically cannot filter them.
- **already-bookmarked**: frontend state, and the bookmark object is going private, so the backend won't know them.
- **own collections**: the frontend trivially knows the current user.
- **DELETED**: whether a tombstoned post is hidden or rendered as a "deleted" card is a rendering decision, so it's applied on the frontend.

Pagination therefore has to be resilient to client-side filtering by design.

The skip-paginated feeds advance their page cursor by the number of items that **survive** filtering, not by the number the backend actually returned. So when a page is mostly (or fully) filtered out, the offset barely moves, and the next request asks for the same slice again. The fill-to-limit backfill that should compensate can't advance the skip offset either, so it re-reads the same page. Discover papers over this with its own shallow, capped retry loop, which is why it "works on the 4th click" instead of stalling outright.

Score/timestamp-paginated feeds (Home, following) aren't affected, they resume from the last item's score, which advances regardless of filtering. (The separate case of the All feed going momentarily empty under a burst of collections lives on that score-paginated path and already backfills correctly.)

## The fix (implemented + verified)

One invariant resolves the whole class: **advance the pagination cursor by what the backend returned, never by the post-filter visible count.**

1. Advance every feed's cursor by the raw items the backend sent, so a fully-filtered page still moves forward and the next fetch never re-requests it.
2. Collapse the two skip-pagination code paths (the shared feed path and Discover's hand-rolled loop) into that one primitive, one consistently filter-resilient path.
3. Make Discover auto-load like the main feed (drop the manual "Show more" and its shallow retry cap), so a filtered page silently advances instead of leaving a dead button.
4. Bound the auto-loader with a stall guard, so pagination can never spin forever even under adversarial filtering.

Server-side filtering is **not** the fix: Nexus can never filter muted users (never indexed) and has no "exclude mine / exclude already-bookmarked" parameter, so it can't remove the client-side filtering that makes filter-resilient pagination mandatory in the first place. A dedicated backend collections stream would, at best, reduce filter density, not remove the need.

Verified end-to-end against the full frontend test suite (green), including the exact failure cases above: a 100%-filtered page advances to the next slice, the Hot feed scrolls past filtered regions, and Discover auto-fills without dead clicks.

## Expected behavior after the fix

- A heavily-filtered region scrolls/loads through to real content instead of stalling.
- Discover fills without repeated "Show more" clicks; the empty state appears only when the stream is genuinely exhausted.
- Unfiltered and score-paginated feeds behave exactly as before.

secondl1ght:
## 🎯 Summary

Fixes #2170. Supersedes, replaces, and closes #2171 (proof of concept) — this PR keeps its core cursor fix and stream-layer refactor, drops its unapproved UX changes, and adds failure-state feedback plus two adjacent hardening fixes found during review.

**The core bug** — two related defects in the shared stream layer:

- `useStreamPagination` derived skip cursors from the count of already-loaded (post-filter) posts, so Hot/engagement feeds and collection-item streams re-requested slices they had already seen whenever filtering removed anything.
- `PostStreamQueue` never advanced its cursor for skip streams: Nexus returns `last_post_score: null` for them, which the queue happily adopted as the next cursor — so within a single load, its fill-to-limit backfill issued up to 20 backend requests for the same page.

Discover Collections avoided the first defect via its own hand-rolled cursor loop; the second could still cost it wasted backend requests, and its skip accounting was blind to stream-layer filtering, so its offset could drift from the true backend position. But the symptom QA actually reported there — "Show more" surfacing nothing — was mostly not a cursor bug at all: the own/followed/empty filters can legitimately empty a bounded scan. That is inherent to filtering on the client against a backend that paginates blind, and it remains possible after both fixes. What changes is that progress is now guaranteed — the per-load request budget (bounded before, bounded now; Discover's rises slightly, see below) buys strictly deeper scanning instead of repeat requests — and the user gets explicit feedback (see toast below) instead of a silent dead click.

**The invariant this PR establishes:** pagination cursors advance by raw backend data, never by the post-filter visible count — enforced once, in the stream layer, for every feed.

### Core cursor fix (stream layer)

- New `advanceCursor()` primitive in `postStream.types.ts`: skip streams advance by the raw page length (any score is ignored — Nexus sends `null` for them); score streams resume from the last raw `last_post_score`, holding position when absent.
- `PostStreamQueue.collect` advances via `advanceCursor` on every fetch iteration; the overflow-buffer path resumes skip streams by the saved raw offset.
- `NexusPostsKeyStream.last_post_score` is now correctly typed `number | null` and normalized to `undefined` at the fetch boundary.
- `useStreamPagination` no longer derives skip cursors from the loaded-post count; it always resumes from the `nextCursor` threaded back by the previous page, and advances it even on empty-after-filter pages.

### API rename: `timestamp` → `nextCursor`

The stream-slice response field is renamed to reflect what it actually is: an opaque resume cursor (raw `skip` offset for skip streams, score for score streams). Purely mechanical across controllers, hooks, and tests.

### Discover Collections filtering moved into the stream layer

- `filterStreamPosts` drops empty collections for the Discover stream (fail-open on missing details, re-applied by the controller's post-hydration second pass).
- New `filterDiscoverOwnAndBookmarked` drops the viewer's own and already-bookmarked collections inside `getOrFetchStreamSlice`'s filter callback — Discover's cursor accounting now happens where the raw backend page is visible, eliminating the offset drift. Malformed ids are kept rather than throwing.

### Discover Collections component — adapted, not rewritten ⚠️

Unlike #2171, this PR **keeps the product-approved UX**: the "Show more" button (no infinite scroll — consistent with the other Collections sections) and the live-reactive overlays (following a collection removes it from Discover instantly, without reload — behavior explicitly asserted by the `collections.cy.ts` E2E suite, which the original rewrite would have failed). The component's now-redundant fetch-time filtering and hand-rolled backfill loop (~100 lines) are removed; it consumes the already-filtered `nextPageIds` and resumes from `nextCursor`.

### 🍞 New: failure feedback + a deliberate scan-cost setting

- When a "Show more" click yields nothing new while the stream isn't exhausted (the bounded scan was fully filtered), a toast (`collections.discover.noNewResults`, translated in all 9 locales) tells the user instead of silently rendering nothing. The next click resumes from the advanced cursor — always digging deeper.
- New `COLLECTIONS_DISCOVER_MAX_FETCHES_PER_LOAD = 5` sets Discover's per-load scan bound: a marginal, deliberate increase over dev's previous component-level cap of 3 (extra headroom in filtered regions), and far below the shared queue default of 20 that #2171 silently opted Discover into. Every scanned page also triggers detail fetches for its cache-miss posts — even filtered-out ones — so deep scans multiply data usage and latency on slow connections. Implemented as a per-call `maxIterations` override on `collect`; all other feeds keep the shared default.

### 🔖 Bonus: bookmark-time cursor on the queue's buffered path (missed site of #2100)

#2100 made bookmark streams paginate by bookmark time at the three app-layer cache-served sites, but the queue's overflow-buffer early return still synthesized cursors from the post's `indexed_at`. Reachable when a buffered surplus (created by collections being filtered out of the `:all` bookmarks feed) serves a later smaller-limit request — the seam would then skip or repeat bookmarks. The queue now takes a stream-aware `cursorForPost` resolver from the application layer. An integration test reproduces the exact scenario and fails without the fix.

Also: `useReplyStream` now uses a null-check instead of truthiness for cursor advancement, so a legitimate cursor of `0` can't read as "no advancement".

## 🧪 Testing

- Full unit suite green: **11,163 passed** (687 files), typecheck and lint clean.
- New coverage: `advanceCursor` unit tests (including "never uses lastScore for skip streams"); skip-stream and score-stream filter-resilience tests in the queue (a fully-filtered page advances 0 → 20 instead of re-requesting); Discover filter tests (unit + end-to-end through `getOrFetchStreamSlice`); per-load cap bound test; overflow bookmark-cursor integration test (verified to fail without the fix); component tests for the toast, raw-offset resume, and both live overlays.
- Manual QA guide: watch the network tab on skip-paginated feeds — the `skip` param must strictly increase and never repeat. Mute top Hot-feed authors and scroll through them; bookmark old posts + collections and check bookmark-feed page seams; follow a Discover card and confirm instant removal.
- `collections.cy.ts` E2E covers the Discover follow/unfollow flows and runs in CI on push to `dev`.

## 📋 Follow-up

- **Infinite-scroll feeds have an effectively unbounded fetch loop — and this PR makes it quieter, not gone.** When a load comes back empty-but-not-exhausted, `hasMore` stays true and the viewport sentinel immediately refires, chaining round after round with no user feedback until the true end of the stream; the per-load cap (20 fetches) bounds each round, not the total. Before the cursor fix, rounds advanced only by the visible (post-filter) count — re-downloading the overlap each time and stalling entirely on fully-filtered pages; now every round advances by the full raw page, which hides the flaw better while the unbounded data/latency cost remains. The remedies we discussed (toast + backoff, or swapping the sentinel for a "Show more" button) are all product-visible UX changes, so they were deliberately kept out of scope — this needs its own ticket and design input.

- The deeper architecture issue — frontend-only filters over backend-blind pagination can still legitimately produce zero visible results — is out of scope; backend has confirmed no server-side filtering is coming. Where a discrete user action exists (Discover's "Show more"), this PR adds explicit feedback. Infinite-scroll feeds still fail silently when filtering empties them (see above).

- This isn't theoretical: it has already caused user-facing failures on production, e.g. a feed rendering empty because a prolific muted user dominated the fetched pages and client-side filtering discarded everything the backend returned. Fixing the cursor accounting reduces how often that surfaces, but no frontend change can eliminate it — only backend-aware filtering could.


secondl1ght:
@greptile resolved your comment with the latest commit please do final review/update

src/core/application/stream/posts/muting/post-stream-queue.ts
Nirzhuk, line 1: P1 — valid posts can be stranded when Nexus reaches the end.
can save overflow while still returning reachedEnd: true. For example, 19 visible results followed by a final two-result page returns 20, queues one, but consumers hide “Show more.” That last item becomes unreachable.

Nirzhuk (dismissed):

src/core/application/stream/posts/muting/post-stream-queue.ts
secondl1ght: Nice one 🙏🏼

Nirzhuk (approved):

catch-21:
Should fix #2134... but doesn't seem to in the extreme case where 200 consecutive posts are filtered out in front end.

secondl1ght:
@catch-21 interesting... currently it should just literally keep fetching forever until visible posts are found. To rule out the obvious, you were testing either on this branch or with the latest from dev pulled? I can take #2134 and see why this case was not resolved if so.

catch-21:
This introduced the "No new collections found right now. Try again later." toast but it can still view more collections by repeated button clicking, covered by bug issue https://github.com/pubky/pubky-app/issues/2200.

https://github.com/user-attachments/assets/d1de25f8-5c9b-4d6e-aaad-5f20062db72b

PR #2252, opened 2026-07-22
[Locks] feat(#2003): unlock and view locked content (reader side)

### LINKED ISSUE #2003: Locks: reader — access locked content
Part of #1998

## Goal
Reader discovers a lock file, unlocks content with a password, then reads the guarded bytes through the Lock server. No Pubky/Ring auth for readers.

## Tasks (updated 2026-07-31 — shipped on `feat/2003-locks-reader-integration`)
- [x] Load & render the public lock file; resolve the Lock server (`override` or creator `config.json`)
- [x] Generate a `bundle_id` per unlock — ~~durably store~~ not persisted; see "Resolved" below
- [x] Submit proof (via `@pubky/locks-sdk`)
- [x] Track verification status (polling)
- [x] Issue access credential (after `completed`)
- [x] Proxy-read guarded content (the SDK session carries the Bearer; never in a query string)
- [x] ~~Store reader state in IndexedDB~~ — replaced: unlocked content is **replicated to the reader's own homeserver** at `/priv/social/unlocked/<lockId>/` (attachments first, `post.json` last as the completion marker); revisits load the replica with no credential. No IndexedDB reader state.

## Resolved
- `bundle_id` storage/recovery — resolved by content replication (2026-07-17 call): the reader keeps the unlocked content itself in their own `/priv`, so `bundle_id` needs no durable storage. #2028 closed accordingly.
- creator-authority-unavailable — surfaced through the generic unlock error state; no dedicated UX.

## Also shipped beyond the original scope
- Creator own-read path: a creator reads their **own** locked content straight from their homeserver (`/priv/locks.app/content/`), authorship-gated — no unlock, no credential.

## 🔁 Follow-up (2026-07-10) — shared lock card: DONE
`PostContentLock` was dissolved; `LockedPostCard` is the single shared shell for reader and composer (`onUnlock` / `disabled` props).

Representative commits: `f2fdef3e6` (lock-post reader, #2027), `fd0488701` (reader unlock & read flow), `4ea3a1729` (unlocked content display + inline lock title).

Taewa:
Reader side of Locks: unlocking a locked post and reading its content.

Two accounts matter for a locked post: **the author** (who posted it) and **the lock creator**
(who put the content behind the lock). Today they must be the same account.

**What it does**

1. **Someone else's post** — enter the password to unlock it. The content is read from the
   Lock Server, then copied into your own homeserver. Later visits read that copy from your
   homeserver, so no password and no Lock Server call.
2. **Your own post** — read straight from your own homeserver, where the original lives.
   No password, nothing sent to the Lock Server.
3. **Author ≠ lock creator** — stays locked. The original lives on the other account's
   homeserver and this session can't read it. To be revisited in phase 2.
4. The lock title is edited inline on the post card.

Tests cover each of the three cases above, and `docs/locks.md` is updated to match.


Taewa:
@greptile-apps

Review plz

src/core/services/locks/locks.ts
```diff
@@ -74,6 +90,90 @@ export class LocksService {
...
+  // TODO:[Locks] #2040 — lock-sdk returns `any`; validate this response with Zod instead of casting.
+  // TODO:[Locks] #2040 — `password` reaches here but isn't forwarded to lock-sdk (no password verifier yet).
+  static async submitProofBundle(bundle: TSubmittedProofBundle, _password: string): Promise<TVerificationTask> {
+    try {
+      const viewer = await this.getViewer();
+      return (await viewer.submitProofBundle(bundle)) as TVerificationTask;
+    } catch (error) {
+      throw toAppError(error, ErrorService.Locks, 'LocksService.submitProofBundle');
+    }
+  }
```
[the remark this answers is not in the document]
Taewa: Bypassing password is intended. The password verification feature will be removed in follow-up PR.

src/core/application/locks/locks.ts
```diff
@@ -42,6 +68,231 @@ export class LocksApplication {
...
+            throw Err.validation(ValidationErrorCode.INVALID_INPUT, 'guarded attachment has no content descriptor', {
+              service: ErrorService.Locks,
+              operation,
+              context: { path },
+            });
+          }
+          return { id: path.slice(path.lastIndexOf('/') + 1), contentType, bytes: await readBytes(path, uri) };
+        } catch {
+          return null; // already reported (Err factory / service `toAppError`); skip so the rest render
+        }
```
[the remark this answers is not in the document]
Taewa: It's updated.

Taewa:
@greptile-apps
Review again plz

Taewa:
@greptile-apps

Intentional and out of scope for this PR: the Lock Server has no password verifier yet — VerifierType only has DevStatic, and any unknown verifier string is rejected at parse time, so password cannot be sent from the client today. Password validation is server-side work tracked in #2040; the dev-static placeholder and its MUST be replaced before ship comment stay until the Lock Server ships the verifier.

Review again.

src/core/controllers/locks/locks.ts
tipogi: There is not need for that, you could do it
```JS
void LocksApplication.setLockServiceConfig().catch(() => {});
```

src/hooks/usePostLock/usePostLock.ts
tipogi, line 1: I think, we can delete that hook, it seems there is not need it. The pipes, usually are inside the controllers as a validator layer, not in the hooks. Some suggestion
```js
static getLockContent(content: string): LockPostContent | null {
  return LockContentParser.parse(content);
}
```
after in the `fetchLockFile` controller function
```js
static async fetchLockFile(params: TFetchLockFileParams): Promise<TFetchLockFileResult> {
  const lockFile = await LocksApplication.fetchLockFile(params);

  return {
    lockFile,
    verifierType: LockFileParser.resolveVerifierType(lockFile),
  };
}
```

src/core/controllers/locks/locks.ts
```diff
@@ -109,6 +126,13 @@ export class LocksController {
       // Malformed/stale secret — already reported by the service Err factory; clear it so the UI
       // shows unauthenticated rather than a broken session.
       store.reset();
+      return;
+    }
+
+    try {
+      await LocksApplication.setLockServiceConfig();
+    } catch (error) {
+      if (isAppError(error) && isAuthError(error)) store.reset();
```
tipogi: restore two (also in 128 line) store.reset() → this.clearSession()

src/core/application/locks/locks.ts
```diff
@@ -42,6 +69,243 @@ export class LocksApplication {
...
+    const bundle = LockProofBundler.build(lockFile, lockUrl, bundleId);
+
+    // Submit the proof; the server verifies asynchronously and returns a pending task.
+    let task = await LocksService.submitProofBundle(bundle, password);
+    // Poll the task status until it's no longer verifying (or the attempt ceiling is hit).
+    for (let attempt = 0; isVerifying(task.status) && attempt < MAX_POLL_ATTEMPTS; attempt++) {
+      await this.wait(POLL_INTERVAL_MS);
+      task = await LocksService.lookupVerificationTask(creator, bundleId);
+    }
+
```
tipogi, line 92: Why are we not adding that status checks?
```js
if (task.status === 'failed' || task.status === 'expired') {
  throw Err.validation(ValidationErrorCode.INVALID_INPUT, `Unlock verification ${task.status}`, {
    service: ErrorService.Locks,
    operation: 'unlockContent',
    context: { status: task.status, failureMessage: task.failure_message },
  });
}
```

src/core/services/homeserver/homeserver.ts
```diff
@@ -512,6 +520,30 @@ export class HomeserverService {
...
+   * Bytes of an owned resource, or null when it is absent (404) — WITHOUT logging. For existence
+   * checks (e.g. unlock detection) where "not there" is an expected outcome, not an error to report.
+   * Calls the SDK directly so a 404 bypasses `handleError`/Sentry, mirroring `list`'s 404 fallback.
+   */
+  static async getBytesIfExists(url: string): Promise<Uint8Array | null> {
+    const owned = this.resolveOwnedSessionPath(url);
+    // Unreadable without a session — return null rather than fire an unauthenticated request.
+    if (!owned) return null;
+    try {
+      const response = await owned.session.storage.get(owned.path);
```
tipogi, line 543: what about if the response that we get is 404?

```js
try {
  const response = await owned.session.storage.get(owned.path);
  if (response.status === HttpStatusCode.NOT_FOUND) return null;
  if (!response.ok) {
     throw httpResponseToError(response, ErrorService.Homeserver, 'getBytesIfExists', url);
  }
  return new Uint8Array(await response.arrayBuffer());
} catch (error) {
  if (extractStatusCode(error) === HttpStatusCode.NOT_FOUND) return null;
  return handleError({ error, additionalContext: { url, method: HttpMethod.GET } });
}
```

src/core/services/locks/locks.ts
```diff
@@ -74,6 +90,90 @@ export class LocksService {
...
+    } catch (error) {
+      throw toAppError(error, ErrorService.Locks, 'LocksService.readContentLock');
+    }
+  }
+
+  /** A fresh reader-generated bundle id — the public handle for the unlock's verification task. */
+  static async generateBundleId(): Promise<string> {
+    await ensureLocksSdkReady();
+    return BundleId.generate().toString();
+  }
```
tipogi, line 133:
```suggestion
static async generateBundleId(): Promise<string> {
  try {
    await ensureLocksSdkReady();
    return BundleId.generate().toString();
  } catch (error) {
    throw toAppError(error, ErrorService.Locks, 'LocksService.generateBundleId');
  }
}
```

src/core/services/locks/locks.utils.ts
tipogi, line 55: Reset the cached promise when init fails so the next call can retry
```suggestion
export function ensureLocksSdkReady(): Promise<void> {
  if (!sdkReady) {
    sdkReady = import('@pubky/locks-sdk')
      .then(async ({ default: init }) => {
        await init();
      })
      .catch((error) => {
        sdkReady = null;
        throw error;
      });
  }
  return sdkReady;
}
```

src/core/controllers/locks/locks.ts
Taewa: Code updated

src/hooks/usePostLock/usePostLock.ts
Taewa: Code updated

src/core/controllers/locks/locks.ts
Taewa: Code updated

src/core/application/locks/locks.ts
Taewa: Code updated

src/core/services/homeserver/homeserver.ts
Taewa: Code updated

src/core/services/locks/locks.ts
Taewa: Code updated

src/core/services/locks/locks.utils.ts
Taewa: Code updated

tipogi (approved):

secondl1ght (approved):
Failing CI is due to the SDK not being bundled with the app yet during development.

PR #2260, opened 2026-07-23
[WoT V1] Restructure profile-tag filtering as Tagged as reach

### LINKED ISSUE #2258: [WoT V1] Restructure profile-tag filtering as Tagged as reach
## Parent

- #2111

## Context

The WoT V1 product direction has changed: profile/domain tags are no longer exposed as filters layered onto Network, Following, Friends, or Me in the Home sidebar. Instead, **Tagged as** becomes a standalone Reach option representing posts from people the viewer's network tagged with any selected profile labels.

The main feed navigation tab deliberately remains **Home**.

References:

- Product alignment: https://synonymworkspace.slack.com/archives/C031F2DC1JM/p1784729638285889
- Updated Figma: https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=43556-201119&m=dev
- Integration branch/PR: #2247

### Home

- Add **Tagged as** as a synthetic Reach option without changing the model-layer `REACH` values.
- Order the Home options as: My network, Tagged as, Following, Friends, Me, All.
- Map Tagged as to `wot_domain` depth 2.
- Keep the previously selected base feed active until the first profile tag is entered.
- Retain entered profile tags when moving between Reach options, but apply them only while Tagged as is active.
- Replace the per-reach tooltip with a feed headline such as: `Posts from people my network tagged as 'bitcoiner'`.
- Format multiple profile tags using explicit OR wording.
- Keep the main feed navigation tab labeled **Home**.
- Default fresh signed-in users with at least three followed accounts to My network; preserve every existing persisted Reach choice through a versioned store migration.

### Custom feeds

- Add Tagged as to the existing Reach dropdown.
- Enable profile-tag editing only while Tagged as is selected.
- Save Tagged as as `reach=Wot` with `domain_tags`.
- Preserve legacy Following/Friends/Me + domain-tag feeds and their original depth during rename-only edits.
- Convert a legacy feed to Wot/depth 2 only when the user explicitly selects Tagged as.
- Display legacy tagged feeds using their actual Reach while retaining their profile-tag chips.

### Compatibility

- Preserve the existing depth-0 and depth-1 stream/API plumbing for foreign and legacy feeds and the post-MVP ideal state.
- Revert the temporary per-reach tooltip implementation.
- Keep Hot and SinglePost Reach ordering unchanged.
- Update the #2111 body and #2247 description to reflect the final V1 direction.

## Acceptance criteria

- [ ] Selecting Tagged as without a profile tag leaves the current base feed unchanged.
- [ ] Adding the first profile tag switches Home to a depth-2 `wot_domain` stream.
- [ ] Leaving Tagged as restores the selected base Reach without deleting parked profile tags.
- [ ] Persisted pre-migration Reach choices remain user-selected; old Network + profile tags migrate to Tagged as without changing stream semantics.
- [ ] Following/Friends/Me + domain-tag custom feeds retain their Reach and depth on rename-only edits.
- [ ] Tagged-as custom feeds save as Wot + `domain_tags` and require at least one profile tag.
- [ ] The profile-tag tooltip is removed and the Home feed headline uses localized OR semantics.
- [ ] Focused tests, typecheck, lint, and formatting checks pass without browser automation.

Orlandohub:
## Summary

- replace per-reach profile-tag combinations with a standalone **Tagged as** reach in Home and custom feeds
- persist and migrate Home Tagged-as state, gate startup streams on auth/Home hydration, and default eligible fresh users to **My network** without overriding user input
- add a localized Tagged-as headline that remains visible in Visual layout
- preserve legacy custom domain feeds and depth-0/depth-1 stream compatibility on read-only and rename-only paths

## Behavior

- selecting Tagged as without tags keeps the underlying Home base feed active
- adding the first profile tag resolves Home to a depth-2 `wot_domain` stream
- leaving Tagged as parks Home profile tags; explicitly selecting a non-Tagged custom-feed reach clears that feed's domain tags
- new custom Tagged-as feeds persist as `Wot` plus `domain_tags`; legacy Following, Friends, and Me domain feeds remain unchanged unless explicitly converted

## Verification

- focused Vitest suites: 16 files, 473 tests passed
- `npm run typecheck`
- `npm run lint`
- changed-file Prettier check
- `git diff --check`

The repository-wide formatting check still reports three pre-existing generated `.expect/replays/*.ndjson.js` files; changed feature files pass Prettier. No browser automation was run, per scope.

Refs #2258


secondl1ght (approved):
Tested on https://github.com/pubky/pubky-app/pull/2260/changes/4782700a394e496570a076fe686b639a976c487f, I found the UX a bit confusing. Specifically the part where you click the Tagged As reach but the feed stays on the previous one (this is inconsistent because it only happens with empty tags). Similarly if you have a profile tag feed and then remove the tags, the feed falls back to something but it's not clear what in the UI. The Tagged As button is still highlighted but an unknown feed is displayed. It would be better IMO if the left sidebar always matches what is in the feed, it doesn't matter if that would be empty when there are no tags selected.

Other than that I'll list a view Claude comments here that can be addressed if you want (none are important but thought I'd share anyway so you can assess):

 ```
1. Signed-out "All" clicks permanently opt the device out of the fresh-user default — HomeFeedSidebar.tsx / home.actions.ts (moderate, product-visible)

HomeFeedFilters.handleReachChange deliberately lets REACH.ALL bypass requireAuth, and setReach unconditionally sets hasUserSetReach: true. Two compounding problems:

- useControlledState.setValue always fires onChange, even when the clicked radio is already selected. So a signed-out explorer clicking "All" — the only enabled option, and already selected — persists hasUserSetReach: true.
- The home store is device-persisted and is only reset on logout, not on first sign-up. A visitor who idly clicked "All" while exploring will never receive the My-network default after creating an account.

Suggested fix: in the REACH.ALL branch, skip the store write when the value is unchanged, or only mark hasUserSetReach when authenticated (e.g., a separate action for the unauthenticated path).

2. LeftSidebar organism has drifted from the new Home reach model — src/components/organisms/LeftSidebar/LeftSidebar.tsx (minor)

It still binds FilterReach (without showTaggedAs, without auth gating) directly to the home store's reach/setReach. It appears to be dead code (only referenced by a ContentLayout test mock), but if it's ever re-wired it will show a stale reach list and highlight the parked base reach while a Tagged-as stream is actually active. Since this PR restructures exactly that contract, delete it or update it here rather than leaving a trap.

3. Late-arriving default: a user who never touches reach and later crosses 3 follows will be silently switched from All to My network on a subsequent Home mount. Matches the spec as written, but there's no messaging when it happens.

4. "Display legacy tagged feeds using their actual Reach while retaining their profile-tag chips" — met on the read-only feed-filters surface (CustomFeedFilters shows the actual reach with the chips). But in the edit dialog, legacy feeds show their actual Reach with chips hidden — the tags survive in state and on save, they're just invisible while editing (tests assert this hiding deliberately). If the requirement was meant to cover the edit dialog too, that's a deviation worth confirming with design; a user renaming a legacy feed can't see its profile tags at all.
```

Orlandohub:
Addressed the review feedback in `9d3e5883`:

- The highlighted Reach now always matches the effective feed. Opening an empty Tagged as editor keeps the base Reach highlighted; adding the first tag selects Tagged as, and removing the last tag restores the base highlight while leaving the editor open.
- Signed-out All selections are now no-ops, so visitors remain eligible for the post-signup My network default.
- Removed the unused legacy `LeftSidebar` implementation and tests.
- Legacy custom-feed profile tags are now visible as read-only chips in edit mode.

I left the late-arriving default behavior unchanged because it matches the current requirement.

PR #2262, opened 2026-07-24
feat: in app url rich post embeds

### LINKED ISSUE #1792: [Bug]: Retweet with a manually pasted Pubky post is not expanded but represented as a link
### Description

I pasted a link to a post into a reply. Instead of a message preview similar to a quote post, the post is represented as a regular link. Example: https://pubky.app/post/ihgjy51fdnaingcp8rum1omfzd6p8bhm7usune41grd97dho5cwy/00352D5XJT9DG

### Expected Behaviour

The post link should be expanded like a quote post.
Image

### Steps to Reproduce

Past a post link to a new post

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

### LINKED ISSUE #2202: [Task]: Show embedded preview when linking to a post, article or collection
When creating a post with a pubky.app link to a post, article or collection then the embedded preview is always placeholder.
To allow users to share content this way, we want it to embed the post/article/collection as a preview.

For example, this post attempts to share a collection but it shows a generic preview https://pubky.app/post/ihgjy51fdnaingcp8rum1omfzd6p8bhm7usune41grd97dho5cwy/0035D4D144DJ0:

Image

secondl1ght:
# ✨ Rich embeds for in-app post & collection URLs

## Summary

URLs pointing back into the app (e.g. `https://pubky.app/post/<pubky>/<id>`) previously rendered as generic external link previews — the server would scrape our own page's OG metadata over HTTP and display a flat website card. This PR gives in-app URLs the same rich embed treatment as quote-reposts: the full `PostPreviewCard`, with kind-specific rendering (regular / article / collection), local-first hydration, click-to-navigate, and missing/deleted states.

## How it works

- **New `InApp` embed provider** (`PostLinkEmbeds/Providers/InApp/ProviderInApp.tsx`) matches URLs on the **current origin** against the `/post/[userId]/[postId]` and `/collections/[userId]/[postId]` route shapes, validates the pubky identifier, and renders `PostPreviewCard` with the composite `author:postId`.
- The provider is checked **before** the static `PROVIDER_MAP` hostname lookup (the current origin is dynamic, so it can't be a map key). Any non-match — different host, wrong route shape, invalid pubky, `/collections/bookmarks` — falls through to the existing Generic OG preview untouched.
- Current-origin-only matching is intentional: cross-environment links (e.g. a production URL pasted on staging) can't resolve against that environment's Nexus index anyway.
- `matchPostRoute()` added to `src/app/routes.ts` (mirrors `matchSingleCollectionRoute`); `isPostRoute()` refactored to use it.
- The embed passes `bg-muted` so it gets the same contrast as repost previews.

## 🔄 Recursion guard (one level of nesting)

Teaching `PostLinkEmbeds` to render `PostPreviewCard` creates a potential cycle (`PostPreviewCard → PostContentBase → PostLinkEmbeds → PostPreviewCard`): post A linking to post B linking back to A would recurse infinitely. A new boolean context (`PostPreviewNestingContext`, provided by `PostPreviewCard` itself) suppresses in-app embeds inside any embedded preview — the URL renders as plain link text instead. External embeds (YouTube etc.) inside previews are unchanged.

Note: only regular-post embeds can actually recurse (`PostArticle` and `CollectionCard` never render `PostLinkEmbeds`), but both `PostPreviewCard` branches provide the context anyway — it's positional ("this subtree is inside an embedded preview") and keeps the invariant future-proof if those components ever render post content.

## 🐛 Bug fixes surfaced during QA

1. **Deleted posts in preview cards were inconsistent with the rest of the app.** `PostMain` and the single post page render a header-less `PostDeleted` state, but preview cards showed the author header above the deleted notice. `PostPreviewCard` now follows the same `missing → deleted → content` ladder at the card level (deleted collections included).
2. **Latent app-wide race: author info could flash on deleted posts.** `PostHeader` runs its own `usePostDetails` instance, independent of its parent's. When the child's queries resolved before the parent's, the real avatar/username rendered briefly before the parent swapped to the deleted UI. This affected every `PostHeader` site under a deleted-swapping parent (feed cards, list rows, preview cards) and was intermittent — scheduling luck decided whether it appeared. `PostHeader` now checks the deleted tombstone itself and holds its skeleton, so author data can never appear for a deleted post while the full loading state (header skeleton included) is preserved for normal posts.

## 🧪 Testing

- New `ProviderInApp.test.tsx`: URL → composite-id parsing (trailing slash, query/hash, bookmarks exclusion, invalid pubky, wrong host, non-http protocols), render type guard, `bg-muted`, snapshots.
- `PostLinkEmbeds.test.tsx`: in-app post/collection embeds, generic fallback cases, nested suppression (in-app URL renders nothing inside a preview; YouTube still embeds).
- `routes.test.ts`: `matchPostRoute` coverage.
- `PostPreviewCard.test.tsx`: deleted post and deleted collection render `PostDeleted` without header/content.
- `PostHeader.test.tsx`: deleted post renders skeleton, never author info.
- Full suite green: **696 files, 11,290 tests passed**; typecheck + lint clean.


src/components/molecules/PostLinkEmbeds/Providers/InApp/ProviderInApp.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Parse an in-app URL into a composite post id
+   * Returns null for any URL that isn't a current-origin post/collection route,
+   * so callers fall through to the generic provider.
+   */
+  parseEmbed: (url: string): EmbedData | null => {
+    const compositeId = parseInAppCompositeId(url);
+    return compositeId ? { type: 'post', value: compositeId } : null;
+  },
```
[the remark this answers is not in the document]
secondl1ght: CollectionCard is self-navigating — the whole card is wrapped in a <Link
  href={getCollectionRoute(...)}> (CollectionCard.tsx:255), so the embed
  navigates via a native anchor with built-in keyboard support. The wrapper
  intentionally has no handlers; this is the same path collection reposts have
  always used.

secondl1ght:
@greptile review plz

src/components/molecules/PostLinkEmbeds/Providers/InApp/ProviderInApp.tsx
Nirzhuk, line 1: `parsed.host !== window.location.host` disagrees with the app's existing same-origin helper.

  `isSameDomain()` (`src/libs/utils/utils.ts:501`) strips `www.` and compares **hostname**; this compares raw **host**. So for
  `https://www.pubky.app/post/<pubky>/<id>` viewed on `pubky.app`:

  - `shouldBypassLinkConfirmation` → **internal** (no "you're leaving" dialog)
  - `parseInAppCompositeId` → **external** → falls through to the generic OG card

  Same URL, two verdicts — and the generic card is the exact placeholder this PR set out to remove (#2202).

  Repro is local, no deploy needed — jsdom takes a per-file origin:

  ```ts
  /**
   * @vitest-environment jsdom
   * @vitest-environment-options { "url": "https://pubky.app/" }
   */
  it('www.pubky.app is internal to link-confirmation but external to the embed', () => {
    const url = `https://www.pubky.app/post/${PUBKY}/${POST_ID}`;
    expect(isSameDomain(url)).toBe(true);         // passes
    expect(InApp.parseEmbed(url)).not.toBeNull(); // fails — returns null
  });
  ```

  Suggest reusing `isSameDomain` here, or at minimum applying the same `.replace(/^www\./, '')` normalization, so the two paths can't drift again.

src/components/molecules/PostLinkEmbeds/Providers/InApp/ProviderInApp.tsx
```diff
@@ -0,0 +1,64 @@
...
+ * (`/collections/[userId]/[postId]`) route. Host (not origin) comparison:
+ * linkify-it normalizes protocol-less text to `http://`, which would fail
+ * strict origin equality on https deployments.
+ */
+const parseInAppCompositeId = (url: string): string | null => {
+  if (typeof window === 'undefined') return null;
+  try {
+    const parsed = new URL(url);
+    if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') return null;
+    if (parsed.host !== window.location.host) return null;
```
Nirzhuk: www. disagrees with the app's own helper

  isSameDomain() (utils.ts:501) strips www. and compares hostname. This compares raw host. For https://www.pubky.app/post/x/y on pubky.app:
  link-confirmation says internal, embed provider says external. Reproduced locally — see the failing test from earlier.

src/components/molecules/PostLinkEmbeds/Providers/InApp/ProviderInApp.tsx
```diff
@@ -0,0 +1,64 @@
...
+    // URL.pathname excludes query/hash; the route matchers handle trailing slashes.
+    const match = matchPostRoute(parsed.pathname) ?? matchSingleCollectionRoute(parsed.pathname);
+    if (!match) return null;
+    if (!isPubkyIdentifier(match.userId) || match.postId.length === 0) return null;
```
Nirzhuk, line 25: postId unvalidated

  Only the pubky shape is checked. /post/<valid-pubky>/typo → PostPreviewCard → 404 → PostMissing card. That URL previously got a generic OG preview; now
  it gets a permanent "Post not found". Same for a real-but-unindexed post. No test covers it.

Create tests for these cases

Nirzhuk:
Small question related to embed result, we're showing the preview but I have the following questions:

- We wanna show actually the link even we do those preview?

- We wanna open those urls into new tabs even being in the same website? (Demo video below)

https://github.com/user-attachments/assets/2451263a-e65a-4f31-a0a1-7eefc53f7b71

Nirzhuk (approved):

PR #2277, opened 2026-07-30
feat: add visual layout support for collections

### LINKED ISSUE #2207: [UI]: Visual Layout for Collections page (v24)
v24 design shows Visual Layout on Collections page.

https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=34188-139704&t=uzj6rQ1I7WHCTbfB-1

Image

We also need to add the Layout option when creating/editing a Collection if it doesn't exist already: https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=40341-335749&t=uzj6rQ1I7WHCTbfB-1

secondl1ght:
## 📋 Summary

Adds **Visual** as a third creator-selected layout for Collections, alongside Grid and List (#2248). Visual reuses the Home feed's mosaic renderer (`VisualTimelinePosts`) to display a collection's media posts as packed image/video tiles.

Per product decision, media-only content is **not enforced** — any post can be added to any collection regardless of layout. When a Visual collection contains non-media posts, the single-collection view shows a subtle info banner:

> Some items in this collection are hidden due to the selected layout type.

No dependency changes were needed: `pubky-app-specs@0.6.2` (already pinned) supports `'visual'` as a `PubkyAppCollectionLayout`, and the entire persistence pipeline funnels through the `CollectionLayout` config, so widening the type + zod enum covered create/edit/parse round-trips.

### Layout selection
- **Create/Edit dialog** (`DialogCollectionForm`): third radio option with the `LayoutGrid` icon (matching Home's Visual toggle). Shown on all viewports — the choice is the collection's default for every viewer.
- **Viewer override picker** (`CollectionLayoutPicker`, non-owners): refactored from two-way ternaries to an options table; Visual override is temporary and resets on reload, same as Grid/List.
- **Edit round-trip**: prefill, save, and re-save all preserve `'visual'`.

### Rendering
- `useFeedLayoutResolution`: Visual enabled for the `COLLECTION` variant by mirroring the existing List pattern — Wide remains unavailable for collections, and the desktop-only gate is kept: **phones fall back to the Grid rendering** (matches Home behavior).
- `CollectionItems`: maps the collection layout to `LAYOUT.VISUAL` and passes the hidden-items notice + Add Content trigger to the feed.
- `TimelineFeedContent`: fixed `shouldRenderChildren` so the **collection hero stays visible** while the Visual mosaic is active (children on Home/Search/Custom are the composer/filters and stay hidden); threads `emptyState`, `trailingSlot`, `visualHiddenItemsNotice`, and `showEndMessage` to the visual renderer (also aligns Visual with Grid/List in suppressing the end-of-feed message on collection surfaces).

### Hidden-items banner
- New `CollectionHiddenItemsNotice` (card surface, info icon, `role="status"`), rendered above the mosaic for owners and visitors alike.
- Driven by a new `hiddenPostCount` from `useVisualFeedTiles` that only counts **definitively resolved** non-media posts — posts still fetching details or file metadata never count, so the banner cannot flicker on and retract. Deleted/not-found posts are excluded (they render as placeholders, see below).
- Collection-only: Home/Search Visual feeds keep hiding non-media posts silently.

### Owner Add Content CTA
- New `appendVisualTrailingCell` helper places the dashed Add Post tile **inline in the last mosaic row when spacer space exists**, otherwise on its own medium-tile row.
- Reuses the grid trigger component (`GRID_DASHED_CTA_TRIGGER_CLASS`) via a new `'visual'` trigger variant on `DialogAddContent`.
- Empty Visual collections show the empty-state copy plus the CTA tile (grid parity).

### Deleted / not-found items (collection context only)
- Visual now keeps deleted and not-found items in the mosaic as placeholder cards using the same `PostDeleted` / `PostMissing` copy as Grid/List, so item counts stay consistent across all three layouts. Cards are clickable and keyboard-accessible (navigate to the post page) like their Grid/List counterparts.
- "Not found" is only declared after the post's ensure-fetch **settles** (the batch equivalent of `PostMain`'s `null && !isLoading` check) — a new `hasPendingPostDetails` flag keeps the skeleton up until then.
- Explicitly scoped via a `showUnavailablePosts` prop enabled only for the `COLLECTION` variant — **Home/Search/Custom Visual behavior is unchanged** (deleted/failed posts are still silently dropped there).
- Note: placeholder cards carry no remove-from-collection affordance in any layout today; author removal of dead items is a planned follow-up feature.

### i18n
- `layoutVisual` (create/edit + picker namespaces) and `hiddenItemsNotice` added to all 9 locales, reusing each locale's existing "Visual" translation and layout terminology.

## 🐛 Fixes bundled from QA of this feature

1. **Pre-existing: edit-collection dialog reopened with stale values** (shipped with #2035/#2248). After saving, reopening the dialog showed pre-save values until a page refresh: the Dexie live query delivers the committed envelope *while the save is still awaited*, so the close handler's `reset()` — holding a stale render closure — wrote the old values back, and the prefill effect never re-fired. Fixed in `useEditCollection` by reading the latest envelope values through a ref; covers name, description, and layout. Regression test reproduces the exact race (verified to fail against the unfixed code). Surfaced by Visual because a layout change makes the stale highlight obvious.
2. **Dead infinite-scroll sentinel spacing**: List and Visual kept a 20px sentinel mounted after the feed fully loaded, producing extra bottom spacing vs Grid (which conditionally unmounts it). Both now mirror Grid. *Reviewer note: this also removes the dead space in end-of-feed states on Home/Search List/Visual.*
3. **Visual skeleton padding**: the mosaic skeleton carried a `py-4` that neither the Grid/List loading states nor the loaded mosaic have — extra space under the hero during load plus a 16px layout jump when tiles arrived. Removed. *Reviewer note: also affects Home/Search Visual loading states (aligns them with their own loaded state).*
4. **"No posts found" flash when switching to Visual**: on a fully-loaded collection (`hasMore: false`), the first frames before the Dexie live query's first emission were indistinguishable from a resolved-empty snapshot. The tiles hook now surfaces `hasPendingSnapshot` (no default snapshot value) and the renderer treats it as loading — toggle transitions show the skeleton, never the empty state. Applies to both viewer overrides and owner layout edits.

## 🔍 Post-QA automated review

An adversarial multi-angle code review of the full diff was run before opening the PR; three low-risk correctness fixes in new code were applied afterward (each provably scoped outside QA'd flows, covered by unit tests):
- `VisualTimelinePosts` now passes `emptyState` to `TimelineStateWrapper`'s `emptyComponent` (Grid/List parity for the visitor zero-post edge case).
- `shouldBackfillInitialRows` gained the `hasPendingSnapshot` guard its sibling expressions already had (prevents a spurious page fetch during layout switches).
- The placeholder tile's hardcoded `aria-label` was removed so screen readers announce the localized deleted/not-found copy instead of a machine id.

## 🧪 Testing

- **Unit/snapshot**: full suite green — 701 files / 11,398 tests. New coverage mirrors #2248's shape: layout resolution (desktop/phone/wide), persistence round-trips, form/picker options, CTA placement helper, hidden-count rules, banner gating, hero visibility, placeholder rendering + settlement transitions, and Home-regression guards (no-prop renders byte-identical; default-off placeholder behavior pinned).
- **Typecheck + lint**: clean.
- **Manual QA**: full checklist passed (create/edit flows, mosaic + banner, CTA placement, viewer overrides, layout-switch transitions, spacing parity, deleted/not-found placeholders, phone fallback, Home/Search regression sweep, locale spot-checks).
- **E2E**: intentionally not included — QA team owns Cypress coverage for this feature.


secondl1ght:
@greptile this is how visual layout already behaves in other areas of the app (pre-existing but now collections support visual layout). Arguably showing a loading state if all the posts metadata cannot be resolved is correct. Another option would be to add a new `MetadataMissing` state in addition to our deleted/missing post render paths, but it's an edge case and should be addressed app-wide not just for this collections feature. Please review again.

secondl1ght:
@greptile are you updating the confidence description? It still has the outdated comments from before the re-review.

secondl1ght:
@greptile OK I guess not, thanks anyway. 🙂

Orlandohub (approved):
I've tested it in desktop and mobile and all seems correct. I was not able to upload a banner image on mobile, I believe it has to do with the mobile session browser cookies not setting something important on preview links. This error is tracked here on Sentry:
https://synonym-52.sentry.io/issues/137899332

Here is some AI review which I don't find that relevant but will leave it here for context:

[1] medium | simplification | src/components/organisms/Timeline/Feed/TimelineFeed/VisualTimelinePosts.tsx:288
What's wrong: VisualTimelinePlaceholderTile copies handleNavigate/handleKeyDown verbatim from VisualTimelineTile (lines 229–243) — same event.target !== event.currentTarget guard, same Enter/' ' + preventDefault activation logic. Why it matters: Keyboard-accessibility behavior for mosaic tiles now lives in two identical implementations in the same file; a fix to one silently diverges from the other. Fix: Extract a shared helper, e.g. useVisualTileActivation(postId, onNavigate) returning { handleNavigate, handleKeyDown }, and use it in both tile components.

[2] low | simplification | src/components/organisms/Collections/DialogAddContent/DialogAddContent.tsx:113
What's wrong: The 'visual' trigger variant renders the exact same DialogAddContentGridTrigger as 'grid' (the PR's own comment admits "no dedicated trigger needed"), yet costs a union member, a switch case, doc lines, and a dedicated test + snapshot. Why it matters: The only per-layout difference is the data-cy suffix, which the caller already passes independently via the dataCy prop — the variant adds a layer without adding behavior. Fix: Drop 'visual' from DialogAddContentTriggerVariant and the switch; in CollectionItems pass triggerVariant={isListLayout ? 'list' : 'grid'} while keeping dataCy={collection-add-content-${addContentVariant}}.

[3] low | simplification | src/components/organisms/Timeline/Posts/Posts.tsx:93
What's wrong: The conditional scroll-sentinel block ({hasMore && <Container overrideDefaults className="h-5" ref={sentinelRef} />}) plus its 3-line rationale comment is added identically to Posts.tsx and VisualTimelinePosts.tsx, mirroring a third copy already in GridPosts.tsx. Why it matters: Three copies of the same pattern and comment must be kept in sync across the three feed renderers. Fix: Extract a small shared component (e.g. TimelineScrollSentinel({ hasMore, sentinelRef })) carrying the comment once, and use it in all three renderers.

[4] low | simplification | src/components/organisms/Collections/CollectionLayoutPicker/CollectionLayoutPicker.tsx:36
What's wrong: The new COLLECTION_LAYOUT_PICKER_OPTIONS list (value + labelKey + icon) duplicates the layoutOptions array built in DialogCollectionForm.tsx — same three layouts, same icons, same labels. Why it matters: The layout→icon→label mapping is now declared in two places; this PR had to touch both, and any future layout or icon change must be made twice. Fix: Move a single COLLECTION_LAYOUT_OPTIONS definition next to COLLECTION_LAYOUT in src/config/collections.ts and consume it from both the picker and the form.

[5] low | simplification | src/components/organisms/Collections/CollectionItems/CollectionItems.tsx:69
What's wrong: Two parallel chained ternaries branch on the same 3-way condition (isListLayout ? … : isVisualLayout ? … : …) to compute requestedLayout and addContentVariant. Why it matters: The three-way layout dispatch is expressed twice and the two must stay aligned by hand. Fix: Replace with one lookup keyed by viewerLayout, e.g. LAYOUT_VIEW = { [COLLECTION_LAYOUT.LIST]: { requested: LAYOUT.LIST, variant: 'list' }, … }, and read both values from it.

PR #2285, opened 2026-07-31
feat(posts): Post Input Improvement with animations - Related to different content unfocus and focus

### LINKED ISSUE #2086: [UI]: [Improvement] Update design for Post Input in Wide / List Layout
### Problem Statement

We need to update the design of PostInput in wide layout and list layout

Image

### Proposed Solution

Update the component with the following desgins:

https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=34188-137485&m=dev
https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=34188-140009&m=dev

### Suggested Priority

None

### LINKED ISSUE #2219: [UI]: Avatar and placeholder in inline reply area are vertically off-centre
When viewing replies in feed the option to reply have your avatar and placeholder text that is vertically off-centre. The design shows it central.

**Actual**

Column Layout
Image

List Layout
Image

**Design**

Image

https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=35688-283315&t=uzj6rQ1I7WHCTbfB-1

Nirzhuk:
resolves #2086 
resolves #2219 

## Description 
Unifies the Post Input around all the layouts and creates a smoother transition between the 2 modes.

We center the three thread SVG too due not having extra space in the input, the extra space was creating un need it spacing.

## Demo
Screenshot 2026-08-03 at 12 12 30

### Column Desktop
https://github.com/user-attachments/assets/1222a780-6749-48cb-9961-81c7d561f576

### Column Mobile

https://github.com/user-attachments/assets/e8d50c29-fd06-4734-9f1c-0d2f62c22a62

### Wide Layout
https://github.com/user-attachments/assets/553ca6f0-e59e-4e09-ba7a-f2e16b064146

https://github.com/user-attachments/assets/167e8194-fdf4-4e17-b6ae-4c4286b30697


src/components/organisms/PostHeader/PostHeader.tsx
```diff
@@ -27,7 +32,11 @@ export function PostHeader({
   const { postDetails } = usePostDetails(isReplyInput ? null : postId);
 
   // Fetch user details for avatar and name
-  const { userDetails, isLoading: isLoadingUserDetails } = useUserDetails(userId);
+  const hasProvidedUserDetails = providedUserDetails !== undefined;
+  const { userDetails: queriedUserDetails, isLoading: isLoadingUserDetails } = useUserDetails(
+    hasProvidedUserDetails ? null : userId,
+  );
+  const userDetails = hasProvidedUserDetails ? providedUserDetails : queriedUserDetails;
```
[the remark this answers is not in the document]
Nirzhuk: @greptile-apps its been fixed

catch-21:
This PR must fix the following VRT failures for quick reply:

Screenshot 2026-08-03 at 17 16 11
Screenshot 2026-08-03 at 17 16 05

Please update baseline images before merging.

src/components/organisms/PostInput/PostInput.tsx
```diff
@@ -230,109 +311,208 @@ export function PostInput({
...
+        className={cn(
+          'min-w-0 contain-inline-size',
+          '[&_textarea::placeholder]:transition-opacity [&_textarea::placeholder]:duration-150',
+          'focus-within:[&_textarea::placeholder]:opacity-0',
+          'motion-reduce:[&_textarea::placeholder]:transition-none',
         )}
+      >
+        <motion.div
+          data-testid="post-input-state-height"
+          className="overflow-hidden"
```
infin1t3: Image

`overflow-hidden` wrapper clips the non-portaled popover to ~110–130px; with 3+ matches the rest is invisible and unclickable. Affects PostInput, QuickReply, and all four dialogs.

olivert2025:
Design reviewed across all views (Column, List, Visual, and Wide). The animation feels smooth, doesn't distract the user, and achieves its intended purpose effectively.

infin1t3:
https://github.com/user-attachments/assets/e75a1e61-3a5c-4745-80e3-87088e9b2edb

The PR redefines what size="large" means for PostHeader. Before, large mapped to a 64px avatar + text-2xl name. The PR remaps large → 48px/text-xl (in PostHeaderUserInfo.utils.ts) and adds a new extraLarge tier that preserves the old 64px rendering. The wide-layout call site in PostMain.tsx:112 was migrated to extraLarge, but PostArticleDetail.tsx:69 still passes size="large" — it's the only remaining caller — so article pages silently pick up the new, smaller rendering.

Change large ---> extra large

src/components/organisms/PostMain/PostMainLayoutRules.ts
```diff
@@ -17,6 +18,13 @@ export function getTagsLayoutForSurfaceLayout(layout: LayoutType): TagsLayout {
   return 'inline';
 }
 
+/** PostInput header size by tags layout: list = small, wide = large, column = normal. */
+export const POST_INPUT_HEADER_SIZE_BY_TAGS_LAYOUT: Record<TagsLayout, PostHeaderSize> = {
+  list: 'normal',
+  side: 'large',
```
infin1t3: **Wide composer avatar renders 48px — Figma and neighboring wide posts use 64px**

This PR remapped `PostHeader` size `'large'` to a 48px avatar (`lg`) and introduced `'extraLarge'` (64px) to preserve the old rendering — but the composer size map still assigns `side → 'large'`. Measured live: the wide-layout PostInput/QuickReply avatar renders 48px while the regular wide posts next to it render 64px (`PostMain.tsx:112` uses `extraLarge`).

Both designs linked from the tickets show 64px for the composer:
- #2086 wide node (`34188-137485`): 64×64 header instance in a 160px-tall input (48+64+48 — the input currently renders 146px)
- #2219 replies node (`35688-283315`): the same 64px avatar in the reply row

Suggested fix: map `side → 'extraLarge'` here. One catch: Figma's wide gap is 20px (`gap-5`, currently paired with `'large'`), while `'extraLarge'` pairs with `gap-6` in `GAP_CLASS_BY_HEADER_SIZE` — the gap entry needs a matching tweak to stay on-spec.

src/components/organisms/QuickReply/QuickReply.tsx
```diff
@@ -149,14 +170,23 @@ export function QuickReply({
...
       <Container
         ref={containerRef}
+        data-state={isExpanded ? 'expanded' : 'collapsed'}
         className={cn(
           'relative w-full cursor-pointer rounded-md border border-dashed transition-colors duration-200',
-          isWideLayout ? 'p-12' : 'p-4',
+          '[&_textarea::placeholder]:transition-opacity [&_textarea::placeholder]:duration-150',
+          'focus-within:[&_textarea::placeholder]:opacity-0',
+          'motion-reduce:[&_textarea::placeholder]:transition-none',
+          isWideLayout ? 'p-12' : 'p-6',
```
infin1t3: **Wide in-thread reply input is 146px tall — the #2219 design shows 112px**

`isWideLayout ? 'p-12' : 'p-6'` reuses the main composer's 48px padding for the in-thread reply. The #2219 design (node `35688-283315`) differentiates the two: the standalone main input is 160px tall with 48px padding, while the in-thread reply row is 112px with 24px padding (+ 64px avatar). Measured live, the wide reply card renders 146px.

Related: the thread-connector corner currently lands ~25px above the avatar center in wide layout — the retuned `QUICK_REPLY_CONNECTOR_SPACER_HEIGHT` assumes p-6 geometry (column/list measure ~−5px). Worth fixing the reply padding and the connector constant together so the corner points at the avatar in all layouts.

src/components/organisms/QuickReply/QuickReplyContent.tsx
```diff
@@ -16,18 +24,60 @@ const BODY_TEXT_CLASS_BY_LAYOUT: Record<TagsLayout, string | undefined> = {
   inline: undefined,
 };
 
+const QUICK_REPLY_REVEAL_EASE = [0.19, 1, 0.22, 1] as const;
```
infin1t3: **Animation timing now lives in four places**

The dissolve variants, reduced-motion twins, and easing array here (lines 27–66) are byte-identical to `SELECTIVE_DISSOLVE_VARIANTS` / `REDUCED_SELECTIVE_DISSOLVE_VARIANTS` / `STATE_REVEAL_EASE` in `PostInput.tsx:37–77`; the `isExpanded ? 0.28 : 0.22` height-transition ternary repeats in both composers; and `QuickReply.tsx:132–135` re-encodes the same numbers a fourth time as Tailwind arbitrary classes (`duration-[280ms]`/`duration-[220ms]`, `ease-[cubic-bezier(0.25,1,0.5,1)]`) so the connector's CSS transition matches the card's framer tween by convention only.

Any retune applied to one copy silently desyncs the two composers this PR set out to unify (and detaches the connector from the card mid-animation).

Suggestion: extract a shared `composerMotion.ts` — variants + eases + expand/collapse durations, with the `cubic-bezier(...)`/`ms` strings derived from the same constants — imported by `PostInput`, `QuickReplyContent`, and `QuickReply`.

src/components/organisms/PostMain/PostMainLayoutRules.ts
```diff
@@ -17,6 +18,13 @@ export function getTagsLayoutForSurfaceLayout(layout: LayoutType): TagsLayout {
...
+  inline: 'normal',
+};
+
 /** Wide and list layouts opt into expanded PostInput treatment. List applies local compact styling overrides. */
 export function usesWidePostInput(tagsLayout: TagsLayout): boolean {
   return tagsLayout === 'side' || tagsLayout === 'list';
```
infin1t3: **List-layout PostInput renders the wide treatment — Figma's list design is 88px/24px**

Because `usesWidePostInput()` returns true for `'list'`, the list-layout main composer gets `p-12` + the wide `text-xl` class (measured live: 138px tall, 48px padding, 20px text). The #2086 list design (node `34188-140009`) shows 88px total: 24px padding, 40px avatar, ~16px text.

Meanwhile the list-layout QuickReply on the same screen renders `p-6` + `text-base` (90px — matching the design), because `QuickReply.tsx:124` uses a different predicate (`effectiveTagsLayout === 'side'`). So the two composers disagree within the same layout, and PostInput is the off-spec one. `PostInput.tsx:424` also only ever applies `WIDE_POST_BODY_TEXT_CLASS`, never `LIST_POST_BODY_TEXT_CLASS` — unlike QuickReplyContent's `BODY_TEXT_CLASS_BY_LAYOUT`.

Suggested fix: align the predicates (PostInput treats only `'side'` as wide, like QuickReply) — or, if list really should opt into the wide shell, add the compact overrides this function's comment promises. Per the linked design, the compact p-6 treatment is the intended one.

src/components/organisms/PostInput/PostInput.tsx
```diff
@@ -18,17 +19,88 @@ import { canSubmitPost, cn, getCharacterCount } from '@/libs/utils/utils';
...
+    transition: { duration: 0.14, ease: STATE_REVEAL_EASE },
+  },
+  exit: {
+    opacity: 0.6,
+    filter: 'blur(0px)',
+    transition: { duration: 0.1, ease: STATE_REVEAL_EASE },
+  },
+} satisfies Variants;
+
+function useMeasuredHeight() {
```
infin1t3: **Second height-measuring hook — duplicates `useElementHeight`**

This private `useMeasuredHeight` (ref + `getBoundingClientRect` + ResizeObserver) re-implements `src/hooks/useElementHeight`, which `QuickReply.tsx:103` already uses in this same PR to drive the identical measured-height animation — with subtly different semantics (`useLayoutEffect` vs `useEffect`+rAF, `null` vs `0` initial value, `<= 0` filtering).

Two hooks now own one concern, so a measurement/timing fix applied to one composer's expand animation silently won't reach the other.

Suggestion: reuse/extend `useElementHeight` — both call sites already guard the unmeasured state the same way (`stateContentHeight === null` here, `cardHeight <= 0` in QuickReply) — and delete this copy.

infin1t3 (changes requested):

src/components/organisms/QuickReply/QuickReply.tsx
Nirzhuk: was not pushed the change to only wide might use p-12, sorry for that.

src/components/organisms/PostInput/PostInput.tsx
Nirzhuk: MentionPopover now uses createPortal so we can prevent to create it inside our PostInput

infin1t3:
image
image

Some CTA's in Dialogs are cut-off now.

src/components/organisms/QuickReply/QuickReply.tsx
```diff
@@ -149,14 +165,23 @@ export function QuickReply({
...
```
infin1t3: **Wide in-thread reply input is 162px tall — the design shows 112px**

`isWideLayout ? 'p-12' : 'p-6'` reuses the main composer's 48px padding for the in-thread reply. The #2219 design (node `35688-283315`) shows a 112px reply row: 24px padding + 64px avatar. The design differentiates the standalone main input (48px padding — correct in code) from the in-thread reply (24px). Measured live after the latest push: 162px.

**Note:** when fixing the padding, please retune `QUICK_REPLY_CONNECTOR_SPACER_HEIGHT` in the same change — the connector corner currently sits ~17px above the avatar center (masked visually by the 64px avatar), and changing the padding shifts that geometry again (~7px below center with p-6).

src/components/molecules/MentionPopover/MentionPopover.tsx
```diff
@@ -1,23 +1,55 @@
...
+/**
+ * Tracks the anchor's viewport rect so the popover can be rendered in a portal.
+ * The composers live inside clipping ancestors (the height-animating card and
+ * the `lg:overflow-hidden` feed column), so an in-flow absolute popover gets cut
+ * off; a fixed one in `document.body` escapes them all.
+ */
+function useAnchorRect(anchorRef: MentionPopoverProps['anchorRef']) {
+  const [rect, setRect] = useState<{ top: number; left: number } | null>(null);
+
+  useLayoutEffect(() => {
```
infin1t3, line 22: **Portaled popover doesn't follow the textarea when it grows**

`useAnchorRect` re-measures only on scroll/resize, and this effect's only dep is the stable ref — so when autosize grows the textarea by a line mid-mention, no event fires, the effect never re-runs, and the `fixed` popover stays at stale coordinates overlapping the composer text until the user scrolls.

Repro: expanded composer → type until the line is about to wrap → start `@name` so the wrap happens while the popover is open.

Suggested fix: also re-measure when the suggestion list changes (e.g. lift `update` into a callback re-run on `users`), or attach a `ResizeObserver` to the anchor.

src/components/organisms/PostHeader/PostHeader.tsx
```diff
@@ -37,7 +46,7 @@ export function PostHeader({
   // once the query settles we render regardless of whether it found a
   // profile — a missed local read + failed Nexus fetch must not hide the avatar,
   // pubky and character counter behind a permanent skeleton.
-  const isLoading = isLoadingUserDetails || (!isReplyInput && !postDetails);
+  const isLoading = (!hasProvidedUserDetails && isLoadingUserDetails) || (!isReplyInput && !postDetails);
```
infin1t3, line 49: **Avatar-only slots flash a doubled full-width skeleton while the profile loads**

This early-return runs before `showUserInfo={false}` / `visuallyHideAvatar` are considered, so while `currentUserDetails` is still undefined (≥1 frame warm, whole fetch on cold cache) the composers render the full `PostHeaderSkeleton` (avatar + name/pubky bars) twice: once absolutely positioned over the textarea (the stable-avatar slot, z-10) and once in-flow, pushing the placeholder right — then snap back.

Repro: DevTools → Clear site data (stay signed in) → Network throttle Slow 3G → reload Home → watch the composer.

Suggested fix: when `showUserInfo` is false, return an avatar-sized skeleton (or an invisible spacer when `visuallyHideAvatar`).

src/components/organisms/PostInput/PostInput.tsx
```diff
@@ -230,109 +244,213 @@ export function PostInput({
...
+          '[&_textarea::placeholder]:transition-opacity [&_textarea::placeholder]:duration-150',
+          'focus-within:[&_textarea::placeholder]:opacity-0',
+          'motion-reduce:[&_textarea::placeholder]:transition-none',
         )}
+      >
+        <motion.div
+          data-testid="post-input-state-height"
+          className="overflow-hidden"
+          initial={false}
+          animate={{ height: shouldReduceMotion || stateContentHeight <= 0 ? 'auto' : stateContentHeight }}
```
infin1t3: **Card height animates on every line wrap, not just expand/collapse**

The animate target continuously tracks the ResizeObserver-measured height, so any content-height change — a wrapping keystroke, adding an attachment — runs a 220–280ms JS height tween with per-frame reflow of everything below the composer. Same pattern in QuickReply (multiplied per feed post, with the connector's CSS transition animating in parallel).

Repro: expanded composer → type until lines wrap → content below eases down ~250ms on each wrap instead of snapping.

Suggested fix: key the height animation to `isExpanded` transitions and rest at `height: 'auto'` (e.g. via `onAnimationComplete`), letting content-driven growth snap in one native reflow.

src/components/organisms/QuickReply/QuickReplyContent.tsx
```diff
@@ -47,56 +51,127 @@ export function QuickReplyContent({
...
-        setAttachments={setAttachments}
-        handleFilesAdded={onFilesAdded}
-        isSubmitting={isSubmitting}
-      />
+          <Container
+            overrideDefaults
+            className={cn('flex w-full min-w-0 items-center', GAP_CLASS_BY_HEADER_SIZE[headerSize])}
+          >
+            {!isExpanded && currentUserPubky && (
+              <div data-testid="quick-reply-collapsed-avatar-placeholder" className="shrink-0">
```
infin1t3: **Invisible avatar-spacer PostHeaders run the full data stack**

Each composer mounts PostHeader 2–3× (visible absolute overlay + this invisible collapsed placeholder + the expanded ghost). The invisible copies still run `usePostDetails`/`useUserDetails` live queries, store subscriptions, and load a duplicate avatar `` — once per feed post that renders a QuickReply — purely to reserve avatar-width space under the absolutely-positioned real avatar.

Suggested fix: replace the invisible PostHeader instances with a plain `aria-hidden` div sized via `AVATAR_SIZE_BY_HEADER_SIZE` (same pattern in PostInput's placeholder slot).

src/components/atoms/PostThreadConnector/PostThreadConnector.tsx
```diff
@@ -7,6 +7,8 @@ import type { PostThreadConnectorVariant } from './PostThreadConnector.types';
 interface PostThreadConnectorProps {
   height?: number;
   variant?: PostThreadConnectorVariant;
+  className?: string;
```
infin1t3: **`className?` is now a dead prop**

This prop was added earlier in this PR to carry the connector's Tailwind transition classes, but the follow-up switched QuickReply to the new `style` prop — so no caller passes `className` anymore (QuickReply → `style`, PostMain → `height`/`variant`, PostInput → `variant` only). It can be removed along with the `cn(..., className)` threading in `getBaseContainerProps`.

Related nit: both `className` and `style` are silently ignored by the `DIALOG_REPLY` variant (it returns before `baseProps` is built) — harmless with today's callers, but a footgun if someone passes them with that variant.

src/components/organisms/PostInputExpandableSection/PostInputExpandableSection.tsx
```diff
@@ -45,85 +43,31 @@ export function PostInputExpandableSection({
...
-                      },
-                    }
-                  : {}),
-              },
-            }}
-            className="overflow-hidden"
-          >
-            <Container className="gap-4">
-              {hasContent && !isArticle && <PostLinkEmbeds content={content} />}
+      {isExpanded && (
```
infin1t3: **`isExpanded` can only ever be `true` now — prop and conditional are dead**

Pre-PR this prop was live: PostInput passed the actual state and this component animated its own collapse. This PR moved mount control to the parents' `AnimatePresence` blocks, and both remaining callers (`PostInput.tsx:430`, `QuickReplyContent.tsx:157`) render the component only when expanded and pass the literal shorthand `isExpanded` — so this internal `{isExpanded && (...)}` is constant-true.

Suggested cleanup: delete `isExpanded` from `PostInputExpandableSectionProps`, render the container unconditionally, and drop the test that exercises the now-impossible collapsed state.

src/components/organisms/QuickReply/QuickReply.types.ts
```diff
@@ -15,6 +16,7 @@ export interface QuickReplyProps {
 export interface QuickReplyContentProps {
   // Current user header
   currentUserPubky: string | null;
+  currentUserDetails: NexusUserDetails | null | undefined;
```
infin1t3: **Three different shapes for the same profile-bypass value** (nit)

The same value is typed three ways in this PR: here `currentUserDetails: NexusUserDetails | null | undefined` (required but undefined-able), the same in `UsePostInputReturn`, and `userDetails?: NexusUserDetails | null` in `PostHeaderProps`. Suggest picking one spelling (the `?: NexusUserDetails | null` form) so required-ness reads consistently across the chain.

src/components/molecules/MentionPopover/MentionPopover.types.ts
```diff
@@ -12,4 +13,6 @@ export interface MentionPopoverProps {
   onSelect: (userId: string) => void;
   /** Callback when hovering over a user (for keyboard navigation sync) */
   onHover: (index: number) => void;
+  /** Element the popover is positioned under — it renders in a portal, so this is its only anchor */
+  anchorRef: RefObject<HTMLTextAreaElement | null>;
```
infin1t3: **`anchorRef` typed narrower than needed** (nit)

Correctly required — the portal can't position without it, and both callers pass it. But the popover only calls `getBoundingClientRect()` on it, so `RefObject<HTMLElement | null>` would keep it reusable beyond textareas (e.g. the article title `Input` later).

src/libs/motion/composerMotion.ts
```diff
@@ -0,0 +1,89 @@
+import type { Variants } from 'motion/react';
```
infin1t3, line 1: Missing tests

infin1t3 (changes requested):

src/components/organisms/PostHeader/PostHeader.tsx
Nirzhuk: Screenshot 2026-08-05 at 04 55 52
I changed the skeleton to be less impactful.

src/components/molecules/MentionPopover/MentionPopover.tsx
Nirzhuk: This is prod:
Screenshot 2026-08-05 at 05 15 17



Changed to keep the position of the end of the textarea even we scroll
Not going to follow the caret tho.

Nirzhuk:

infin1t3 (approved):
All good from my end.

Checked the code in several rounds - also tested the functionality.

PR #2289, opened 2026-07-31
fix: user tag count shows tags applied, not tags received (#1990)

### LINKED ISSUE #1990: [Bug]: User tag count reflects tags received, not tags applied by the user
### Description

There is an issue with how user tag counts are calculated and displayed: The tag count shown for a user reflects the number of tags received on their posts or profile (i.e., tags applied to them by others).

Image

### Expected Behaviour

The tag count should reflect the number of tags applied by the user themselves (i.e., tags they create/assign to posts, profiles, or other content), not the number of tags they receive.

Context

In Pubky’s semantic social graph, tagging is a user action: users apply tags to content and other profiles to express context, curation, and relationships. The current metric appears to measure inbound tagging activity rather than outbound tagging activity, which is misleading when interpreting a user’s profile or activity stats.

### Steps to Reproduce

1. Go to "Active users"
2. Tag somebody
3. See

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

secondl1ght:
## Description

Closes #1990

User list tag counts displayed inbound tagging activity (tags received on the user's profile/posts) instead of outbound (tags applied by the user). Per the issue, tagging is a user action in Pubky's semantic social graph, so the stat now reads Nexus `counts.tagged`.

### Changes

- `useUserStream`: map the `tags` stat from `counts.tagged` instead of `counts.tags` — fixes the **Active Users sidebar** (issue screenshot), the **Hot page Active Users list**, and **Who to Follow**
- `useProfileConnections`: map the `tags` stat from `counts.tagged` instead of `counts.unique_tags` — keeps the TAGS stat consistent on **Followers / Following / Friends** lists
- No data-layer changes needed: the local write path already increments the tagger's `tagged` count optimistically on tag create/delete, so counts update reactively
- The profile page's own "Tagged" tab count still uses `unique_tags` intentionally (it labels the list of tags the profile *received*)
- Removed the unused `FollowerItem` molecule (dead code superseded by `UserListItem`'s full variant)

### Tests

- New regression test for the `useUserStream` counts mapping (previously uncovered)
- `useProfileConnections` test fixtures now give `tagged` / `tags` / `unique_tags` distinct values so the assertions prove which field is read


Taewa (approved):
LGTM.
I tested on Ephemeral link (d33ca33).
All correctly applied including `/profile/tagged` page which should not be changed at all.

PR #2290, opened 2026-08-01
fix: stream pagination livelock on long filtered runs

### LINKED ISSUE #2134: [Bug]: Feed breaks as many consecutive Collections are being created

secondl1ght:
# fix: stream pagination livelock on long filtered runs (advance cache-walk anchor by raw scan)

## 🎯 Summary

Follow-up to #2251. QA reported that publishing ~200 collection posts still breaks the home feed: no new posts ever load, and the bottom loading indicator flickers rapidly. Reproduced and confirmed on latest — this is a second resume cursor that #2251's fix did not cover.

**The core bug:** pagination has *two* resume positions, and #2251's invariant ("cursors advance by raw backend data, never by the post-filter visible count") was only enforced for one of them:

- `streamTail` — the backend cursor. Fixed by #2251. ✅
- `lastPostId` — the anchor for walking the **local stream cache**. `getStreamFromCache` pages exclusively by `indexOf(lastPostId)` and ignores `streamTail` entirely, and the hook only advanced this anchor when a round returned *visible* posts. ❌

After a fully-filtered load round (zero visible posts), the next round restarted the cache walk at the head. One round scans at most `limit (10) × MAX_FETCH_ITERATIONS (20) = 200` raw posts, and cache full-hits always report `reachedEnd: false` — so **≥200 consecutive filtered-out posts cached at the anchor** meant every round rescanned the same 200 ids, returned nothing, and never advanced: an infinite livelock. Rounds are pure IndexedDB reads (milliseconds), so the auto-firing infinite-scroll sentinel re-triggered instantly — the reported flicker. QA's "~200 posts" sits exactly on the threshold.

This was **not** home-feed-specific and **not** collection-specific: the same livelock applied to every score-paginated, cache-served feed (home timelines, timeline-sorted custom/search feeds, profile timeline/replies, bookmarks, collections sections) and to every filter type that can form a run (collection-kind, deleted tombstones, muted authors).

### Core fix — thread the raw scan anchor (`lastRawPostId`)

- `TPostStreamChunkResponse` / `TReadPostStreamChunkResponse` gain an optional `lastRawPostId`: the id of the last **raw** post scanned in a round, visible or filtered. The application layer already tracked this internally (`lastReturnedPostId`); it is now returned and threaded through both controller branches (the cache-miss second-pass filter shrinks only the visible page — the raw anchor passes through untouched).
- `useStreamPagination` now advances `lastPostId` from the raw anchor **before** the empty-page early return, so fully-filtered rounds resume instead of restarting. Falls back to the last visible id when no anchor is threaded; never clobbers a defined anchor with `undefined`.
- New `resolveResumeAnchor()` helper (`controllers/stream/posts/posts.utils.ts`, with direct unit tests) owns the "prefer raw anchor over last visible id" rule once. All three anchor-threading callers use it: `useStreamPagination`, `FollowedCollections`, `DiscoverCollections`.
- **No queue changes** — `PostStreamQueue` is untouched.

### Companion hardening (found during regression analysis)

- **Coordinator head-polls bypass the shared queue** (`streamHead > 0` → straight to `fetchStreamFromNexus`). Polls only need the fetch side effects (unread persist + counts); routing them through `collect()` consumed and rewrote the UI's overflow buffer every ~9s, which under a raw anchor would have silently skipped buffered posts. Side effects verified identical for every polled stream type (home timelines, tag-filtered custom feeds, reply streams). Bonus: this also fixes a latent poll bug where an existing overflow buffer caused an inverted request window that silently starved the unread "new posts" pill.
- **Mute/unmute no longer clears the queue** (`LocalMuteService`). The clear was redundant — `collect()` re-filters the buffer against a fresh Dexie read of the mute list on every access — and under a raw anchor it would have skipped up to `limit − 1` buffered posts on any mute action mid-scroll.
- Controller `getOrFetchStreamSlice` collapsed to a single return so future response fields can't be silently dropped on one branch.

## 🧪 Tests

All new correctness tests were written **red-first** and confirmed failing against the unfixed code.

- **Integration (real Dexie), `mute-pagination.test.ts`** — new describe reproducing the QA scenario end-to-end: 200-collection run at the cache head (round 1 empty but anchored, round 2 surfaces the real posts, Nexus never consulted), plus deleted-tombstone run, interior-run (anchor mid-stream), sub-threshold contrast case (190 posts recover in one round), and mute-mid-scroll buffer survival (buffer re-filtered, muted posts dropped, remainder delivered).
- **Hook, `useStreamPagination.test.tsx`** — anchor adopted from `lastRawPostId` on empty pages, preferred over the last visible id, fallback when absent, no clobber on anchor-less empty pages.
- **Application, `post.test.ts`** — anchor advances through fully-muted scans, holds at the caller's position on buffer-only rounds, and head-polls leave the overflow buffer untouched.
- **Controller, `post.test.ts`** — anchor threaded through both branches and preserved when the second-pass filter empties the page.
- **Components** — `FollowedCollections` / `DiscoverCollections` Show More anchors on `lastRawPostId`.

Gates: typecheck ✅ · lint ✅ · full unit suite ✅ (11,466 passed). An adversarial multi-angle review + final verification pass over the complete diff returned zero regressions.

## 📋 Affected surfaces

| Surface | Status |
| --- | --- |
| Home timelines (all/following/friends × kinds) | Fixed (reported case) |
| Custom feeds & search (timeline-sorted, incl. tags) | Fixed |
| Profile timeline & replies | Fixed (deleted/collection runs) |
| Bookmarks feed | Fixed (bookmarked-collection runs) |
| My Collections / Followed Collections | Fixed (deleted-collection runs) |
| Hot, Discover, single-collection items, engagement feeds | Never susceptible (skip-paginated, cache walk bypassed — #2251 territory) |

## ⚠️ Known limitations (unchanged, documented)

- **Cold cache**: a first-ever scan through a long filtered run still chains bounded empty rounds (~20 fast rounds for 200 posts) before content appears — the feedback-less auto-load limitation documented in #2251, tracked as follow-up. No livelock: progress is now guaranteed every round.
- **Unmute mid-scroll** doesn't resurface already-scanned posts on Discover Collections until refresh; timeline surfaces already auto-refresh on unmute.

## 🔎 QA re-test notes

Beyond the original repro (publish ~200 collection posts → home feed), the regression matrix should include: a profile with a long run of deleted posts, and a bookmarks feed containing many bookmarked collections. Warm cache should recover within 1–2 rounds; cold cache within ~20 fast bounded rounds.


catch-21:
Tested with over 200 collections with the 'All', 'Following', 'Friends', 'Me' feeds and posts in profile page all reliably show "No posts found". This happens even if the posts were not already in IndexedDb, and also on sign in.

With as little as 60 collections created I get this on sign in, although it sorts itself out with some filter-switching or navigating around.

These observations closely match previous observations, athough now I don't see the flickering.

Deleted posts doesn't appear to impact this, I think nexus doesn't serve them up at all so nothing extra for pubky-app to filter.

Screenshot 2026-08-05 at 11 42 14

Screenshot 2026-08-05 at 11 42 05

catch-21:
Tested again, with over 500 collections created consecutively. The feed does load on a fresh sign in or explore. See videos. There are a lot of requests but the overall amount of data is not huge. This is probably the best outcome we can hope for.

**Explore mode**

https://github.com/user-attachments/assets/ccce72a6-821a-466e-a50f-7dac19f30bcd

**Sign in**

https://github.com/user-attachments/assets/c781d6e9-3539-4013-a3db-16eb4f8b989e

src/core/services/local/mute/mute.ts
Orlandohub: `let statusChanged = false;` is hoisted outside the Dexie transaction solely so the removed `postStreamQueue.clear()` block could read it. It's now only assigned and read inside the callback.

`const statusChanged = isMuting !== isCurrentlyMuted;` inside the transaction and delete the outer declaration.

Orlandohub (dismissed):
`src/core/application/stream/posts/post.ts:329,357,364`

The PR's own head-poll bypass made the `isFirstFetch` mechanism dead. Every call that now reaches `postStreamQueue.collect()` has `streamHead === 0`, so `streamHead: isFirstFetch ? streamHead : SKIP_FETCH_NEW_POSTS` yields 0 on both branches.

Replace the ternary with `streamHead: SKIP_FETCH_NEW_POSTS` and delete the `isFirstFetch` declaration and mutation.

### AI Side Note on the overall architecture assessment

**The fix itself is architecturally correct and conforms to this codebase's rules.** Layer boundaries are respected end-to-end, the new helper lives at the right layer (controllers, where both the hook and components can reach it), the response field is optional and backward-compatible with no-clobber semantics, and the change ships with tests at every layer. For a local-first app, "advance cursors by raw scanned data, never by post-filter visible count" is the textbook-correct invariant — the same rule that governs keyset pagination with post-filtering anywhere.

That said, this PR is the second consecutive fix for the same class of bug, and that pattern points at three structural weaknesses worth follow-up work rather than blocking this merge:

1. **The resume state is fragmented, and the threading protocol is convention-enforced.** A pagination session's position now lives in four places: `streamTail` (backend cursor), `lastPostId` (cache-walk anchor), the `postStreamQueue` buffer, and the Dexie stream cache. Every caller — `useStreamPagination`, `FollowedCollections`, `DiscoverCollections` — must independently remember to advance both cursors, prefer the raw anchor, and never clobber with `undefined`. `resolveResumeAnchor` centralizes one rule, but the protocol is still copy-paste. The standard remedy is an **opaque continuation token**: have the controller return a single `cursor` object (wrapping `streamTail` + `lastPostId`) that callers thread back wholesale. That makes the #2251/#2290 bug class — advancing one cursor and forgetting the other — structurally impossible.

2. **`streamHead > 0` as an implicit call-intent discriminator.** The code itself admits this (`post.ts:668`: "we might need to add some enum param to describe that type of call"). The PR deepens the reliance by adding another branch keyed on it. Polls and pagination are different operations with different contracts sharing one method; an explicit `intent: 'paginate' | 'poll'` parameter (or a separate `pollStreamHead()` controller method) would make the bypass self-evident instead of magic-number-driven.

3. **The module-level `postStreamQueue` singleton is shared mutable state.** The head-poll bug this PR fixes existed *because* an unrelated caller shared the UI's pagination buffer. The bypass fixes the symptom; the deeper fix is scoping the buffer to the pagination session that owns it. Two feeds mounted on the same `streamId` would still contend for one buffer today.

One broader design note: the scan-and-filter-with-iteration-budget approach (cap of 200 raw posts per round) is a pragmatic choice given that mute lists are viewer-local and Nexus can't pre-filter them. But the documented cold-cache limitation (~20 silent skeleton rounds through a long filtered run) is an inherent cost of filtering *after* the walk. If these filtered runs grow common, the durable fix is pushing the cheap filters (kind, deleted) down into the storage layer — e.g., a Dexie compound index so the cache walk skips filtered runs in one indexed query — rather than raising iteration budgets. Worth a line in the follow-up ticket the PR already references.

Orlandohub (approved):

PR #2292, opened 2026-08-03
VRT: Collections

catch-21:
- Collections overview (/collections)
- Single collection — grid layout (/collections/[userId]/[postId])
- Single collection — list layout (/collections/[userId]/[postId])
- Single collection — visual layout (/collections/[userId]/[postId])
- Bookmarks collection (/collections/bookmarks)

-----

Made one class change to production app here https://github.com/pubky/pubky-app/blob/0e99ff8a0549eadc60c684bc3b229a45bf321224/src/components/organisms/Collections/CollectionHero/CollectionHero.tsx#L193 to ensure background image is always displayed on collection header. This seems to only show up with some headless browsers. The reason given is:

> `isolate` keeps the `-z-10` cover inside this card's stacking
context (same fix as CollectionCard). Without it the cover joins the
page root's stacking context, where engines disagree on whether it
paints above or below the page background once desktop ancestors add
`lg:overflow-hidden` — leaving the hero blank in some browsers.

Screenshot 2026-08-03 at 23 09 06

As a good side-effect, this change appears to have made the background looking less like it has a grey-filter:

Previous:
old

Now:
new


src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
```diff
@@ -190,7 +190,7 @@ function CollectionHeroContent({
     <Card
       data-cy="collection-hero"
       className={cn(
-        'relative gap-0 overflow-hidden rounded-md py-0',
+        'relative isolate gap-0 overflow-hidden rounded-md py-0',
```
infin1t3, line 198: I would add a code comment explaining isolate here too, not just in the PR description.

vitest.config.ts
```diff
@@ -88,7 +88,7 @@ export default defineConfig({
                 },
                 // Home feed desktop (WebKit/Linux) needs extra headroom for
                 // layout to settle after fonts/images decode.
-                timeout: 10_000,
+                timeout: 15_000,
```
infin1t3, line 91: nit: the comment above still attributes this headroom to the Home feed, but the 10s→15s bump came from the image-heavy Collections suite (5f0e059e). Worth generalizing, e.g. `// Image-heavy suites (Home, Collections) on WebKit/Linux need extra headroom for layout to settle after fonts/images decode.` so the next reader doesn't wonder why Home needs 15s.

src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
```diff
@@ -199,6 +199,7 @@ function CollectionHeroContent({
         <Container
           overrideDefaults
           aria-hidden="true"
+          data-testid="collection-cover"
```
infin1t3: nit: `collection-cover` has no references left — it was added for the VRT cover-wait helper that 31f9516d later removed, so nothing selects it now. Fine to keep as a self-documenting hook, but if that's not deliberate, it can be dropped.

src/test/vrt/feed/Collections.vrt.test.tsx
```diff
@@ -0,0 +1,743 @@
...
+async function renderSingleCollection(
+  layout: keyof Awaited<typeof fixtures>['singleCollections'],
+  viewport: { width: number; height: number },
+) {
+  const f = await fixtures;
+  const collection = f.singleCollections[layout];
+  routeState.pathname = `/collections/${collection.details.author}/${collection.postId}`;
+  routeState.params = { userId: collection.details.author, postId: collection.postId };
+
+  const screen = await renderForVRT(<CollectionWithHeader postId={collection.compositeId} />, { viewport });
```
infin1t3, line 677: optional: unlike `renderCollectionsOverview`, this doesn't preload the hero cover, and `renderForVRT` only awaits `` elements — CSS `background-image` isn't covered. So the hero background settles via `toMatchScreenshot`'s retry window instead of an explicit wait. It's stable today (ran the grid test solo 3× with cold caches across all three engines), but an `await preloadImages(Object.values(f.collectionCoverUrls))` here would make readiness explicit and claw back some of the 15s assertion timeout on slow WebKit.

catch-21: Done in ab67b293c7e1473774f1ce0917a72507fa23b1a8 and added note in visual-regression-testing.md

src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
catch-21: Done in ab67b293c7e1473774f1ce0917a72507fa23b1a8

vitest.config.ts
catch-21: Done in ab67b293c7e1473774f1ce0917a72507fa23b1a8

src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
catch-21: Done in ab67b293c7e1473774f1ce0917a72507fa23b1a8

PR #2300, opened 2026-08-04
fix(ui): allow emoji picker scroll without prior click (#1871)

### LINKED ISSUE #1871: [Bug]: Cannot scroll emoji selector for add tag
### Description

When adding a tag to post or profile, if the emoji button is pressed and the emoji selector appears then it is not possible to scroll it until it has been clicked on, which could accidentally select an unwanted emoji.

### Expected Behaviour

Emoji selector is scrollable as soon as it is opened - the profile status one works correctly.

### Steps to Reproduce

1. Go to add a tag
2. Click the emoji button
3. When emoji selector appears, place cursor within the area and try to use mouse wheel to scroll 
4. Observe no scrolling happens

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
Stop wheel/touchmove from bubbling out of EmojiPickerDialog so Radix RemoveScroll no longer blocks emoji-mart's Shadow DOM scroller until the picker is clicked. Covers add-tag and other shared consumers.

https://github.com/user-attachments/assets/0d96a658-9c46-4d1c-bfe6-1f1f23e66fe9

Resolves #1871


catch-21 (approved):
Verified using issue steps. Works on MacOS chrome and safari. Still working on android chrome. Not tried on other platforms.

New tests and test changes look good.

PR #2113, opened 2026-07-01
Fix reply modal long username overflow

Orlandohub:
## Summary
- Constrain the shared post header user-info slot so long display names can truncate inside tight layouts.
- Add width constraints to the reply dialog input wrapper and PostInput containers.
- Add regression assertions and refresh snapshots for the affected components.

## Context
Slack thread: https://synonymworkspace.slack.com/archives/C09HQTLAARF/p1782924462493009

Screenshot 2026-07-01 at 14 54 24

Fixes #2112

Issue #2112, opened 2026-07-01
[Bug]: Reply modal overflows with long usernames


Orlandohub:
## Description
Long display names are not constrained inside the reply modal/reply composer. When a post author has a very long username, the username text in the replied-to post preview and in the reply composer header spills horizontally out of the modal on both desktop and mobile.

Slack context: https://synonymworkspace.slack.com/archives/C09HQTLAARF/p1782924462493009

## Expected Behaviour
Long usernames should truncate or wrap within the available reply modal width without causing horizontal overflow, clipped controls, or text extending past the modal bounds.

## Steps to Reproduce
1. Use or create an account with a very long display name, for example `BobiWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWW`.
2. Create a post from that account.
3. Open the post and click reply.
4. Observe the reply modal on desktop and mobile widths.

## Screenshots
Screenshots are attached in the linked Slack thread and show the long username overflowing in the reply preview and composer.

## Console Log
_No console errors reported._

## Browser & Version
_Not specified in the Slack thread._

PR #2304, opened 2026-08-05
fix(ui): scope post header popover to the user profile block

### LINKED ISSUE #2303: [Bug]: Post header popover opens when hovering anywhere on the header line
## Description

Hovering **anywhere** on the post header line opens the user info popover, instead of only when hovering the user profile (avatar / display name / pubky).

This is a regression introduced in #2113 (fix for #2112). To make long display names truncate, that PR changed the `PostHeaderUserInfo` root container — which is also the `UserInfoPopover` trigger (`PopoverTrigger asChild`) — from a shrink-to-fit box to a full-width one:

```diff
- flex w-fit min-w-0 items-center gap-3
+ grid w-full max-w-full min-w-0 grid-cols-[auto_minmax(0,1fr)] items-center gap-3
```

Because the trigger element now stretches across the whole header row, the empty space between the display name and the timestamp is part of the hover target.

## Expected Behaviour

The user info popover should only open while hovering the user profile block (avatar, display name, pubky). Hovering the empty space on the rest of the post header line should do nothing.

The long-username truncation fixed in #2112 must keep working.

## Steps to Reproduce

1. Open the home feed.
2. Hover the empty area of a post header, to the left of the `1h` timestamp and far away from the display name.
3. The user info popover opens.

Orlandohub:
## Summary

- fix #2303
- The user info popover was opening from anywhere on the post header line, not just the user profile. Regression from #2113.
- #2113 made the `PostHeaderUserInfo` root container `w-full` so long display names could truncate — but that container is also the `UserInfoPopover` hover target (`PopoverTrigger asChild`), so the hover area grew to span the whole header row. The username `<a>` grew with it, making the empty space clickable through to the profile too.

## Changes

- `PostHeaderUserInfo.tsx`: root container `w-full` → `w-fit` (keeping `max-w-full min-w-0`), so the popover trigger hugs the avatar + name and still falls back to the available width — and truncates — when the name is long.
- `PostHeaderUserInfo.tsx`: username link `w-full` → `w-fit` for the same reason, so its click target ends with the text.
- Added a comment on the container explaining that it doubles as the hover target, so it doesn't get reverted to `w-full` by the next layout fix.
- Added two regression tests: one asserting the popover trigger is not full-width, one asserting the username link is not full-width.
- Updated the existing long-name truncation assertions and 9 snapshots.

## Test plan

- [x] `npm run test` — full unit suite green (737 files, 11,758 tests)
- [x] `npm run test:vrt` — 30 files / 63 tests green, no baseline changes needed (the container is invisible; only its box width changed)
- [x] `npm run typecheck`, `npm run lint`, Prettier — clean
- [x] Manual browser check on `/home` at 1440px: post header popover trigger went from **696px wide → 143px**; a point 20px left of the timestamp is no longer inside the trigger
- [x] Real hover: hovering the timestamp no longer opens the popover; hovering the display name and avatar still does
- [x] Long-name regression (the #2112 fix): forced a 96-char name and a 74-char unbreakable name into the header — truncates with no overflow and no horizontal page scroll, on the feed, in the reply dialog (both the composer header and the replied-to preview), and at mobile width
- [x] Checked the `size="large"` / `timeAgoPlacement="bottom-left"` variant on the article detail page

## Notes

- Fixing the username link width is slightly beyond the reported hover bug, but it's the same root cause on the same element tree: pre-#2113 that link was `w-fit`, and while it was `w-full` a click in the empty header space navigated to the author's profile.
- `w-fit` is `min(max-content, available)`, and `max-w-full` caps the pathological case of a single unbreakable word whose `min-content` exceeds the column — both verified in the browser.

### DIFF
--- a/src/components/molecules/PostHeaderUserInfo/PostHeaderUserInfo.test.tsx
+++ b/src/components/molecules/PostHeaderUserInfo/PostHeaderUserInfo.test.tsx
@@ -468,10 +468,28 @@ describe('PostHeaderUserInfo - Navigation', () => {
     const usernameLink = profileLinks[1];
     const userInfoRoot = usernameLink.parentElement?.parentElement;
 
-    expect(userInfoRoot).toHaveClass('grid', 'w-full', 'max-w-full', 'min-w-0', 'grid-cols-[auto_minmax(0,1fr)]');
+    expect(userInfoRoot).toHaveClass('grid', 'w-fit', 'max-w-full', 'min-w-0', 'grid-cols-[auto_minmax(0,1fr)]');
     expect(usernameLink.parentElement).toHaveClass('max-w-full', 'min-w-0');
-    expect(usernameLink).toHaveClass('block', 'w-full', 'min-w-0', 'max-w-full', 'overflow-hidden');
+    expect(usernameLink).toHaveClass('block', 'w-fit', 'min-w-0', 'max-w-full', 'overflow-hidden');
     expect(screen.getByText(longName)).toHaveClass('w-full', 'truncate', 'max-w-full');
+  });
+
+  it('keeps the popover hover target hugging the user info instead of the whole header row', () => {
+    render(<PostHeaderUserInfo userId="testuser123" userName="Test User" />);
+
+    const userInfoRoot = screen.getByTestId('popover-trigger').firstElementChild;
+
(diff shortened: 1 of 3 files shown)


infin1t3 (approved):
https://github.com/user-attachments/assets/6597eef9-33b8-4326-b5a0-aa80003a0db2

PR #2306, opened 2026-08-05
refactor(i18n): make English the only supported language

### LINKED ISSUE #2145: [Task]: Remove internationalization part 1 - Make English the only supported language while retaining i18n infrastructure
# Summary

The app currently allows users to switch languages through Settings, with translations available across the UI. These translations are largely AI-generated and unvalidated, which creates a risk of mistranslations and UI breakage. One recent example is #2013.

Maintaining high-quality translations across all supported locales requires ongoing implementation, review, and testing. Without an intentional localization process, we cannot reliably guarantee a good experience in non-English languages.

For now, the app should support English only. However, the existing i18n infrastructure should remain in place so that additional languages can be added again in the future without requiring a larger architectural change.

# Solution

- Remove the Language section and language selection controls from Settings.
- Remove all non-English translation files, keeping only the English translations.
- Configure the app to always use English.
- Keep the existing i18n library, configuration, providers, and translation-key usage.
- Remove code that is only needed for runtime language selection.
- Ensure previously stored non-English language preferences are ignored or migrated safely to English.
- Users who need translation can rely on browser-level or OS-level translation tools. This may not be consistently available in a PWA context, although OS-level translation support is improving.

# Acceptance Criteria

- The Language section is no longer visible in Settings.
- Users cannot select or switch the application language.
- The application always renders in English, including for users with a previously stored non-English preference.
- Only the English translation files remain.
- Existing i18n infrastructure and translation calls remain functional.
- There are no missing translation keys, broken routes, or visible UI regressions.

# Out of Scope

- Removing the i18n framework or its dependencies.
- Replacing translation calls with hardcoded English strings.
- Adding or validating translations for additional languages.

infin1t3:
## Summary

Closes #2145

Removes runtime language selection and all non-English translations while keeping the i18n infrastructure (next-intl, providers, translation-key usage) in place, so additional languages can be re-added later without architectural changes. Users needing translation can rely on browser/OS-level translation tools.

## Changes

- **Settings UI**: removed the Language section — `/settings/language` route, `Language` template, `LanguageSelector` organism, and the Language items in both settings menus (desktop + mobile)
- **Locale resolution**: `routing.ts` declares `locales: ['en']`; `request.ts` no longer reads the `locale` cookie and always resolves English, so previously stored non-English preferences are ignored
- **Selection plumbing removed**: `SettingsController.setLanguage`, `useSettingsActions.setLanguage`, the store's `language` field/action, `setLocaleCookie` (+ its callers in the auth and migration controllers), and RouteGuard's locale-divergence `router.refresh()` effect
- **Homeserver `settings.json`**: the `language` field is no longer written. Backward-compatible in both directions: this client ignores a legacy `language` field from older clients (covered by a new normalizer test), and older clients fall back to `'en'` via their existing `?? 'en'` default when the field is absent
- **Logout**: `clearCookies` no longer excludes the `locale` cookie (the exclusion existed only to preserve the language across logout), so stale locale cookies are actively dropped
- **Dead code**: removed the orphaned `Locale` type export, the now-unused `exclude` param on `clearCookies`, the `useLocale` entry in the global next-intl test mock, and the unused `language`/`settings.language`/`settings.menu.language` keys from `en.json`
- **Kept (per issue scope)**: next-intl dependency and config, `IntlProvider`, all `useTranslations()` calls, RTL constants, and `force-dynamic` in the root layout (its comment now states the real remaining reason: per-request runtime config serialization)

## Testing

- Full unit suite: 704 files, 11,671 tests passed
- VRT: 63 passed, no baseline changes (settings surfaces have no VRT coverage)
- Lint, typecheck, and production build pass; build route list confirms `/settings/language` is gone
- Settings menu snapshots regenerated; removed the skipped language Cypress test

https://github.com/user-attachments/assets/8ec03188-4305-46b7-b1b3-f6641db0c5b8

image


Orlandohub (approved):
LGTM

PR #2314, opened 2026-08-07
feat(ui): group repeated deleted and edited notifications

### LINKED ISSUE #1570: [Feature]: Group same notifications such as 'deleted a post'
### Problem Statement

When a user deleted or edits a bunch of posts that I've interacted with I'll see a bunch of notifications that all look the same.

### Proposed Solution

Deletion notifications have no links so when lots of notifications occur for deletion, group them into a single list item in notification list.

As for edits, they each list to the edited post so aggregating would impact the UX here. Seek input from design on this one - perhaps use a popover list.

### Suggested Priority

Nice to have (would be cool)

### LINKED ISSUE #2287: [Bug]: Profile page crops descender letters in username
### Description

Descender letters have their decending part cropped on the profile page.

### Expected Behaviour

All characters display in UI without being cropped.

### Steps to Reproduce

1. Edit your username to have some decender letters (j, y, g, p)
2. View your profile page
3. Observe the parts of those letters that hang down are cropped

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
Closes #1570

Bursts of post-change notifications (an author editing or deleting many posts you interacted with) flooded the list with one row per post. This groups them.

## What changed

- **Grouping** — consecutive deleted/edited notifications from the same actor and content kind collapse into one row; repeated notifications about the same post keep only the newest occurrence.
- **Deleted groups** render as a flat summary (`John deleted 8 posts you interacted with`) with no post links; single deleted rows are informational too — the deleted post has no destination.
- **Edited groups** list every affected post title behind a `Show`/`Hide` disclosure on desktop (pill inline in the row while collapsed, under the list when expanded) and render permanently expanded on mobile, per the [Figma design](https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=44799-221063).
- **Titles** resolve live and local-first (article title / collection name / content with mentions resolved), quoted and truncated at 40 graphemes; a member deleted after its edits renders a plain unlinked notice, and a post that no longer exists drops its link.
- **Freshness** — edited posts are refetched when their notification arrives, so rows show post-edit content even when the post was already cached.
- **Refresh** — poll-driven refreshes merge new notifications in place instead of swapping the list for a skeleton, so expanded groups, scroll position and loaded pages survive.
- **Pagination** — grouping can make a fetched page add zero rows, so auto-loading budgets unproductive loads and hands control back via a manual `Load more` button; a failed load keeps the accumulated rows and offers an inline `Try again`.
- **Shared chrome** — avatar, heading and timestamp/icon extracted into `NotificationRowChrome`, used by both single and grouped rows.

## Tests

Unit coverage across grouping rules, the group row (disclosure, focus, icons, lazy fetches), title resolution, the silent refresh and the load budget. The two e2e "…that you reposted" scenarios now create two posts each and assert the grouped row — including the Show/Hide disclosure on desktop and the permanently expanded list on the mobile config.
image
image
image

Edit: Also closes #2287

### DIFF
--- a/src/core/application/notification/notification.ts
+++ b/src/core/application/notification/notification.ts
@@ -13,7 +13,8 @@ import { PostStreamApplication } from '@/application/stream/posts/post';
 import { UserStreamApplication } from '@/application/stream/users/users';
 import { HttpMethod } from '@/libs/http/http.types';
 import { Logger } from '@/libs/logger/logger';
-import type { Pubky } from '@/models/models.types';
+import { CompositeIdDomain, type Pubky } from '@/models/models.types';
+import { buildCompositeIdFromPubkyUri } from '@/models/models.utils';
 import { type FlatNotification, NotificationType } from '@/models/notification/notification.types';
 import { NotificationNormalizer } from '@/pipes/notification/notification.normalizer';
 import { HomeserverService } from '@/services/homeserver/homeserver';
@@ -300,6 +301,10 @@ export class NotificationApplication {
   /**
    * Fetches posts and users referenced in notifications that are not yet persisted in cache.
    *
+   * Edited posts are refetched even on a cache hit: the notification itself proves the
+   * cached copy predates the edit, and the local-first read path never revalidates a
+   * cached post, so without this the notification row would render the pre-edit content.
+   *
    * @param notifications - Array of flat notifications to extract post and user references from
    */
   static async fetchMissingEntities({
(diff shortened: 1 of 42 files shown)


aldertnl:
Left comments on the Figma, the design solution needs a bit of work @olivert2025 @infin1t3

aldertnl:
Reviewed again, can you shorten the action buttons to 'Show' and 'Hide' ? 

image

image

aldertnl:
And maybe on mobile we should render the expanded state always without the show and hide buttons, since it renders vertically anyway?

aldertnl (dismissed):
LGTM

infin1t3:
@greptileai review

src/core/application/notification/notification.ts
```diff
@@ -300,6 +301,10 @@ export class NotificationApplication {
   /**
    * Fetches posts and users referenced in notifications that are not yet persisted in cache.
    *
+   * Edited posts are refetched even on a cache hit: the notification itself proves the
+   * cached copy predates the edit, and the local-first read path never revalidates a
+   * cached post, so without this the notification row would render the pre-edit content.
+   *
    * @param notifications - Array of flat notifications to extract post and user references from
    */
   static async fetchMissingEntities({
```
Orlandohub, line 310: `fetchMissingEntities` no longer only fetches missing entities. Callers reading `fetchMissingEntities` will assume a cheap no-op when everything is cached, but pages containing PostEdited notifications always issue a Nexus fetch.
Updating to `syncNotificationEntities` or `fetchMissingAndStaleEntities` would be better imo

Orlandohub (approved):
Small comment on `fetchMissingEntities` but LGTM otherwise

src/core/application/notification/notification.ts
infin1t3: Will do in a follow-up.

PR #2318, opened 2026-08-08
feat: reorder posts in collections

### LINKED ISSUE #2195: [Task]: Collections reordering (v25)
The design for Collections shows a reorder button that enables the ability to drag-drop to reorder posts in your collection.

https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=35688-285514&t=uzj6rQ1I7WHCTbfB-1

secondl1ght:
# ✨ Collections: Reorder Mode (drag-and-drop item ordering)

## Summary

Adds **reorder support** to the single collection page (`/collections/[userId]/[postId]`). Collection owners get a new **Reorder** button in the hero action bar that switches the item feed into a drag-and-drop grid of every collection post. Saving commits the new order to the `items` array of the collection's content envelope via the existing local-first post-edit path.

## 🖱️ UX

- **Reorder button** (Move icon) in the owner action row, between Share and Edit. Enabled only when the collection has 2+ items; disabled while a delete is in flight. Not rendered for non-owners or blurred (moderated) collections.
- **Active state** per the Figma design: the button swaps to **"✓ Save order"** (brand `default` variant, spinner while committing) plus a **Cancel** button (new `destructive-soft` Button variant — the brand-soft recipe in destructive red).
- While reordering, all other hero actions (Add Post, Share, Edit, Delete, tag toggle) are disabled and the global **FAB is hidden** — reorder mode is for reordering only.
- **Drag support on all inputs**: mouse (8px activation distance), touch (250ms long-press so scrolling still works), and keyboard (Space to lift, arrows to move, Space to drop) via `@dnd-kit`.
- Cards render with the design's **white dashed border**, are fully **inert** (no post actions reachable — the whole cell is the drag surface), and use translate-only drag transforms so the preview never distorts across rows of different heights.
- Save with no changes exits without a network write. A failed save keeps the mode and draft intact (error toast) so the user can retry or cancel.

### New core pieces
- `CollectionPostContent.reorderItems` (pipe): pure reorder with **merge semantics safe against concurrent edits** — items added since the draft was taken are kept (prepended, matching `addItem`), removed items are dropped rather than resurrected; reference-equality no-op contract like `addItem`/`removeItem`.
- `PostController.commitReorderCollectionItems`: mirrors `commitUpdateCollectionItem` guards (tombstone → not-found, invalid envelope, author enforcement via `PostNormalizer.toEdit`) → `PostApplication.commitEdit` (Dexie-first write, homeserver PUT, rollback on failure).
- `StreamPostsController.fetchMissingPostsByUris`: warms the local cache for envelope URIs in one `by_ids` request (≤100 ids = 1 fetch; only cache misses are sent).
- `collectionReorder` zustand store: minimal `activeCollectionId` flag bridging the page ↔ the globally-mounted FAB; cleared on save/cancel/unmount.

### UI
- `useReorderCollection` hook owns the draft (a deduped enter-time snapshot of envelope `items`), move/save/cancel, cache warming, and toasts.
- `CollectionReorderGrid` / `CollectionReorderCard` organisms: `DndContext` + `SortableContext` (`rectSortingStrategy`), sortable ids = envelope URIs, reusing the standard grid classes so the footprint matches the normal grid exactly. Deleted/missing posts render their standard placeholders and stay draggable; envelope items whose URI can't resolve to a post keep their slot with a fallback card (never silently dropped on save).
- `CollectionItems` branches to the reorder grid instead of `TimelineFeed` while active (which drops pagination, pull-to-refresh, and the trailing Add Post cell for free).

### Feed ordering correctness (post-save display)
Two complementary changes make the saved order display **immediately**, despite Nexus re-indexing the collection stream asynchronously:
1. **Envelope-order client sort** — the collection feed's post ids are sorted by the local-first envelope's `items` order (`sortPostIdsByCollectionOrder`, threaded via a `transformPostIds` prop). Order authority is local (instant after any commit); membership/existence authority stays with the stream.
2. **Eager full-membership loading** — collection feeds now fetch at the Nexus max page size (50) and eagerly load all pages (≤2–3 requests for a spec-max 100-item collection), so the sort always has complete membership. A defensive rounds cap (derived from `COLLECTION_ITEMS_MAX_COUNT / NEXUS_STREAM_MAX_LIMIT`) guards against a misbehaving backend; if it ever trips, the feed degrades to normal scroll-to-load. Applies only to the single collection page — all other feed variants keep lazy pagination (test-pinned). Heavy media stays viewport-lazy (`next/image` + `loading="lazy"` iframes), so eager loading costs metadata JSON, not a media waterfall.

### Robustness (from two review rounds)
- **Zero-drag saves can't clobber concurrent edits**: "did anything change" compares against the enter-time snapshot, not the live envelope — a save without drags never writes, even if another tab/device reordered meanwhile. A dragged save still wins (explicit intent) with the merge protecting concurrent adds/removes.
- **Third-party envelope tolerance**: duplicate item URIs are deduped on entry (they'd otherwise become duplicate React keys / sortable ids).
- **Moderation**: if the collection becomes blurred mid-reorder, the mode auto-cancels (the blurred hero has no Save/Cancel).
- Grid drop handler, cap, and reset behavior are all test-pinned; dnd-kit's global aria id counter is normalized in snapshots to keep baselines order-independent.

## 🌍 i18n

5 new keys (`collections.single.reorder` / `saveOrder` / `cancelReorder`, `toast.collection.orderSaved` / `orderSaveFailed`) added to all 9 locales.

## 📦 Dependencies

`@dnd-kit/core@6.3.1`, `@dnd-kit/sortable@10.0.0`, `@dnd-kit/utilities@3.2.2` (~10kb gz; purpose-built React sortable with keyboard a11y and touch sensors).

## ✅ Testing

- Unit + snapshot coverage across every layer: pipe merge semantics, controller guards, sort helper, hook state machine (enter/move/save/cancel/failure/concurrency/unmount), grid/card components, hero button states, `CollectionItems` branching, FAB gating, feed limit/eager-load/cap behavior.
- Full suite green: **712 files / 11,719 tests**, lint + typecheck clean.
- Manually QA'd: drag on desktop + touch, cross-page reorder on a 50+ item collection, save/cancel/error flows, disabled states, FAB behavior.
- No Cypress additions (e2e owned by QA); no VRT baselines exist for collection surfaces yet.

---


src/core/controllers/post/post.ts
```diff
@@ -469,6 +470,46 @@ export class PostController {
     });
   }
 
+  static async commitReorderCollectionItems({ collectionId, items }: TReorderCollectionItemsParams): Promise<void> {
+    const currentUserPubky = useAuthStore.getState().selectCurrentUserPubky();
+    const collection = await PostApplication.getDetails({ compositeId: collectionId });
```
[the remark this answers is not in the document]
secondl1ght: Valid observation, but the window is narrower than stated and pre-existing: the merge protects against all concurrent edits visible locally at save time (multi-tab is fully covered — IndexedDB is shared per-origin); the residual is remote-device edits not yet synced. That same window exists identically in every current envelope write (`commitUpdateCollectionItem`, and `commitEditCollection` carries `items` verbatim on any metadata edit) and in post editing generally — homeserver PUTs are last-writer-wins today. The planned app-wide fix is reading current state from the homeserver before writing, which resolves this for all edit paths at once; out of scope for this PR and tracked as a known limitation.

secondl1ght:
@greptile replied to comment plz review

infin1t3 (approved):
LGTM.

src/components/organisms/Collections/CollectionReorderGrid/CollectionReorderCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const t = useTranslations('post');
+  const { attributes, listeners, setNodeRef, transform, transition, isDragging } = useSortable({
+    id: entry.uri,
+    disabled,
+  });
+
+  return (
+    <Container
+      ref={setNodeRef}
+      {...attributes}
```
infin1t3, line 39: **A11y: every sortable card exposes as an unnamed button.**

dnd-kit's `attributes` put `role="button"` + `tabIndex=0` on this container, but the card's entire content sits inside the `inert` wrapper below — and inert subtrees are removed from the accessibility tree, so the button's accessible name computes to empty (visible in the committed snapshot: `role="button"` with no name). A screen-reader user tabbing through the grid hears N indistinguishable "button, sortable" stops with no way to tell which post is which — a WCAG 2.1 SC 4.1.2 (Name, Role, Value) failure.

The `inert` wrapper itself is good design (it's what keeps reply/repost/tags unreachable while reordering) — the fix is just to name the button: add a per-card `aria-label` on this outer `Container` (dnd-kit's `attributes` doesn't include `aria-label`, so there's no spread conflict). Minimally positional, e.g. `aria-label={"Post " + (index + 1) + " of " + total}`; ideally author + content snippet.

src/components/organisms/Collections/CollectionReorderGrid/CollectionReorderGrid.tsx
```diff
@@ -0,0 +1,63 @@
...
+  // inert, while the touch long-press delay leaves one-finger scrolling
+  // working. Keyboard: Space/Enter lifts, arrows move, Space/Enter drops.
+  const sensors = useSensors(
+    useSensor(MouseSensor, { activationConstraint: { distance: 8 } }),
+    useSensor(TouchSensor, { activationConstraint: { delay: 250, tolerance: 8 } }),
+    useSensor(KeyboardSensor, { coordinateGetter: sortableKeyboardCoordinates }),
+  );
+
+  return (
+    <DndContext
```
infin1t3, line 43: **A11y: default screen-reader drag announcements read the raw `pubky://` URI.**

With no `accessibility` prop on `DndContext`, dnd-kit's default announcements interpolate the sortable ids — here full ~70-char `pubky://…/pub/pubky.app/posts/…` URIs — so every pickup/move/drop is announced as a long character salad, twice per drag-over (`active.id` and `over.id`).

Suggested fix: pass `accessibility={{ announcements }}` with position-based text derived from `entries.findIndex`, e.g. "Picked up post. Post is in position 4 of 12." / "Post moved to position 2 of 12." Pairs with the per-card `aria-label` on `CollectionReorderCard` — the two together make keyboard reordering genuinely usable with a screen reader.

infin1t3 (commented):
Accessibility pass on the reorder grid — two should-fix findings, both small (~15 lines combined):

PR #2320, opened 2026-08-09
fix(image): keep aspect ratio of EXIF-rotated images during compression

### LINKED ISSUE #2268: [Bug]: Portrait images can stretch when attached to post
### Description

When attaching a portrait image to a post it can become stretched. This is probably caused by the image sanitation/compression performed in the FE.

Affected post: https://pubky.app/post/9634ggqes1jusuny1tfoewosmcny4kt1hr7ksy8cq1t95gm8kmeo/0035F3JP4Q1W0

Internal discussion: https://synonymworkspace.slack.com/archives/C09HQTLAARF/p1785127530867609

### Expected Behaviour

Image is sanitised and compressed but retains same dimensions.

### Steps to Reproduce

1. Take portrait image (or use the one attached)
2. Attach image to post
3. Observe the post in feed has stretched image (which was uploaded to hs this way)

### Screenshots

Original image:

Image

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
Portrait photos from phone cameras store sensor-native landscape pixels plus an EXIF orientation tag. The decode-time resize computed its target from the raw SOF dimensions while createImageBitmap applied the EXIF rotation, so 90-degree-rotated photos were squeezed into swapped-aspect frames and the distorted pixels were uploaded to the homeserver.

Swap the resize target for orientations 5-8 via a JPEG EXIF orientation parser, re-decode without resizing when the returned bitmap does not match the requested dimensions, and skip decode-time resizing for WebP that flags EXIF in VP8X. Consolidate the duplicated RIFF chunk and JPEG segment walkers into shared helpers.

Before state:
https://staging.pubky.app/post/8yb68y6n3enzq1tygzczdh5rowgjf8txm61greuifpb5wey86b8y/0035H8A6M7ER0

After state:
https://staging.pubky.app/post/8yb68y6n3enzq1tygzczdh5rowgjf8txm61greuifpb5wey86b8y/0035H8AF1SZJG

Resolves #2268


secondl1ght (approved):
Approving to fix this bug but we should do a follow-up at some point. All the custom compression/stripping code should be ripped out and replaced with a battle-tested and popular public lib.

catch-21:
Raised https://github.com/pubky/pubky-app/issues/2330 off the back of @secondl1ght's comment.

PR #2326, opened 2026-08-10
chore: bump deps (not pubky or pubky-app-specs)

catch-21:
## Intentionally not updated

- **`@synonymdev/pubky` (0.8.0) and `pubky-app-specs` (0.6.2)** — left for a follow-up. These are protocol/SDK packages; bumping them is a behaviour and compatibility change, not a routine patch bump.

- **ESLint 10** — `eslint-plugin-import@2.32.0` and `eslint-plugin-react@7.37.5` (both still latest) only declare peers through ESLint 9. `eslint-plugin-react` also crashes on ESLint 10 (`context.getFilename` was removed). Stay on `eslint@9.39.5` until those plugins (and ideally `eslint-config-next`) support v10.

- **TypeScript 7** — `typescript-eslint` 8.x still requires the 6.x compiler API (`>=4.8.4 <6.1.0`) and throws on TS 7.0. Native TS 7 support is expected around 7.1. Stay on `typescript@6.0.3`. `@typescript-eslint/parser` and `@typescript-eslint/eslint-plugin` are now direct devDependencies (8.67.0) so they are no longer a nested hoist from `eslint-config-next`.

- **`eslint-plugin-import` / `eslint-plugin-react`** — already at latest; the blocker is ESLint 10, not our pins.

## Compat notes (not product changes)

- `linkify-it` v6 defaults `fuzzyLink` to `false`. We pass `{ fuzzyLink: true }` in `PostLinkEmbeds` and `post.kind` so protocol-less URLs (`youtube.com/…`) still embed as before.
- Snapshots updated for Radix (closed triggers no longer set `aria-controls`) and Lucide v1 icon class/path changes (`lucide-smile` → `lucide-face-slightly-smiling`, etc.).


catch-21:
superceded by https://github.com/pubky/pubky-app/pull/2461

PR #2327, opened 2026-08-10
feat(users): add safe moderation bot default follow

### LINKED ISSUE #2255: [Task]: Ensure every user follows moderation bot
## Problem

Pubky App blurs moderated content only when the client receives a configured moderation tag authored by the configured moderation bot.

The reported failure mode is caused by tag selection, not by the introduction of Web of Trust (WoT): Nexus returns a limited set of tags per post (approximately five in the affected flow). If other tags rank ahead of the moderation tag, or the moderation tag arrives later, Pubky App may not receive it and therefore may fail to blur the content. Following the moderation bot prioritizes its tags and makes the existing blur behavior reliable.

This is a pre-existing moderation/blur reliability concern that is independent of WoT reach and domain feeds. Trust-aware tag selection in [pubky-nexus#1001](https://github.com/pubky/pubky-nexus/issues/1001) is related longer-term work, but it should not block the App-side fix.

## Desired behavior

- New accounts follow the environment-configured moderation bot automatically after successful account setup, regardless of the account's homeserver.
- Existing accounts receive a one-time backfill/migration so they are covered too.
- The follow remains ordinary user-controlled state: it is visible, synced through the normal local-first follow flow, and can be unfollowed.
- An explicit unfollow is not silently reversed on every sign-in. If mandatory recurring re-follow is desired, that must be approved as a separate product decision.
- The global blur setting continues to work independently of follow creation.

## Acceptance criteria

- [ ] The moderation bot Pubky is environment-configurable; the implementation does not hardcode a single environment's identity.
- [ ] Newly created accounts follow the configured moderation bot exactly once after the account becomes usable.
- [ ] Existing accounts, including identities signing in from another homeserver, receive an idempotent one-time migration.
- [ ] Accounts already following the moderation bot do not create duplicate follow state.
- [ ] A user who explicitly unfollows after the default follow or migration is not automatically re-followed on later sign-ins.
- [ ] Follow creation uses the existing local-first follow path and synchronizes normally with the homeserver.
- [ ] A follow or migration failure does not break account creation or sign-in and can be retried safely.
- [ ] Tests cover new-account follow creation, existing accounts, already-following accounts, explicit opt-out, and failure handling.

## References

- [Internal Slack discussion](https://synonymworkspace.slack.com/archives/C09HQTLAARF/p1784740485029789)
- [Earlier WoT-scoped issue and clarification: #2153](https://github.com/pubky/pubky-app/issues/2153#issuecomment-5049134283)
- [Related Nexus trust-aware tag selection: pubky-nexus#1001](https://github.com/pubky/pubky-nexus/issues/1001)

Orlandohub:
## Summary

- fix #2255
- New and existing accounts follow the moderation bot once during shared bootstrap, covering onboarding and sign-in without a separate migration pass.
- A durable homeserver marker preserves later explicit unfollows.
- An unset deployed `PUBKY_RUNTIME_MODERATION_ID` now means moderation is disabled instead of silently using the staging account.

## Changes

- `UserApplication.ensureModerationFollow` uses canonical homeserver state, validates the normalized URL with `followUriBuilder`, writes local follow → homeserver follow → marker, and cooperatively stops when its bootstrap signal is aborted.
- Manual moderation-bot unfollow writes the marker before local or remote deletion. This is intentionally network-gated: if the marker PUT fails (including while offline), the unfollow fails without changing local or remote follow state.
- `BootstrapApplication` owns the active `AbortController`; new bootstraps and all auth lifecycle transitions cancel older detached work.
- `PubkySpecsSingleton` now rebuilds when the requested identity changes.
- `HomeserverService.exists` uses `session.storage.exists` and `publicStorage.exists`; HTTP URLs retain status/session-expiration handling.
- Runtime moderation IDs are trimmed and validated as 52-character z-base-32 Pubkys. Deployed startup warns once when moderation is unset; lenient dev/test parsing retains the validated staging default.
- `detectModerationFromTags` returns `false` when no moderation ID is configured.

## Test plan

- [x] Full unit suite — 715 files; 11,906 passed, 2 skipped
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `git diff --check`
- [ ] Sign in with the staging recovery phrase on a fresh profile; confirm the moderation bot appears in Following and the marker exists
- [ ] Sign in again; confirm no duplicate follow or second marker write
- [ ] Unfollow the bot, sign out, and sign back in; confirm the unfollow sticks
- [ ] Run a production-mode deployment with the moderation ID unset; confirm startup warns and moderation remains disabled

## Notes

- **Historical unfollows:** users who unfollowed the bot before this marker existed cannot be distinguished from never-migrated users and may be followed once during rollout. This is an accepted one-time tradeoff.
- **Backend impact:** rollout can generate one follow/indexer event and corresponding notification per eligible existing account.
- **Following cache:** `activeStreamId: undefined` is intentional. Dexie is cleared before bootstrap, bootstrap does not populate Following caches, and follow invalidation does not touch `TIMELINE_ALL_ALL`.
- **Deployment manifests:** current production explicitly uses `euarj183st16psbzusb44usk8i1k5foubnw7zt3n3j8h9zxg6hro`; staging and preview use `nto4u7kkagk5hfjk4wgueemzy61nssic811hid1ty9u81uatmqzy`.
- **Private testnet:** its moderation-account configuration is being handled by the `pubky-stack`/`pubky-docker` owner; no app-side Compose override is included.
- **Marker namespace:** `pub/pubky.app/migrations/moderation-follow/v1/<moderationId>.json` is app-defined and accepted by `pubky-app-specs` as an unknown resource, matching existing app-owned-resource precedent.
- No UI surfaces changed, so no VRT baselines need regenerating.
- Shared Pubky-validator consolidation remains intentionally out of scope.


src/core/application/user/user.ts
```diff
@@ -256,6 +262,37 @@ export class UserApplication {
...
+        eventType: HttpMethod.PUT,
+        followUrl: meta.url,
+        followJson: follow.toJson(),
+        follower,
+        followee: moderationId,
+        activeStreamId: undefined,
+      });
+    }
+
+    await HomeserverService.request({
```
infin1t3: **The "already done" note can be lost, which later turns into an unwanted re-follow.**

The follow (line 279) and this note write are two separate server calls, and the note goes last.

If the follow lands but this call fails (flaky connection, tab closed mid-way), the account ends up following the bot with no note. 

If the user then unfollows the bot, their next sign-in finds no note and no follow — which looks exactly like an account that was never processed — so the bot gets followed again. That silently undoes the user's explicit choice, which the ticket rules out ("An explicit unfollow is not silently reversed").

**Suggested fix:** record the user's choice at the moment they make it — when someone unfollows the moderation bot, write this same note (or an equivalent "user opted out" note) as part of the unfollow. Then even if this write failed earlier, an explicit unfollow durably closes the door on re-following. Retrying this write a couple of times here also shrinks the window, but only the unfollow-side note removes it completely.

src/core/application/user/user.ts
```diff
@@ -256,6 +262,37 @@ export class UserApplication {
...
+    const { meta, follow } = FollowNormalizer.to({ follower, followee: moderationId });
+    const followExists = await HomeserverService.exists(meta.url);
+    if (!followExists) {
+      await this.commitFollow({
+        eventType: HttpMethod.PUT,
+        followUrl: meta.url,
+        followJson: follow.toJson(),
+        follower,
+        followee: moderationId,
+        activeStreamId: undefined,
```
infin1t3, line 332: **Passing nothing here makes the cleanup delete the feed the user is currently looking at.**

Creating a follow clears the saved "following" feeds so they can rebuild with the new account included. The button-click path (`UserController.commitFollow`) passes the id of the feed currently on screen so that one is kept and the user's view survives. Here nothing is passed, so every saved feed is deleted — including the one that was just rendered on /home seconds earlier. The feed goes blank and re-downloads from Nexus, which likely doesn't know about this follow yet, so the work sign-in just did is thrown away and the user sees an unexplained blank/flickering feed. Since every existing user goes through this once after release, everyone will see it.

**Suggested fix:** pass the currently-viewed feed id here the same way the button path does — thread the value the controller gets from `getActiveStreamId()` into this call, or accept it as a parameter from the caller.

src/core/application/user/user.ts
```diff
@@ -256,6 +262,37 @@ export class UserApplication {
...
+   */
+  static async ensureModerationFollow({ follower, moderationId }: TEnsureModerationFollowParams): Promise<void> {
+    if (!moderationId || follower === moderationId) return;
+
+    const markerUrl = `${baseUriBuilder(follower)}migrations/moderation-follow/v1/${moderationId}.json`;
+
+    if (await HomeserverService.exists(markerUrl)) return;
+
+    const { meta, follow } = FollowNormalizer.to({ follower, followee: moderationId });
+    const followExists = await HomeserverService.exists(meta.url);
```
infin1t3, line 322: I think we can go with this and that it's not a big deal, just wanted to call it out. Feel free to just dismiss.

**People who unfollowed the bot before this release will be re-followed by the rollout.**

The note this feature relies on only starts existing with this release. So for anyone who followed the moderation bot and deliberately removed it in the past, there is nothing that distinguishes them from an account that was never processed: no note, no follow. The one-time migration will therefore follow the bot for them again, silently reversing a choice they already made.

This may be an acceptable one-time cost of rolling the feature out — but the ticket asks for exactly this kind of thing to be an explicit product decision ("If mandatory recurring re-follow is desired, that must be approved as a separate product decision"), and right now it isn't mentioned in the PR description or anywhere else.

**Suggested fix:** either get the trade-off acknowledged in the PR description / ticket so it's a deliberate decision, or soften it in-app — for example a one-time notice ("we added the moderation bot to your follows") so the reversal at least isn't silent.

src/core/application/user/user.ts
```diff
@@ -256,6 +262,37 @@ export class UserApplication {
...
+   * explicit unfollows, while the canonical follow resource makes retries idempotent.
+   */
+  static async ensureModerationFollow({ follower, moderationId }: TEnsureModerationFollowParams): Promise<void> {
+    if (!moderationId || follower === moderationId) return;
+
+    const markerUrl = `${baseUriBuilder(follower)}migrations/moderation-follow/v1/${moderationId}.json`;
+
+    if (await HomeserverService.exists(markerUrl)) return;
+
+    const { meta, follow } = FollowNormalizer.to({ follower, followee: moderationId });
```
infin1t3, line 312: Edge case

**The note and the follow can end up belonging to two different accounts.**

The note address (line 272) is built directly from the `follower` value passed into this function. The follow address, though, comes from `FollowNormalizer.to`, which relies on `PubkySpecsSingleton` — a shared helper that keeps answering for whichever account it was first set up with, ignoring the `follower` it is handed. Because this routine runs in the background and nothing stops it when someone signs out (see the `void` call in `bootstrap.ts`), it can still be mid-flight when a different account signs in. At that point the two addresses point at two different people, and there are two bad endings depending on timing: the new account gets marked "done" without ever following the bot (and since the note now exists, nothing ever retries — moderation quietly never works for them), or a follow is recorded locally that the server then rejects, leaving the UI showing a follow that doesn't really exist.

**Suggested fix:** build both addresses from the same source — the `follower` passed in. If the shared helper can't be made to honor that argument, at least check that the address it returns starts with `pubky://<follower>/` and bail out of the whole routine if it doesn't. Cancelling the background task on sign-out (bootstrap/auth side) would close the remaining gap.

src/core/application/bootstrap/bootstrap.ts
```diff
@@ -98,12 +100,25 @@ export class BootstrapApplication {
       LocalStreamTagsService.upsert(TagStreamTypes.TODAY_ALL, bootstrapData.ids.hot_tags),
     ]);
     onProgress?.('dataPersisted'); // Step 4 complete (80%)
+    void this.ensureModerationFollow(pubky);
```
infin1t3: **This background job keeps running after the user signs out — and can write into the next account's data.**

`void` here means: start the job and don't wait for it. Nothing stops that job when the user signs out, and signing in as someone else first wipes and then refills the local database. So a job started for account A can still be mid-flight while account B is signing in, and A's follow gets recorded into B's freshly prepared local data. B's screen then shows the moderation bot as "Following" even though B never followed anyone — and the server write that follows fails, so the display never matches reality.

**Suggested fix:** before each step of this job (and especially before saving anything), check that the account it was started for is still the signed-in account, and quietly stop if not. Cancelling the job on sign-out achieves the same from the other side; doing both is cheap insurance.

src/libs/runtime-config/runtime-config.schema.ts
```diff
@@ -231,7 +234,7 @@ export const runtimeConfigValueSchema = networkConfigValueSchema.extend({
   ttlPostMaxBatchSize: positiveIntValue.default(APP_RUNTIME_DEFAULTS.ttlPostMaxBatchSize),
   ttlUserMaxBatchSize: positiveIntValue.default(APP_RUNTIME_DEFAULTS.ttlUserMaxBatchSize),
   ttlRetryDelayMs: positiveIntValue.default(APP_RUNTIME_DEFAULTS.ttlRetryDelayMs),
-  moderationId: nonEmptyStringValue.default(APP_RUNTIME_DEFAULTS.moderationId),
+  moderationId: pubkyValue.optional(),
```
infin1t3, line 239: **Deployments that relied on the old built-in fallback will silently lose moderation.**

Until now, an environment that never set `PUBKY_RUNTIME_MODERATION_ID` still got a working value, because the code shipped with a built-in fallback. This change removes that fallback for deployed environments: unset now means "no moderation bot at all". The app boots normally, but blurring of flagged content stops working and the new default follow never happens — with no error or warning anywhere. Whoever operates an environment that relied on the fallback won't find out until users see unblurred content.

**Suggested fix:** before release, make sure every deployment manifest sets this variable (the preview one now does — staging/production need the same check), and add a loud startup warning when the value is missing in a deployed environment, so a gap like this shows up in logs instead of staying invisible.

src/libs/runtime-config/runtime-config.schema.ts
```diff
@@ -34,6 +34,9 @@ const testnetValue = z.boolean();
 const sampleRateValue = z.number().min(0).max(1);
 const positiveIntValue = z.number().int().positive();
 const nonEmptyStringValue = z.string().min(1);
+const pubkyValue = z
```
infin1t3, line 37: **A nearly-right value stops the app from starting, while a blank one silently turns moderation off.**

Two edge cases around this new check. First: the value arrives untrimmed, and this 52-character check rejects anything with extra whitespace — so a correct id with a stray trailing newline or space (easy to pick up from secret managers and env files) now prevents the whole app from starting, where the previous release booted fine. Second, the opposite: a value that is only whitespace gets quietly converted to "not set" before reaching this check, so the app boots and moderation is just... off — even though the docs for this variable say a configured value must be a valid Pubky, and every other wrong value fails loudly.

**Suggested fix:** trim before validating — `z.string().trim().regex(...)` — so padded-but-correct values work. After trimming, a blank value behaves like an unset one; pair that with a startup warning when the value ends up unset in a deployed environment (see the comment below on the removed fallback) so "off" is at least visible.

src/libs/runtime-config/runtime-config.schema.ts
```diff
@@ -257,14 +260,19 @@ export const runtimeConfigValueSchema = networkConfigValueSchema.extend({
   playStoreUrl: urlValue.default(APP_RUNTIME_DEFAULTS.playStoreUrl),
 });
 
+const lenientRuntimeConfigValueSchema = runtimeConfigValueSchema.extend({
+  // Zod v4 `.default()` bypasses validation; `.prefault()` sends the staging fallback through pubkyValue.
+  moderationId: pubkyValue.prefault(APP_RUNTIME_DEFAULTS.moderationId),
```
infin1t3, line 268: **Every fresh dev/test account will follow an account that doesn't exist in that environment.**

In non-deployed environments (dev, tests, E2E), an unset value falls back to the staging moderation bot id — but the private test stack has no such account. The follow write still succeeds (it only writes into the new account's own storage), so every freshly created account in E2E ends up "following" a ghost: following count of 1, an unresolvable profile in the Following list, and any test assuming a brand-new account follows nobody breaks.

**Suggested fix:** don't fall back to the staging id where that account can't exist — leave the value unset in testnet/E2E setups so the default follow simply skips itself (the code already does nothing when there's no id), or have the test stack create and configure its own moderation account.

src/libs/runtime-config/runtime-config.schema.ts
infin1t3, line 37: **This is now the third place in the codebase that defines what a Pubky id looks like.**

The same alphabet string already exists in `sentry.constants.ts` (as `Z32_ALPHABET`, used to build the redaction patterns), and a third, looser check lives in `utils.ts` (`isPubkyIdentifier`). Three independent definitions of one format will drift: if it ever needs adjusting, someone has to remember all three, or config validation, error-report scrubbing, and UI checks start disagreeing about what a valid id is.

**Suggested fix:** move the alphabet into one shared constants file with no imports of its own and build all three from it (`sentry.constants.ts` currently imports nothing, so it could become — or feed — that shared spot).

src/core/services/homeserver/homeserver.ts
```diff
@@ -512,6 +512,37 @@ export class HomeserverService {
     }
   }
 
+  /**
+   * Checks whether a homeserver resource exists without treating an expected 404 as an error.
+   * Handles both SDK rejections and resolved non-OK Responses before either can reach Err.*.
+   */
+  static async exists(url: string): Promise<boolean> {
```
infin1t3, line 519: **This hand-builds something the SDK already provides.**

The pubky SDK has a direct way to ask "does this exist?" — `session.storage.exists(path)` and `publicStorage.exists(address)` both return a plain yes/no. This method instead downloads the resource and then interprets two different shapes of "not found" by hand (a thrown error carrying a status code, and a response object with a status field) — about 30 lines that must keep matching the SDK's behavior, and that transfer the file's contents just to answer yes or no.

**Suggested fix:** call the SDK's `exists` for the two storage branches, and keep only the plain-URL branch and the genuine-error handling here.

src/core/services/homeserver/homeserver.test.ts
```diff
@@ -924,6 +926,62 @@ describe('HomeserverService', () => {
         });
       });
     });
+
+    describe('exists', () => {
```
infin1t3, line 936: **The riskiest path of `exists()` has no test.**

These tests cover the case where the user's own signed-in storage answers. But the whole reason `exists()` gets called from the background follow is the other case — no session, or a resource belonging to someone else — where the code falls back to public reads. That branch, and the plain-URL branch, are untested (compare `get()`, which does test its public branch). If the public branch ever mishandles a "not found" answer, the background follow would misread it as a real failure, and nothing here would catch it.

**Suggested fix:** add cases where the session is absent or doesn't own the address so the public-read branch runs — one for "exists", one for "not found", one for a genuine error — plus one for the plain http URL branch.

src/core/application/user/user.types.ts
```diff
@@ -13,3 +14,8 @@ export type TUserApplicationFollowParams = TFollowParams & {
   followJson: Record<string, unknown>;
   activeStreamId?: PostStreamId | null;
 };
+
+export type TEnsureModerationFollowParams = {
```
infin1t3, line 19: **This type has exactly one user — it can live where it's used.**

Only `user.ts` imports it; the caller in `bootstrap.ts` just passes a plain object and never references the type. The convention for these sibling type files is to hold types shared by several places; a shape used once reads easier written right at the function that takes it. (To be fair, `TUserApplicationFollowParams` above is in the same situation, so the file already carries that precedent — take or leave.)

**Suggested fix:** inline the two-field shape into the function signature in `user.ts`.

src/core/application/user/user.test.ts
```diff
@@ -143,6 +147,225 @@ describe('UserApplication.commitFollow', () => {
...
+      asOpaque<ReturnType<typeof FollowNormalizer.to>>({
+        meta: { url: followUrl },
+        follow: { toJson },
+      }),
+    );
+    return { normalize, toJson };
+  };
+
+  beforeEach(() => {
+    vi.restoreAllMocks();
```
infin1t3: **Two of these three reset calls do nothing.**

The `afterEach` below already returns every replaced function to its original after each test. So by the time `beforeEach` runs there is nothing left to restore — its `restoreAllMocks` is a no-op, and the `clearAllMocks` right after it clears call history on functions that are no longer replaced. Harmless today, but it muddies which reset actually matters and tends to get copy-pasted into the next test file.

**Suggested fix:** keep only `afterEach(() => vi.restoreAllMocks())` and delete the `beforeEach` block.

infin1t3 (changes requested):

src/core/application/user/user.ts
Orlandohub: Kept `activeStreamId: undefined` intentionally. Dexie is cleared before bootstrap, bootstrap does not populate Following caches, and this invalidation does not touch `TIMELINE_ALL_ALL`, so there is no rendered Following stream to preserve.

src/core/application/user/user.ts
Orlandohub: Acknowledged and documented in the PR: historical pre-marker unfollows may be followed once during rollout. This is an accepted one-time tradeoff.

src/core/application/user/user.ts
Orlandohub: Fixed in 41014f1e: moderation-bot unfollow writes the marker before either deletion, and marker failure leaves both states unchanged. Added the failed-marker → manual-unfollow → no-refollow regression.

src/core/application/user/user.ts
Orlandohub: Fixed in 41014f1e: the specs builder now follows the requested identity, the normalized URL is checked with `followUriBuilder`, and auth-lifecycle cancellation prevents cross-account continuation.

src/core/application/bootstrap/bootstrap.ts
Orlandohub: Fixed in 41014f1e: bootstrap owns an `AbortController`, invalidates older runs, and all five auth lifecycle transitions cancel it. The workflow rechecks cancellation after probes and before writes.

src/libs/runtime-config/runtime-config.schema.ts
Orlandohub: Fixed in 41014f1e: deployed startup warns once when moderation is unset. I also verified and recorded the explicit production, staging, and preview moderation IDs in the PR.

src/libs/runtime-config/runtime-config.schema.ts
Orlandohub: Fixed in 41014f1e with trim-before-validation plus padded-value and deployed-unset warning coverage.

src/libs/runtime-config/runtime-config.schema.ts
Orlandohub: Agreed. The `pubky-stack`/`pubky-docker` owner is handling private-testnet moderation configuration on his side, so no app-side Compose override is included. The zero-follow Cypress expectation remains.

src/libs/runtime-config/runtime-config.schema.ts
Orlandohub: Agreed on the duplication, but keeping validator consolidation out of this PR to avoid broad routing, identity, Sentry, and avatar-validation scope.

src/core/services/homeserver/homeserver.ts
Orlandohub: Fixed in 41014f1e: owned/public resources now use the SDK `exists()` methods; fetch/status handling remains only for HTTP URLs.

src/core/services/homeserver/homeserver.test.ts
Orlandohub: Added owned, public, and HTTP coverage for exists, missing resources, genuine failures, and session expiration.

src/core/application/user/user.types.ts
Orlandohub: Kept the named type to match the adjacent application parameter contract and keep the signal-bearing workflow shape explicit. No functional change here.

src/core/application/user/user.test.ts
Orlandohub: Fixed in 41014f1e: restore cleanup is centralized at file scope and the redundant suite-level reset block is removed.

infin1t3 (approved):
https://github.com/user-attachments/assets/40872b1d-b2a4-452c-920f-cdca26badda0
https://github.com/user-attachments/assets/52315f09-c5a3-4d14-a9a4-a7850fdf212e
https://github.com/user-attachments/assets/0c13f334-3108-44e5-8773-9dabeee34cb9

Code looks ok, multiple tested scenarios also. Approved.

PR #2333, opened 2026-08-11
fix(posts): skip optimistic feed prepend on WoT-sourced streams

### LINKED ISSUE #2308: [Bug]: Post optimistically added at top of feed when not matching filter
### Description

Same as previosuly fixed https://github.com/pubky/pubky-app/issues/1554

A newly created post shows at the top of the feed when the 'My network' and 'Tagged as' Reach options applied, which do not normally show this. Then changing filters from and back to this Reach option doesn't show the new post.

### Expected Behaviour

New post is only optimistically put at top of feed when filter would show it.

### Steps to Reproduce

1. Have 'Tagged as' Reach applied
2. Create a new post without any tags from the quick post area at the top
3. Observe post is optimistically displayed at top of the feed

### Screenshots

Image

https://github.com/user-attachments/assets/aa26200e-cc52-4e1f-a9d3-a13606635d9b

### Console Log

```shell

```

### Browser & Version

_No response_

Orlandohub:
## Summary
- fix #2308
- A newly created post flashed at the top of the feed with the 'My network' or 'Tagged as' Reach filter active, then vanished after toggling filters — the optimistic prepend gate only checked content kind, never the stream source.
- WoT-sourced streams (`wot`, `wot_domain`) never contain the viewer's own posts: Nexus excludes them server-side (`author.id <> observer_id`) and the local create path never writes into those streams, so the optimistic entry was guaranteed inconsistent.

## Changes
- `postStream.types.ts`: add `isWotStream` (symmetric with `isWotDomainStream`) and `isViewerExcludedWotStream`, documenting why WoT streams never carry the viewer's own posts.
- `usePostInput.ts`: skip the optimistic prepend when the active stream is WoT-sourced, before the kind lookup; composer collapse and `onSuccess` behavior unchanged. Covers both POST and REPOST variants.
- `NewPostsSection.tsx`: comment-only — note the intentional divergence (unread ids come from polling the WoT stream itself, so they belong and are not source-gated).
- Tests: `isWotStream`/`isViewerExcludedWotStream` cases; `usePostInput` no longer prepends on `timeline:wot:all` / `timeline:wot_domain:2:all:bitcoin` even when the kind matches, still prepends on All/Following/Me streams, and still collapses when skipped. Removed the old wot_domain kind-mismatch test, superseded by the stronger source gate.

https://github.com/user-attachments/assets/b2c58779-3626-4ef0-89c2-ca6db42f508d

## Test plan
- [ ] Manual: apply 'Tagged as' (or 'My network') Reach, create a post from the quick post area — post must NOT appear at top of feed; switch to All/Following/Me — post appears

## Notes
- Own posts still show optimistically on All, Following, Friends, and Me — Following/Friends are persisted into local streams by `LocalPostService.updatePostStream`, so they are self-consistent. Whether own posts *should* appear on Following/Friends at all is a separate product question worth its own issue.

### DIFF
--- a/src/hooks/usePostInput/usePostInput.test.ts
+++ b/src/hooks/usePostInput/usePostInput.test.ts
@@ -724,16 +724,22 @@ describe('usePostInput', () => {
       expect(mockOnSuccess).toHaveBeenCalledWith('created-post-id');
     });
 
-    it('does not prependPosts when the created post kind does not match a wot_domain stream kind', async () => {
+    /*
+      WoT-sourced streams are gated before the kind lookup (#2308), which also
+      covers the former kind-mismatch case on `timeline:wot_domain:…:image`.
+    */
+    it.each([
+      ['wot (My network)', 'timeline:wot:all'],
+      ['wot_domain (Tagged as)', 'timeline:wot_domain:2:all:bitcoin'],
+    ])('does not prependPosts on a %s stream even when the kind matches (#2308)', async (_label, streamId) => {
       mockContent = 'Test content';
       mockPost.mockImplementation(async ({ onSuccess }) => {
         onSuccess('created-post-id');
       });
       vi.mocked(useTimelineFeedContext).mockReturnValue({
         ...mockTimelineFeedContext,
-        streamId: 'timeline:wot_domain:2:image:bitcoin' as PostStreamId,
+        streamId: streamId as PostStreamId,
       });
-      vi.mocked(PostController.getDetails).mockResolvedValue({ kind: 'short' } as never);
 
       const mockOnSuccess = vi.fn();
       const { result } = renderHook(() =>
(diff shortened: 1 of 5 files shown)


catch-21:
With the current nexus configuration, this seems functionally correct to me.

https://github.com/user-attachments/assets/cce3c8e0-dbcb-40d0-bc83-529d4be6fb1d

With the proposed new config for nexus in https://github.com/pubky/pubky-nexus/issues/1025, we'll need to change pubky-app again for `Following`, `Friends` and `Tagged as`

src/hooks/usePostInput/usePostInput.test.ts
```diff
@@ -748,12 +754,75 @@ describe('usePostInput', () => {
...
-        expect(PostController.getDetails).toHaveBeenCalledWith({ compositeId: 'created-post-id' });
+        expect(mockOnSuccess).toHaveBeenCalledWith('created-post-id');
       });
+      // The gate short-circuits before the kind lookup
+      expect(PostController.getDetails).not.toHaveBeenCalled();
       expect(mockPrependPosts).not.toHaveBeenCalled();
+    });
+
+    it.each([
+      ['following', PostStreamTypes.TIMELINE_FOLLOWING_ALL as string],
```
catch-21, line 765: Test isn't covering 'Friends'

catch-21 (approved):

PR #2334, opened 2026-08-12
fix (Bug): Add tag focus lost when moving mouse cursor #2291

### LINKED ISSUE #2291: [Bug]: Add tag focus lost when moving mouse cursor
### Description

When typing into an add tag input field, if the mouse cursor is moved (possibly only when ~to a different component that also has a tag input field~ a popover is triggered, such as hovering over profile name in post) then the focus is lost from the input field currently being used.

Happens at least on MacOS with Chrome.

### Expected Behaviour

Tag input field focus never lost just by moving the mouse cursor. Only when clicking away.

### Steps to Reproduce

1. View feed
2. Click on the `+` to add a tag to either a new post (quick post area at top) or an existing post in the feed
3. Type some characters
4. Move mouse cursor away, over a profile name in post to trigger popover
5. Observe blinking text cursor disappears and text can no longer by typed into the add tag input field

### Screenshots

https://github.com/user-attachments/assets/fbb51ec6-7787-4328-9a19-485db5a1f984

### Console Log

```shell

```

### Browser & Version

_No response_

Abhishek7Tech:
Closes #2291

When cursor is moved to popover the cursor lose focus as the blur function in the handleMouseLeave get triggered. 

### Screenrecording

https://github.com/user-attachments/assets/c0e28740-96c7-4134-9dbc-b9cf42a6c0ab

### DIFF
--- a/src/components/atoms/Popover/Popover.tsx
+++ b/src/components/atoms/Popover/Popover.tsx
@@ -40,15 +40,6 @@ function Popover({
     }
   };
 
-  const blurActiveElement = () => {
-    setTimeout(() => {
-      const activeElement = document.activeElement as HTMLElement;
-      if (activeElement?.blur) {
-        activeElement.blur();
-      }
-    }, 0);
-  };
-
   const handleMouseEnter = () => {
     // Cancel any pending close
     if (closeTimeoutRef.current) {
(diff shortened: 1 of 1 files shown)


infin1t3:
Closing in favor of #2361

PR #2339, opened 2026-08-12
feat: add JSON-LD for AI agents and crawlers to learn of Nexus-Scout

catch-21:
## Summary

Publishes schema.org JSON-LD so AI agents/crawlers that land on any pubky.app page — via a
plain HTTP fetch, no JS execution required — learn that the Pubky social graph can be queried
through [nexus-scout](https://nexus-scout.pubky.app), and where its full usage guide lives.

- `StructuredData` (new molecule) emits a `<script type="application/ld+json">` with a
  `@graph` of `WebSite` + `Organization` + `WebAPI` entities. `WebAPI` is the schema.org type
  for exactly this — an API accessible over web technologies, described via its
  `documentation` property (same pattern Google uses for its own Knowledge Graph API).
- Rendered from the root layout as a **sibling** of `DatabaseProvider` / `RouteGuardProvider`,
  not a descendant. Those are client components that gate `{children}` behind
  IndexedDB/auth readiness (rendering a loading spinner during SSR), so anything placed inside
  them is missing from the initial server-rendered HTML — verified by curling the dev server
  before and after moving it. Sitting outside that gate puts it in the same guaranteed-
  synchronous tier as the existing `<head>` metadata, on every route.
- Added `nexusScoutUrl` as a new deployer-overridable runtime config value (`PUBKY_RUNTIME_NEXUS_SCOUT_URL`,
  defaults to `https://nexus-scout.pubky.app`), following the exact same pattern as the
  existing `githubUrl` / `pubkyCoreUrl` external links.

## Why

nexus-scout already documents itself well for agents (`/llms.txt`, `/v1/schema`), but nothing
on pubky.app pointed at it. JSON-LD was chosen over (or alongside) `/llms.txt` because it's
consumed far more widely — Google and other engines already parse schema.org markup — whereas
`llms.txt` adoption/consumption is still narrow and mostly limited to dev-tooling agents.

## Changes

- `src/components/molecules/StructuredData/StructuredData.tsx` — new component
- `src/app/layout.tsx` — render `<StructuredData />` outside the DB/auth-gated tree
- `src/libs/runtime-config/runtime-config.schema.ts` / `runtime-config.ts` — add `nexusScoutUrl`
  (schema, default, env name, getter)
- `src/config/externalLinks.ts` — re-export as `getNexusScoutLink`
- `.env.example` — document `PUBKY_RUNTIME_NEXUS_SCOUT_URL`

## Test plan

- [x] Fetched the dev server's raw HTML (`curl`, no JS) and confirmed the `<script
      id="pubky-structured-data" type="application/ld+json">` tag is present verbatim in the
      initial response on `/` and on a second route (`/sign-in`), and that its content is
      valid, parseable JSON.
- [x] `npm run lint` / `npm run typecheck` / `npm run test`
- [ ] Confirm `PUBKY_RUNTIME_NEXUS_SCOUT_URL` override behaves as documented for a
      self-deployed instance pointing at a different nexus-scout/graph


Orlandohub (dismissed):
Looks good! Two small findings from AI review

[1] medium | simplification | src/config/externalLinks.ts:30
What's wrong: Comment references a non-existent LandingStructuredData component and says the URL is surfaced “on the landing page”. Why it matters: The only consumer is StructuredData in the root layout, so it ships on every route. The stale name sends readers looking for a file that does not exist. Fix: Replace with something like: “surfaced to AI agents via StructuredData in the root layout”.

[2] low | simplification | src/app/layout.tsx:41
What's wrong: layout.tsx and StructuredData.tsx both explain why the script sits outside DatabaseProvider / RouteGuardProvider. Why it matters: The same placement rationale in two places will drift — the layout comment already omits the spinner/no-JS detail that the component JSDoc includes. Fix: Keep the layout JSX comment (it sits at the placement decision) and drop the second JSDoc paragraph in StructuredData.tsx, leaving the schema.org / WebAPI explanation.

Orlandohub (approved):

PR #2352, opened 2026-08-14
feat: dynamic open-graph images for collections

### LINKED ISSUE #2321: [Task]: OG image preview for Collection url
### Task Description

When a url link to a Collection is shared in Slack, etc.  the generic pubky.app image is used.
We should show the Collection's background image instead. If no background image, one from a post?

### Affected Layer(s)

_No response_

### Affected Domain(s)

Posts

### Acceptance Criteria

- [ ] Sharing url link in Slack (at least) shows relevant image when one is available for the Collection, not placeholder.

### Technical Details

_No response_

### Testing Strategy

_No response_

secondl1ght:
# ✨ Dynamic Open Graph image for collections

Follow-up to the dynamic OG images PR (#2233): fills in the deferred `renderCollectionOg` now that the Figma design has landed. Shared collection links (and collection-kind posts shared via `/post` URLs) now render a real preview card instead of the static fallback. Both entry points — the `/collections/[userId]/[postId]` image routes and the `renderPostOg` short-circuit for `kind: collection` posts — were already wired to this renderer, so no route changes were needed.

## 🖼️ What renders

The standard OG header (avatar + display name + brand mark, with the shortened-pubky and Facehash fallbacks shared with posts) over a scaled-up version of the app's `CollectionCard`:

- **Cover image** — `object-cover`, centered, behind the content, with the card's exact darkening gradient (`to right, 0.7 → 0.35`); cover-less collections fall back to the muted card background
- **Title row** — `Library` icon + collection name (single line, ellipsized) | "N POSTS" count pill (compact notation, `CollectionCountBadge`'s tone contract: `background` on covered chrome, `card` on the muted shell) + owner avatar
- **Description** — single line, Regular (400) weight, ellipsized (matches the Figma card)
- **Tag chips** — top 3 tags by tagger count from Nexus, colored with the app's label-derived palette pre-blended to the opaque color `PostTag` produces in-app, with compact counts

Interactive chrome from the app card (Follow/Unfollow, the add-tag button) is intentionally omitted — the Figma's inner card was a screenshot of the live component, and a static image can't host controls.

## 🧠 Behavior & decisions

- **Dynamic height** — the card collapses to fit whichever rows exist, like the app card: title-only collections render a compact card, description/tags each add their fixed-height row. Bottom-anchored 64px off the frame per the design.
- **Kind gate** — `renderCollectionOg` requires `post.kind === 'collection'`; probing the collections URL with an arbitrary post whose content happens to parse as a collection envelope gets the static fallback, mirroring the page's `generateMetadata` and `renderPostOg`'s positive check.
- **Chip width budget** — satori neither wraps nor shrinks flex children, so chips go through the app's own `getDisplayTags` (36-char label budget, max 3) and trailing chips are dropped rather than clipped; the row also clips at the content box as a belt.
- **Tags fetch** — one extra non-fatal Nexus call (`postApi.tags`, `limit_tags=3`, server-ordered by popularity — same order the app shows); failures collapse to an empty list and the card renders without the row. Kicked off before the user/post await so it overlaps the first round-trip on cold renders.
- **SSRF stance maintained** — covers resolve through the same hardened attachment path from #2233: only `pubky://` file URIs (our own CDN) are fetched server-side; a legacy absolute `http(s)` cover renders in-app but the OG card falls back to the cover-less background.
- **Fallbacks** — missing post/user, non-collection kind, deleted posts, and unparsable envelopes all serve the default static preview; caching (segment `revalidate`, upstream fetch revalidate, `stale-while-revalidate` headers) is unchanged from #2233.

## 🔬 satori constraints discovered (documented in code)

This card surfaced three new satori/yoga behaviors, now encoded as comments for future OG work:

1. **Absolutely-positioned children need explicit pixel dimensions** — inset-anchored or percentage-sized absolute layers silently collapse inside content-sized parents (why the card height is computed up front from fixed row heights instead of left to flex).
2. **`background-size: cover` mis-computes nested elements' boxes** — it scales against a pre-layout box and tiles the remainder (a visible repeat seam), which is why the cover is an `` with `object-cover` rather than a CSS background.
3. **Explicitly-`undefined` style values crash at stream time** — satori parses every present style key, and the failure bypasses the renderer's try/catch (hard 500 to crawlers); conditional styles must omit keys via spread.

Also: the parent's rounded `overflow: hidden` clip doesn't reach absolute children, so each cover layer rounds its own corners.

## 🔧 Supporting changes

- **`ogData.ts`** — new `fetchPostTags(userId, postId, limit)` helper on the shared `fetchWithValidation` path
- **`ogConstants.ts`** — new `background` token (`#05050a`, converted from the `--background` oklch value) + truncation entries for the collection description and tag labels
- **`OgIcons.tsx`** — inlined lucide `Library` icon
- **`ogFonts.ts` / assets** — bundled Inter Tight Regular (400) for the description weight, sourced like the existing Medium/Bold
- Chip colors reuse the pure `generateRandomColor` from the app; the opaque blend (`0.3·tag + 0.7·background`) reproduces `PostTag`'s browser-composited result

## ✅ Testing

- **New unit tests**: `fetchPostTags` (endpoint path + author/post order + limit asserted, 404 → `[]`, network failure → `[]`); `solidTagColor` blend math against known label colors; `TagChip` structural render + snapshot (label, compact count, opaque background, grapheme truncation); `getDisplayTags` (previously uncovered — count cap, char budget, truncated-length accounting, first-tag-always-kept)
- 253 tests passing across the OG and utils suites; typecheck and lint clean
- **Verified live against staging**: cover + description + tags card, cover-less title-only card (collapsed height, muted-tone pill), `/post/…` short-circuit rendering the identical card, unknown/bad ids 307-ing to the static preview, and pixel-measured top/bottom frame padding symmetry

## 📋 Review round

A full code review was run on the diff; confirmed findings applied: the kind gate, the chip width budget, the chip blend base (was blending over `card` instead of `background`), the count pill tone on cover-less cards, the parallelized tags fetch, and the hardened endpoint-path test assertion. Compact tag counts (vs the app's raw counts) are kept as an intentional, documented divergence to fit the fixed-width card.


infin1t3 (approved):
https://github.com/user-attachments/assets/0b9627c7-b3e8-41c8-add1-8d68854ed323

Tested and approved from a functional aspect. Claude had some comments that I'll paste here. IMO nothing worth blocking a PR over cc @secondl1ght 


### 1. Wide characters (Chinese/Japanese/Korean) overflow the tag row — [[renderCollectionOg.tsx:24](https://claude.ai/epitaxy/src/libs/og/renderCollectionOg.tsx:24)](src/libs/og/renderCollectionOg.tsx:24)

**Why it's happening:** The image is a fixed 1200px wide, and the tag chips at the bottom can't wrap to a second line — the drawing library just doesn't do that. So the code decides *in advance* which chips fit, by counting characters: it allows up to 36 characters total across all chips, on the assumption that one character is roughly 15px wide. That assumption is only true for Latin letters. Chinese, Japanese and Korean characters are roughly twice as wide (~26px at this font size).

**Where the problem is:** The `TAG_CHAR_BUDGET = 36` constant and the comment claiming it "leaves headroom even for wide glyphs". For CJK text it doesn't: three 12-character Japanese tags pass the 36-character check but need about 1160px of space in a row that's only 976px wide.

**How to reproduce:** Create a collection, tag it with three long Japanese or Chinese tags (around 12 characters each), then request its share-preview image (`/collections/<user>/<post>/opengraph-image`). The third chip will be visibly cut in half at the card's right edge. Nothing crashes — there's a safety net that clips instead of letting chips spill out — but the image looks broken.

**Suggested fix:** Count wide characters as 2 instead of 1 when checking the budget (there's a standard Unicode notion of "full-width" characters you can test for with a small regex), or simply lower the budget enough that even all-wide-character tags fit.

---

### 2. Emoji tags get unfairly dropped — [[utils.ts:465](https://claude.ai/epitaxy/src/libs/utils/utils.ts:465)](src/libs/utils/utils.ts:465)

**Why it's happening:** Two different pieces of code measure text length in two different ways. The function that decides *which* chips fit (`getDisplayTags`) uses JavaScript's `.length`, which counts an emoji as 2 (that's just how JavaScript stores emoji internally). The function that *shortens* a chip's label counts an emoji as 1, the way a human would. So an emoji-heavy tag looks twice as long to the "does it fit?" check as it does everywhere else.

**Where the problem is:** The line `Math.min(tag.length, maxTagLength)` inside `getDisplayTags` — `tag.length` is the internal count, not the visible character count.

**How to reproduce:** Tag a collection with two tags of ~10 emoji each. Each one counts as 20 toward the 36 budget, so the second tag gets dropped from the preview image — even though both chips would physically fit in the row with room to spare.

**Suggested fix:** Measure both places the same way — count visible characters (the truncation helper `truncateByGraphemes` already does this correctly; reuse its counting approach in `getDisplayTags`). Two softening notes: this errs on the safe side (a chip disappears rather than being clipped), and this measuring quirk existed before this PR — the PR just added a new caller that exposes it.

---

### 3. Two preview renderers format the same number differently — [[renderCollectionOg.tsx:59](https://claude.ai/epitaxy/src/libs/og/renderCollectionOg.tsx:59)](src/libs/og/renderCollectionOg.tsx:59)

**Why it's happening:** This is now the fourth place in the codebase that creates its own "shorten big numbers" formatter (the thing that turns 10500 into "10.5K"). They were each written by hand with slightly different settings, and the settings actually disagree: the profile preview image keeps one decimal place, the new collection preview image rounds to whole numbers.

**Where the problem is:** The `compactNumber` formatter in this file vs. the one in [[renderProfileOg.tsx:12](https://claude.ai/epitaxy/src/libs/og/renderProfileOg.tsx:12)](src/libs/og/renderProfileOg.tsx:12) — verified concretely: 10,500 renders as "10.5K" on a profile card but "11K" on a collection card. The "1 POST / N POSTS" wording is also a hand-copy of what [[CollectionCountBadge.tsx](https://claude.ai/epitaxy/src/components/molecules/CollectionCountBadge/CollectionCountBadge.tsx:8)](src/components/molecules/CollectionCountBadge/CollectionCountBadge.tsx:8) does in the app.

**How to reproduce:** Look at the share-preview image of a profile with 10,500 followers and a collection with 10,500 posts — the same quantity is written two different ways by two files sitting in the same folder.

**Suggested fix:** Create one shared formatter (and a small `formatCollectionCount(count)` helper for the "N POSTS" text) in a plain shared file, and have the app badge, the profile preview, and the collection preview all use it. The badge component itself can't be shared (it's browser-only code), but a plain function can be.

---

### 4. The card's background color is a hand-copied duplicate — [[ogConstants.ts:31](https://claude.ai/epitaxy/src/libs/og/ogConstants.ts:31)](src/libs/og/ogConstants.ts:31)

**Why it's happening:** The app defines its colors in CSS, which the image renderer can't read, so the OG code keeps its own list of colors as plain hex strings. Fine — except for this particular color, the codebase *already has* a JavaScript copy: [[theme.ts:21](https://claude.ai/epitaxy/src/config/theme.ts:21)](src/config/theme.ts:21) exports `COLORS.background = '#05050A'`, created for exactly this purpose (it even carries the identical explanatory comment).

**Where the problem is:** `background: '#05050a'` in `OG_TOKENS` re-types the value instead of importing the existing constant.

**How to reproduce:** Not a visible bug today — it's a trap. Change the app's background color: you'd update the CSS and `COLORS` (the documented pair), and the preview image would silently keep the old color. Worse, the tag chip colors in the app are computed *from* `COLORS.background`, so app chips and preview chips would stop matching.

**Suggested fix:** One-line change: `background: COLORS.background` with an import from `@/config/theme` (that file has no dependencies, so it's safe to use in server code). The other entries in `OG_TOKENS` have no existing constant to import, so they're fine as-is.

---

### 5. The chip color math secretly depends on a number defined elsewhere — [[renderCollectionOg.tsx:45](https://claude.ai/epitaxy/src/libs/og/renderCollectionOg.tsx:45)](src/libs/og/renderCollectionOg.tsx:45)

**Why it's happening:** In the app, a tag chip's color is made by laying a 70%-opaque dark layer over the tag's color. The image renderer can't stack translucent layers, so this PR pre-computes the final mixed color instead: 30% tag color + 70% background. That 30% is written as a local constant `TINT_ALPHA = 0.3` — which is really "1 minus the 0.7 used in [[PostTag.tsx:20](https://claude.ai/epitaxy/src/components/molecules/PostTag/PostTag.tsx:20)](src/components/molecules/PostTag/PostTag.tsx:20)", but nothing in the code connects the two numbers. The same file also re-implements "turn a hex color into red/green/blue numbers" — the third copy of that snippet in the codebase.

**Where the problem is:** `solidTagColor` in this file: the standalone `0.3`, and the duplicated hex-parsing lines.

**How to reproduce:** Another trap rather than a visible bug: if design ever softens the app chips (say 0.7 → 0.6), the app changes but the preview image doesn't, and no search for "0.6" or "0.7" would lead anyone to this file.

**Suggested fix:** Put one shared constant for the overlay strength in the utils file (next to the color helpers both sides already use), have `PostTag` use it directly and `solidTagColor` use `1 -` it. While there, add a tiny shared "hex to red/green/blue" helper so the parsing exists once.

---

### 6. Collection posts fetch the same data twice — [[renderCollectionOg.tsx:113](https://claude.ai/epitaxy/src/libs/og/renderCollectionOg.tsx:113)](src/libs/og/renderCollectionOg.tsx:113)

**Why it's happening:** There are two ways to reach the collection preview: the collections URL directly, or a regular post URL where the post turns out to be a collection. On the second path, [[renderPostOg.tsx:30](https://claude.ai/epitaxy/src/libs/og/renderPostOg.tsx:30)](src/libs/og/renderPostOg.tsx:30) fetches the user and post from the backend, sees "this is a collection", and hands off to `renderCollectionOg` — which fetches the exact same user and post again, because it takes only IDs, not the data.

**Where the problem is:** `renderCollectionOg`'s parameters — `{ userId, postId }` — force the refetch.

**How to reproduce:** Hard to see with the naked eye because a one-hour cache usually absorbs the second request. But request a collection post's preview image with caching off (e.g. a dev server): the backend logs will show the user and post endpoints hit twice for one image.

**Suggested fix:** Let `renderCollectionOg` accept an optional, already-fetched `{ user, post }`. The post path passes what it has; the direct collections path passes nothing and fetches as it does today. Rated "plausible" rather than "confirmed" because in production the cache makes the waste rare — but the fix is tiny.

---

### 7. Four lines of set-building where a one-line slice would do — [[renderCollectionOg.tsx:138](https://claude.ai/epitaxy/src/libs/og/renderCollectionOg.tsx:138)](src/libs/og/renderCollectionOg.tsx:138)

**Why it's happening:** The code asks `getDisplayTags` which tag labels fit, builds a Set of those labels, then filters the original tag list by membership. But `getDisplayTags` can only ever *drop tags from the end* — whatever it returns is always the first N of the input, in order. So the whole Set-and-filter dance reduces to "keep the first N tags".

**Where the problem is:** The `visibleLabels` Set and the `.filter(...)` that follows it.

**How to reproduce:** Not a visible bug in practice — it's extra machinery. There is one theoretical wrinkle it introduces: if two tags ever shared the same label, the filter would match both (the app's backend combines same-label tags into one, so this shouldn't occur — but the simpler code wouldn't even raise the question).

**Suggested fix:** `const displayTags = tags.slice(0, getDisplayTags(...).length)` — same result, one step, no edge case.

---

### 8. The darkening gradient is copy-pasted from the app card — [[renderCollectionOg.tsx:224](https://claude.ai/epitaxy/src/libs/og/renderCollectionOg.tsx:224)](src/libs/og/renderCollectionOg.tsx:224)

**Why it's happening:** To keep white text readable over a cover photo, the app's collection card darkens the photo with a left-to-right shading (70% dark fading to 35%). The preview image reproduces the same effect by pasting the identical CSS string from [[CollectionCard.tsx:281](https://claude.ai/epitaxy/src/components/organisms/Collections/CollectionCard/CollectionCard.tsx:281)](src/components/organisms/Collections/CollectionCard/CollectionCard.tsx:281). The only thing tying the copies together is a comment.

**Where the problem is:** The `linear-gradient(...)` string literal, existing in two files.

**How to reproduce:** A trap again: tweak the app card's shading and the share image keeps the old look — the two cards that are supposed to match slowly stop matching.

**Suggested fix:** Export the gradient string as a named constant from one shared file and use it in both places. Only the string itself can be shared (the app combines it with the photo in one CSS rule; the image renderer draws the photo separately), which is fine. This is the mildest of the duplication findings.

---

### 9. The photo and its shading layer repeat the same six style lines — [[renderCollectionOg.tsx:215](https://claude.ai/epitaxy/src/libs/og/renderCollectionOg.tsx:215)](src/libs/og/renderCollectionOg.tsx:215)

**Why it's happening:** The cover photo and the darkening layer on top of it must occupy exactly the same box. Instead of defining that box once, both style blocks spell out the same six properties (position, top, left, width, height, corner rounding) word for word.

**Where the problem is:** The two adjacent `style={{...}}` blocks inside the cover section — identical except the photo adds "crop to fill" and the overlay adds the gradient.

**How to reproduce:** No visible bug today — the risk is a future edit touching one block and not the other, which would leave the shading misaligned with the photo. The risk is modest because the actual numbers come from shared constants, so routine size changes already flow to both.

**Suggested fix:** Define the box once — `const coverLayerStyle = { position: 'absolute', top: 0, left: 0, width: CARD_WIDTH, height: cardHeight, borderRadius: CARD_RADIUS }` — and spread it into both blocks, each adding its own extras.

PR #2354, opened 2026-08-17
refactor: use pubkyring auth handler until multiple apps are supported

secondl1ght:
# fix(auth): target Pubky Ring exclusively for auth deeplinks and QR codes

## 🐛 Problem

On Android, tapping **"Authorize with Pubky Ring"** could open **Bitkit** instead of Pubky Ring. The button (and the sign-in/signup QR codes) handed the generic `pubkyauth://` URL to the OS, and Bitkit also registers that scheme in its Android manifest. Worse, Bitkit's deeplink handler silently drops `pubkyauth://` URIs when its Paykit feature flag is off (the default), so affected users landed on Bitkit's home screen with **no error and no auth sheet** — a complete dead end. The same ambiguity applied to scanning our QR codes with the system camera app.

The Bitkit team has been notified about their side of this (unconditional scheme registration + silent drop).

## 🔧 Fix

Convert the SDK's authorization URL to a **Pubky-Ring-exclusive deeplink** at the single point where it enters UI state (`useAuthUrl`):

```
pubkyauth://signin?caps=...  →  pubkyring://signin?caps=...
pubkyauth://signup?caps=...  →  pubkyring://signup?caps=...
```

Only Pubky Ring registers the `pubkyring` scheme, so OS scheme resolution can no longer route the link elsewhere. Because the conversion happens at the source, every surface inherits it consistently: the tap-to-open button, both QR codes (desktop sign-in + onboarding signup), and copy-link.

### Why scheme *replacement* instead of a `pubkyring://` prefix

A prefix wrapper was tried before and reverted in #748: Android URI normalization strips the inner colon of a nested scheme (`pubkyring://pubkyauth://signin?...` → `pubkyring://pubkyauth//signin?...`), which Pubky Ring cannot parse. Scheme replacement produces the exact deeplink forms Ring's input parser explicitly supports and unit-tests (`pubkyring://signin?...` / `pubkyring://signup?...`) — there is no nested scheme left to mangle.

## ✅ Verification

- **Bitkit collision confirmed**: both Pubky Ring and Bitkit register `pubkyauth` in their Android manifests; Bitkit's `processDeeplink` silently returns when `isPaykitEnabled` is false (default).
- **SDK URL shape confirmed**: `@synonymdev/pubky` 0.8.0 emits `pubkyauth://signin?caps=` / `pubkyauth://signup?caps=` (verified against the WASM binary).
- **Ring compatibility confirmed**: Ring's parser normalizes `pubkyring://signin|signup?...` and `pubkyauth://signin|signup?...` onto the same code path with byte-identical query parsing, for all input sources (deeplink, QR scan, clipboard paste). Support predates January 2026, so no minimum-app-version concern.
- **The relay handshake is untouched**: internal layers still produce and use the canonical `pubkyauth://` URL; only the user-facing string changes.
- Adversarial code review completed; findings addressed in this PR (test pinning, doc contract, stale copy) or deferred with rationale (see below).

## 📝 Changes

- `src/libs/deeplink/deeplink.ts` — new `generatePubkyRingAuthDeeplink()` helper (scheme replacement, pass-through for non-`pubkyauth://` input)
- `src/hooks/useAuthUrl/useAuthUrl.tsx` — apply the conversion where the URL enters React state
- `src/hooks/useAuthUrl/useAuthUrl.types.ts` — document the actual `url` contract
- `src/components/organisms/Scan/Scan.tsx` — footer copy updated: the QR is Ring-targeted now, so it no longer claims "any other Pubky Core–powered keychain" support (+ regenerated text snapshot)
- Tests — fixtures now mirror the real SDK/service URL shapes (signin unencoded, signup percent-encoded), and the conversion is pinned across mount-fetch, manual fetch, signup, and clipboard-copy paths

## 📌 Notes

- **Behavior when Ring is not installed**: nothing opens (the scheme has exactly one possible handler). The pre-existing "Opening Pubky Ring…" spinner without a timeout fallback is a known issue and will be addressed separately — it predates this change.
- The `OnboardingScanPage` VRT baseline will regenerate in CI (footer copy change).
- If broader signer support ships later, the conversion is one presentation-layer call site to revisit.

### DIFF
--- a/src/components/organisms/Scan/Scan.test.tsx.snap
+++ b/src/components/organisms/Scan/Scan.test.tsx.snap
@@ -86,7 +86,7 @@ exports[`Scan Components - Snapshots > ScanFooter - Snapshots > matches snapshot
   >
     Pubky Ring
   </a>
-   or any other 
+  , powered by 
   <a
     data-testid="link"
     href="https://pubky.org"
@@ -94,7 +94,7 @@ exports[`Scan Components - Snapshots > ScanFooter - Snapshots > matches snapshot
   >
     Pubky Core
   </a>
-  –powered keychain.
+  .
 </div>
 `;
 
--- a/src/components/organisms/Scan/Scan.tsx
+++ b/src/components/organisms/Scan/Scan.tsx
@@ -121,11 +121,11 @@ export const ScanFooter = () => {
       <Link href={getPubkyRingLink()} target="_blank">
         {'Pubky Ring'}
       </Link>
-      {' or any other '}
+      {', powered by '}
       <Link href={getPubkyCoreLink()} target="_blank">
         {'Pubky Core'}
       </Link>
-      {'–powered keychain.'}
+      {'.'}
     </FooterLinks>
   );
 };
--- a/src/hooks/useAuthUrl/useAuthUrl.test.tsx
+++ b/src/hooks/useAuthUrl/useAuthUrl.test.tsx
@@ -65,7 +65,7 @@ describe('useAuthUrl', () => {
   const createCancelAuthFlow = (): (() => void) => vi.fn();
 
   it('fetches auth URL on mount when autoFetch is enabled', async () => {
-    const mockAuthUrl = 'pubkyring://authorize?token=test123';
+    const mockAuthUrl = 'pubkyauth://signin?token=test123';
     const mockAwaitApproval = new Promise<Session>(() => {});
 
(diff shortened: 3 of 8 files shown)


secondl1ght:
Closing this one (discussion happened in Slack), TLDR: Bitkit removed the handler and we will revisit this flow again once we support multiple auth apps.

Issue #1746, opened 2026-04-28
UX: decide what to show when notification filter finds no matches across pages


Taewa:
## Problem

When a user enables only certain notification types (for example, only "New Friend"), the app filters out everything else. This can create a confusing experience during "load more."

### Example scenario

1. A user goes to Settings and enables only "New Friend" notifications
2. They haven't received a "New Friend" notification in months
3. The app scans up to 10 pages of older notifications (30 items per page = 300 notifications checked)
4. None of those 300 notifications are "New Friend" type — so zero results are shown
5. But the system knows there are still older notifications beyond those 300
6. The UI shows an empty list, yet "load more" is still available
7. The user taps "load more," another 300 are scanned, still nothing — and "load more" appears again

The user can keep tapping "load more" indefinitely without ever seeing any results.

## What we need to decide

This is a product/UX question, with some technical constraints from the local-first model.

The key decision is what should happen after the app has automatically scanned the current budget: **10 pages / 300 raw notifications**. At that point, the app may have found zero matching notifications, while older matching notifications may still exist.

Options include:

- **Keep "load more" visible** — simple, but the user may keep tapping and repeatedly get empty results.
- **Show a clearer empty state** — for example, "No matching notifications found in the latest 300 notifications." This is clearer, but still needs a next action.
- **Auto-hide "load more" after 300 scanned items** — avoids repeated empty loads, but can make older matching notifications unreachable.
- **Let the user explicitly search older matching notifications** — after the 300-item scan returns no matches, show a CTA such as "Search older matching notifications." This could open/show a separate filtered result list and treat those results as transient search results rather than canonical timeline coverage.
- **Automatically keep scanning beyond 300 items** — not preferred by default because it can create too many server requests, long waits, or runaway loops when the selected notification type is rare.

Preferred direction to discuss: after the normal 300-item scan finds no matches, avoid making the user repeatedly press "load more" for another empty 300-item scan. Instead, show a clear message and offer an explicit action to search older matching notifications, potentially backed by a type-filtered backend API if available.

## Suggested solution

Keep the automatic notification scan bounded to the current budget, for example 10 pages / 300 raw notifications. This protects the client and backend from accidental runaway pagination, especially when the enabled notification types are rare.

If the app reaches that scan budget and still finds no matching notifications, the UI should stop presenting the same normal "load more" action as if another page is likely to help. Instead, show a clear empty state and provide an explicit action such as "Search older matching notifications."

[...]

PR #2355, opened 2026-08-17
feat(search): add people and collections sections to search results

### LINKED ISSUE #1841: [Task]: Collections in search results + v25
Update Search results page to be according to the v25 handoff. 

Results should now include Collections.

https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=35688-283355&m=dev

https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=35688-284378&m=dev

infin1t3:
Closes #1841

Search results page redesign per the v25 handoff. Tag search now shows a People section and a Collections section above the posts feed, which sits under its own "Posts" heading. The old "Results for: ..." header (`SearchHeader`) is removed. Collections in the autocomplete is a separate discussion and not part of this PR.

I noticed #2399 while working on this PR, but given it's preexisting issue, I just opened a ticket and will address separately.

## People

- New Nexus endpoint `GET /v0/search/users/by_tags` (pubky/pubky-nexus#1031) returns scored user ids for the searched tags. Wired through the usual search api/service/application/controller layers.
- `useSearchPeople` fetches ids with skip pagination, hydrates them in one `POST /v0/stream/users/by_ids` call and reads them back reactively from Dexie. It dedupes ids within and across pages, advances the cursor by raw response length so index drift can't stall it, clamps to the endpoint's 5-tag limit, drops muted and unhydrated (deleted) users, and holds `loading` until the read-back emits so the section doesn't flash on settle.
- Cards are a new `card` variant of `UserListItem`: self-contained card with stats, tag chips and follow on desktop, avatar + name below `lg`. The whole card links to the profile; tag chips and the follow button suppress their own events so they don't navigate (same contract as `CollectionCard`).

## Collections

`SearchCollections` reuses the existing tagged collection stream (`<sort>:all:collection:<tags>`) through `useStreamPagination`, rendered with the existing `CollectionCard`. No new API.

## Behavior

- Both sections: preview of 4 cards, "See all" expands in place, "Show more" paginates. No matches means the section doesn't render. If a fully filtered page settles empty while the stream has more, "Show more" stays visible so the cursor can still advance.
- Sections render only with the Content filter on All and visual layout off. Any narrower filter shows the bare feed exactly as before, so nothing duplicates the Collections content filter.
- No Discover-style filtering: your own profile and own/followed/empty collections stay in the results on purpose.
- Existing `data-cy` selectors are untouched; the section grids live outside `post-search-results`.

## Testing

- Unit tests for the hook, both section organisms, the card variant, template gating and the search layers. Full suite green.
- VRT: tagged search baselines cover both sections on desktop and mobile, darwin and linux.
- typecheck, lint, format, unit, VRT and build all pass. A max-effort review pass ran over the People diff and everything actionable was fixed in-branch.

## Video walkthrough

https://github.com/user-attachments/assets/58ad524b-2d08-4076-8238-81fae57a0b9a


aldertnl:
Collection results are looking nice. I don't see any People results though on the provided ephemeral.

secondl1ght (approved):
This follows the right precedents — both sections use a Show more button rather than an infinite-scroll sentinel, and the cursor discipline is correct throughout (advance by raw scanned length, never by visible count — exactly the class of bug we've had to fix previously). One code issue at the end; the bulk of this comment is an architectural concern that predates this PR and is **not blocking**, but that this PR extends, so it should be on the record.

### The backend-pagination / client-filtering pattern, now on two more surfaces

Nexus paginates by raw cursor and doesn't support viewer-aware filtering (mutes, deleted posts) — a known constraint, and server-side filtering isn't currently planned. That means every stream surface filters client-side and then reconciles raw cursors against visible counts, and that reconciliation is where a long line of bugs has come from. We now have **at least four divergent implementations** of the same workaround:

1. **Infinite-scroll feeds**: `postStreamQueue.collect()` bounds one call (20 fetches, `post-stream-queue.ts:16`), but its own comment notes an auto-loading sentinel "will chain collects until the true stream end" — an effectively unbounded fetch loop through a filtered region.
2. **DiscoverCollections** — the most hardened variant: fetch-time filtering in the stream layer, a *tighter* per-click scan cap, and a warning toast ("No new collections found right now…") when a round surfaces nothing (`DiscoverCollections.tsx:166-174`).
3. **SearchCollections (this PR)**: inherits the shared refill via `useStreamPagination`, but when the scan budget exhausts inside a filtered region the click is a silent no-op — spinner clears, nothing appears, button remains.
4. **SearchPeople (this PR)**: a new bespoke variant *below* the existing bar — skip pagination with render-time mute filtering and **no refill at all**. One click = one raw page of 20; a filtered page renders nothing, silently. The preview can under-fill below 4 cards while "See all" still shows, and muted ids still cost a `by_ids` hydration before being dropped.

We already have an open bug report for "Show more button doesn't load more results," and given the backend constraint above, the frontend's ceiling is a bounded scan plus user feedback — the Discover pattern. Every new filtered surface will keep re-paying this cost until we converge on one hardened client-side pattern instead of re-deciding it per file.

### One code issue: People section can hang on skeletons (non-blocking, ticket please)

`src/hooks/useSearchPeople/useSearchPeople.ts:241` — the `detailsHydrated` gate pins the returned `loading` true if none of the fetched ids ever hydrate. The hydration call's failure is swallowed two layers down (`users.ts:106-108`, `Logger.warn` only), so a non-retryable failure (e.g. malformed response) or an all-deleted result page (stale search index — the hook's own comment calls unhydrated ids expected) leaves the section on 4 skeleton cards permanently: no error state, no retry, invisible to monitoring. Rare, but unbounded when it fires and cheap to fix — treat "hydration settled with nothing" as loaded-empty.

secondl1ght:
@infin1t3 I summarized my feedback above, non-blocking, but this issue has been encountered several times by different developers all working on different parts of the app (it actually came up again during last week's triage meeting). This needs an ADR at some point.

secondl1ght:
Oh just one other thing I forgot to mention, the UI can be a bit confusing because the tag you search may not be visible on the search result cards. Posts handles this by having a button to display all the tags, but no button exists on the user and collection search results.

infin1t3:
Opened #2417 to track this — the convergence plus the ADR. Captured all four surfaces you listed, and notifications (#1746) turns out to be a fifth instance of the same pattern, so that's in there too.

infin1t3:
> This follows the right precedents — both sections use a Show more button rather than an infinite-scroll sentinel, and the cursor discipline is correct throughout (advance by raw scanned length, never by visible count — exactly the class of bug we've had to fix previously). One code issue at the end; the bulk of this comment is an architectural concern that predates this PR and is **not blocking**, but that this PR extends, so it should be on the record.
> 
> ### The backend-pagination / client-filtering pattern, now on two more surfaces
> Nexus paginates by raw cursor and doesn't support viewer-aware filtering (mutes, deleted posts) — a known constraint, and server-side filtering isn't currently planned. That means every stream surface filters client-side and then reconciles raw cursors against visible counts, and that reconciliation is where a long line of bugs has come from. We now have **at least four divergent implementations** of the same workaround:
> 
> 1. **Infinite-scroll feeds**: `postStreamQueue.collect()` bounds one call (20 fetches, `post-stream-queue.ts:16`), but its own comment notes an auto-loading sentinel "will chain collects until the true stream end" — an effectively unbounded fetch loop through a filtered region.
> 2. **DiscoverCollections** — the most hardened variant: fetch-time filtering in the stream layer, a _tighter_ per-click scan cap, and a warning toast ("No new collections found right now…") when a round surfaces nothing (`DiscoverCollections.tsx:166-174`).
> 3. **SearchCollections (this PR)**: inherits the shared refill via `useStreamPagination`, but when the scan budget exhausts inside a filtered region the click is a silent no-op — spinner clears, nothing appears, button remains.
> 4. **SearchPeople (this PR)**: a new bespoke variant _below_ the existing bar — skip pagination with render-time mute filtering and **no refill at all**. One click = one raw page of 20; a filtered page renders nothing, silently. The preview can under-fill below 4 cards while "See all" still shows, and muted ids still cost a `by_ids` hydration before being dropped.
> 
> We already have an open bug report for "Show more button doesn't load more results," and given the backend constraint above, the frontend's ceiling is a bounded scan plus user feedback — the Discover pattern. Every new filtered surface will keep re-paying this cost until we converge on one hardened client-side pattern instead of re-deciding it per file.
> 
> ### One code issue: People section can hang on skeletons (non-blocking, ticket please)
> `src/hooks/useSearchPeople/useSearchPeople.ts:241` — the `detailsHydrated` gate pins the returned `loading` true if none of the fetched ids ever hydrate. The hydration call's failure is swallowed two layers down (`users.ts:106-108`, `Logger.warn` only), so a non-retryable failure (e.g. malformed response) or an all-deleted result page (stale search index — the hook's own comment calls unhydrated ids expected) leaves the section on 4 skeleton cards permanently: no error state, no retry, invisible to monitoring. Rare, but unbounded when it fires and cheap to fix — treat "hydration settled with nothing" as loaded-empty.

Code issue will be addressed in #2377

PR #2359, opened 2026-08-17
fix(ui): restore post header popover hover target

Orlandohub:
## Summary
- Restore the #2304 post-header popover hover target after #2285 stretched it across the full header row.
- Feed posts use `w-fit` on the trigger so the profile card opens only over the avatar and name, not empty space next to the timestamp.
- Composer headers keep `w-full` (`showPopover={false}`) so the character count still sits on the trailing edge.
- Restore the regression tests that assert the trigger and username link are not full-width.

## Test plan
- [ ] Hover a feed post header in the gap next to the timestamp — the profile popover should not open
- [ ] Hover the avatar and display name — the profile popover should still open
- [ ] Click empty space in the header row — it should not navigate to the author's profile
- [ ] Confirm long display names still truncate without overflowing the header
- [ ] Confirm the composer character count still aligns to the trailing edge when expanded


infin1t3 (approved):
👍

PR #2361, opened 2026-08-18
fix(ui): keep tag input focus when hovering user info popovers

### LINKED ISSUE #2291: [Bug]: Add tag focus lost when moving mouse cursor

infin1t3:
Closes #2291

## Problem

While typing in an add-tag input, moving the mouse over any hover-popover trigger (post header profile name, `@mention` link, feed row avatar) drops focus from the input — the text cursor disappears and typing stops working.

## Root cause

Two independent focus-stealing mechanisms in the `Popover` atom:

1. **Unconditional blur on mouse-leave** — `handleMouseLeave` blurred `document.activeElement` whenever the cursor left a hover trigger or its content, even when the popover never opened (just sweeping the cursor across a profile name was enough).
2. **Radix close auto-focus** — non-modal Radix popovers focus their trigger on close by default. When a hover popover with a focusable trigger (mention/avatar `Link`s) opens and closes, the trigger steals focus from the input; the trigger's hover-mode focus handler then blurs it, dropping focus entirely.

## Fix

In `src/components/atoms/Popover/Popover.tsx`:

- Removed the `blurActiveElement` helper and its call sites (mechanism 1).
- `PopoverContent` now prevents Radix's `onOpenAutoFocus`/`onCloseAutoFocus` **in hover mode only** (mechanism 2). Hover popovers are mouse-driven and never take or move focus; click-mode popovers keep Radix's default focus behavior for accessibility. Caller-provided handlers still run first.

## Tests

- Hover trigger crossed without opening → outside input keeps focus.
- Full hover open/close cycle → outside input keeps focus throughout.
- Click-mode regression guard: Escape-close still returns focus to the trigger.

Full unit suite: 716 files, 11,922 tests passing. Typecheck, ESLint, Prettier clean.

## Video walkthrough

https://github.com/user-attachments/assets/66a7bcc2-b366-4292-add0-10d7835f57ff


Taewa (approved):
Tested on ephemeral link and works well

PR #2363, opened 2026-08-18
fix(posts): improves article tags rendering in column view

aldertnl:
## Summary

**Improves visual rendering and layout of actions and tags for Articles in Columns layout.**
- Reuses the shared inline post tags and actions for article details in Columns layout.
- Places tags and actions directly below the user header and before the article cover image or body.
- Preserves the existing side tags panel in Wide view.
- Updates focused tests and snapshots for the new rendering.

## Why

Article details in Columns view used an always-expanded tags panel and behaved differently from regular post pages. This change makes both surfaces consistent while reusing the existing tag interaction logic.

## Validation

- 31 focused unit tests passed
- ESLint passed
- TypeScript typecheck passed
- Git diff hygiene check passed

## Before
image

## After
image


Orlandohub (approved):
LGTM

PR #2377, opened 2026-08-20
feat(search): add full-text content search

### LINKED ISSUE #1840: [Task]: Collections in live search result suggestions (v24)
v24 design handoff includes: Added collections to live search results

This issue covers the core and UI work.

https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/shadcn_ui-PUBKY?node-id=19729-9886&t=uugCeTBfdCBTp9wh-1

infin1t3:
## Summary

**Full-text content search**

- Integrate Nexus `GET /v0/search/posts/by_content` through a relevance-ordered, offset-paginated content-search stream.
- `content_search` is a first-class stream family (`StreamSource.CONTENT_SEARCH`) routed through `breakDownStreamId`/`createPostStreamParams`/`NexusPostStreamService.fetch`. Stream-id format is `content_search:q~<encodedQuery>:<kind>` — the marker makes reserved-word queries (`bookmarks`/`author`/`wot`) unmatchable by legacy segment classifiers, so mute/deleted filtering cannot be bypassed by a search term.
- `?q=` routing with `useSearchCriteria()` as the single source of q-vs-tags precedence (`content | tags | invalid | none`); invalid shared `?q=` URLs render an inline explanation and seed the query into the input for editing.
- Content-kind filtering, query result headers, and full-text-only Sort suppression.
- `by_content` fires only on explicit submit (Enter / "Show all results"), never per keystroke — live suggestions remain prefix-based tag/user lookups in the default rate-limit bucket.

**Tags pivot row**

- Full-text results show a "Tags" row above the feed: prefix matches of the query terms (exact matches first), one `by_prefix` lookup per term on submit — a one-click pivot from content results to tag search. Parts-of-word matching needs backend support and stays out of scope.

**Search input and recents**

- The input seeds from `?q=` for refinement: typed drafts survive navigation, same-query resubmits keep the value, and Enter / Show-all / recent-query clicks share one submit path. The X clears the whole search — draft, active query, and tag chips — landing on `/search`. No longer readOnly at max active tags.
- Typed searches are recorded in a new `recentQueries` store slice; recent tag and full-text searches share one row interleaved by recency (capped at 5), with full-text pills outlined (bg `background` / border `input` / text `foreground`) so they are distinguishable from tag chips.
- The dropdown hugs its content (no internal scrolling; viewport-relative safety cap only), "Show all results" is a left-aligned hug-content pill, and user rows use 32px avatars with the right-sidebar opacity hover.

## Testing

- `npx vitest run --project unit` — 741 files passed; 12,644 tests passed, 2 skipped
- `npm run typecheck`, `npm run lint`, `npm run format:check`; Next.js build verified in CI
- Cypress `search.cy.ts`: tag search via autocomplete, chip removal, and the full-text submit flow
- VRT: macOS baselines regenerated (content pages include the Tags row); Linux baselines will be regenerated via the VRT Update Baselines workflow once #2424 merges to dev

## Scope

Live collection autocomplete, keyboard-arrow navigation, trust-rank sorting, author-scoped search, and backend changes remain out of scope. Collections stay out of live results — confirmed by Chris: both live collection search and as-you-type full text hit Nexus's expensive rate-limit bucket. Accordingly, `by_content` fires only on explicit submit (Enter / Show all results), never per keystroke; live suggestions remain prefix-based tag/user lookups in the default bucket.

Closes #1840


infin1t3:
@greptileai review

aldertnl:
LGTM

Ideally we want the tag results to be broad, so this example search for 'bit' would also result in 'bitcoin' tag being part of the results. However, I would say this is not a blocker. More of an improvement suggestions for a next iteration. 

image

Also, I am not not seeing any People or Collections page search results in this ephemeral, but assume that is because it's out of scope for this particular PR.

secondl1ght (dismissed):
Reviewed the full diff with a focus on the stream-id round-trip, mute/deleted filtering, and the new search UI flows. Core plumbing looks solid — stream-id encode/decode round-trips exactly (including reserved words and `:`/`~` in queries), filtering applies to content-search pages, and the offset pagination math checks out. Approving, with four small fixes I'd ask for before merge and a couple of follow-ups.

**Please fix before merge (all ~one-line)**

1. **`useContentSearchTags.ts:28` — stale Tags pivot row.** The effect only clears `tags` when the query becomes `null`, not when it changes, so on a slow connection the previous query's chips stay rendered (and clickable) above the new query's results until the new `by_prefix` lookups resolve. Clearing state on query change fixes it.
2. **`useContentSearchTags.ts:41` — punctuated terms never match.** Terms are sent to `by_prefix` verbatim, but tags can't contain `,`/`:` (`tagInvalidChars`), so `"bitcoin, wallets"` looks up `by_prefix('bitcoin,')` and the natural `bitcoin` chip silently vanishes. Stripping tag-invalid chars from terms fixes the common case. (Same line: `term.length` counts UTF-16 units where tag validation counts code points — `[...term].length` aligns them.)
3. **`postStream.ts:75` — raw `Error` throw.** New code in `src/core`; AGENTS.md and docs/error-handling.md mandate `Err.validation(ValidationErrorCode.*, ...)` for exactly this input-validation case (see `userStream.utils.ts:19` in the same layer). The neighboring raw throws are documented migration debt — this one's new.
4. **`useSearchInput.ts:41` — no IME composition guard on Enter.** Committing an IME candidate fires Enter with `isComposing === true`, which now submits `/search?q=<half-composed text>` and records a bogus recent query. The repo already has the guard pattern in `useEnterSubmit.ts:60`. The gap pre-exists, but free-text search makes CJK input a first-class case here.

**Non-blocking follow-ups**

- **`contentSearch.ts:24` — term cap vs tokenizer:** the 4-term cap counts spaces, but `"one,two,three,four,five"` passes as one term while the index tokenizer sees five. What does `by_content` do with >4 tokens — 400 like `by_tags`, or truncate? If it errors, the validator should split the way the tokenizer does so users get the friendly invalid-mode message.
- **`contentSearch.ts:13` — case normalization:** `"Bitcoin"` and `"bitcoin"` produce separate recents chips, stream ids, and caches (tags are lowercased everywhere by contrast). If `by_content` is case-insensitive, normalizing at the stream-id/dedupe layer would consolidate them while keeping display case in `?q=`.
- **`postStream.utils.ts:123` — skip-pagination predicate duplicated:** this disjunction and `isSkipPaginatedStream` were both extended in parallel for this family; they're equivalent today and `streamId` is in scope, so calling the classifier here would keep the next skip-paginated family from silently getting timestamp/offset-mismatched pages.

One pre-existing bug surfaced during review that isn't this PR's to fix: the URL tag cap (`getMaxStreamTags()` = 5) vs chip cap (`MAX_ACTIVE_SEARCH_TAGS` = 3) desync, where removing a chip from a shared 5-tag URL silently drops the hidden tags. Exists identically on `dev` — will file separately.

secondl1ght:
Filed the pre-existing tag-cap desync from the review notes as #2446 (URL honors 5 tags, chips clamp to 3, chip edits rebuild the URL from the truncated list). Not this PR's regression, so no need to grow the diff here — but since `useSearchCriteria` is new in this PR, feel free to fold the fix in and close #2446 with it if you'd rather. Otherwise it's tracked as a follow-up.

infin1t3:
@secondl1ght thanks for the thorough review — all four blocking items are fixed in b4fc6194:

1. **Stale Tags pivot row** — tags now clear on every query change, so previous chips never linger while the new `by_prefix` lookups are in flight.
2. **Punctuated terms** — terms are stripped of tag-banned chars via the existing `sanitizeTagInput` helper (then re-deduped, since stripping can merge terms), and length is counted in code points via `getCharacterCount`.
3. **Raw `Error` throw** — replaced with `Err.validation(ValidationErrorCode.INVALID_INPUT, ...)` following the `userStream.utils.ts` pattern.
4. **IME guard** — `handleKeyDown` now returns early when `e.nativeEvent.isComposing`, covering both Enter (candidate commit) and Escape (composition cancel).

**Re: term cap vs tokenizer (`contentSearch.ts:24`)** — I tested this against staging Nexus, and `by_content` counts terms exactly the way our validator does (whitespace split), so there's no divergence to fix:

```
# 5 comma-joined tokens — one "term" to both our validator and Nexus
GET /v0/search/posts/by_content?q=one,two,three,four,five&limit=3
→ HTTP 200, results returned

# 4 space-separated terms (our cap)
GET /v0/search/posts/by_content?q=one two three four&limit=3
→ HTTP 200, results returned

# 5 space-separated terms
GET /v0/search/posts/by_content?q=one two three four five&limit=3
→ HTTP 400 {"error":"Invalid input: Failed to deserialize query string: q: Invalid input: Search query exceeds maximum of 4 terms"}
```

So the backend's own term counter treats comma-joined text as a single term (no 400, unlike `by_tags`), and its 4-term boundary matches the frontend cap exactly. Nothing slips through the validator that the backend would reject.

The other two follow-ups are filed and assigned: case normalization → #2447, skip-pagination predicate dedup → #2448.

secondl1ght:
@Orlandohub you pushed a commit here to @infin1t3's PR and now tests are failing, was that intentional?

secondl1ght (dismissed):
Re-reviewed after b4fc6194: all four requested fixes verified — the Tags pivot row now clears on every query change (regression test asserts chips are gone before the new lookup resolves), terms are sanitized with `sanitizeTagInput` and length-gated in code points, the raw `Error` is now `Err.validation` with proper context, and `handleKeyDown` guards `isComposing` for both Enter and Escape. All 154 unit tests in the touched modules pass locally. Re-approving.

Orlandohub:
@secondl1ght 

> `Orlandohub` you pushed a commit here to @infin1t3's PR and now tests are failing, was that intentional?

Yes, it was intentional, but it was a scope mistake on my side. #2396 is stacked on #2377, and the review asked us to correct the icon in both the shared search bar and profile filter bar. I should have kept that change on #2396 or asked before updating your branch. The initial push also missed parent snapshots, which caused the test failures; those were subsequently fixed. Sorry for changing your PR without coordinating first.

infin1t3:
Should be ok now @secondl1ght

secondl1ght (approved):

PR #2382, opened 2026-08-20
feat: attachment edit support for posts and articles

### LINKED ISSUE #1062: [Feature]: Add/Remove attachments whilst Editing post

secondl1ght:
# ✨ feat(posts): editable post & article attachments

Closes #1062.

## Summary

Post and article edits were text-only — attachments were frozen at creation. This PR completes edit support: users can now **add and remove attachments** on existing posts and **add, remove, or replace article cover images**, with full homeserver file cleanup and optimistic rendering.

The creation flow's validation, upload pipeline, and components are reused wholesale — no new designs, no duplicated rules.

## 🖱️ UX

- The edit dialog renders the post's current attachments (previews per media type, skeleton while metadata resolves) alongside the standard composer attachment UI — picker, drag-and-drop, and paste, all newly enabled in edit mode.
- Limits span existing + new attachments (10 per post, 1 image cover per article). Submit validity matches create: content **or** attachments.
- Removing an article cover reveals the create flow's "Add image" card; replace = remove + add.
- Owner sees edited attachments instantly (local blob cache shields the Nexus indexing window); viewers converge via the existing TTL refresh.

## 🔧 Implementation

- **`PostController.commitEdit`** accepts optional `attachments: { original, kept, added }` — `undefined` keeps the old content-only path byte-identical (collections unaffected). Removals are diffed against `original` (the snapshot the dialog was seeded from), **never the live row**, so concurrent row changes can't delete files the user never saw.
- **`PostNormalizer.toEdit`** injects the new attachment list and recomputed `kind` into the reconstructed post handed to `builder.editPost` (the specs edit API is content-only — spec-team follow-up filed). Post ID/URL stay stable; spec validation still runs.
- **`PostApplication.commitEdit`** orchestrates with full rollback: upload new files → optimistic Dexie write (content + attachments + kind, TTL bumped) → homeserver PUT → on any failure, restore the row and delete the uploads (via their known blob/record URLs, so partial uploads can't orphan) → on success, best-effort delete of removed files and old-kind stream cleanup.
- **Kind is recomputed** when the attachment set changes (image post with images removed → `short`, etc.); articles/collections always keep theirs. Cached kind-filtered feeds are purged of the old kind via the canonical stream-ID parser.
- **File deletion hardening**: blob-before-record order (retriable partial failures), 404-tolerant + retrying deletes (`HomeserverService.deleteIdempotent`), and a shared-blob guard — content-hash blobs shared by byte-identical uploads are skipped while still referenced.
- **New `useEditAttachments` hook** seeds and resolves existing attachments for the composer (local blob cache first, then Dexie metadata/CDN with Nexus backfill) and snapshots the seeded URI set for submit-time change detection.

## 🐛 Pre-existing bugs fixed along the way

1. Editing any image/video/link/file post silently rewrote its `kind` to `short` on the homeserver (`mapKindToEnum` gap; now maps all kinds and fails closed on unknown ones).
2. Edits never bumped `post_ttl`, letting stale Nexus data clobber fresh edits.
3. Attachment renderers never handled a shrinking attachment set — removed attachments/covers kept rendering (`PostAttachments`, `usePostArticle`, lightbox carousel index, error paths included).
4. `FileApplication.commitDelete` orphaned blobs in its fallback path and aborted post deletion on file-cleanup failures (now best-effort).
5. Deleted posts leaked their optimistic attachment cache (blob URLs) for the session.

## 🛡️ Verification process

Three hardening rounds beyond unit coverage: a 3-agent adversarial sweep of every attachment-state consumer (surfaced the carousel, stream-ghost, and change-detection issues), manual owner/viewer QA across local + staging clients (surfaced the shrink-render bug and the shared-blob 404), and a full `/code-review` (15 findings: 10 fixed, 5 deferred with rationale).

## ⚠️ Known constraints

- Viewer-side CDN variant-generation lag can briefly show a broken image (pre-existing, shared with post creation).
- The shared-blob guard uses local-cache knowledge only; homeserver-side ref-counting is the complete fix.
- Deferred items and backend asks (specs edit API, blob GC, CDN readiness, bootstrap freshness guard, e2e specs) are tracked in `.prs/2026/aug/feat-post-attachment-editing-followups.md`; full implementation narrative in `.prs/2026/aug/feat-post-attachment-editing.md`.

## ✅ Testing

- Full unit suite green: **720 files / 12,108 tests** (~+200 for this feature across pipes, services, application, controllers, hooks, and components), typecheck/lint/format clean.
- Manual QA: owner add/remove/replace flows for posts and article covers, viewer convergence via TTL and relaunch, old-client compatibility against the shared staging backend.
- E2E (Cypress) and VRT deliberately out of scope (tracked as follow-up).


src/hooks/useConfirmableDialog/useConfirmableDialog.ts
```diff
@@ -34,25 +34,38 @@ export function useConfirmableDialog({
...
   const internalHasContent = () => {
     return (
       contentRef.current.content.trim().length > 0 ||
       contentRef.current.tags.length > 0 ||
       contentRef.current.attachments.length > 0 ||
-      contentRef.current.articleTitle.trim().length > 0
+      contentRef.current.articleTitle.trim().length > 0 ||
+      contentRef.current.existingAttachments.length > 0
     );
   };
```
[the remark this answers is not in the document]
secondl1ght: Both observations are mechanically accurate, resolving as known/deferred with rationale:

1. **Always-prompting on unchanged attachments** is the pre-existing `useConfirmableDialog` semantics — it tracks *has content*, not *has changes*, so an edit dialog with seeded text already always prompts today. This PR extends attachments into that heuristic consistently. The real fix is delta-dirty semantics (compare against the seeded baseline), which touches all four composer dialogs and is deliberately deferred post-QA — tracked in the feature's follow-ups list ("Discard-prompt semantics").

2. **The silent close after removing every attachment discards an *unsubmittable* state**: an edit with no content and no attachments fails `canSubmitPost` (a post cannot become empty), so no savable change is lost — the prompt would offer to preserve a draft that cannot be submitted anyway.

src/components/molecules/PostInputAttachments/PostInputAttachments.tsx
```diff
@@ -20,49 +22,96 @@ type PostInputAttachmentsProps = {
...
-  if (file.type.startsWith('image/')) return 'image';
-  if (file.type.startsWith('video/')) return 'video';
-  if (file.type.startsWith('audio/')) return 'audio';
-  if (file.type === 'application/pdf') return 'pdf';
+const getAttachmentType = (mimeType: string): AttachmentType | undefined => {
+  if (mimeType.startsWith('image/')) return 'image';
+  if (mimeType.startsWith('video/')) return 'video';
+  if (mimeType.startsWith('audio/')) return 'audio';
+  if (mimeType === 'application/pdf') return 'pdf';
+};
```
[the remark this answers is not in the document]
secondl1ght: Keeping this one deliberately — it's a semantic memo, not a performance memo: `URL.createObjectURL` allocates a browser-held resource whose **identity** drives the revoke-on-change cleanup effect directly below. React Compiler memoization is an optimization, not a semantic guarantee — if the compiler skips or invalidates this value on an unrelated re-render, each render would allocate fresh object URLs (leak) and the cleanup would revoke URLs still bound to rendered previews (broken images). The memo also predates this PR (the diff renamed `attachmentsWithPreviews` → `newFilePreviews` while splitting existing/new previews).

Fair point on the missing documentation though — added a code comment above the `useMemo` explaining exactly this, so the exception to the no-manual-memoization rule is explicit.

secondl1ght:
@greptile all comments addressed, ready for re-review.

catch-21:
Done some basic checks so far. Can edit an existing post with text and image to remove, add image, replace image. Every time the PUT and DELETE with homeserver is correct.

Todo:  non-image attachment, max bound, article, e2e test.

src/hooks/usePost/usePost.tsx
```diff
@@ -172,19 +174,34 @@ export function usePost(): UsePostReturn {
     }
   };
 
-  const edit = async ({ editPostId, onSuccess }: UsePostEditOptions) => {
-    // requires content if normal edit and title if article
-    if (!content.trim() || (isArticle && (!content.trim() || !articleTitle.trim())) || !editPostId || !currentUserId)
+  const edit = async ({ editPostId, originalAttachmentUris, onSuccess }: UsePostEditOptions) => {
+    // allow empty content when attachments remain; articles require content and title
+    const totalAttachments = existingAttachments.length + attachments.length;
+    if (
```
Orlandohub, line 181: the `edit()` guard re-derives submit eligibility inline instead of calling `canSubmitPost` from `src/libs/utils/utils.ts`. `usePostInput.ts:208` does the same, so the rule lives in three places.

src/components/molecules/PostInputAttachments/PostInputAttachments.tsx
```diff
@@ -20,49 +22,101 @@ type PostInputAttachmentsProps = {
...
+// Mirrors the feed's variant rule (PostAttachmentsImagesAndVideos): FEED for
+// images so the preview paints from the browser cache the feed already warmed
+// (MAIN is the full-res lightbox/detail variant), except GIFs, where the feed
+// itself uses MAIN to keep the animation.
+const getExistingPreviewUrl = (attachment: ExistingAttachment): string | null => {
```
Orlandohub: `getExistingPreviewUrl` reimplements the "GIF uses urls.main, otherwise urls.feed ?? urls.main" variant-selection rule that already exists in `PostAttachmentsImagesAndVideos.tsx:156`, `PostListMediaThumbnail.tsx:24`, and `useVisualFeedTiles.ts:102`

Maybe extract `getPreviewUrl(attachment)` into `src/libs/utils/` and call it from all four sites?

src/core/controllers/post/post.types.ts
```diff
@@ -66,9 +66,35 @@ export interface TDeletePostParams {
   compositePostId: string;
 }
 
+export interface TEditPostAttachments {
```
Orlandohub, line 69: `TEditPostAttachments` is exported but never imported anywhere

Orlandohub (dismissed):
Removing an attachment can accidentally delete the underlying file even when another post still uses it, causing that post’s attachment to break. Even though this is a know constraint I would like to know if it's something we could/should mitigate before moving on?

secondl1ght:
@Orlandohub yes this is a pre-existing issue, the same thing can happen when deleting a post already. We need to discuss this with the homeserver team about different options to handle this. The frontend does not have full visibility into all the users posts, this should ideally be handled on the backend which knows about all the file/blob/post relationships. 

EDIT: I can reach out to them and create a follow-up ticket to address this app-wide (not just edits), if we decide to try and handle it on the frontend.

src/components/molecules/PostInputAttachments/PostInputAttachments.tsx
secondl1ght: Done — extracted `getAttachmentPreviewUrl` and swapped all the sites you listed (plus this one) onto it. One placement note: I put it in `src/libs/file/attachmentPreviewUrl.ts` next to the other file helpers rather than growing `utils.ts` — shout if you'd rather have it there. Two scope notes: `PostListMediaThumbnail` keeps its explicit video branch in front of the helper (identical behavior, but I didn't want to silently rely on videos never carrying a `feed` URL), and `useVisualFeedTiles`' remote-tile builder wasn't converted since it constructs variant URLs from `FileController` rather than selecting from an existing `urls` object — different inputs, same rule; noted for a cleanup follow-up.

src/core/controllers/post/post.types.ts
secondl1ght: Fair — it was only referenced within its own file. Now imported and used to type the construction site in `usePost.edit` (the built `editAttachments` object), which also gives the hook compile-time drift protection against the controller contract.

src/hooks/usePost/usePost.tsx
secondl1ght: Agreed it's a real duplication, deliberately deferring this one: the hook-level guards are early-returns that intentionally mirror the UI-level `canSubmitPost` (defense in depth — the hook must not trust that the button was correctly disabled), and unifying the three sites means touching the QA'd create/reply/repost submit guards too. It's on the tracked cleanup list for a follow-up PR ("submit-predicate triplication") so it lands with its own focused review rather than riding along here post-QA.

catch-21:
There's duplication of
```
  if (hasSupportedUrl(content)) {
    return PubkyAppPostKind.Link;
  }
```
Create and edit both apply the same remaining kind rule after their own guards: URL → Link, else attachment MIME types, else Short. That is fine today, but a priority change (e.g. image beats link) would need updating in both. Can we extract that sequence into one helper and keep the article/collection guards in the callers?

catch-21 (dismissed):

catch-21 (approved):

Issue #2390, opened 2026-08-21
[Task]: Starter Packs — swap to live starter_pack source + E2E happy path


Orlandohub:
## Problem Statement

The Starter Packs FE ships against a mocked stream source (`most_followed` alias from #2386, mock-namespaced cache keys) until `source=starter_pack` is live on staging. Once pubky/pubky-nexus#1024 is merged and deployed, we need to swap to the real source and validate the full flow end-to-end.

## Proposed Solution

- [ ] Flip `STARTER_PACK_SOURCE_LIVE` to true, then remove the flag, the `most_followed` mock dispatch, and all `starter_pack_mock` handling
- [ ] Purge orphaned `starter_pack_mock:*` rows from the Dexie `user_streams` table
- [ ] Keep the mock-cache regression test green: a pre-existing mock-keyed row must never satisfy a live request (first live load calls `source=starter_pack` and writes the live-keyed row)
- [ ] Verify against staging: ranked results, tag-order sensitivity (`travel,music` ≠ `music,travel`), viewer + already-followed exclusion, and the new 400 behavior for `tags` on other sources (we send none today — confirm nothing regressed)
- [ ] Cypress E2E happy path: select tags → suggestions render → Follow All → land on My network feed. Account creation needs SMS/invite against staging — reuse the existing onboarding E2E fixtures/approach
- [ ] Cover the two edge paths: no tags → most-active-this-month suggestions; no follows → land on All feed

## Gated on

- pubky/pubky-nexus#1024 **merged** at `14610363` — still gated on the staging deploy (staging currently returns `unknown variant starter_pack`)
- Ranking quality depends on the trust job running in prod (seeds + cron); without it the API falls back to endorsement counting — acceptable for testing, verify before launch

## Depends on

- #2386, #2388 (and transitively #2387)

Orlandohub:
Closing as unnecessary: pubky/pubky-nexus#1024 deployed to staging before PR #2391 merged (verified 2026-08-21 — `source=starter_pack` returns 200 with ordered comma-separated tags, 400 when tags are missing).

Since no build containing the mock ever shipped, the swap this issue tracked never needs to happen:

- The `STARTER_PACK_SOURCE_LIVE` flag, `starter_pack_mock:*` namespace, and `most_followed` aliasing were removed from PR #2391 directly (57f2a7e1) — the stream layer targets the live source from the start.
- The Dexie purge of orphaned `starter_pack_mock:*` rows is moot — no user ever had them.
[...]

Issue #2388, opened 2026-08-21
[Feature]: Starter Packs — onboarding "Follow your best matches" step + Follow All & landing feed


Orlandohub:
## Problem Statement

As a new user who picked interests, I want a ranked list of people to follow — and a way to follow them all in one tap — so my feed starts out useful. After finishing onboarding I should land on the right feed: "My network" if I followed people, "All" if I didn't (an empty My network feed would be a dead end).

## Proposed Solution

Second screen of the onboarding "Experience" step: **Follow your best matches**, including Follow All and the post-onboarding landing feed.

Design:
- [Follow your best matches](https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=45738-571825&m=dev)
- [Spec annotations](https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=45738-581938&m=dev)

### Suggestions

- [ ] Fetch suggestions via the starter-pack stream ID from #2386 using the ordered tags selected in #2387, `limit` 10 (per design: max 10 suggestions)
- [ ] No tags picked → fallback stream `influencers:this_month:all` ("most active users this month", already supported)
- [ ] Card grid per design: avatar, name, truncated pubky, TAGS / POSTS counts, per-card follow toggle — reuse `UserListItem` patterns + `useFollowUser` + `useWhoToFollowFollowPreservation`
- [ ] Card tag chips: **max 2 profile tags that match the chosen interest tags** (client-side intersection over hydrated user tags)
- [ ] "Load more" is marked "eventually" in design — out of MVP scope (`useUserStream` paginated mode covers it later; note the API caps skip via `STARTER_PACK_MAX_SKIP`)

### Follow All + Finish

- [ ] `useFollowAll` hook orchestrating N × `UserController.commitFollow(HttpMethod.PUT, …)`: per-user progress, partial-failure handling. Note each commit awaits local Dexie **and** the homeserver PUT, and `useFollowUser` only tracks one `loadingUserId` — this needs its own orchestration
- [ ] "Follow all (N)" button per design — N counts only not-yet-followed suggestions, button reflects progress
- [ ] Finish: followed ≥ 1 → `useHomeStore.getState().setReach(REACH.NETWORK)` ("My network" = WoT reach, per flow spec); followed 0 → leave `REACH.ALL`
[...]

Issue #2387, opened 2026-08-21
[Feature]: Starter Packs — onboarding "Tags of interest" step


Orlandohub:
## Problem Statement

As a new user, I want to pick topics I care about during onboarding so the app can suggest people to follow and my feed isn't empty on day one. Onboarding currently ends at profile creation (`useProfileForm` create-success → `/home`) with no interest selection or follow suggestions.

## Proposed Solution

New onboarding step directly after **Create your profile**, shown as step 4 under the "Experience" stepper label. This issue covers the scaffolding plus the first of two screens: **Tags of interest**.

Design:
- [Tags of interest](https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=45738-569396&m=dev)
- [Create your profile (preceding step / stepper reference)](https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=41492-354599&m=dev)

### Scaffolding

- [ ] New route in `ONBOARDING_ROUTES` + stepper entry in `pathToStepConfig` — reconcile numbering with design (design shows profile = 3, experience = 4; current config maps 5 routes) + RouteGuard allow-lists
- [ ] Redirect change in `useProfileForm` create-success: `/home` → new step. Must cover both entry paths (browser keys and Pubky Ring — Ring enters onboarding at profile creation)
- [ ] Persisted "completed/skipped" flag so users are never re-prompted

### Screen

- [ ] "Popular interests" chips from `useHotTags` (`THIS_MONTH` timeframe), selectable, header shows "(N selected)"
- [ ] "Your interests" free entry: `TagInput` + prefix autocomplete (`useTagSuggestions`), removable chips
- [ ] Combined cap of 5 tags (reuse the `getMaxStreamTags()` pattern), existing sanitize/lowercase utils
- [ ] Selected tags (ordered) handed to the next screen via onboarding state
- [ ] Back / Continue per design; Continue with 0 tags is allowed (fallback handled in the Follow your best matches issue)

VRT baselines for the new template are **out of scope** — handled later by QA.

## Acceptance Criteria

- After profile creation the user lands on Tags of interest with stepper step 4 active (both browser-keys and Ring paths)
- Max 5 tags across both sections, selection order preserved
- A user who completed or skipped the step is never re-prompted (refresh, back-nav, re-login)

## References

[...]

PR #2391, opened 2026-08-21
feat(streams): add starter_pack user stream source with tag-aware IDs

### LINKED ISSUE #2386: [Task]: Starter Packs — starter_pack user stream source with tag-aware stream IDs
## Problem Statement

Nexus added `source=starter_pack` to `/v0/stream/users/ids` (pubky/pubky-nexus#1024, **merged and live on staging** — verified 2026-08-21): pass 1–5 **ordered** tags, get one ranked list of users to follow. With `viewer_id` it excludes the viewer and people they already follow. The frontend user-stream layer cannot consume it yet:

- `UserStreamSource` has no `starter_pack` member, and `NexusUserStreamService.fetch` throws on unknown sources
- The stream cache key (Dexie `user_streams` row id, `source:timeframe:reach`) has **no slot for a tag list** — two different tag sets would collide on one cache row (this was called out explicitly by the Nexus side)
- Tag order matters on the API (`travel,music` ≠ `music,travel`), so the key must preserve order

## Proposed Solution

Core/data layer only — no UI in this issue.

- [ ] Add `STARTER_PACK = 'starter_pack'` to `UserStreamSource` (`src/core/services/nexus/stream/users/userStream.types.ts`)
- [ ] Add `userStreamApi.starterPack` URL builder sending the ordered `tags` param (`userStream.api.ts`)
- [ ] Handle the new source in the `NexusUserStreamService.fetch` switch (`userStream.ts`)
- [ ] Extend the stream ID convention with a tag segment, e.g. `starter_pack:all:all:bitcoin,music` (mirrors post streams' `${baseStreamId}:${tags.join(...)}`); add a `buildStarterPackStreamId(tags)` helper (`src/core/models/stream/user/`)
- [ ] Teach `createUserStreamParams` to parse tag-bearing IDs — it currently throws unless the ID has 2 or 3 parts (`userStream.utils.ts`)
- ~~Temporary mock: until staging serves `starter_pack`, a `STARTER_PACK_SOURCE_LIVE = false` flag makes `buildStarterPackStreamId` emit a mock-namespaced stream ID that dispatches to `most_followed`.~~ **Dropped:** Nexus #1024 went live on staging before this issue's PR merged, so the layer targets `source=starter_pack` directly and no mock/flag ships. This also made #2390 (the mock-to-live swap) unnecessary — closed; its E2E checklist moved to #2388.
- [ ] Unit tests: cache isolation per ordered tag set, tag order preservation, param mapping

## Acceptance Criteria

- Two different ordered tag lists never share a Dexie `user_streams` row
- `useUserStream` consumes the new stream ID without changes
- Tag order is preserved end-to-end (cache key and query param)

## References

- Nexus PR: pubky/pubky-nexus#1024 · design rationale: pubky/pubky-nexus#1022
- Slack context: [Nexus plan](https://synonymworkspace.slack.com/archives/C0BNBJT321X/p1785953814788869) · [status + FE guidance](https://synonymworkspace.slack.com/archives/C0BNBJT321X/p1787009391801879)

Orlandohub:
## Summary
- fix #2386
- Adds the core stream plumbing for Starter Packs: a 4-part user stream ID format (`source:all:all:tag1,tag2`) so recommendations for an ordered tag list get their own Dexie cache row.
- Targets the live `source=starter_pack` directly: pubky/pubky-nexus#1024 is deployed on staging (verified 2026-08-21 — 200 with ordered comma-separated `tags`, 400 when `tags` is missing). An earlier revision of this PR shipped a temporary `starter_pack_mock` namespace behind a `STARTER_PACK_SOURCE_LIVE` flag; that machinery was removed in 57f2a7e1 once staging went live, which also made #2390 (the swap issue) unnecessary.
- Unblocks the onboarding UI issues (#2387, #2388), which consume `buildStarterPackStreamId` without touching service internals.

## Changes
- `buildStarterPackStreamId(tags)` (model helper): canonicalizes labels (trim/lowercase) and validates the full tag contract (1–5 tags, 1–20 chars, no banned chars) via `Err.validation`; emits `starter_pack:all:all:<tags>`.
- `createUserStreamParams`: parses 4-part starter pack IDs, enforces `all:all` timeframe/reach segments, rejects noncanonical hand-built IDs; invalid-format fallthrough converted from raw `Error` to `Err.validation`.
- `NexusUserStreamService.fetch`: dispatches `starter_pack` → new `userStreamApi.starterPack` (`source=starter_pack&tags=…`); switch default is now a `never`-checked `Err.validation`.
- Config: `STARTER_PACK_MAX_TAGS = 5` (Nexus contract) in `src/config/nexus.ts`.
- Types: `StarterPackStreamId` extends the `UserStreamId` union; `STARTER_PACK` enum member and `TUserStreamStarterPackParams` in the Nexus service types.

## Test plan
- [x] `npm run typecheck` — clean
- [x] ESLint + Prettier on touched files — clean
- [x] Helper suite: canonicalization, tag ordering, every rejection class (empty, over-limit, whitespace, overlength, delimiter-bearing, noncanonical)
- [x] Parser suite: ordered tags param, `all:all` segment enforcement, invalid timeframe/reach/source/label cases with `AppError` category/code assertions
- [x] Service suite: starter pack ID → `source=starter_pack&tags=…` via mocked `queryNexus`; endpoint count 10 → 11
- [x] Application suite (fake-indexeddb): reversed tag orders persist as distinct Dexie rows
- [x] Staging smoke: `GET /v0/stream/users/ids?source=starter_pack&tags=bitcoin,art,music` → 200 with user IDs; missing `tags` → 400

## Notes
- One pre-existing test was updated: the old "4+ parts is invalid" case now asserts the starter-pack-specific rejection; a 5-part case covers the format fallthrough.

### DIFF
--- a/src/core/application/stream/users/users.test.ts
+++ b/src/core/application/stream/users/users.test.ts
@@ -2,7 +2,7 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { UserStreamApplication } from '@/application/stream/users/users';
 import type { Pubky } from '@/models/models.types';
 import { buildUserCompositeId } from '@/models/stream/user/userStream.helper';
-import { UserStreamTypes } from '@/models/stream/user/userStream.types';
+import { type UserStreamId, UserStreamTypes } from '@/models/stream/user/userStream.types';
 import { UserDetailsModel } from '@/models/user/details/userDetails';
 import { LocalStreamUsersService } from '@/services/local/stream/users/users';
 import type { NexusUser } from '@/services/nexus/nexus.types';
@@ -287,6 +287,50 @@ describe('UserStreamApplication', () => {
         params: { skip: 0, limit: 3, viewer_id: DEFAULT_VIEWER_ID },
       });
       expect(result.nextPageIds).toEqual(['influencer-1', 'influencer-2', 'influencer-3']);
+    });
+
+    describe('starter pack cache isolation', () => {
+      const LIVE_STREAM_ID = 'starter_pack:all:all:bitcoin,music' as UserStreamId;
+      const MOCK_STREAM_ID = 'starter_pack_mock:all:all:bitcoin,music' as UserStreamId;
+
+      it('should persist reversed tag orders as distinct Dexie rows', async () => {
+        const forwardId = 'starter_pack_mock:all:all:travel,music' as UserStreamId;
+        const reversedId = 'starter_pack_mock:all:all:music,travel' as UserStreamId;
+
+        await LocalStreamUsersService.upsert({ streamId: forwardId, stream: ['user-a'] });
+        await LocalStreamUsersService.upsert({ streamId: reversedId, stream: ['user-b'] });
+
+        expect((await LocalStreamUsersService.findById(forwardId))?.stream).toEqual(['user-a']);
+        expect((await LocalStreamUsersService.findById(reversedId))?.stream).toEqual(['user-b']);
+      });
+
+      it('should never serve cached mock rows for live starter pack requests (mock-to-live swap regression)', async () => {
+        // Seed a mock-namespaced row exactly as a pre-swap session would have left it
+        const mockCachedIds: Pubky[] = ['mock-user-1', 'mock-user-2'];
+        await LocalStreamUsersService.upsert({ streamId: MOCK_STREAM_ID, stream: mockCachedIds });
+        await createUserDetails(mockCachedIds);
+
+        const liveUserIds: Pubky[] = ['live-user-1', 'live-user-2'];
+        const fetchSpy = vi.spyOn(NexusUserStreamService, 'fetch').mockResolvedValue(liveUserIds);
+
+        const result = await UserStreamApplication.getOrFetchStreamSlice({
+          streamId: LIVE_STREAM_ID,
+          skip: 0,
+          limit: 2,
+          viewerId: DEFAULT_VIEWER_ID,
+        });
+
+        // The live key must miss the cache and hit the network
+        expect(fetchSpy).toHaveBeenCalledWith({
+          streamId: LIVE_STREAM_ID,
+          params: { skip: 0, limit: 2, viewer_id: DEFAULT_VIEWER_ID },
+        });
+        expect(result.nextPageIds).toEqual(liveUserIds);
+
+        // Both rows stay isolated: live written fresh, mock untouched
+        expect((await LocalStreamUsersService.findById(LIVE_STREAM_ID))?.stream).toEqual(liveUserIds);
+        expect((await LocalStreamUsersService.findById(MOCK_STREAM_ID))?.stream).toEqual(mockCachedIds);
+      });
     });
 
     it('should pass viewerId to Nexus API for relationship data', async () => {
--- a/src/core/application/stream/users/users.test.ts
+++ b/src/core/application/stream/users/users.test.ts
@@ -1,8 +1,8 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { UserStreamApplication } from '@/application/stream/users/users';
 import type { Pubky } from '@/models/models.types';
-import { buildUserCompositeId } from '@/models/stream/user/userStream.helper';
-import { type UserStreamId, UserStreamTypes } from '@/models/stream/user/userStream.types';
+import { buildStarterPackStreamId, buildUserCompositeId } from '@/models/stream/user/userStream.helper';
+import { UserStreamTypes } from '@/models/stream/user/userStream.types';
 import { UserDetailsModel } from '@/models/user/details/userDetails';
 import { LocalStreamUsersService } from '@/services/local/stream/users/users';
 import type { NexusUser } from '@/services/nexus/nexus.types';
@@ -291,12 +291,33 @@ describe('UserStreamApplication', () => {
 
     describe('starter pack caching', () => {
       it('should persist reversed tag orders as distinct Dexie rows', async () => {
-        const forwardId = 'starter_pack:all:all:travel,music' as UserStreamId;
-        const reversedId = 'starter_pack:all:all:music,travel' as UserStreamId;
+        const forwardId = buildStarterPackStreamId(['travel', 'music']);
+        const reversedId = buildStarterPackStreamId(['music', 'travel']);
+        const fetchSpy = vi.spyOn(NexusUserStreamService, 'fetch').mockImplementation(async ({ streamId }) => {
+          return streamId === forwardId ? ['user-a'] : ['user-b'];
+        });
 
-        await LocalStreamUsersService.upsert({ streamId: forwardId, stream: ['user-a'] });
-        await LocalStreamUsersService.upsert({ streamId: reversedId, stream: ['user-b'] });
+        await UserStreamApplication.getOrFetchStreamSlice({
+          streamId: forwardId,
+          skip: 0,
+          limit: 1,
+          viewerId: DEFAULT_VIEWER_ID,
+        });
+        await UserStreamApplication.getOrFetchStreamSlice({
+          streamId: reversedId,
+          skip: 0,
+          limit: 1,
+          viewerId: DEFAULT_VIEWER_ID,
+        });
 
+        expect(fetchSpy).toHaveBeenNthCalledWith(1, {
+          streamId: forwardId,
+          params: { skip: 0, limit: 1, viewer_id: DEFAULT_VIEWER_ID },
+        });
+        expect(fetchSpy).toHaveBeenNthCalledWith(2, {
+          streamId: reversedId,
+          params: { skip: 0, limit: 1, viewer_id: DEFAULT_VIEWER_ID },
+        });
         expect((await LocalStreamUsersService.findById(forwardId))?.stream).toEqual(['user-a']);
         expect((await LocalStreamUsersService.findById(reversedId))?.stream).toEqual(['user-b']);
       });
--- a/src/core/application/stream/users/users.test.ts
+++ b/src/core/application/stream/users/users.test.ts
@@ -289,47 +289,16 @@ describe('UserStreamApplication', () => {
       expect(result.nextPageIds).toEqual(['influencer-1', 'influencer-2', 'influencer-3']);
     });
 
-    describe('starter pack cache isolation', () => {
-      const LIVE_STREAM_ID = 'starter_pack:all:all:bitcoin,music' as UserStreamId;
-      const MOCK_STREAM_ID = 'starter_pack_mock:all:all:bitcoin,music' as UserStreamId;
-
+    describe('starter pack caching', () => {
       it('should persist reversed tag orders as distinct Dexie rows', async () => {
-        const forwardId = 'starter_pack_mock:all:all:travel,music' as UserStreamId;
-        const reversedId = 'starter_pack_mock:all:all:music,travel' as UserStreamId;
+        const forwardId = 'starter_pack:all:all:travel,music' as UserStreamId;
+        const reversedId = 'starter_pack:all:all:music,travel' as UserStreamId;
 
         await LocalStreamUsersService.upsert({ streamId: forwardId, stream: ['user-a'] });
         await LocalStreamUsersService.upsert({ streamId: reversedId, stream: ['user-b'] });
 
         expect((await LocalStreamUsersService.findById(forwardId))?.stream).toEqual(['user-a']);
         expect((await LocalStreamUsersService.findById(reversedId))?.stream).toEqual(['user-b']);
-      });
-
-      it('should never serve cached mock rows for live starter pack requests (mock-to-live swap regression)', async () => {
-        // Seed a mock-namespaced row exactly as a pre-swap session would have left it
-        const mockCachedIds: Pubky[] = ['mock-user-1', 'mock-user-2'];
-        await LocalStreamUsersService.upsert({ streamId: MOCK_STREAM_ID, stream: mockCachedIds });
-        await createUserDetails(mockCachedIds);
-
-        const liveUserIds: Pubky[] = ['live-user-1', 'live-user-2'];
-        const fetchSpy = vi.spyOn(NexusUserStreamService, 'fetch').mockResolvedValue(liveUserIds);
-
-        const result = await UserStreamApplication.getOrFetchStreamSlice({
-          streamId: LIVE_STREAM_ID,
-          skip: 0,
-          limit: 2,
-          viewerId: DEFAULT_VIEWER_ID,
-        });
-
-        // The live key must miss the cache and hit the network
-        expect(fetchSpy).toHaveBeenCalledWith({
-          streamId: LIVE_STREAM_ID,
-          params: { skip: 0, limit: 2, viewer_id: DEFAULT_VIEWER_ID },
-        });
-        expect(result.nextPageIds).toEqual(liveUserIds);
-
-        // Both rows stay isolated: live written fresh, mock untouched
-        expect((await LocalStreamUsersService.findById(LIVE_STREAM_ID))?.stream).toEqual(liveUserIds);
-        expect((await LocalStreamUsersService.findById(MOCK_STREAM_ID))?.stream).toEqual(mockCachedIds);
       });
     });
 
--- a/src/core/models/stream/user/userStream.helper.test.ts
+++ b/src/core/models/stream/user/userStream.helper.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect, it } from 'vitest';
 import { AppError } from '@/libs/error/error';
 import { ValidationErrorCode } from '@/libs/error/error.codes';
-import { ErrorCategory } from '@/libs/error/error.types';
+import { ErrorCategory, ErrorService } from '@/libs/error/error.types';
 import { buildStarterPackStreamId, USER_STREAM_TAG_DELIMITER } from './userStream.helper';
 
 const expectValidationError = (fn: () => unknown) => {
@@ -13,6 +13,7 @@ const expectValidationError = (fn: () => unknown) => {
     const appError = error as AppError;
     expect(appError.category).toBe(ErrorCategory.Validation);
     expect(appError.code).toBe(ValidationErrorCode.INVALID_INPUT);
+    expect(appError.service).toBe(ErrorService.Local);
   }
 };
 
@@ -37,6 +38,16 @@ describe('buildStarterPackStreamId', () => {
 
       expect(fromMixedCase).toBe(fromCanonical);
       expect(fromMixedCase).toBe('starter_pack:all:all:bitcoin,music');
+    });
+
+    it('should deduplicate canonical labels while preserving first-seen order', () => {
+      expect(buildStarterPackStreamId(['Bitcoin ', 'music', 'bitcoin', 'MUSIC', 'art'])).toBe(
+        'starter_pack:all:all:bitcoin,music,art',
+      );
+    });
+
+    it('should enforce the tag cap after canonical duplicates are removed', () => {
+      expect(buildStarterPackStreamId(['a', 'A', 'b', 'B', 'c', 'C'])).toBe('starter_pack:all:all:a,b,c');
     });
 
     it('should accept the maximum of 5 tags', () => {
@@ -81,7 +92,7 @@ describe('buildStarterPackStreamId', () => {
     });
 
     it('should accept a label at exactly 20 chars', () => {
-      expect(buildStarterPackStreamId(['a'.repeat(20)])).toContain(`:${'a'.repeat(20)}`);
+      expect(buildStarterPackStreamId(['a'.repeat(20)])).toBe(`starter_pack:all:all:${'a'.repeat(20)}`);
     });
   });
 });
--- a/src/core/models/stream/user/userStream.helper.test.ts
+++ b/src/core/models/stream/user/userStream.helper.test.ts
@@ -1,0 +1,121 @@
+import { describe, expect, it, vi } from 'vitest';
+import { AppError } from '@/libs/error/error';
+import { ValidationErrorCode } from '@/libs/error/error.codes';
+import { ErrorCategory } from '@/libs/error/error.types';
+import { buildStarterPackStreamId, USER_STREAM_TAG_DELIMITER } from './userStream.helper';
+
+const expectValidationError = (fn: () => unknown) => {
+  try {
+    fn();
+    expect.fail('Should have thrown');
+  } catch (error) {
+    expect(error).toBeInstanceOf(AppError);
+    const appError = error as AppError;
+    expect(appError.category).toBe(ErrorCategory.Validation);
+    expect(appError.code).toBe(ValidationErrorCode.INVALID_INPUT);
+  }
+};
+
+describe('buildStarterPackStreamId', () => {
+  // The checked-in STARTER_PACK_SOURCE_LIVE flag is false until Nexus deploys the
+  // starter_pack source to staging (#2390 flips it and removes the mock namespace).
+  describe('with the shipped flag value (mock namespace)', () => {
+    it('should join ordered tags under the mock namespace', () => {
+      expect(buildStarterPackStreamId(['bitcoin', 'music'])).toBe('starter_pack_mock:all:all:bitcoin,music');
+    });
+
+    it('should preserve tag order (reversed lists yield distinct IDs)', () => {
+      const forward = buildStarterPackStreamId(['travel', 'music']);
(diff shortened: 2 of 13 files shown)


src/core/services/nexus/stream/users/userStream.ts
```diff
@@ -44,8 +48,21 @@ export class NexusUserStreamService {
       case 'most_followed':
         url = userStreamApi.mostFollowed(apiParams as TUserStreamBase);
         break;
-      default:
-        throw new Error(`Invalid reach type: ${reach}`);
+      case 'starter_pack':
```
infin1t3, line 51: (Anchored on the new case because the 2/3-part parser branches it exposes aren't touched by this PR — they live at `userStream.utils.ts:62` and `:104`.)

A 2-part `user-1:starter_pack` or 3-part `starter_pack:all:all` ID bypasses every new starter-pack validation and dispatches a tag-less request.

The 2-part branch returns `reach: reach as ReachType` unchecked, and the 3-part branch does the same with `source as ReachType` — `streamRequiresUserId` only matches followers/following/friends/recommended, so `starter_pack:*` falls through with `apiParams: baseParams` and no `tags`. This new case then matches (pre-PR both shapes hit the default throw before any I/O), and `buildUserStreamUrl` silently omits the absent key, so `source=starter_pack&skip=0&limit=20` goes on the wire — with a stray `user_id` in the 2-part case. `'starter_pack:today:friends'` even typechecks as `UserStreamId` with no cast (`Pubky = string`), and it's asymmetric with this PR's own tests, which reject the 4-part `starter_pack:today:friends:bitcoin`. The rewritten default branch below also ships untested, though it stays runtime-reachable through these same casts.

Fix: reject `starter_pack` in the 2/3-part branches (or route 3-part `starter_pack:*` into the 4-part validation), and add tests for both shapes plus the default branch.

src/core/models/stream/user/userStream.helper.ts
```diff
@@ -46,6 +57,49 @@ export function parseUserCompositeId(compositeId: string): UserStreamIdParts {
...
+ * Dexie row, then validated against the canonical tag contract (1-20 chars, no banned characters)
+ * and the Nexus starter pack limit (1-5 tags). Order is preserved: Nexus interleaves per-tag
+ * rankings in the order given, so ['travel','music'] and ['music','travel'] are different streams.
+ *
+ * @example
+ * buildStarterPackStreamId(['Bitcoin ', 'music'])
+ * // Returns: 'starter_pack:all:all:bitcoin,music'
+ */
+export function buildStarterPackStreamId(tags: string[]): StarterPackStreamId {
+  const canonical = tags.map((tag) => tag.trim().toLowerCase());
```
infin1t3: Canonicalization collapses casing/whitespace variants of one tag into duplicates that then pass validation.

`buildStarterPackStreamId(['Bitcoin ', 'bitcoin'])` returns `'starter_pack:all:all:bitcoin,bitcoin'` (reproduced by executing the code), the parser accepts it and forwards `tags=bitcoin,bitcoin` verbatim, and the Dexie row is distinct from `'starter_pack:all:all:bitcoin'` for the same logical stream. Duplicates also burn the budget: `['a','A','b','B','c','C']` throws the >5-tags error with only 3 distinct tags. `normalizeTagList` (`feed.utils.ts`) already pairs the identical trim+lowercase with an order-preserving `Set` dedupe, so order — which matters here by design — survives. #2394's `useInterestTags` dedupes upstream, so the onboarding path won't hit this, but the builder is the public contract other callers get.

Fix: `[...new Set(canonical)]` before the count check.

src/core/services/nexus/stream/users/userStream.types.ts
```diff
@@ -47,6 +48,13 @@ export type TUserStreamUsernameParams = Omit<TUserStreamBase, 'preview'> & {
   username: string;
 };
 
+// tags: comma-joined ordered interest labels (1-5); Nexus rejects `tags` on any other source.
```
infin1t3, line 51: (Anchored on the exclusion note because `follow.ts` isn't touched by this PR. The problem lives at `follow.ts:190`.)

Following someone from a starter pack doesn't invalidate the cached pack, so the session keeps serving people the viewer already followed.

This comment documents that Nexus drops the viewer and everyone they follow — the same class of stream as RECOMMENDED. But `LocalFollowService`'s stream update removes the followee from `UserStreamTypes.RECOMMENDED` only; nothing removes from, deletes, or expires `starter_pack:*` rows, `user_streams` has no TTL, and `getOrFetchStreamSlice` is cache-first. Follow a suggested user (the primary onboarding action), re-read the slice, and the cached row still contains them where a fresh fetch would exclude them — the identical sequence on RECOMMENDED removes them at exactly that site. Cross-account is fine (`clearDatabase()` on auth transitions). Latent until #2388 wires a consumer.

Fix: at `follow.ts:190`, also remove the followee from (or drop) rows whose id starts with `starter_pack:`.

src/core/models/stream/user/userStream.helper.ts
```diff
@@ -46,6 +57,49 @@ export function parseUserCompositeId(compositeId: string): UserStreamIdParts {
   };
 }
 
+/**
+ * Build a starter pack stream ID from ordered interest tags.
+ *
+ * Labels are canonicalized (trimmed + lowercased) so 'Bitcoin' and 'bitcoin' resolve to the same
+ * Dexie row, then validated against the canonical tag contract (1-20 chars, no banned characters)
+ * and the Nexus starter pack limit (1-5 tags). Order is preserved: Nexus interleaves per-tag
```
infin1t3, line 65: (Anchored on the interleaving note because `users.ts` isn't touched by this PR. The problem lives at `users.ts:145`.)

If the interleave can rank one user under two requested tags, a single page carries them twice and nothing on our side dedupes it.

`fetchStreamFromNexus` filters only against `cachedStream.stream`, and only when appending — the initial page is stored and returned raw. `useUserStream`'s `loadMore` builds its `Set` from `prev` only, so two copies inside one page both pass, and consumers key rows by `user.id` (`WhoToFollow.tsx:84`). Whether Nexus dedupes the interleave server-side isn't verifiable from this repo — no fixture or contract text settles it, and the removed mock aliased `most_followed`, a single ranking that couldn't exhibit this. Duplicate React keys and doubled rows if it can happen.

Fix: uniq the fetched page before storing/returning it; cheap regardless of what Nexus guarantees.

src/core/models/stream/user/userStream.helper.ts
```diff
@@ -46,6 +57,49 @@ export function parseUserCompositeId(compositeId: string): UserStreamIdParts {
...
+ */
+export function buildStarterPackStreamId(tags: string[]): StarterPackStreamId {
+  const canonical = tags.map((tag) => tag.trim().toLowerCase());
+
+  if (canonical.length === 0 || canonical.length > STARTER_PACK_MAX_TAGS) {
+    throw Err.validation(
+      ValidationErrorCode.INVALID_INPUT,
+      `Starter pack streams require 1-${STARTER_PACK_MAX_TAGS} tags`,
+      {
+        service: ErrorService.Nexus,
```
infin1t3: These two validation errors are tagged `service: ErrorService.Nexus` from a models-layer function that performs no Nexus I/O.

docs/error-handling.md defines Services as "WHERE it originated" with `Local` for client-side operations; the other 37 error throws under `src/core/models/` use `ErrorService.Local` (one `PubkyAppSpecs`), making these the models layer's only Nexus attribution.

Fix: `ErrorService.Local` at both sites. The parser's twins can stay `Nexus` — that file is the service layer.

src/core/models/stream/user/userStream.helper.ts
```diff
@@ -46,6 +57,49 @@ export function parseUserCompositeId(compositeId: string): UserStreamIdParts {
...
+      'Starter pack tags must be 1-20 characters without banned characters',
+      {
+        service: ErrorService.Nexus,
+        operation: 'buildStarterPackStreamId',
+        context: { invalidLabels },
+      },
+    );
+  }
+
+  return `${STARTER_PACK_STREAM_SOURCE}:all:all:${canonical.join(USER_STREAM_TAG_DELIMITER)}` as StarterPackStreamId;
```
infin1t3: The `as StarterPackStreamId` cast is unnecessary and actively masks grammar drift.

The return expression typechecks with the cast deleted (verified with the repo's tsc), and with it present, editing the `StarterPackStreamId` template while the builder still emits `:all:all:` compiles silently; cast-free, that's a TS2322. The `:all:all:` shape is also spelled independently in the type (`userStream.types.ts:28`), here, and the parser's `timeframe !== 'all' || reach !== 'all'` gate (`userStream.utils.ts:124`), and this template spells `:` literally with `USER_STREAM_ID_DELIMITER` defined 85 lines up in the same file.

Fix: delete the cast — two tokens, restores compile-time coupling. Longer term, a shared `'starter_pack:all:all:'` prefix constant can feed the type, the builder, and a `startsWith` check in the parser.

src/core/services/nexus/stream/users/userStream.utils.ts
```diff
@@ -102,7 +106,65 @@ export function createUserStreamParams(
...
+        {
+          service: ErrorService.Nexus,
+          operation: 'createUserStreamParams',
+          context: { streamId },
+        },
+      );
+    }
+
+    const tags = tagSegment.split(USER_STREAM_TAG_DELIMITER);
+    const hasInvalidLabel = tags.some((label) => label !== label.trim().toLowerCase() || !isValidTagLabel(label));
```
infin1t3: The canonical-form rule itself is hand-written twice, and nothing ties builder output to parser acceptance.

The count bound and charset are already shared (`STARTER_PACK_MAX_TAGS`, `isValidTagLabel`), so those can't drift — but `trim().toLowerCase()` is written out here and in the builder (`userStream.helper.ts:73`), and #2394 adds a third copy (`canonicalizeInterestTag`). The reject-vs-normalize asymmetry is documented intent (the "instead of silently normalizing" test), but the kernel isn't: add Unicode normalization or `sanitizeTagInput` to one side only and builder-produced, Dexie-persisted IDs start throwing at fetch time, with no test catching it — the parser suite uses only hand-written literals.

Fix: one exported `canonicalizeTagLabel` used by all three sites, plus a single round-trip test `createUserStreamParams(buildStarterPackStreamId(tags))`.

src/core/services/nexus/stream/users/userStream.ts
```diff
@@ -44,8 +48,21 @@ export class NexusUserStreamService {
...
+        url = userStreamApi.starterPack(apiParams as TUserStreamStarterPackParams);
+        break;
+      default: {
+        const exhaustiveCheck: never = reach;
+        throw Err.validation(
+          ValidationErrorCode.INVALID_INPUT,
+          `Unsupported user stream reach: ${String(exhaustiveCheck)}`,
```
infin1t3: The rewritten error keeps calling a source a "reach", and Sentry will now say "reach: starter_pack" — a value that matches no reach in the Nexus API.

`ReachType` keys four `UserStreamSource` values (recommended, influencers, most_followed, now starter_pack) alongside true reaches, in a codebase that deliberately separates the two vocabularies. The wart predates this PR, but the PR rewrote every other word of this throw and kept the noun, and the new 4-part branch binds `reach` to the ID's genuine reach segment (`'all'`) and then returns `reach: 'starter_pack'` 39 lines later. A nit, not a bug — telemetry wording and readability only.

Fix: rename the discriminant (`endpoint` or `dispatch`) — it touches only files this PR already modifies — or at least drop "reach" from the message.

src/core/models/stream/user/userStream.helper.test.ts
```diff
@@ -0,0 +1,87 @@
+import { describe, expect, it } from 'vitest';
+import { AppError } from '@/libs/error/error';
+import { ValidationErrorCode } from '@/libs/error/error.codes';
+import { ErrorCategory } from '@/libs/error/error.types';
+import { buildStarterPackStreamId, USER_STREAM_TAG_DELIMITER } from './userStream.helper';
+
+const expectValidationError = (fn: () => unknown) => {
```
infin1t3, line 7: `expectValidationError` catches its own `expect.fail`, and it's copy-pasted byte-identical into both new test files.

When `fn()` stops throwing, `expect.fail`'s AssertionError lands in the `catch` and the reported failure becomes "expected AssertionError to be an instance of AppError" (reproduced with the repo's vitest) instead of "Should have thrown" — no false pass, but a misleading diagnostic. The same idiom already exists inline in ~24 test files, so this follows convention; these are just two more copies where the first shared helper could be (`src/test-utils` has no error-assertion helper today).

Fix: one shared helper — `expect(fn).toThrowError(AppError)` plus category/code checks — in `src/test-utils`, imported by both files.

src/core/application/stream/users/users.test.ts
```diff
@@ -289,6 +289,19 @@ describe('UserStreamApplication', () => {
       expect(result.nextPageIds).toEqual(['influencer-1', 'influencer-2', 'influencer-3']);
     });
 
+    describe('starter pack caching', () => {
+      it('should persist reversed tag orders as distinct Dexie rows', async () => {
```
infin1t3, line 293: The 'starter pack caching' test sits in the `UserStreamApplication` suite but never calls the application layer.

It exercises `LocalStreamUsersService.upsert`/`findById` directly, restating two already-pinned facts: distinct ID strings → distinct rows (local-service suite) and reversed tag orders → distinct IDs (helper suite). The application layer is ID-agnostic, so the risk is low — the cost is a false coverage signal under `getOrFetchStreamSlice`.

Fix: move it to the local-service suite or drop it.

src/core/services/nexus/stream/users/userStream.utils.test.ts
```diff
@@ -278,20 +393,19 @@ describe('createUserStreamParams', () => {
       );
     });
 
-    it('should throw error for invalid format (1 part)', () => {
+    it('should throw a validation error for invalid format (1 part)', () => {
       const streamId = 'invalid-stream-id' as UserStreamId;
 
+      expectValidationError(() => createUserStreamParams(streamId, baseParams));
       expect(() => createUserStreamParams(streamId, baseParams)).toThrow(
```
infin1t3: The updated invalid-format test keeps the exact-message string assertion the migration checklist says to replace.

docs/error-handling.md, Migration Checklist item 4: "Update tests: assert on `error.category`/`code` rather than string matching" — and .greptile/rules.md repeats it ("Check `category`/`code` not message strings"). This PR performed exactly that migration on `createUserStreamParams` and added `expectValidationError` on the line above, then retained the `toThrow('Invalid stream ID: …')` string match alongside it.

Fix: drop the string match — or if something is worth pinning here, assert `context.streamId`, the only place the offending ID now survives for telemetry.

src/core/models/stream/user/userStream.helper.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    it('should reject labels containing the stream ID delimiter', () => {
+      expectValidationError(() => buildStarterPackStreamId(['bit:coin']));
+    });
+
+    it('should reject overlength labels (>20 chars)', () => {
+      expectValidationError(() => buildStarterPackStreamId(['a'.repeat(21)]));
+    });
+
+    it('should accept a label at exactly 20 chars', () => {
+      expect(buildStarterPackStreamId(['a'.repeat(20)])).toContain(`:${'a'.repeat(20)}`);
```
infin1t3: The 20-char boundary test asserts a substring every regression also contains.

`toContain(':' + 'a'.repeat(20))` passes for a 21-`a` output too, and for trailing junk after the label; every sibling construction test pins the full ID with `toBe`.

Fix: `toBe('starter_pack:all:all:' + 'a'.repeat(20))`.

src/core/services/nexus/stream/users/userStream.utils.ts
```diff
@@ -102,7 +106,65 @@ export function createUserStreamParams(
...
+          context: { streamId },
+        },
+      );
+    }
+
+    return {
+      reach: 'starter_pack',
+      apiParams: {
+        ...baseParams,
+        tags: tags.join(USER_STREAM_TAG_DELIMITER),
```
infin1t3: `tags.join(USER_STREAM_TAG_DELIMITER)` rebuilds the exact string that was split 18 lines up.

`tags` is `const` and only read between split and join, and `split(d).join(d)` is the identity for any string, so this implies a normalization that never happens — unlike the builder's join, which joins a genuinely transformed array.

Fix: `tags: tagSegment`.

src/core/services/nexus/stream/users/userStream.utils.ts
```diff
@@ -102,7 +106,65 @@ export function createUserStreamParams(
...
+      },
+    } as NexusParamsResult<'starter_pack'>;
+  }
+
+  throw Err.validation(
+    ValidationErrorCode.INVALID_INPUT,
+    `Invalid stream ID: expected 2, 3, or 4 parts separated by "${DELIMITER}"`,
```
infin1t3, line 185: (Anchored on the new error because the JSDoc it contradicts, at line 36, isn't touched by this PR.)

The function's JSDoc still says "Handles two formats" while this error says "expected 2, 3, or 4 parts" and the branch above accepts the third format.

An IDE hover tells a caller 4-part IDs are invalid input, 120 lines above the branch that accepts them.

Fix: add the 4-part `source:timeframe:reach:tags` line to the JSDoc's format list.

src/core/models/stream/user/userStream.helper.ts
```diff
@@ -46,6 +57,49 @@ export function parseUserCompositeId(compositeId: string): UserStreamIdParts {
...
+ * Labels are canonicalized (trimmed + lowercased) so 'Bitcoin' and 'bitcoin' resolve to the same
+ * Dexie row, then validated against the canonical tag contract (1-20 chars, no banned characters)
+ * and the Nexus starter pack limit (1-5 tags). Order is preserved: Nexus interleaves per-tag
+ * rankings in the order given, so ['travel','music'] and ['music','travel'] are different streams.
+ *
+ * @example
+ * buildStarterPackStreamId(['Bitcoin ', 'music'])
+ * // Returns: 'starter_pack:all:all:bitcoin,music'
+ */
+export function buildStarterPackStreamId(tags: string[]): StarterPackStreamId {
```
infin1t3, line 73: `buildStarterPackStreamId` is exported with no importer outside its own test file, which rules.md flags unless the export carries an exception comment.

.greptile/rules.md, "Audit redundant exports": a symbol never imported by another file "should be removed or demoted to a non-exported local", with a carve-out for external/test consumers that requires "a short comment like `// exported for integration tests`". This one is deliberate staging — #2386 scopes UI out and names #2387/#2388 as the consumers — but nothing in the code says so, and the next reader (or the next automated review) sees a stale export.

Fix: one comment on the export, e.g. `// exported for the onboarding UI (#2387/#2388)`. Removal would be wrong here; the annotation is the rule's own escape hatch.

infin1t3 (changes requested):

src/core/services/nexus/stream/users/userStream.ts
Orlandohub: Fixed in `81665993`: `starter_pack` is now rejected in both the 2-part reach branch and the 3-part source branch before dispatch. Tests cover `user-abc:starter_pack`, `starter_pack:all:all`, and the type-admitted `starter_pack:today:friends`; the service suite also exercises the runtime default and asserts `queryNexus` is not called.

src/core/models/stream/user/userStream.helper.ts
Orlandohub: Fixed in `81665993`: the builder now uses an order-preserving `Set` after canonicalization and before enforcing the cap. I also tightened the parser to reject duplicate labels in hand-built IDs, so `bitcoin,bitcoin` cannot create an alternate Dexie key outside the builder. Both behaviors have focused tests.

src/core/services/nexus/stream/users/userStream.types.ts
Orlandohub: Leaving `starter_pack:*` cache membership stable intentionally. #2388 requires per-card follow/unfollow state to survive list refresh via the preservation behavior; removing a followee from the pack row would make that candidate disappear and undermine toggle/unfollow and Follow All accounting. `viewer_id` still excludes users followed before the initial pack fetch. `RECOMMENDED` is a continuously refreshed discovery stream, while this is the fixed onboarding candidate set.

src/core/models/stream/user/userStream.helper.ts
Orlandohub: Not adding frontend deduplication here because Nexus already guarantees it. pubky/pubky-nexus#1024 implements `interleave_unique` in `get_starter_pack_ids` and has the focused integration test `test_starter_pack_returns_each_user_once`: https://github.com/pubky/pubky-nexus/blob/09075e83c4c922feede6c93556a67d42595b115c/nexus-common/src/models/user/stream.rs#L245-L281 and https://github.com/pubky/pubky-nexus/blob/09075e83c4c922feede6c93556a67d42595b115c/nexus-webapi/tests/stream/user/starter_pack.rs#L131-L139. Additional FE uniq would duplicate an upstream contract.

src/core/models/stream/user/userStream.helper.ts
Orlandohub: Fixed in `81665993`: both model-helper validation paths now use `ErrorService.Local`, and the helper tests pin that attribution. The service-layer parser remains `ErrorService.Nexus`.

src/core/models/stream/user/userStream.helper.ts
Orlandohub: Fixed in `81665993`: removed the `as StarterPackStreamId` cast. The return expression now has to satisfy the template-literal return type directly.

src/core/services/nexus/stream/users/userStream.utils.ts
Orlandohub: Fixed in `81665993`: added neutral `canonicalizeTagLabel` in `@/libs/utils/utils`, reused it in the builder and parser, and added a builder-to-parser round-trip test using mixed-case/whitespace input. #2394 can replace its local `canonicalizeInterestTag` with this utility when it rebases onto the updated base.

src/core/services/nexus/stream/users/userStream.ts
Orlandohub: Fixed the actionable wording in `81665993`: the error now says `Unsupported user stream source`. I left the pre-existing `ReachType` discriminant rename out of this focused fix.

src/core/models/stream/user/userStream.helper.test.ts
Orlandohub: I agree the failure diagnostic is suboptimal, but I’m leaving the shared-helper extraction out of #2391. It follows an existing repo-wide pattern, does not create a false pass, and centralizing the helper across the existing call sites is better handled as a dedicated test-infrastructure cleanup.

src/core/application/stream/users/users.test.ts
Orlandohub: Fixed in `81665993`, but kept the intended application-level coverage: the test now calls `UserStreamApplication.getOrFetchStreamSlice` for both ordered IDs, mocks the distinct Nexus responses, asserts both application-to-service calls, and verifies the two persisted Dexie rows.

src/core/services/nexus/stream/users/userStream.utils.test.ts
Orlandohub: Fixed in `81665993`: removed the exact-message assertion. The test now relies on the existing validation category/code checks.

src/core/models/stream/user/userStream.helper.test.ts
Orlandohub: Fixed in `81665993`: the 20-character boundary test now pins the complete ID with `toBe`.

src/core/services/nexus/stream/users/userStream.utils.ts
Orlandohub: Fixed in `81665993`: the validated request now passes `tagSegment` directly instead of rebuilding it with `split().join()`.

src/core/services/nexus/stream/users/userStream.utils.ts
Orlandohub: Fixed in `81665993`: the JSDoc now documents all three accepted formats, including `source:timeframe:reach:tags`.

src/core/models/stream/user/userStream.helper.ts
Orlandohub: Fixed in `81665993`: added an explicit comment that `buildStarterPackStreamId` is exported for the starter-pack onboarding consumer in #2388.

infin1t3 (approved):

PR #2396, opened 2026-08-21
feat(profile): add "Filter posts" to the profile Posts tab

### LINKED ISSUE #2234: [Task] Add 'Filter posts' to profile page (v25)
Design v25 has added a 'Filter posts' search bar to the Posts section on Profile page. This will use nexus new "search by content' query with its profile pubky filter (already in staging and prod).

https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=40260-125814&t=auh06NoXIArMYibz-1

Image

Orlandohub:
## Summary

Stacked on #2377 (full-text content search) — linked as a native GitHub stack (#2377 → this PR), so this diff shows only the new commits and GitHub retargets this PR to `dev` automatically when #2377 merges.

Closes #2234.

https://github.com/user-attachments/assets/82688b7b-9608-4935-92d8-08f8a5d62a07

### Shared stream machinery hardening (first 3 commits)

- **Stale in-flight fetch guard** (`useStreamPagination`): a fetch generation counter invalidates in-flight requests on stream switch/refresh, so late-resolving responses can no longer overwrite posts/cursor/`hasMore`, surface stale errors, or strand `loadingMore`.
- **Nexus skip bound** (`PostStreamApplication`): content-search streams stop paginating at Nexus's inclusive `BoundedSkip<1000>` — requests that would exceed the bound are short-circuited and `reachedEnd` is reported when the next offset would cross it.
- **Transactional delete** (`useDeletePost`): deletes go through `removePostsOptimistically`, committing when the local row is tombstoned/gone and rolling back when it's still live or unverifiable — keeping skip-stream cursors correct after deletions.

### Profile "Filter posts" feature (last 2 commits)

- **Author-scoped content search** end to end: `content_search:q~<query>:<kind>:<author>` stream ids, `author` forwarded to `GET /v0/search/posts/by_content`, profile-tab semantics inherited (no mute filter, collections excluded).
- **Reply exclusion** via local relationships rows: fail-open before hydration, strict in the controller's post-hydration second pass; relationships-aware cache-miss detection hydrates details-cached posts that lack a relationships row.
- **UI**: `useProfilePostsFilter` hook (500ms debounce, shared 2–30 char / max 4 term validation, immediate cancel-on-clear), `FilterPostsBar` pill above the profile Posts feed (stays mounted across stream swaps, preserving input focus), and a search-specific `FilterPostsEmpty` no-results state.

### Decisions needing sign-off

- **Replies excluded** from filtered results (decided in-session; the endpoint cannot exclude them server-side). Unclassifiable posts are hidden, never shown as potential replies.
- **Collections excluded** from the scoped search — inferred from the profile Posts tab contract ("all posts except collections"); Nexus does index collection content, so please confirm.
- **No-results copy/artwork** ("No posts match your search") is a placeholder pending design confirmation — the issue/Figma document no empty state.
- No content/kind filter UI (decided in-session); relevance order only (endpoint has no sort param).

## Test plan

- [x] `useStreamPagination` race tests: late-resolving/rejecting fetches for a previous stream are ignored; `loadingMore` is not stranded
- [x] Skip-bound tests: request at offset 1000 issued; past-bound short-circuits without a Nexus call; full page crossing the bound reports `reachedEnd`
- [x] `useDeletePost` transaction tests: commit on success/tombstone/hard-delete, rollback on live/unverifiable local state
- [x] Stream id builder/parser round-trip with author segment; malformed-id cases
- [x] Service tests: `author` carried through `breakDownStreamId`/`createPostStreamParams` into the `by_content` URL; omitted for global search
- [x] Application tests: collections + reply exclusion (fail-open/strict passes), relationships-aware cache-miss detection
- [x] Hook tests: debounce, validator integration, immediate cancel-on-clear
- [x] Component tests: `FilterPostsBar`, profile stream swap idle ↔ searching, conditional empty state
- [x] Full typecheck + 890 tests across the 31 affected suites pass
- [ ] Manual pass on staging (own profile + other-user profile)

### DIFF
--- a/src/components/molecules/FilterPostsBar/FilterPostsBar.test.tsx
+++ b/src/components/molecules/FilterPostsBar/FilterPostsBar.test.tsx
@@ -31,6 +31,29 @@ describe('FilterPostsBar', () => {
     fireEvent.change(input, { target: { value: 'bitcoin wallet' } });
     expect(onValueChange).toHaveBeenCalledWith('bitcoin wallet');
   });
+
+  it('renders the validation message below the pill with alert semantics', () => {
+    render(
+      <FilterPostsBar
+        value="one two three four five"
+        onValueChange={vi.fn()}
+        validationMessage="Search can contain up to 4 terms"
+      />,
+    );
+
+    const message = screen.getByRole('alert');
+    expect(message).toHaveTextContent('Search can contain up to 4 terms');
+    const input = screen.getByRole('textbox', { name: 'Filter posts' });
+    expect(input).toHaveAttribute('aria-invalid', 'true');
+    expect(input).toHaveAttribute('aria-describedby', message.id);
+  });
+
+  it('omits the message and invalid state when the query is fine', () => {
+    render(<FilterPostsBar value="bitcoin" onValueChange={vi.fn()} />);
+
+    expect(screen.queryByRole('alert')).not.toBeInTheDocument();
+    expect(screen.getByRole('textbox', { name: 'Filter posts' })).toHaveAttribute('aria-invalid', 'false');
+  });
 });
 
 describe('FilterPostsBar - Snapshots', () => {
--- a/src/components/molecules/FilterPostsBar/FilterPostsBar.test.tsx
+++ b/src/components/molecules/FilterPostsBar/FilterPostsBar.test.tsx
@@ -1,0 +1,39 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { CONTENT_SEARCH_QUERY_MAX_LENGTH } from '@/config/search';
+import { FilterPostsBar } from './FilterPostsBar';
+
+describe('FilterPostsBar', () => {
+  it('renders the pill container with the input and search icon', () => {
+    const { container } = render(<FilterPostsBar value="" onValueChange={vi.fn()} />);
+
+    expect(screen.getByTestId('filter-posts-bar')).toBeInTheDocument();
+    expect(screen.getByRole('textbox', { name: 'Filter posts' })).toHaveAttribute('placeholder', 'Filter posts');
+    expect(container.querySelector('svg.lucide-search')).toBeInTheDocument();
+  });
+
+  it('caps the input at the content-search max query length', () => {
+    render(<FilterPostsBar value="" onValueChange={vi.fn()} />);
+
+    expect(screen.getByRole('textbox', { name: 'Filter posts' })).toHaveAttribute(
+      'maxLength',
+      String(CONTENT_SEARCH_QUERY_MAX_LENGTH),
+    );
+  });
+
+  it('displays the controlled value and reports changes as plain strings', () => {
+    const onValueChange = vi.fn();
+    render(<FilterPostsBar value="bitcoin" onValueChange={onValueChange} />);
+
+    const input = screen.getByRole('textbox', { name: 'Filter posts' });
+    expect(input).toHaveValue('bitcoin');
+
+    fireEvent.change(input, { target: { value: 'bitcoin wallet' } });
+    expect(onValueChange).toHaveBeenCalledWith('bitcoin wallet');
+  });
+
+  it('matches snapshot', () => {
+    const { container } = render(<FilterPostsBar value="" onValueChange={vi.fn()} />);
+    expect(container.firstChild).toMatchSnapshot();
+  });
+});
--- a/src/components/molecules/FilterPostsBar/FilterPostsBar.test.tsx
+++ b/src/components/molecules/FilterPostsBar/FilterPostsBar.test.tsx
@@ -31,7 +31,9 @@ describe('FilterPostsBar', () => {
     fireEvent.change(input, { target: { value: 'bitcoin wallet' } });
     expect(onValueChange).toHaveBeenCalledWith('bitcoin wallet');
   });
+});
 
+describe('FilterPostsBar - Snapshots', () => {
   it('matches snapshot', () => {
     const { container } = render(<FilterPostsBar value="" onValueChange={vi.fn()} />);
     expect(container.firstChild).toMatchSnapshot();
--- a/src/components/molecules/FilterPostsBar/FilterPostsBar.tsx
+++ b/src/components/molecules/FilterPostsBar/FilterPostsBar.tsx
@@ -3,38 +3,51 @@
 import { Search } from 'lucide-react';
 import { Container } from '@/atoms/Container/Container';
 import { Input } from '@/atoms/Input/Input';
+import { Typography } from '@/atoms/Typography/Typography';
 import { CONTENT_SEARCH_QUERY_MAX_LENGTH, SEARCH_CLOSED_STYLE } from '@/config/search';
 import type { FilterPostsBarProps } from './FilterPostsBar.types';
+
+const MESSAGE_ID = 'filter-posts-bar-message';
 
 /**
  * FilterPostsBar
  *
  * Pill-shaped inline search input for the profile Posts tab ("Filter posts",
  * issue #2234). Presentation-only: debounce/validation live in
- * `useProfilePostsFilter`. Modeled on `SearchInputBar` minus tags/dropdown.
+ * `useProfilePostsFilter`. Modeled on `SearchInputBar` minus tags/dropdown;
+ * the validation message mirrors the `InputField` message pattern.
  */
-export function FilterPostsBar({ value, onValueChange }: FilterPostsBarProps) {
+export function FilterPostsBar({ value, onValueChange, validationMessage }: FilterPostsBarProps) {
   return (
-    <Container
-      data-testid="filter-posts-bar"
-      data-cy="profile-filter-posts"
-      className="flex h-12 min-w-0 shrink-0 items-center gap-3 rounded-full border border-border px-6 py-3"
-      style={SEARCH_CLOSED_STYLE}
-      overrideDefaults
-    >
-      <Input
-        type="text"
-        placeholder={'Filter posts'}
-        value={value}
-        onChange={(event) => onValueChange(event.target.value)}
-        maxLength={CONTENT_SEARCH_QUERY_MAX_LENGTH}
-        data-cy="profile-filter-posts-input"
-        aria-label={'Filter posts'}
-        className="h-auto min-w-20 flex-1 border-none bg-transparent pr-0 pl-0 text-base font-medium text-foreground md:text-base"
-      />
-      <span className="pointer-events-none -mr-2 flex size-8 shrink-0 items-center justify-center" aria-hidden="true">
-        <Search className="size-4 text-muted-foreground" />
-      </span>
+    <Container overrideDefaults className="flex min-w-0 shrink-0 flex-col gap-2">
+      <Container
+        data-testid="filter-posts-bar"
+        data-cy="profile-filter-posts"
+        className="flex h-12 min-w-0 shrink-0 items-center gap-3 rounded-full border border-border px-6 py-3"
+        style={SEARCH_CLOSED_STYLE}
+        overrideDefaults
+      >
+        <Input
+          type="text"
+          placeholder={'Filter posts'}
+          value={value}
+          onChange={(event) => onValueChange(event.target.value)}
+          maxLength={CONTENT_SEARCH_QUERY_MAX_LENGTH}
+          data-cy="profile-filter-posts-input"
+          aria-label={'Filter posts'}
+          aria-invalid={!!validationMessage}
+          aria-describedby={validationMessage ? MESSAGE_ID : undefined}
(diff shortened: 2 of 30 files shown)


src/components/molecules/FilterPostsBar/FilterPostsBar.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+    expect(input).toHaveValue('bitcoin');
+
+    fireEvent.change(input, { target: { value: 'bitcoin wallet' } });
+    expect(onValueChange).toHaveBeenCalledWith('bitcoin wallet');
+  });
+
+  it('matches snapshot', () => {
+    const { container } = render(<FilterPostsBar value="" onValueChange={vi.fn()} />);
+    expect(container.firstChild).toMatchSnapshot();
+  });
```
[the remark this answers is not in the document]
Orlandohub: Addressed in 91bbb2bf — both snapshot cases moved into dedicated `FilterPostsBar - Snapshots` / `FilterPostsEmpty - Snapshots` blocks; snap keys regenerated with identical content.

src/core/application/stream/posts/post.ts
```diff
@@ -128,19 +130,57 @@ export class PostStreamApplication {
...
+  }): Promise<string[]> {
+    if (postIds.length === 0 || !isAuthorScopedContentSearchStream(streamId)) {
+      return postIds;
+    }
+
+    const relationships = await LocalPostService.readRelationshipsByIds(postIds);
+    return postIds.filter((_postId, index) => {
+      const relationship = relationships[index];
+      if (!relationship) {
+        return !strict;
```
infin1t3, line 181: On a cold cache every result is a miss. `fetchMissingPostsFromNexus` swallows a failed `by_ids` call, the controller still sees the miss list and runs this strict pass, and every id without a relationships row gets dropped. The user then sees "No posts match your search" even though the search did not succeed. That also fights the rule in this PR that an unclassified post must never render as a possible reply. An empty page looks like a successful search with zero hits.

src/core/controllers/stream/posts/posts.ts
```diff
@@ -83,7 +83,13 @@ export class StreamPostsController {
       // Second-pass: cache-miss details are now resolved,
       // re-filter to catch posts that were fail-open in the first pass.
       // Only the visible page shrinks here — the raw anchor passes through untouched.
-      visibleIds = await PostStreamApplication.filterStreamPosts({ streamId, postIds: nextPageIds });
+      // Strict reply classification: after hydration, an author-scoped content-search
+      // result that still cannot be classified is hidden, never risked as a reply.
+      visibleIds = await PostStreamApplication.filterStreamPosts({
```
infin1t3, line 109: This second pass only runs when `cacheMissPostIds.length > 0`. Overflow early-return sends empty `cacheMissIds`, so buffered ids never get the strict pass. If hydration already succeeded, the non-strict filter still drops replies because the relationships rows are there. A reply only leaks when hydration failed and load-more then serves the buffer fail-open. Same hole as the strict drop on the application filter.

src/hooks/useProfilePostsFilter/useProfilePostsFilter.ts
```diff
@@ -0,0 +1,47 @@
...
+ * id and `useStreamPagination` does the rest.
+ */
+export function useProfilePostsFilter(): UseProfilePostsFilterResult {
+  const [inputValue, setInputValue] = useState('');
+  const [activeQuery, setActiveQuery] = useState<string | null>(null);
+
+  const debouncedApplyRef = useRef(
+    debounce((rawValue: string) => {
+      const validation = validateContentSearchQuery(rawValue);
+      setActiveQuery(validation.isValid ? validation.query : null);
```
infin1t3: The validator returns a message. The hook discards it and sets `activeQuery` to `null`, which the UI reads as no filter. Five terms, or a single character after the debounce, still looks like a filter is on while the whole profile feed is showing. There is no signal to the user.

src/core/application/stream/posts/post.ts
```diff
@@ -128,19 +130,57 @@ export class PostStreamApplication {
...
+  }: {
+    streamId: PostStreamId;
+    postIds: string[];
+    strict: boolean;
+  }): Promise<string[]> {
+    if (postIds.length === 0 || !isAuthorScopedContentSearchStream(streamId)) {
+      return postIds;
+    }
+
+    const relationships = await LocalPostService.readRelationshipsByIds(postIds);
```
infin1t3, line 177: A Dexie failure in `readRelationshipsByIds` aborts pagination. Same uncaught read at the miss-detection call later in this file. The comparison to line 436 is the wrong analogy: that call is an optional fetch of repost originals, not a filter. Sibling filters are not consistent. The deleted filter is fail-open; the collections filter also throws. Fail-open here would let replies through, which this PR does not want.

src/hooks/useDeletePost/useDeletePost.tsx
```diff
@@ -57,14 +57,21 @@ export function useDeletePost(options?: UseDeletePostOptions): UseDeletePostResu
...
     setIsDeleting(true);
 
-    // Optimistically remove post from timeline feed
-    timelineFeed?.removePosts(postId);
+    // Optimistically remove the post as a transaction: the commit path also decrements
+    // skip-stream offsets (content search, engagement, collections), so the next page
+    // doesn't skip one live row once Nexus drops the deleted post. The plain removePosts
+    // fallback covers providers that don't expose the transactional API.
+    const removal = timelineFeed?.removePostsOptimistically?.(postId);
+    if (!removal) {
```
infin1t3, line 65: `TimelineFeedContent` already always provides the transactional API. This still branches to `removePosts` / `prependPosts`, and the comment above says that covers providers that omit the transactional API. Optional on the type is a capability flag. Those fallbacks are not a production path after this change. The transactional path does the skip-cursor work it was added for.

src/core/application/stream/posts/post.ts
```diff
@@ -128,19 +130,57 @@ export class PostStreamApplication {
...
     postIds: string[];
+    strictReplyClassification?: boolean;
   }): Promise<string[]> {
     // Bookmark and single-collection item feeds keep deleted posts so the post
     // component can render its deleted-state placeholder; all other streams drop them.
     const visiblePostIds = isDeletedRetainingStream(streamId)
       ? postIds
       : await LocalPostService.filterDeletedPosts(postIds);
     const afterCollections = await this.filterCollectionsFromStream({ streamId, postIds: visiblePostIds });
+    const afterReplies = await this.filterRepliesFromAuthorScopedSearch({
```
infin1t3, line 146: This adds a relationships read in the reply filter, and miss detection reads relationships again, then the controller's second pass reads details and relationships on top of the existing deleted and collections filters. That's about nine sequential IndexedDB reads per search page.

src/core/application/stream/posts/post.ts
```diff
@@ -727,6 +797,11 @@ export class PostStreamApplication {
 
   /** Collections appear only on streams whose id encodes `kind=collection` (e.g. timeline:all:collection). */
   private static shouldExcludeCollectionsFromStream(streamId: PostStreamId): boolean {
+    // Profile search inherits the Posts tab contract: collections have their own tab.
+    if (isAuthorScopedContentSearchStream(streamId)) {
```
infin1t3, line 815: Mute skip, exclude collections, reply filter, and relationships-aware miss all re-derive the Posts-tab contract. Fine. Split by concern, not a bug.

src/core/application/stream/posts/post.ts
```diff
@@ -710,6 +763,23 @@ export class PostStreamApplication {
...
+   * `relationships.replied`, so a details-cached post without one must still be hydrated —
+   * otherwise it would stay fail-open (or be dropped by the strict second pass) forever.
+   */
+  private static async getCacheMissPostIds(streamId: PostStreamId, postIds: string[]): Promise<string[]> {
+    const missingDetailsIds = await this.getNotPersistedPostsInCache(postIds);
+    if (postIds.length === 0 || !isAuthorScopedContentSearchStream(streamId)) {
+      return missingDetailsIds;
+    }
+
+    const relationships = await LocalPostService.readRelationshipsByIds(postIds);
```
infin1t3: The new miss path treats a missing relationships row as a miss. The tombstone still has details; the relationships row is gone for good, so those ids go to `by_ids` and the deleted filter drops them anyway.

src/hooks/useProfilePostsFilter/useProfilePostsFilter.ts
```diff
@@ -0,0 +1,47 @@
...
+ * typing, validated through the shared content-search rules (2–30 chars, max
+ * 4 terms). Invalid or empty input yields `null`, keeping the ordinary
+ * profile feed. No pagination logic lives here — the caller swaps the stream
+ * id and `useStreamPagination` does the rest.
+ */
+export function useProfilePostsFilter(): UseProfilePostsFilterResult {
+  const [inputValue, setInputValue] = useState('');
+  const [activeQuery, setActiveQuery] = useState<string | null>(null);
+
+  const debouncedApplyRef = useRef(
```
infin1t3, line 25: Lodash debounce in a ref, with cancel on unmount and cancel on clear. Same pattern as `useSearchAutocomplete`. `useDebounceCallback` exists in the repo, but sibling search already uses lodash.

src/components/molecules/FilterPostsBar/FilterPostsBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * Pill-shaped inline search input for the profile Posts tab ("Filter posts",
+ * issue #2234). Presentation-only: debounce/validation live in
+ * `useProfilePostsFilter`. Modeled on `SearchInputBar` minus tags/dropdown.
+ */
+export function FilterPostsBar({ value, onValueChange }: FilterPostsBarProps) {
+  return (
+    <Container
+      data-testid="filter-posts-bar"
+      data-cy="profile-filter-posts"
+      className="flex h-12 min-w-0 shrink-0 items-center gap-3 rounded-full border border-border px-6 py-3"
```
infin1t3: The icon span below is copied verbatim from `SearchInputBar`. The container and input are not: no `relative`, has `shrink-0`, always `rounded-full`. Container and input already differ, so there is no copy-paste drift to clean up.

src/components/molecules/FilterPostsBar/FilterPostsBar.types.ts
```diff
@@ -0,0 +1,6 @@
+export interface FilterPostsBarProps {
```
infin1t3, line 1: `FilterPostsBar.types.ts` and `useProfilePostsFilter.types.ts` each have a single consumer. The repo convention is a sibling `*.types.ts`. Inlining is not the rule.

src/components/molecules/FilterPostsBar/FilterPostsBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+      style={SEARCH_CLOSED_STYLE}
+      overrideDefaults
+    >
+      <Input
+        type="text"
+        placeholder={'Filter posts'}
+        value={value}
+        onChange={(event) => onValueChange(event.target.value)}
+        maxLength={CONTENT_SEARCH_QUERY_MAX_LENGTH}
```
infin1t3: This is the first input that puts HTML `maxLength` and the code-point validator on the same field. `maxLength` counts UTF-16 units. A pasted 32-character ASCII phrase is truncated to 30 and searched as 30, so the validator max never surfaces in the UI.

src/core/models/stream/post/postStream.types.ts
```diff
@@ -204,25 +208,38 @@ export function buildCollectionItemsStreamId(authorPubky: Pubky, postId: string)
...
     return null;
   }
 
   try {
     const query = decodeURIComponent(markedQuery.slice(CONTENT_SEARCH_QUERY_MARKER.length));
-    return query ? { query, kind: parsedKind } : null;
+    if (!query) {
+      return null;
+    }
+    return author ? { query, kind: parsedKind, author } : { query, kind: parsedKind };
```
infin1t3, line 242: Two object literals only to avoid an `author: undefined` key. Nothing reads that as a present key.

src/hooks/useProfilePostsFilter/useProfilePostsFilter.constants.ts
```diff
@@ -0,0 +1,2 @@
+/** Debounce delay before an edited filter query hits Nexus (same cadence as autocomplete). */
```
infin1t3: `PROFILE_POSTS_FILTER_DEBOUNCE_MS` is 500 and the comment says it is the same cadence as autocomplete, but it does not derive from `AUTOCOMPLETE_DEBOUNCE_MS`. Both are 500. The comment overclaims.

infin1t3 (changes requested):
I have tested the happy path and looks good to me. 

Comments under are from the AI, checked afterwards most of them are edge cases/hygiene. Only one is high prio.

infin1t3 (approved):

PR #2411, opened 2026-08-25
fix: mention popover clickthrough #2406

### LINKED ISSUE #2406: [Bug]: Unable to select profile/username mention in popover when creating a post or in dialog
### Description

On both desktop and mobile, when creating a post, editing, replying, or reposting it isn't possible to tap on the profile/username mention list. The tap appears to be going through to the dialog and hitting buttons there instead of selecting the profile from the list.

### Expected Behaviour

When mentioning a username, it should always be possible to select them from the list.

### Steps to Reproduce

1. Reply to a post
2. Type `@abcd` (where 'abcd' is the start of some profile's username)
3. When the list dropdown appears, tap their username
4. Observe their pubky is not inserted into the post area

### Screenshots

https://github.com/user-attachments/assets/343d8a10-8fd7-46f6-a27c-0942a941cb60

### Console Log

```shell

```

### Browser & Version

_No response_

catch-21:
Fixes #2406 

Can now click on profile mention in dialog on desktop and mobile. Can't see any negative side-effects.

https://github.com/user-attachments/assets/98e5dab3-9dc7-42eb-9914-f829193b130f

### DIFF
--- a/src/components/molecules/MentionPopover/MentionPopover.test.tsx
+++ b/src/components/molecules/MentionPopover/MentionPopover.test.tsx
@@ -1,5 +1,8 @@
+import { useRef } from 'react';
 import { act, fireEvent, render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


infin1t3 (approved):
Looks ok.

Issue #2369, opened 2026-08-19
Locks: remove dev-static verifier and password lock UI


Taewa:
Sub-issue of #2364. **Do this last**, after the payment flow works.

## Problem

Phase 1 ships a "password" lock that does not verify anything: the UI collects a password, the app writes a `dev-static` criterion that always passes, and the password is never sent anywhere. Phase 2 replaces it with payment, so both the placeholder verifier and the password UI go away.

## Scope

- Remove the `dev-static` verifier constant and the code paths that assume it
- Remove the password tab, its input, and its copy
- Payment becomes the only lock method

Related: #2040 tracks removing the `TODO:[Locks] #1998` markers, several of which point at exactly this placeholder.

**Why last:** removing the placeholder before payment works leaves locking broken with nothing to replace it.

## Definition of Done

- [ ] Creating a lock offers payment only — no password option anywhere in the flow.
- [ ] No user-facing copy still refers to password-protected content.
- [ ] Existing paid locks keep working.
- [ ] The `dev-static` verifier is gone from the codebase.

Issue #2365, opened 2026-08-19
Locks: local dev setup for the Paykit payment stack


Taewa:
Sub-issue of #2364. Prerequisite for the rest of Phase 2.

## Problem

Phase 1 dev runs against a 5-service stack with no payment infrastructure. Paykit Server needs its own database, a Bitcoin regtest chain, and an Electrum indexer to observe payments — 14 services total. Without this, none of the Phase 2 work can be run or reviewed locally.

**Long-running (9)**

| service | role | vs Phase 1 |
| --- | --- | --- |
| `postgres` | Locks database | existing |
| `pubky-testnet` | identity, homeserver, PKARR relay | existing |
| `locks-server` | Lock Server | existing |
| `creator-demo` | locks creator browser demo | existing |
| `reader-demo` | locks reader browser demo + native Paykit reader worker | existing (worker added) |
| `paykit-server` | derives addresses, delivers payment requests, watches the chain | new |
| `paykit-postgres` | Paykit Server database, separate from Locks | new |
| `bitcoin` | Bitcoin Core regtest — throwaway chain, no real money | new |
| `fulcrum` | Electrum indexer — lets Paykit Server ask "did this address get paid?" | new |

A Bitcoin node stores the ledger in time order and cannot answer per-address queries. Fulcrum builds the address index that makes payment detection possible.

**One-shot jobs (5)** — these exit after succeeding; `exited` is their healthy state

| job | role |
| --- | --- |
| `compose-bootstrap` | generates local credentials, service env files, state directories |
| `bitcoin-bootstrap` | creates the miner wallet and mines 101 blocks so coinbase funds are spendable |
| `electrum-readiness` | verifies Fulcrum answers the Electrum protocol, not just that the port is open |
| `paykit-config` | waits for the Lock Server's runtime public key, then writes Paykit Server config trusting that exact identity |
| `demo-config` | writes the shared browser config from the runtime Locks identity and testnet endpoints |

The two browser demos (`creator-demo`, `reader-demo`) belong to the locks repo, not pubky-app. They are the reference implementation and the fallback way to exercise the flow while pubky-app is being wired up.

## Scope

- Run the Paykit demo stack from pubky/locks (`compose.paykit-local-demo.yaml`)
- Point pubky-app at it: local Lock Server, PKARR relay, homeserver
- Link `@pubky/locks-sdk` to the locks local build — not published to npm, and Phase 2 needs API added in pubky/locks#10
[...]

PR #2413, opened 2026-08-25
feat(locks): connect a Bitkit payout account before locking content (#2366)

### LINKED ISSUE #2366: Locks: creator — connect a Paykit payout account
Sub-issue of #2364.

## Problem

A creator cannot charge for content without telling Paykit where the money goes. Nothing in the app connects a payout account yet.

## Where this lives

Not the publish dialog's payment tab. That tab holds the price input, and both belong to #2367.

This ticket adds a step to the Locks auth dialog (#2080), between "Enable Locks" and "Locks Enabled":

```
Intro -> Enable Locks (Pubky Ring) -> Enable Payments (Bitkit) -> Locks Enabled
```

Design: [Enable Payments, desktop](https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=43039-191109&m=dev). The mobile pattern (a button instead of a QR) is [here](https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=43039-280871&m=dev).

When the creator already has a Locks session but has not connected Bitkit in this browser session, the dialog opens straight at Enable Payments.

## Scope

- A step that embeds Paykit Server's `/setup` page in an iframe, where the creator approves with Bitkit
- Accept the completion message and remember it for the current browser session
- Turning the lock switch on leads to this step until Bitkit is connected
- A Paykit Server origin in app config (`PUBKY_RUNTIME_PAYKIT_SERVER_URL`)

### The iframe

Same split as the Phase 1 Locks auth step (#2080): pubky-app owns the modal, and the content inside it is Paykit Server's own `/setup` page.

Paykit Server's design doc (`docs/plans/0001-receiver-only-prototype-design.md`) fixes this split:

> - Pubky.app browser starts setup.
> - Paykit Server hosts setup iframe and secret-bearing QR/deep link.
> - Iframe/postMessage is the only completion mode; no redirect fallback.
> - auth URL, session secret, xpub, and other long-lived credentials never cross postMessage.

The last line is why pubky-app cannot render the QR itself: it is never given the auth URL.

FE work here:

- Entry URL `GET /setup?return_to=<our origin>&state=<opaque>`
- Accept `{ type: "paykit-setup-callback", state }` only from that exact origin, that exact iframe window, and the pending state
- States: waiting for approval, connected (advance to the next step), failed (with retry), dismissed

pubky-app has no Content-Security-Policy today, so no `frame-src` change is needed here. Per-environment Paykit config and deployment are #2372.

**Paykit Server does not render the QR yet.** Its `/setup` page returns the auth URL as text plus CLI commands. Checked again on 2026-08-20 against `f38c791`, the current `origin/master` and the commit the local demo stack pins: no QR or SVG code in the repo, and no open PR or issue about it. The page also sends no height message, so the modal cannot size the iframe to its content.

**This does not block the work.** Build against the current text page and leave the QR out. It appears inside Paykit Server's iframe, so it lands on its own once implemented, with no change on our side. The postMessage contract is identical whether the approval came from Bitkit or from the CLI, so every FE path is testable today.

**Mobile shows a button, not a QR.** That button opens Bitkit through a deep link, which again needs the auth URL, so it belongs to the Paykit page as well. The Lock Server `/connect` shell already branches the same way. Ask the Paykit team for both together.

### Remembering the connection

Keep it in memory for the current browser session. It is not written to storage, for two reasons:

- There is no way to ask Paykit whether a creator is connected. The completion message is the only signal we get.
- A stored flag can be wrong. It can be edited by hand, or left over from a Paykit database reset, or absent on another device. A wrong "connected" does not fail at publish time. It fails much later, when a reader tries to unlock and Paykit refuses to create the invoice, and the creator never sees it.

Approving again is harmless: setup creates or re-authenticates exactly one creator account. So reloading the page means approving in Bitkit once more. That is accepted for now. A durable check is possible later by reading the Receiver Marker that Paykit publishes on the creator's homeserver, but it needs the Locks creator pubky (#2283) and per-environment config (#2372).

### Local approval stand-in

In place of scanning with Bitkit:

```bash
npm --prefix examples/js-sdk run generate-paykit-account-tpub

printf '%s\n%s\n%s\n' '<auth URL shown in the iframe>' '<tpub>' '0' \
  | docker compose -f compose.paykit-local-demo.yaml exec -T creator-demo \
      npm --prefix examples/js-sdk run authenticate-paykit -- --role content-creator
```

The iframe polls, sees the approval, and posts the completion message exactly as it will with Bitkit. See #2365 for the stack this runs against. The Paykit demo config only allows the demo origin, so add the pubky-app dev origin to `[setup] allowed_origins` before testing.

## Definition of Done

The two things a creator must be able to do:

- [ ] **Connect Bitkit.** After authorizing Locks, the dialog shows a step for connecting the payout account, and approving there completes it. (The QR comes from Paykit Server's page — see the note above; until it renders one, the local CLI stand-in proves the same path.)
- [ ] **Move on once connected.** The step advances on its own to "Locks Enabled", and the creator reaches the lock dialog.

Supporting behaviour:

- [ ] While waiting for approval, the step says so rather than looking stuck.
- [ ] Until Bitkit is connected, turning the lock switch on leads to this step instead of the lock configuration dialog.
- [ ] Cancelling or failing the step leaves the creator able to retry.
- [ ] Signing in as a different Locks creator does not inherit the previous creator's connected state.
- [ ] Reloading the page asks for Bitkit again. This is intended, for the reasons above.

Note for Phase 1 password locks: because the step gates the lock switch, connecting Bitkit becomes a prerequisite for any lock, including a password one, until #2369 removes the password method.

Taewa:
Closes #2366.

Before a creator can lock content, they now connect the Bitkit account that will receive payments. The Locks auth dialog gains an **Enable Payments** step after **Enable Locks**: it loads Paykit's setup page in an iframe, the creator approves in Bitkit, and the dialog moves on to the lock configuration on its own.

The lock switch stays hidden until both the Lock Server and the Paykit Server are configured. That means a deploy with only `PUBKY_RUNTIME_LOCK_SERVER` set loses the lock feature entirely, password locks included — worth setting `PUBKY_RUNTIME_PAYKIT_SERVER_URL` alongside it before this lands anywhere.

Unlike the Locks sign-in, no session comes back from the Bitkit approval — the app only learns that it succeeded. That fact is kept in memory for the browser session, so a reload asks for the QR again. Approving twice is harmless: setup re-authenticates the same account.

Also in here, since the same dialogs and the composer row were involved:

- Both auth steps now match Figma — header spacing, the QR block, and the wordmark plus store badges under it. The Lock Server shell hides its own wordmark on desktop, so the modal carries one.
- On mobile the Post button moves to its own full-width row. The lock switch had widened the action bar enough to push it off the edge.

## Testing

Run against the local Paykit stack (#2365). The Bitkit approval was done through the CLI stand-in described in #2366 rather than a real wallet, which exercises the same callback the app receives.

Both steps were checked on desktop and mobile widths, including the deep-link layout touch devices get in place of the QR.

## Note

Password locks go away as Phase 2 finishes (#2369) — payment is the only unlock method that ships. Until then the Enable Payments step gates both, so connecting Bitkit is a prerequisite for a password lock too.

https://github.com/user-attachments/assets/71f948b3-e193-4166-a747-fce0b59415b7


src/hooks/usePaykitSetupFlow/usePaykitSetupFlow.utils.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (typeof data !== 'object' || data === null) return null;
+
+  const message = data as Record<string, unknown>;
+  if (message.type !== PAYKIT_SETUP_MESSAGE_TYPE) return null;
+
+  const { state, error } = message;
+  if (typeof state !== 'string' || state.length === 0) return null;
+
+  if (typeof error === 'string' && error.length > 0) return { state, error };
+  return { state };
```
secondl1ght, line 44: Non-blocking, but worth tightening: a malformed failure payload reads as a success here.

Success only requires a non-empty `state`, so `{ type, state, error: { code: 'declined' } }` — or `error: 0` — falls through to `return { state }`. The hook then treats that as an approval and calls `markPaykitConnected()`, and the creator is marked connected with no payout account behind it. The failure is silent in exactly the direction that costs money later.

The sibling `readLocksAuthBridgeMessage` is biased the other way: success there needs a `code`, so a malformed payload returns `null` and the flow just keeps waiting. Suggest matching that — an `error` key that is present but isn't a usable string is a signal something went wrong, not a signal that nothing did:

```ts
if (error !== undefined && error !== null && error !== '') {
  return typeof error === 'string' ? { state, error } : { state, error: 'unknown' };
}
return { state };
```

(or just `return null` in the non-string case, so the one-shot listener stays armed rather than erroring out).

Either way `usePaykitSetupFlow.utils.test.ts:47` needs updating — the `0` and `{}` cases in `'treats a blank or non-string error as no error'` currently pin the unsafe reading. `''` and `null` as "no error" seem right to keep.

---

src/libs/runtime-config/runtime-config.ts
```diff
@@ -171,6 +171,9 @@ export const getDeployEnv = (): DeployEnv => getRuntimeConfig().deployEnv;
 // Optional Lock Server pubky the composer's lock flow signs into (undefined = Locks disabled).
 export const getLockServer = (): string | undefined => getRuntimeConfig().lockServer;
 
+// Optional Paykit Server address for the creator's payout setup (undefined = Locks disabled).
+export const getPaykitServerUrl = (): string | undefined => getRuntimeConfig().paykitServerUrl;
```
secondl1ght, line 175: Follow-up suggestion rather than a change request: this new value makes Locks fail silently on a partial deploy, and nothing says so at startup.

As the PR description notes, an existing deploy that has `PUBKY_RUNTIME_LOCK_SERVER` set but not `PUBKY_RUNTIME_PAYKIT_SERVER_URL` now loses the lock switch entirely — password locks included. Both are optional in the schema, so the config parses clean and the only symptom is a control that quietly stopped rendering. That is a rough thing to debug from the outside.

`warnIfModerationDisabled` in this file is the precedent for exactly this shape of problem — an optional value whose absence disables a feature, warned about once on deployed startup so operators can tell "intentionally off" from "someone missed an env var". A matching warn would fit:

```ts
export function warnIfLocksDisabledByPartialConfig(config: RuntimeConfig): void {
  // ...same didWarn / isRuntimeConfigRequired / build-phase guards
  if (!config.lockServer || config.paykitServerUrl) return;
  console.warn(
    '[runtime-config] PUBKY_RUNTIME_LOCK_SERVER is set but PUBKY_RUNTIME_PAYKIT_SERVER_URL is not; the lock feature is disabled.',
  );
}
```

Only worth firing on the asymmetric case (lock server set, Paykit missing) — a deploy with neither set has simply not opted into Locks and doesn't need telling. Happy for this to land as a separate PR if you'd rather keep this one scoped.

---

secondl1ght (approved):
Two non-blocking comments inline: one on the bridge treating a malformed failure payload as a success, one suggesting a startup warning for the partial-config case the PR description already flags.

PR #2422, opened 2026-08-26
fix(ui): prevent profile name descender clipping

aldertnl:
## Summary

- Prevent profile-name descenders from being clipped on desktop
- Preserve `lg:leading-none` and the compact 60px header height
- Keep long-name truncation and status-emoji positioning unchanged
- Add focused unit, snapshot, and cross-browser VRT coverage

## Why
Profile names use `truncate`, which applies `overflow: hidden` to keep long names from pushing the status emoji out of the header. Combined with the tight `lg:leading-none` line box, that overflow boundary clips descenders in letters such as g, j, p, q, and y.

Changing back to `lg:leading-normal` would prevent the clipping but expand the line box from 60px to 90px, reintroducing excess header height and spacing.

This fix adds 4px of internal room for descenders and offsets it with an equal negative margin. The glyphs remain visible without changing the heading’s layout footprint, truncation behavior, or surrounding gaps.

## Verification

- ProfilePageHeader unit tests: 30 passed
- Browser VRT passed in Chromium, Firefox, and WebKit
- TypeScript passed
- ESLint passed
- Prettier passed
- Staged diff check passed

PR #2424, opened 2026-08-26
fix(ui): make profile name descender fix work in Safari

infin1t3:
## Summary

- #2422 fixed profile-name descender clipping with `overflow-clip-margin`, but Safari does not support that property (through v27 — [caniuse](https://caniuse.com/mdn-css_properties_overflow-clip-margin)), so descenders (`g`, `y`, `p`) stayed clipped there — and its webkit VRT assertion (`overflowClipMargin === '9px'` → `undefined`) has kept the dev VRT gate red since the merge ([failing run](https://github.com/pubky/pubky-app/actions/runs/32996318538)) and blocks the baseline-regeneration workflow for every branch off dev.
- Swaps the mechanism for one every browser honors: `pb-[0.15em]` gives the descenders the same ~9px of room inside the clip box (overflow clips at the padding edge), and the matching `-mb-[0.15em]` cancels it out of layout so the 60px line and the status-emoji alignment don't move. Same em-scaling as before.
- The VRT test now asserts the padding pair (cross-browser computed styles) instead of `overflowClipMargin`; the layout guards (line height, truncation, emoji boundary) are unchanged.

## Test plan

- [x] `Profile.vrt.test.tsx` passes locally in chromium, firefox and webkit (57/57); darwin baselines unchanged (pixel diff below threshold in chromium/firefox; webkit now paints descenders)
- [x] `ProfilePageHeader.test.tsx` 30/30, snapshots updated
- [x] typecheck / lint / format clean
- [ ] dev `vrt.yml` run goes green after merge

image

### DIFF
--- a/src/components/organisms/ProfilePageHeader/ProfilePageHeader.test.tsx
+++ b/src/components/organisms/ProfilePageHeader/ProfilePageHeader.test.tsx
@@ -148,11 +148,10 @@ describe('ProfilePageHeader', () => {
     expect(screen.getByText('Satoshi Nakamoto')).toHaveClass(
       'leading-8',
       'lg:leading-none',
-      'overflow-x-clip',
-      'overflow-y-clip',
-      'text-ellipsis',
-      'whitespace-nowrap',
-      '[overflow-clip-margin:0.15em]',
+      'truncate',
+      // Descender room that Safari honors too (it lacks overflow-clip-margin support)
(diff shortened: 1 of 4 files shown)


catch-21 (approved):
LGTM. The existing test "renders name and bio correctly" clearly comments the classes for the descender so hopefully we don't see regression.

PR #2426, opened 2026-08-27
Fix MDXEditor popup focus handling in article dialog

secondl1ght:
## Summary
Fixes an issue where MDXEditor popups (like the link dialog) lose focus when opened inside the article composer dialog. The problem occurs because MDXEditor was portaling popups to `document.body`, placing them outside the Radix Dialog's focus trap, which would immediately pull focus back to the editor.

## Changes
- Added an `overlayContainer` state to `InitializedMDXEditor` that serves as an in-tree host for MDXEditor popups
- Created an absolutely-positioned div with `data-testid="mdx-editor-overlay-container"` to host popup contents
- Passed the `overlayContainer` ref to MDXEditor's `overlayContainer` prop so popups render inside the dialog's focus trap
- Updated test snapshots to reflect the new overlay container div
- Added test case verifying that the overlay container is properly passed to MDXEditor

## Implementation Details
- The overlay container div uses `absolute` positioning with zero size so it doesn't affect layout or flex gaps
- Popup contents position themselves as `fixed` relative to this container, keeping them visually correct while staying in-tree
- This ensures MDXEditor's link dialog URL input receives autofocus and keyboard input works as expected


infin1t3 (approved):
Looks good. This matches MDXEditor's `overlayContainer` API and the official within-dialog pattern: popups that used to portal to `document.body` now mount in-tree, so the article dialog's focus trap can keep focus on the link URL field.

The callback-ref + `useState` host is correct — MDXEditor's effect depends on `overlayContainer`, so the first-render `null` then swaps the popup host off `body`. The empty `absolute` div stays out of the flex gap.

Issue #2368, opened 2026-08-19
Locks: reader — pay with Bitkit to unlock content


Taewa:
Sub-issue of #2364. Absorbs the locked-post display work. #2297 is a sub-issue of this one.

## Problem

The reader unlock pipeline exists end to end, but it was built for a lock that resolves immediately:

- it builds a `dev-static` proof, which always passes
- it polls for at most 60 seconds (`1500ms × 40`)
- running out of polls throws a server error, even though the purchase is still `pending` and progressing normally

The reader leaves the app for Bitkit in the middle of the payment, so none of those assumptions hold.

## Reader flows

The reader never authenticates with the Lock Server. What they do need is a Paykit wallet — Bitkit — set up under **the same identity they use in pubky.app**. The wallet publishes a receiver marker on its own homeserver in the background as soon as it has an identity; that marker is how the platform knows a reader can be paid to, and without it a payment cannot even be requested.

### A. The reader has Bitkit

1. Reader opens a locked post and presses Unlock.
2. A modal shows the price and a **Pay with Bitkit** button.
3. Pressing it submits the proof bundle. The button gives way to a spinner reading *awaiting payment*.
4. A payment request arrives in the reader's Bitkit. They pay it there — the app is not involved.
5. The app sees the purchase complete, opens the content, and replicates it to the reader's own storage.

### B. The reader does not have Bitkit

1. Reader opens a locked post and presses Unlock.
2. The modal explains that Bitkit is needed and lists the steps: install it, set up the profile with the same identity, fund the wallet. Store links are shown.
3. Nothing is submitted and nothing is polled while they are away — getting a wallet and funding it takes ten minutes at best, and there is nothing to watch for yet.
4. The reader comes back and presses **I completed the steps**. That is the start of the payment, the same role Pay with Bitkit plays in flow A: it submits and the modal switches to the spinner. Nothing is detected on its own.
5. If the wallet still is not ready the submission cannot go through, so the modal stays where it is and says why.

Paying from a wallet other than Bitkit is out of scope here: it needs Paykit embedded in pubky.app so the app can read the payment request and render a standard QR. The control for it stays hidden until that lands.

## Scope

**Unlock flow**
- Build a payment proof instead of a dev-static one
- A waiting state that survives the reader leaving for Bitkit and coming back
- Stop treating a timeout as a failure
- Resume a purchase already in flight when the reader reopens the post

The proof bundle gains a top-level field and empties its payload:

```jsonc
{
[...]

PR #2436, opened 2026-08-27
Feat/2367 locks price paid publish

### LINKED ISSUE #2367: Locks: creator — set a price and publish a paid lock
Sub-issue of #2364. Follows the payout account work (#2366).

## Problem

Publishing works, but every lock is written with the `dev-static` placeholder verifier — the constants sit hardcoded in the publish path. There is no way to attach a price.

## Scope

Three steps, matching how the composer already works:

1. **Price input (sats)** on the payment tab of the lock dialog
2. **Hold it in composer state.** Applying the lock only *configures* it — the same place the current `password` value is handed up (and today dropped). Nothing is published at this point.
3. **Assemble the criterion at publish time**, replacing the hardcoded `dev-static` constants, when the Post button publishes the announcement and the locked content.

The upload and publish pipeline is unchanged, and publishing still happens exactly once from the Post button. Only the criterion written into `lock.json` changes:

```jsonc
// Phase 1
{ "verifier_type": "dev-static",      "params": { "satisfied": true } }

// Phase 2
{ "verifier_type": "paykit-payment",  "params": {
    "recipient_pubky": "pubky<locks creator>",  // must equal the lock's creator
    "amount": "1234",                            // sats, positive integer STRING
    "asset": "BTC"
} }
```

No new field: `verifier_type` and `params` already exist, and `params` is verifier-specific by design.

Constraints enforced by the Lock Server: exactly one payment criterion, referenced once by the lock logic, and `recipient_pubky` equal to the creator. `amount` is a string, not a number.

The price lands in the public `lock.json`, which cannot be edited after publishing — so the price is fixed once the post goes out.

~~**Blocker:** pinning the recipient needs the Locks session to expose its creator pubky. The SDK gains this in pubky/locks#10; the gap is already marked as a TODO in `src/core/stores/locksAuth/locksAuth.types.ts`. Related: #2283.~~

**Not a blocker after all.** The response to the guarded content upload already names the account that owns it, in the form the criterion needs. That account is the lock creator, so the recipient can be pinned there without waiting for the SDK. Related: #2283.

## Definition of Done

The two things a creator must be able to do:

- [ ] **Set a price.** Enter an amount in sats in the lock dialog and apply it to the post.
- [ ] **Post once, with the price attached.** Pressing Post publishes the announcement and the locked content exactly as today, now carrying the price — no extra step, no second publish.

Supporting behaviour:

- [ ] Zero, blank, and non-numeric prices are rejected before posting.
- [ ] The creator cannot change who gets paid — the payout goes to the authenticated Locks creator.
- [ ] A creator who has not connected a payout account (#2366) cannot reach this step.
- [ ] Abandoning the lock discards the price along with the rest of the lock config, as it does today.

Whether a reader can actually pay and open the post is #2368; that is the cross-check for the whole flow, not this ticket.

Taewa:
Closes #2367.

https://github.com/user-attachments/assets/bac3bbd5-4056-40ae-ad19-9a819b9c63a9

Stacked on #2413. Its commit shows up here until that one lands on `epic/locks-v1`.

A creator can now put a price on a lock and publish it with the same single Post.

## What a creator can do

- Enter a price in sats on the **Payment** tab of the Lock Content dialog, and see the dollar value as they type
- Apply the lock, which only configures it — nothing is published until Post, exactly as before
- Press Post once. The announcement, the locked content, and the price all go out together

## What a reader sees

- A paid lock shows a wallet icon and the price beside Unlock
- A password lock keeps the masked dots. So does a lock whose file has not loaded yet, so the slot is never empty

## Guardrails

- Zero, blank, and non-numeric prices are rejected. Apply stays disabled, and the price is checked once more before anything is published, so a rejected value cannot leave a half-created lock behind
- The payout always goes to the signed-in Locks creator. Nothing in the UI can change who gets paid
- The lock switch still requires a connected payout account (#2366), so this step cannot be reached without one
- Abandoning the lock discards the price along with the rest of the configuration

## Testing

Run against the local Paykit stack (#2365).

- Published a paid lock and read the resulting lock file back from the homeserver: it carries the price as a string, the payment verifier, and a recipient equal to the lock creator
- Checked the announcement card for a paid lock, a password lock, and a lock that fails to load
- Confirmed the composer preview shows the price the creator just applied

## Notes

- The issue listed an SDK gap as a blocker. It turned out not to be one: the upload response already names the account that owns the guarded content, which is the lock creator, so the recipient can be pinned from there.
- The dollar value comes from an exchange-rate service the browser could not reach directly. It is now requested through the app instead, which also repaired two dollar figures elsewhere in the app that had been silently blank.
- This PR covers the creator side only. Paying for a lock is #2368, so pressing Unlock on a paid post still opens the password dialog today.
- Password locks are still here. They go away with #2369, once payment is the only unlock method.


secondl1ght (approved):
Since this targets `epic/locks-v1` rather than `dev`, none of the below needs to block the merge — address whatever you like here and spin the rest out as follow-ups on the epic.

---

### 1. The BTC-rate request fires on every composer

`DialogLockContent.tsx:46` calls `useBtcRate()` in the component body, and `PostInput.tsx:559` mounts the dialog unconditionally:

```tsx
{lockServerPubky && (
  <DialogLocksAuth ... />
)}
<DialogLockContent open={isLockDialogOpen} onOpenChange={closeLockDialog} onApplied={handleLockApplied} />
```

`DialogLocksAuth` is guarded by `lockServerPubky`; `DialogLockContent` isn't. The component body — and so `useBtcRate`'s mount effect — runs whether or not `open` is true, so loading the home feed, a post detail, or any reply composer issues a `/api/btc-rate` request, including for users with Locks disabled who will never see the Payment tab. Worth deferring until the dialog is actually open, or the Payment tab is selected.

### 2. `/api/btc-rate` response is trusted unvalidated, and there's no unmount guard

`useSatUsdRate.ts:31`:

```ts
.then((json) => setResult({ rate: { ...json, lastUpdatedAt: new Date(json.lastUpdatedAt) }, status: 'ready' }))
```

Any 2xx body becomes `status: 'ready'`. If the route ever returns a shape without `satUsd` — a proxy/edge JSON error page, a future response change — then `rate.satUsd` is `undefined` and `DialogLockContent`'s `usdValue = amountSats * undefined` is `NaN`, which is `!== null`, so the field renders `$NaN` next to the price instead of the "dollar value can't be shown" message that `status: 'failed'` would have given. `lastUpdatedAt` likewise becomes `Invalid Date` with no signal. A shape check before `status: 'ready'` fixes both.

Separately there's no `cancelled` flag in that effect, so the resolve can `setResult` on an unmounted composer — `useLockFile` already does this correctly and is a good local pattern to copy.

### 3. Price-field reformatting fights the caret, and can rewrite what was typed

`DialogLockContent.tsx:208-209`:

```tsx
value={amount ? satsFormatter.format(amountSats) : ''}
onChange={(event) => setAmount(toSats(event.target.value))}
```

I checked this against the real component rather than eyeballing it, and it's narrower than "the field is broken" — the caret survives fine when an edit doesn't change the grouping. It's specifically edits that add or remove a separator:

- **Crossing a thousands boundary.** Field shows `100`, caret after the `1`, type `2`. Browser gives `1200`; React recomputes `1,200`, rewrites the DOM value, and the caret moves 2 → 5. The rest of the correction lands at the end.
- **Deleting the comma.** Field shows `1,000`, backspace over the separator. Browser gives `1000`; React recomputes `1,000`, so the comma reappears and the caret goes 1 → 5 — the deletion looks like it did nothing.
- **Above 2^53 the displayed value gets fed back into state.** Typing `12345678901234567` displays `12,345,678,901,234,568` — a digit that was never entered. `isValidAmount` correctly disables Apply, but the next keystroke sends that *displayed* rounded string through `onChange` → `toSats`, so `amount` becomes `...568` and the typed digits are gone. Backspacing back under 2^53 then leaves a price derived from the rounded value.

Formatting on blur, or keeping a raw buffer and only grouping for display when the field isn't focused, would sidestep all three.

### 4. The new API route has no test

`src/app/api/{copyright,dev/signup-token,feedback,og-metadata,report}` each have a sibling `route.test.ts`; `btc-rate` doesn't. Nothing currently covers the success shape (that `BtcRate.lastUpdatedAt` serializes to a string the client re-hydrates with `new Date(...)`), the `Cache-Control` header the comment says the CDN depends on, or the `handleApiError` path when `ExchangerateService` throws. A rename of the rate field would silently hand the browser `satUsd: undefined` and nothing would catch it — which is the same failure mode as point 2.

### Nits

- `DialogLockContent.tsx:62` hand-rolls the amount check as `amountSats > 0 && Number.isSafeInteger(amountSats)`, but this PR adds `isPositiveIntegerString` for exactly this rule and both `useCreateLockContent.ts:63` and `LockFileParser.resolvePriceSats` call it. The two agree today only because `toSats` pre-strips signs, decimals and leading zeros; if `toSats` changes, or a paste path bypasses it, the Apply gate desynchronizes from the publish-time guard. `isPositiveIntegerString(amount)` states the rule directly on the string that actually gets emitted.
- `const satsFormatter = new Intl.NumberFormat('en-US')` now exists identically in `LockedPostCard.tsx:32` and `DialogLockContent.tsx:28`, and `HumanBitcoinCard.tsx:30` formats the same quantity with `toLocaleString('en-US')`. The `₿` prefix is hand-written at three call sites too. A shared helper would keep the composer preview and the reader card from drifting.

---

PR #2437, opened 2026-08-28
feat: inline article images

secondl1ght:
# ✨ feat: inline images in article bodies

## Overview

Articles (`kind: long`) now support inline images across create, edit, and view — via toolbar button, paste, and drag-and-drop in **both** editor modes (rich text and raw markdown). Images upload to the author's homeserver at insert time and published bodies use a decentralized attachment-mapping scheme: **no static Nexus CDN URLs are ever persisted in article bodies**.

## 📐 The scheme (contract)

| Where | Image destination |
| ----- | ----------------- |
| Composer (draft/edit session) | `pubky://{author}/pub/pubky.app/files/{fileId}` — inserted only after upload succeeds |
| Published body | `alt` — strict grammar, `n` = slot in `post.attachments` |

- `post.attachments = [cover?, ...inline first-appearance, ...unreferenced-originals]`, deduped, max 10 total.
- **Slot-0 rule**: if the body references `attachment:0`, slot 0 is inline and the article has no cover; otherwise `attachments[0]` is the cover (rendered by card/detail chrome, never by markdown). Detection is AST-based and centralized.
- **Ownership invariant**: only author-owned file URIs become managed attachments. Non-author `pubky://` and external `https:` URLs stay in the body verbatim.
- Serialization is span-targeted over mdast positions — the author's markdown formatting is preserved byte-for-byte; images inside code fences are never touched; nesting (lists, quotes, headings, link-wrapped images) is fully supported.

## ✍️ Composer

- **Rich text**: MDXEditor `imagePlugin` with toolbar button; paste and drop handled natively (insert-after-upload invariant). Image resize is disabled deliberately — resized images serialize as raw HTML and would escape the AST rewrite.
- **Custom image dialog** (replaces MDXEditor's default): app Dialog atoms (responsive, themed), file preview panel with remove, URL + alt fields, real upload loading state (spinner, disabled controls, stays open on failure for retry).
- **Markdown mode**: image toolbar button + paste/drop on the textarea, with GitHub-style `Uploading name…` placeholders that swap in place on completion (deleting a placeholder cancels the insertion). Caret and focus preserved throughout.
- **Feedback everywhere**: viewport uploading pill for rich-mode paste/drop (the only dialog-less flows), placeholders in markdown mode, spinner in the dialog — one indicator at a time.
- Publish is blocked while uploads are in flight; mode toggles are disabled mid-upload (prevents editor-pane desync).
- Drop targeting: onto the editor → inline; onto the composer chrome → cover; non-image files get the standard unsupported-type toast.
- Serialize-time validation blocks publishing with actionable errors: hand-typed `attachment:` refs, `blob:`/homeserver-blob URIs, reference-style definitions of managed files, raw-HTML smuggling, over-cap.
- Cap enforcement at insert time too (counts body images + in-flight uploads; whole-batch rejection for a single over-budget drop — no wasted uploads, one toast).

## 💾 Persistence & session lifecycle

- New pure AST lib (`src/libs/post/articleInlineImages.ts`) owns serialize/deserialize/slot-0 detection; server-safe, heavily unit-tested.
- Create carries pre-uploaded URIs via `attachmentUris`; edit uses a new `nextOrder`/`addedUris` contract with author-ownership validation in the controller. `PostApplication` needed no changes.
- **Edits never delete files the user wasn't shown**: original attachments not referenced by the body (other clients, malformed refs) are preserved at the tail of the attachment list; an unrelated text edit commits content-only.
- Upload session cleanup is best-effort and safe by construction: unreferenced uploads swept on publish, all session uploads deleted on discard, deletions suppressed while a commit is in flight (leak-don't-delete on abort), late-resolving uploads after discard clean themselves up.
- Images removed during an edit ride the existing `removedUris` best-effort deletion.
- Composer dialogs refuse to close while a commit is in flight.

## 👀 Viewer

- New `ArticleInlineImage` molecule: validates `attachment:{n}` (index + author ownership), resolves via CDN `/main`; direct `pubky://` file URIs (any owner) resolve via CDN; external `https:` renders with `referrerPolicy="no-referrer"` + `loading="lazy"`; everything else renders an unavailable placeholder with **zero network requests**. Loading skeleton reserves space; load errors degrade to the placeholder and retry when the source changes.
- Fresh publishes render instantly from local object URLs (papering over the Nexus CDN variant-generation window), index-aligned with attachments and guarded against stale-body races.
- `PostText` allow-lists `img` only when the article detail page passes attachment context explicitly — feed previews and embedded article cards strip images entirely (alt text no longer consumes preview budget).
- Slot-0-aware cover handling everywhere: article cards, detail page, and list-view thumbnails/lightbox (inline images never leak into post-media surfaces).
- Extracted shared `pubkyUriToCdnUrl` into `src/libs/file/` (as `collectionCoverImage.ts`'s docblock requested) and refactored the collection cover resolver onto it.

## 📦 Dependencies

Pinned as direct deps (all already in the tree transitively, zero install weight): `mdast-util-from-markdown`, `mdast-util-gfm`, `micromark-extension-gfm`, `mdast-util-to-markdown`, `@mdxeditor/gurx`.

## 🧪 Testing

- ~250 new unit tests: AST lib round-trips (formatting preservation, nesting, code fences, cover offsets), upload session lifecycle (batching, commit latch, discard races), editor interactions (paste/drop/placeholders in both modes), dialog states, viewer resolution policy, slot-0 rule, controller contract validation.
- Full suite green: **742 test files / 12,684 tests**, typecheck and lint clean.
- Three manual QA rounds across create/edit/view/cleanup flows, plus an adversarial multi-agent code review (15 confirmed findings triaged: 9 fixed pre-merge, 1 fix reverted after failing QA and documented as upstream, remainder assessed as deferred with recorded consequences).


src/hooks/usePost/usePost.tsx
```diff
@@ -95,15 +216,39 @@ export function usePost(): UsePostReturn {
...
       });
+
+      if (isArticle) {
+        // Build entries before finalizeSession clears the session map; the
+        // referenced object URLs' ownership moves to the store.
+        seedArticleLocalFiles(createdPostId, [
+          ...attachments.map(fileToLocalAttachment),
+          ...(await completeInlineSeedEntries(inlineUris)),
+        ]);
+        void inlineImageSession.finalizeSession(inlineUris);
```
infin1t3, line 270: Publishing within the editor's 500ms debounce window silently drops a just-inserted image and then deletes the uploaded file.

The submit gate reopens the moment `uploadingCount` hits 0 (`runUpload`'s `finally` runs before the caller inserts the markdown), but the inserted markdown only reaches `content` after the 500ms `useDebounceCallback` in `usePostInput`. Publishing inside that window serializes the stale body: the image is missing from `inlineUris`, so `finalizeSession` sweeps the freshly uploaded file as unreferenced and `commitDelete`s it. In markdown mode the literal `Uploading name…` placeholder can get published verbatim. This is easy to hit by pasting an image and pressing Cmd+Enter in flow, and the cost is a deleted homeserver file. The `edit()` path has the same hole.

src/core/controllers/post/post.ts
```diff
@@ -585,6 +603,62 @@ export class PostController {
...
+          context: { compositePostId },
+        });
+      }
+
+      // Removals are diffed against the seeded snapshot (`original`), never
+      // the live row — only files the user actually saw and removed get
+      // deleted. Session uploads dropped before saving are not our concern
+      // here; the composer session cleans those up itself.
+      const nextOrderSet = new Set(nextOrder);
+      const orderRemovedUris = original.filter((uri) => !nextOrderSet.has(uri));
```
infin1t3, line 641: Removing an image from one article hard-deletes a file that another post may still reference.

`orderRemovedUris` sends every original attachment absent from `nextOrder` to `FileApplication.commitDelete`, which deletes the file record unconditionally, with no cross-post reference check. This PR is what makes shared references constructible in the first place: `serializeArticleBody` collects any author-owned file URI in the body, not just session uploads, and the controller check is ownership-only. Pasting your own `pubky://…/files/{id}` into a second article therefore attaches the same record to both posts, and editing the first article to drop the image breaks the second one permanently. Reproduced manually.

Fix: at publish, reject (or re-upload as a copy) any attachment URI that is neither a this-session upload nor one of the post's original attachments. That restores the invariant that one file belongs to one post, which is what makes deletion safe.

src/components/molecules/MarkdownEditor/InitializedMDXEditor.tsx
```diff
@@ -232,6 +450,26 @@ export default function InitializedMDXEditor({
             codeMirrorExtensions: [oneDark],
           }),
           maxLengthPlugin(ARTICLE_MAX_CHARACTER_LENGTH),
```
infin1t3, line 465: The character cap is enforced on Lexical text content, but the published limit applies to serialized markdown. Images count as 0 characters in the editor and about 110 on the wire.

An article near the cap plus 3 or 4 images passes the editor check and then fails spec validation ("Post content exceeds maximum length for Long kind"), surfaced only as "Could not create post. Try again.". Two follow-on bugs: `updateMaxLengthWarning` only shows "reached" when `remaining === 0`, and remaining goes negative here; and switching to markdown mode imports the over-cap text, where `handleMarkdownTextChange` refuses every keystroke, including single-character deletions that stay over the cap, so the textarea reverts each edit.

Fix: enforce the cap on `getMarkdown().length`, use `remaining <= 0` for the warning, and let markdown mode accept edits that shrink the length.

src/hooks/usePost/usePost.tsx
```diff
@@ -187,22 +338,95 @@ export function usePost(): UsePostReturn {
...
+        // Removal is diffed against the seeded snapshot, never the live row.
+        const originalUris = originalAttachmentUris ?? existingAttachments.map((attachment) => attachment.uri);
+        const originalSet = new Set(originalUris);
+        const referencedOrder = [...(coverUri ? [coverUri] : []), ...serialized.inlineUris];
+        // Original attachments the user was never shown (not the cover, not
+        // referenced by the body at open) ride at the tail so an unrelated
+        // edit never deletes files the user did not see and remove. Slots of
+        // referenced images are unaffected by the tail.
+        const referencedOrderSet = new Set(referencedOrder);
+        const nextOrder = [
```
infin1t3, line 391: The preserved attachment tail is appended after the cap check, so an edit can exceed the 10-attachment limit and fail with only the generic toast.

`serializeArticleForPublish` caps cover plus inline only; `nextOrder` then gains the preserved tail (original attachments the body never referenced, which the UI deliberately never shows). Cover plus 9 inline plus 1 stray makes 11 entries, and `builder.editPost` throws "Too many attachments (max: 10)" (verified against pubky-app-specs 0.7.0), which surfaces as "Could not update post. Try again.". The user is blocked with no visible culprit.

Fix: check `nextOrder.length > ARTICLE_ATTACHMENT_MAX_FILES` before `commitEdit` and show a specific message. Longer term, subtract the preserved-tail count from the inline budget.

src/libs/post/articleInlineImages.ts
```diff
@@ -0,0 +1,386 @@
...
+ * `attachment:`-scheme destination — are removed with a warning instead of
+ * failing, so a damaged article always remains editable.
+ */
+export function deserializeArticleBody(params: {
+  body: string;
+  attachments: string[];
+  authorPubky: string;
+}): DeserializeArticleBodyResult {
+  const { body, attachments, authorPubky } = params;
+  const { images } = collectNodes(parseBody(body));
```
infin1t3, line 320: Reference-style image bodies (`![a][r]` plus `[r]: attachment:1`) render fine but brick the edit flow.

`deserializeArticleBody` rewrites only inline `image` nodes. `collectNodes` never visits `imageReference`, and the collected `definitions` go unused, so the definition survives into the composer verbatim. On save, the definition check fires `HAND_TYPED_ATTACHMENT_REF` and every save is blocked, even a typo fix. That contradicts the module's own contract ("removed with a warning instead of failing, so a damaged article always remains editable"), while the read path fully supports these bodies: react-markdown resolves them, `collectAttachmentRefIndexes` counts definitions, and `remarkStripImages` visits `imageReference`. The trigger is bodies written by other clients, which the edit contract explicitly supports elsewhere (`preservedAttachmentUris`).

Fix: in deserialize, rewrite resolvable `imageReference` plus definition pairs to inline syntax and remove the definition, falling back to the existing remove-with-warning path.

src/components/organisms/DialogNewPost/DialogNewPost.tsx
```diff
@@ -21,20 +21,29 @@ interface DialogNewPostProps {
...
+  const [isSubmitting, setIsSubmitting] = useState(false);
   const title = isArticle ? 'New Article' : 'New Post';
   const { showConfirmDialog, setShowConfirmDialog, resetKey, handleContentChange, handleOpenChange, handleDiscard } =
     useConfirmableDialog({
       onClose: () => onOpenChangeAction(false),
     });
 
+  // Closing mid-commit could strand a publish that references session
+  // uploads; the commit takes a moment and the submit button shows it
+  const handleOpenChangeGuarded = (nextOpen: boolean) => {
```
infin1t3: The mid-commit close guard combined with a timeout-less write path can trap the composer permanently. The same guard is duplicated in DialogEditPost.

Every close affordance (ESC, X, overlay) routes through the guarded `onOpenChange` and gets swallowed with no feedback while `isSubmitting` is true. The homeserver PUT has no timeout at any layer: the app passes no AbortSignal and the pubky WASM SDK explicitly doesn't support one. A hung request therefore leaves the dialog unclosable indefinitely. The guard also doesn't protect anything the session doesn't already protect: `discardSession` no-ops during commits via `committingRef`, and the success continuation works after unmount.

src/hooks/usePostAttachmentsMedia/usePostAttachmentsMedia.tsx
```diff
@@ -35,23 +37,37 @@ export function usePostAttachmentsMedia(postId: string): UsePostAttachmentsMedia
   useEffect(() => {
     let cancelled = false;
 
+    // Articles carry inline body images in `attachments`, but only the cover
+    // (slot-0 rule) is post-level media — inline images render inside the
+    // article body, never as thumbnails or gallery items.
+    const isArticle = postDetails?.kind === 'long';
+    const articleHasCover =
+      isArticle &&
+      Boolean(postDetails.attachments?.length) &&
```
infin1t3, line 46: (Anchored here because the affected file isn't touched by this PR.)

Article inline images fan out as standalone tiles in the Visual layout. The problem lives in `useVisualFeedTiles.ts:346`.

The tile builder there maps every `post.attachments` entry, and every seeded `localPostAttachments` entry, to a tile with no kind or slot-0 awareness. It never learned the rule this hook implements ("inline images render inside the article body, never as thumbnails or gallery items"). An article with a cover and 3 inline images renders 4 tiles, immediately for the author via the seeded store, and the tile overlay shows the raw JSON envelope. Before this PR articles had at most one attachment, so that path never saw the problem.

Fix: in `useVisualFeedTiles`, for `kind === 'long'` slice both local and remote attachments to `articleHasCover ? 1 : 0` (the same slot-0 rule used here), and use the parsed article title for the overlay.

src/libs/post/articleInlineImages.ts
```diff
@@ -0,0 +1,386 @@
...
+  for (const node of images) {
+    if (isAttachmentRefScheme(node.url)) {
+      pushError({ code: 'HAND_TYPED_ATTACHMENT_REF' });
+      continue;
+    }
+    if (isBlobUri(node.url)) {
+      pushError({ code: 'BLOB_URI' });
+      continue;
+    }
+    if (!isAuthorFileUri(node.url, authorPubky)) continue;
```
infin1t3, line 259: `data:` and `http:` image destinations publish without any warning and never render for readers.

Three layers disagree. The image dialog forwards `src.trim()` unvalidated; this serializer only blocks `attachment:`, `blob:`, and hidden author URIs, so `data:` and `http:` pass verbatim; and the viewer renders only https, pubky, and attachment sources, showing a permanent "Image unavailable" placeholder for everything else. The editor previews the image fine (`imagePreviewHandler` passes it through), so the author sees it working and readers never do. Note the inconsistency: `blob:`, equally dead for readers, hard-errors at publish.

Fix: add an `UNRENDERABLE_URI`-style serialize error that shares one scheme predicate with `ArticleInlineImage.utils`, so publish blocks what the viewer won't render. Optionally validate in the dialog too for earlier feedback.

src/components/molecules/MarkdownEditor/InitializedMDXEditor.tsx
```diff
@@ -232,6 +450,26 @@ export default function InitializedMDXEditor({
...
+          ...(inlineImages
+            ? [
+                imagePlugin({
```
infin1t3, line 468: Pasting an image copied from a browser in rich mode hotlinks it instead of uploading.

A browser's "Copy image" puts both the file and text/html on the clipboard. MDXEditor's built-in paste bails on mixed payloads (return false when the payload is mixed), so Lexical's HTML paste path runs and `ImageNode.importDOM` inserts the external `src`. The article then publishes a hotlink that dies whenever the host blocks or removes the image. The same paste in markdown mode correctly uploads, since its handler is explicitly file-first. The two modes diverge on the feature's core flow.

Fix: register a higher-priority paste handler that applies the same file-first rule in rich mode. As a safety net, warn at serialize time about external `` sources that came from paste.

src/hooks/useInlineImageUpload/useInlineImageUpload.ts
```diff
@@ -0,0 +1,277 @@
...
+  const uploadInlineImage = (file: File): Promise<string> => {
+    if (!enabled || !authorPubky) {
+      return rejectWithToast('Images can only be uploaded while composing an article.');
+    }
+    const pubky = authorPubky;
+    // A fresh upload means the composer session is active again (e.g. after
+    // an earlier discard in the same mounted composer)
+    discardedRef.current = false;
+
+    if (!ARTICLE_SUPPORTED_ATTACHMENT_MIME_TYPES.includes(file.type)) {
```
infin1t3, line 148: One invalid file in a multi-image rich-mode drop discards the insertion of every valid sibling.

The per-file rejections here (type and size) feed MDXEditor's all-or-nothing `Promise.all` insertion. The valid files upload and register in the session, the single rejection kills the `.then`, `INSERT_IMAGE_COMMAND` never fires, and the uploads sit orphaned until the session sweep deletes them. The user sees one unsupported-type toast and an unchanged article. Markdown mode handles the identical drop per file.

Fix: filter invalid files before they enter MDXEditor's flow. Toast for the rejects and let the valid ones proceed.

src/hooks/usePostAttachmentsMedia/usePostAttachmentsMedia.tsx
```diff
@@ -35,23 +37,37 @@ export function usePostAttachmentsMedia(postId: string): UsePostAttachmentsMedia
   useEffect(() => {
     let cancelled = false;
 
+    // Articles carry inline body images in `attachments`, but only the cover
+    // (slot-0 rule) is post-level media — inline images render inside the
+    // article body, never as thumbnails or gallery items.
+    const isArticle = postDetails?.kind === 'long';
```
infin1t3, line 43: `postDetails?.kind === 'long'` treats not-yet-loaded as not-an-article, which leaks inline images into post media during interim renders.

Right after publishing, the store already has the seeded `[cover?, …inline]` entries while `usePostDetails` is still `undefined`, so `isArticle` evaluates to false and the hook returns the full entry list as post-level media. The list layout's thumbnail then renders an inline image until details resolve. For slot-0 articles, which should show no media at all, this is visibly wrong.

Fix: don't resolve local media while `postDetails === undefined`. Wait for it instead of defaulting to the non-article branch.

src/components/organisms/PostArticleDetail/PostArticleDetail.tsx
```diff
@@ -55,12 +56,23 @@ export const PostArticleDetail = ({ postId, content, attachments, isBlurred }: P
 
   const localAttachments = useLocalFilesStore((s) => s.posts[postId]);
 
-  const localCoverImage = localAttachments?.[0]?.type.startsWith('image')
-    ? { src: localAttachments[0].urls.main, alt: localAttachments[0].name }
-    : null;
+  // Local entries are index-aligned with attachments; slot 0 is the cover
+  // only when the slot-0 rule says so (otherwise it's an inline image).
+  const localCoverImage =
```
infin1t3, line 61: After an edit commits, the fresh store and the stale body cross for the duration of the PUT and can show the wrong image or cover.

This cover reader, and its twin in `PostArticle.tsx:34`, pairs prop-derived `hasCover` with store-derived `localAttachments[0]` with no alignment guard, although the sibling `ArticleInlineImage` documents exactly why one is needed. The guard that does exist there (`ArticleInlineImage.tsx:34`) compares only counts, so a reorder or replacement that keeps the same count passes it and serves the wrong image under the other image's alt until the liveQuery emission lands. The window is transient and only the author sees it, but it shows wrong content.

Fix: store each seeded entry's source URI and gate per slot on `attachments[i] === entry.uri`. That closes both the missing guard and the same-count hole at all three sites.

src/hooks/usePost/usePost.tsx
```diff
@@ -187,22 +338,95 @@ export function usePost(): UsePostReturn {
...
+        // publish uploads nothing.
+        const serialized = serializeArticleForPublish(Boolean(newCoverFile ?? keptCover));
+        if (!serialized) return;
+
+        let coverUri = keptCover?.uri;
+        if (newCoverFile) {
+          coverUri = await FileController.commitCreate({ file: newCoverFile, pubky: currentUserId });
+          // Joining the session means a failed commit below leaves cleanup to
+          // the session's discard sweep.
+          inlineImageSession.registerSessionUpload(coverUri, newCoverFile);
```
infin1t3, line 383: Replacing a cover leaks one object URL per replacement.

`registerSessionUpload` mints URL #1, and the seed entry gets URL #2 from `fileToLocalAttachment`. `finalizeSession` skips revoking referenced URIs on the assumption that the store took ownership. That holds for inline images, which are seeded through `buildLocalAttachmentEntries` with the same URL, but not for the cover: the store only ever saw #2. URL #1 pins the full-size blob until page unload.

Fix: seed the cover from the session entry, for example by running `buildLocalAttachmentEntries` over the full `nextOrder` including slot 0, instead of minting a second URL.

src/hooks/usePost/usePost.tsx
```diff
@@ -52,6 +88,91 @@ export function usePost(): UsePostReturn {
...
+          },
+        };
+      });
+    } catch {
+      // Best-effort: an incomplete seed just defers rendering to the Dexie/CDN path
+      return sessionEntries;
+    }
+  };
+
+  const seedArticleLocalFiles = (postId: string, entries: (InlineImageLocalEntry | null)[]) => {
```
infin1t3, line 191: The incomplete-seed branch (`complete ? entries : []`) leaks every object URL that was already created.

When any entry is null (the metadata fetch fails, and it's guaranteed whenever a kept cover has `urls: null`), the whole seed is discarded. But the cover URL and the session inline URLs already exist, were never in the store (so its set-difference revoke can't touch them), and `finalizeSession` skips them as referenced. Nothing ever revokes them.

Fix: in the incomplete branch, revoke every blob: URL in the built entries before storing `[]`.

src/components/molecules/MarkdownEditor/InitializedMDXEditor.tsx
```diff
@@ -136,6 +178,144 @@ export default function InitializedMDXEditor({
...
+      // Restore focus (the toolbar button's native file chooser steals it)
+      // and park the caret after the inserted placeholders. Later placeholder
+      // swaps deliberately do NOT focus — the user may be typing elsewhere.
+      textareaElement.focus();
+      textareaElement.setSelectionRange(caretAfterSnippet, caretAfterSnippet);
+    });
+
+    // Sequential uploads keep each placeholder swap operating on the freshest
+    // textarea content
+    for (const [index, file] of files.entries()) {
```
infin1t3, line 280: Multi-file uploads run strictly sequentially, and the per-tick calls defeat the same-tick batch admission. The same pattern is in `usePostInput.ts:545`.

N files means N times the summed latency with submit blocked. The all-or-nothing budget check also degenerates into N batches of one: an over-budget drop uploads the leading files and then toasts once per rejected file, the opposite of the intended single-toast wholesale rejection. The rich-mode plugin path batches same-tick, as designed.

Fix: start all uploads in one tick (`files.map(f => upload(f))`), then swap placeholders or insert in order as each settles.

src/libs/post/articleInlineImages.ts
```diff
@@ -0,0 +1,386 @@
...
+ * Replaces an image node's destination inside `body`. When the destination
+ * appears exactly once in the node's source span, only that substring is
+ * replaced (preserving the author's alt/title formatting verbatim); otherwise
+ * the whole span is rebuilt from the node.
+ */
+function replaceImageDestination(body: string, node: Image, span: NodeSpan, newUrl: string): string {
+  const source = body.slice(span.start, span.end);
+
+  let replacement: string;
+  if (countOccurrences(source, node.url) === 1) {
```
infin1t3, line 187: Two entity-escape edge cases in this module. Both reproduce against the real parser, and both need a foreign client or a deliberate author to trigger.

Here: the single-occurrence heuristic searches the raw span for the decoded URL. A destination written as `…img&#49;` with the decoded URI pasted as alt text rewrites the alt and persists the raw file URI, breaking the no-raw-URI contract with no error raised. Guard the fast path so it only matches inside the destination (past `](`), and otherwise fall back to the `imageMarkdown` rebuild.

Line 384: the `includes('attachment:0')` pre-check in `articleHasInlineSlotZero` misses `attachment:&#48;`, so cover detection disagrees with `collectAttachmentRefIndexes` and with the viewer, and slot 0 renders as both cover and inline. Implement it as `collectAttachmentRefIndexes(body).has(0)`.

src/libs/post/articleInlineImages.ts
```diff
@@ -0,0 +1,386 @@
...
+ */
+export function isAttachmentRefScheme(src: string | null | undefined): boolean {
+  return Boolean(src && ATTACHMENT_SCHEME_PATTERN.test(src.trim()));
+}
+
+/**
+ * Returns the owner pubky of a strict homeserver file URI
+ * (`pubky://{owner}/pub/pubky.app/files/{fileId}`), or null for anything else.
+ */
+export function parseFileUriOwner(uri: string | null | undefined): string | null {
```
infin1t3, line 97: This is a third homeserver-file-URI parser, and it disagrees with the existing two on edge-case URIs.

`parseFileUriOwner` (strict WHATWG prefix match), `isHomeserverFileUri` (loose segment match), and `buildCompositeIdFromPubkyUri` diverge on 7 of 14 tested shapes: `PUBKY://` casing, double slashes, non-canonical paths. The controller uses the strict one to validate attachments while the deletion filter uses the loose one. First-party URIs agree across all three, so this is a drift risk rather than a live bug.

Fix: unify on one canonical parse so validation and deletion cannot drift, for example by having `parseFileUriOwner` and `isHomeserverFileUri` share a single implementation. Placement needs care: `libs/file` already imports from `core/models` (`pubkyFileCdnUrl` uses `buildCompositeIdFromPubkyUri`), so the shared parser has to sit where both sides can reach it without a cycle.

src/hooks/usePost/usePost.tsx
```diff
@@ -17,6 +28,31 @@ import type {
...
+      return 'Uploaded images cannot be used inside raw HTML. Use image markdown instead.';
+    case 'REFERENCE_STYLE_FILE_URI':
+      return 'Uploaded images cannot use reference-style links. Use inline image syntax instead.';
+    case 'UNPROCESSABLE_IMAGE':
+      return 'An image in the article could not be processed. Remove it and insert it again.';
+  }
+}
+
+/** Maps a File to a local-store attachment entry backed by a fresh object URL. */
+function fileToLocalAttachment(file: File): InlineImageLocalEntry {
```
infin1t3, line 49: `fileToLocalAttachment` is a byte-identical copy of `newFileToLocalAttachment` (`usePostInput.ts:242`).

A change to the localFiles entry shape or URL variants will land in one copy and break same-session previews on the other composer path.

Fix: extract one shared helper and import it in both hooks.

src/components/molecules/MarkdownEditor/InitializedMDXEditor.tsx
```diff
@@ -136,6 +178,144 @@ export default function InitializedMDXEditor({
...
+    void uploadAndInsertInMarkdownMode(files);
+  };
+
+  const handleMarkdownTextareaDrop = (event: React.DragEvent<HTMLTextAreaElement>) => {
+    if (!inlineImages || readOnly) return;
+
+    // Image files only — anything else bubbles to the composer container,
+    // whose handler shows the standard unsupported-type toast (uploading
+    // placeholders must never appear for files the upload will reject)
+    const files = Array.from(event.dataTransfer?.files ?? []).filter((file) => file.type.startsWith('image/'));
```
infin1t3: The drop filter accepts `image/*` while the upload allows 5 MIME types, so a HEIC drop flashes a placeholder that gets yanked right back out.

An iPhone `photo.heic` passes the filter, `Uploading photo.heic…` lands at the caret, the upload rejects the type, and the handler removes the placeholder again. The adjacent comment promises exactly this never happens ("uploading placeholders must never appear for files the upload will reject").

Fix: filter on the actual supported-type whitelist instead of `image/*`.

src/components/molecules/MarkdownEditor/MarkdownEditorImageDialog.tsx
```diff
@@ -0,0 +1,260 @@
...
+function ImageDialogForm({
+  isEditing,
+  initialSrc,
+  initialAltText,
+  initialTitle,
+  upload,
+  onSave,
+  onClose,
+}: ImageDialogFormProps) {
+  const [file, setFile] = useState<File | null>(null);
```
infin1t3, line 87: This new form is built on `useState` and a raw `<form>`, while AGENTS.md's Forms standard calls for react-hook-form plus zod for new forms.

The rule even anticipates this exact shape: "Non-text inputs (file pickers, rich text, etc.) live in their own dedicated hooks … and the form hook composes them." Every other submitting form in the repo follows it. Mitigating context: this form feeds MDXEditor's realm publisher rather than an app mutation hook, so the controller-wrapping half of the rule doesn't apply cleanly.

Fix: restructure as a react-hook-form plus zod hook with Controller-wrapped fields, or record an explicit exemption for third-party-editor dialogs.

src/core/controllers/file/file.ts
```diff
@@ -33,6 +34,19 @@ export class FileController {
...
+  /**
+   * Deletes files (blob first, then record) from the homeserver and the local
+   * cache. Shared blobs and already-deleted resources are handled by the
+   * application layer, so callers may treat this as idempotent best-effort
+   * cleanup (e.g. discarding uploads of an unpublished article draft).
+   *
+   * @param params - Parameters for file deletion
+   * @param params.fileUris - Homeserver file URIs (`pubky://…/files/{id}`) to delete
+   */
+  static async commitDelete({ fileUris }: TDeleteFilesParams): Promise<void> {
```
infin1t3, line 46: The new public `FileController.commitDelete` ships without a test.

`file.test.ts` exists and tests even one-line delegations ("delegates to FileApplication.getAvatarUrl", and so on), and the only test references to this method mock it entirely, so the body never executes under test.

Fix: add a `describe('commitDelete')` block mirroring the existing delegation tests: assert `FileApplication.commitDelete` is called with `fileUris` and that errors propagate.

src/libs/utils/utils.ts
```diff
@@ -639,8 +640,9 @@ export function canSubmitPost(
   isSubmitting: boolean,
   isArticle?: boolean,
   articleTitle?: string,
+  hasBlockingUploads?: boolean,
 ): boolean {
-  if (isSubmitting) return false;
+  if (isSubmitting || hasBlockingUploads) return false;
```
infin1t3, line 662: `canSubmitPost` gained `hasBlockingUploads` without a test exercising the new branch.

The existing `canSubmitPost` describe block wasn't extended, and no test anywhere passes the new argument; all fixtures use `uploadingCount: 0`.

Fix: extend the describe block with a case where `hasBlockingUploads: true` blocks an otherwise-valid post and one where `false`/`undefined` preserves prior behavior, ideally plus an end-to-end case asserting submit is disabled while `uploadingCount > 0`.

infin1t3 (changes requested):
One additional comment that I wasn't sure on which line to put:

Type placement is inconsistent across the PR. `ArticleInlineImage` and `MarkdownEditor` get sibling `.types.ts` files, `ImageDialogFormProps` is declared inline in `MarkdownEditorImageDialog.tsx`, `InitializedMDXEditor` types its props anonymously in the function signature, and `articleInlineImages.ts` exports its types directly from the module. Any of these is fine on its own, but the same PR uses all four.

Related: `InlineImageSessionHandle` in `useInlineImageUpload.types.ts` is exported but nothing imports it.

src/hooks/usePost/usePost.tsx
secondl1ght: Known — we hit this in our own review pass. The window is the composer's 500ms debounce and requires publishing mid-flow, but the deleted-file cost makes it worth closing properly: the fix we want is serializing from the editor's current value (flushing the debounce) at submit instead of racing it. Tracked as a follow-up.

src/core/controllers/post/post.ts
secondl1ght: Fixed: publish now rejects any inline file URI that is neither a this-session upload nor (on edit) one of the post's original attachments, with an explanatory toast — on both create and edit, checked before the cover upload. That restores the one-file-one-post invariant that makes edit-time deletion safe. Covered by new tests on both paths.

src/components/molecules/MarkdownEditor/InitializedMDXEditor.tsx
secondl1ght: Fixed the follow-on bugs: the warning is now computed from the serialized markdown in both modes with `remaining <= 0`, switching to markdown mode surfaces the cap state immediately, and markdown mode always accepts edits that shrink an over-cap body. We didn't add hard keystroke blocking on serialized length in rich mode — the accurate warning plus the markdown recovery path covers the trap, and rejecting Lexical input on serialized length is more editor surgery than we want here.

src/hooks/usePost/usePost.tsx
secondl1ght: Fixed: `nextOrder` is now checked against the 10-attachment cap before `commitEdit`, with a message that calls out attachments kept from previous versions as part of the count. Subtracting the preserved tail from the inline budget up front is noted as the longer-term shape. Test added.

src/libs/post/articleInlineImages.ts
secondl1ght: Deferred: this only triggers on bodies written by other clients (our composer can't produce reference-style refs), and the read path renders them fine — so no first-party article can get bricked. Agreed the right fix is rewriting resolvable `imageReference` + definition pairs to inline syntax in deserialize, falling back to remove-with-warning; recorded as a follow-up with the other reference-style handling notes.

src/components/organisms/DialogNewPost/DialogNewPost.tsx
secondl1ght: Removed the guard entirely (both dialogs, plus the `onSubmittingChange` plumbing). As you note, it protected nothing the session doesn't already protect — `discardSession` no-ops during commits and the success continuation survives unmount — and the outcome still announces itself via the publish/failure toast when the commit settles. So all it added was the hang risk.

src/hooks/usePostAttachmentsMedia/usePostAttachmentsMedia.tsx
secondl1ght: Fixed: `useVisualFeedTiles` now slices article attachments to `hasCover ? 1 : 0` (same slot-0 rule) at all three sites — metadata prefetch, local-store tiles, and remote tiles — with a test through the live-query pipeline. The raw-envelope overlay on the remaining cover tile predates this PR (single-attachment articles already hit it); parsing the title for the overlay is a follow-up.

src/libs/post/articleInlineImages.ts
secondl1ght: Deferred: agreed on the inconsistency, but the blast radius is small — it's an author-only foot-gun and the viewer degrades to a placeholder with zero network. A shared scheme predicate between serialize and `ArticleInlineImage.utils` backing an `UNRENDERABLE_URI` serialize error is recorded as a follow-up.

src/components/molecules/MarkdownEditor/InitializedMDXEditor.tsx
secondl1ght: Deferred: fixing this needs a higher-priority Lexical paste handler, which is exactly the kind of editor-internals override we decided not to land at the tail of this PR. Tracked as a follow-up; markdown mode keeps the correct file-first behavior in the meantime.

src/hooks/useInlineImageUpload/useInlineImageUpload.ts
secondl1ght: Deferred: the failure mode is safe — the rejects toast, the orphaned uploads are swept by the session cleanup, and nothing publishes half-inserted — so it's a UX gap on an uncommon mixed drop. Pre-filtering invalid files before they enter MDXEditor's `Promise.all` flow is the recorded follow-up.

src/hooks/usePostAttachmentsMedia/usePostAttachmentsMedia.tsx
secondl1ght: Fixed: the hook now resolves nothing while `postDetails` is undefined instead of defaulting to the non-article branch, with a test covering the seeded-store interim render right after publish.

src/components/organisms/PostArticleDetail/PostArticleDetail.tsx
secondl1ght: Deferred: the window is transient, author-only, and self-heals when the liveQuery emission lands. Per-slot URI gating means storing each seeded entry's source URI in the localFiles store — a store-shape change we didn't want to make at the tail of this PR. Recorded as a follow-up covering all three sites (including the same-count hole in the existing guard).

src/hooks/usePost/usePost.tsx
secondl1ght: Known/deferred: object-URL lifetime is recorded in the follow-ups — the store not tracking source URIs is the same root cause as the alignment-guard finding, and seeding slot 0 from the session entry via `buildLocalAttachmentEntries` over the full `nextOrder` is the shape we'd take. Bounded to one full-size blob per cover replacement, page-lifetime only.

src/hooks/usePost/usePost.tsx
secondl1ght: Same family as the cover-URL leak above — recorded in the object-URL follow-up. It only triggers when the metadata completion fails (or a kept cover has `urls: null`), and is page-lifetime bounded; revoking the built entries before storing `[]` will ride that follow-up.

src/components/molecules/MarkdownEditor/InitializedMDXEditor.tsx
secondl1ght: Deferred: the consequence is comfort-tier only — serial upload latency and one toast per rejected file instead of a single wholesale rejection. The budget is still enforced correctly and nothing orphans (rejected files never insert; the session sweep owns cleanup). Starting all uploads in one tick and swapping placeholders as each settles is the recorded follow-up.

src/libs/post/articleInlineImages.ts
secondl1ght: Both fixed: the fast path now only rewrites a match inside the destination (past `](`), falling back to the `imageMarkdown` rebuild otherwise; and `articleHasInlineSlotZero` now delegates to `collectAttachmentRefIndexes(body).has(0)`, with the `includes` pre-check broadened so entity-escaped refs reach the parser. Regression tests added for both cases.

src/libs/post/articleInlineImages.ts
secondl1ght: Deferred: agreed it's drift risk rather than a live bug — first-party URIs agree across all three parsers. Unifying on one canonical parse is recorded as a follow-up; as you note, placement needs care to avoid a layering cycle, which is exactly why we didn't rush it into this PR.

src/hooks/usePost/usePost.tsx
secondl1ght: Acknowledged — recorded as a small refactor follow-up (extract one shared helper and import it from both hooks). Left as-is here to keep this round free of cross-hook churn.

src/components/molecules/MarkdownEditor/InitializedMDXEditor.tsx
secondl1ght: Fixed: all three drop/paste filters now check the actual supported-type whitelist (`ARTICLE_SUPPORTED_ATTACHMENT_MIME_TYPES`) instead of `image/*`, so a HEIC drop falls through to the standard unsupported-type toast with no placeholder flash. Test added.

src/components/molecules/MarkdownEditor/MarkdownEditorImageDialog.tsx
secondl1ght: Deferred with an exemption: as the mitigating context says, this form feeds MDXEditor's realm publisher rather than an app mutation hook, so the react-hook-form + zod + Controller shape doesn't map cleanly onto it. We'll record the third-party-editor-dialog exemption explicitly rather than force the pattern.

src/core/controllers/file/file.ts
secondl1ght: Fixed: added a `commitDelete` describe block mirroring the existing delegation tests — asserts `FileApplication.commitDelete` is called with `fileUris` and that errors propagate.

src/libs/utils/utils.ts
secondl1ght: Fixed: extended the suite with the `hasBlockingUploads` branch — `true` blocks an otherwise-valid post, `false`/`undefined` preserve prior behavior.

src/components/molecules/ArticleInlineImage/ArticleInlineImage.tsx
```diff
@@ -0,0 +1,94 @@
...
+  // Local entries are index-aligned with attachments; the store only holds
+  // image object URLs for posts created/edited this session. The length guard
+  // mirrors useEditAttachments: right after an edit the store already has the
+  // NEW order while this render still shows the stale body — indexing across
+  // that mismatch would serve the wrong image.
+  const alignedLocalAttachments = localAttachments?.length === attachments.length ? localAttachments : undefined;
+  const localUrl =
+    resolved.kind === 'attachment' && alignedLocalAttachments?.[resolved.index]?.type.startsWith('image')
+      ? alignedLocalAttachments[resolved.index].urls.main
+      : undefined;
```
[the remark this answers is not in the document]
secondl1ght: Known — same finding as the alignment-guard thread in the human review. Practical impact is minimal: it needs a same-count replace/reorder edit, only the author can ever see it (no other reader has local preview entries), and it self-heals when the liveQuery emission lands — typically tens of milliseconds after the Dexie write, so the wrong image is on screen for well under a second. Deferred on that basis; per-slot URI gating (storing each seeded entry's source URI) is recorded as a follow-up, since the same store change also closes the missing cover-reader guards.

infin1t3 (approved):
LGTM.

PR #2438, opened 2026-08-28
refactor(ui): standardize toast module behind single toast() facade

### LINKED ISSUE #2370: Standardize the toast module behind a single public `toast()` interface
### Summary

The toast module currently exposes two producer patterns:

```tsx
const { toast } = useToast();
```

and:

```tsx
import { toast } from '@/molecules/Toaster/use-toast';
```

These are not two distinct capabilities. `useToast()` returns the same module-level `toast()` function while also subscribing the caller to global toast state.

Standardize the module around one public producer interface:

```tsx
import { toast } from '@/molecules/Toaster/toast';

const toastHandle = toast({
  variant: 'error',
  title: 'Could not publish post',
  description: 'Please try again.',
});

toastHandle.dismiss();
```

Toast state should remain an internal rendering concern, not a second application-facing API.

### Problem

The current module has grown beyond the interface the application needs:

- Trigger-only `useToast()` consumers subscribe to every toast state change and can re-render without using that state.
- The direct producer is imported from a file named `use-toast`, which obscures the intended API.
- The public toast input inherits Radix toast props, exposing implementation-specific lifecycle and styling fields such as `open`, `onOpenChange`, and `className`.
- A caller must construct a Radix `ToastAction` component instead of describing an application action.
- Internal store details such as `actionTypes` and `reducer` are exported for tests.
- Public capabilities including toast updates and some renderer exports have no production consumers.
- Some tests mock internal hook state, including states that cannot be reached through the real public API while `TOAST_LIMIT` is `1`.
- Documentation recommends direct `toast()` but also describes it as “hook-bound” and contains stale examples.

Manual dismissal is not a reason to keep the producer hook. The direct `toast()` call can return a handle for dismissing that specific toast, while `dismissButton` continues to provide a close control in the UI.

### Root cause

The module started from the older Shadcn Radix toast pattern, which exposed both an imperative function and a state hook. Both patterns were adopted without an application-level boundary between producers and the renderer.

PR #1892 made direct `toast()` the preferred producer API and migrated part of the codebase, but the hook, implementation-specific types, renderer components, tests, and conflicting documentation remained public.

The application currently renders toasts through Radix primitives. This refactor should preserve that behavior behind the new facade so the renderer can be replaced later without another call-site migration.

### Target interface

Expose an app-owned API from `@/molecules/Toaster/toast`:

- `toast(options)` for triggering a notification.
- `ToastOptions` containing only supported application concepts:
  - `variant`: `default | error | warning | info`
  - `title` and/or `description`
  - `dismissButton`
  - an optional action descriptor containing its label, accessible alternative text, and callback
- `ToastHandle.dismiss()` for dismissing the created toast programmatically.

The input type should require at least one of `title` or `description`. Callers should not be able to set raw Radix lifecycle, open-state, duration, or styling props. Action rendering and styling should be owned by the Toaster renderer.

Use an internal hook named `useToastState()` only inside the Toaster rendering module. It should not be part of the producer API.

### Proposed changes

1. Add `@/molecules/Toaster/toast` as the single public producer facade with explicit, app-specific `ToastOptions` and `ToastHandle` types.

2. Audit current toast producers and migrate them to the facade. Remove application-level `useToast()` usage.

3. Rename and internalize the state subscription as `useToastState()`, used only by the Toaster renderer.

4. Keep the state store and dismissal timers behind the facade. Use `useSyncExternalStore` or an equivalent stable subscription mechanism, and move timer scheduling out of the reducer so the reducer remains pure.

5. Replace component-valued custom actions with the public action descriptor. Preserve the existing Undo behavior while keeping Radix components private to the renderer.

6. Remove unused public surface after confirming there are no remaining consumers, including:
   - per-toast `update()`
   - `actionVariant`
   - `ToastClose`
   - test-only `actionTypes` and `reducer` exports

7. Update Toaster module tests to render the real `Toaster`, trigger notifications through the public `toast()` facade, and assert only reachable behavior. Update call-site tests to mock the public facade rather than hook-shaped state.

8. Align `docs/components.md`, `AGENTS.md`, and `.cursor/rules/toast.mdc` with the single public interface. Remove the inaccurate “hook-bound” wording and stale i18n examples.

9. Add an automated import restriction so application code cannot import the internal toast store, state hook, or Radix renderer components.

10. Remove the unused `sonner` dependency if the repository audit confirms it has no consumer. This issue does not require changing the current renderer to Sonner or Base UI.

### Acceptance criteria

- [ ] Application code triggers notifications only through `toast()` from `@/molecules/Toaster/toast`.
- [ ] `useToastState()` is internal to the Toaster renderer, and no application consumer subscribes to toast state.
- [ ] The public input type requires toast content and does not expose raw Radix lifecycle, duration, open-state, or styling props.
- [ ] Custom actions are passed as application data; callers do not import or construct Radix toast components.
- [ ] `dismissButton`, `ToastHandle.dismiss()`, automatic dismissal, the existing Undo flow, and all four variants retain their current behavior.
- [ ] Unused producer capabilities, renderer exports, and test-only store exports are removed.
- [ ] Toaster tests exercise the public facade with the real renderer and do not manufacture states that violate the configured toast limit.
- [ ] Call-site tests mock the public `toast()` facade.
- [ ] Toast documentation consistently describes the single public API and contains current examples.
- [ ] An automated restriction prevents imports of internal toast implementation modules from application code.
- [ ] The unused Sonner dependency is removed if no consumer is found.

infin1t3:
Standardizes the toast module behind one public producer API. Closes #2370.

## What changed

**Public API** — `toast()` from `@/molecules/Toaster/toast` is the only way to trigger a notification:

```tsx
import { toast } from '@/molecules/Toaster/toast';

const handle = toast({ variant: 'error', title: 'Upload failed', description: 'Try again.' });
handle.dismiss(); // optional programmatic dismissal
```

`ToastOptions` carries app concepts only: `variant`, `title`/`description` (at least one required), `dismissButton`, and an `action` descriptor. No Radix lifecycle, duration, open-state, or styling props. Custom actions are plain `{ label, altText, onClick }` descriptors — the Toaster renders them and dismisses the toast before invoking `onClick`, so the repost → Undo flow behaves exactly as before.

**Store owns the full toast lifecycle** (`toast.store.ts`):

- Pure reducer; all timer scheduling lives in `dispatch`. Auto-dismiss is scheduled on `ADD_TOAST`, cancelled on manual dismiss, and every timer is cleared for toasts evicted by the limit — no timer can fire for an absent toast.
- No-op dispatches (dismissing an absent or already-dismissed toast) return the same state reference and skip listener notification, so the Toaster never re-renders for nothing.
- Radix's own 5s default close timer is disarmed with `duration={Infinity}` — the store is the single source of dismissal timing. (Radix's timer is the one that gets stuck via `isClosePausedRef`, radix-ui/primitives#2233; the store timer is the fix.)
- State subscription is internal: `useToastState()` (`useSyncExternalStore`) is consumed only by the Toaster renderer. Trigger-only components no longer subscribe to toast state.

**Removed** (zero consumers): `useToast()`, per-toast `update()`, `actionVariant`, `ToastClose`, dismiss-all branches, test-only `actionTypes`/`reducer` exports, dead `ToastVariant` re-export, unused store exports, and the unused `sonner` dependency.

**ESLint boundary**: `no-restricted-imports` blocks application imports of `**/Toaster/toast.store`, `**/Toaster/useToastState`, and `**/atoms/Toast/*` (directory glob — new internal files are covered automatically). The Toaster module and Toast atoms are exempt for their own internals.

**Action button styling**: `toastActionVariants` now has a semantic `brand`/`muted` emphasis axis instead of piggybacking on toast-variant names. Rendered class output is unchanged.

**Test mocks consolidated** (second commit, reviews in isolation): all 66 call-site test files use a bare `vi.mock('@/molecules/Toaster/toast')` resolving to a single typed manual mock in `Toaster/__mocks__/toast.ts` (returns a dismiss handle like the real facade), with assertions via `vi.mocked(toast)`. The typed mock surfaced and fixed a latent unsafe destructure in `usePost.test.tsx`.

**Docs**: `docs/components.md`, `AGENTS.md`, `.cursor/rules/toast.mdc`, and `docs/error-handling.md` describe the single public API with current examples.

## Tests

- Toaster module tests render the real Toaster and exercise the real store through the public facade, with fake timers; all timings derive from the exported `TOAST_DURATION` / `TOAST_REMOVE_DELAY` constants.
- Click-dismiss tests assert immediate dismissal with no timer advance, so removed click wiring fails the suite.
- New coverage: auto-dismiss timer cancellation on manual dismiss and limit eviction (via `vi.getTimerCount()`), and no listener notification on no-op dismisses.
- Full unit suite green (12,529 tests); typecheck and lint clean.

## Review guide

Two commits: the refactor itself, then the mechanical mock consolidation (67 files, mostly deletions).

## Notes

- No VRT baselines touched — the Toaster is not mounted in any VRT scene; the five `src/test/vrt/**` changes are mock plumbing only.
- `useEditAttachments.test.ts` fails intermittently on `dev` as well (pre-existing, from the attachment-edit work) — unrelated to this branch.

## Video

Uploading Screen Recording 2026-08-31 at 09.20.47.mov…


Orlandohub (approved):
https://github.com/user-attachments/assets/6089b2d0-e76f-4322-99ea-22f3e5acb7fd

This PR standardized all app notifications behind a single typed `toast()` API, internalized toast state and dismissal timing, preserved existing actions such as Undo, and removed redundant hooks, exports, mocks, and dependencies.

It's quite a beast and I would suggest stacked PRs for future if possible.

PR #2440, opened 2026-08-28
Add form submission support to StatusPickerContent

### LINKED ISSUE #1440: [Bug]: Can't set custom status on android
### Description

Using my real android device, it's not possible to set a custom status because the enter key on the soft keyboard doesn't submit. It works on iOS and desktop.

### Expected Behaviour

Tapping enter applies custom status

### Steps to Reproduce

1. Using a real android device (I am using Gboard keyboard)
2. Go to own profile page
3. Go to change status to a custom one - **must add emoji first and then the text**
6. Observe enter key symbol changes to a right arrow with vertical line
7. Tap enter on software keyboard
8. Observe nothing happens, status is not applied### Screenshots

### Video

https://github.com/user-attachments/assets/d7f5ada8-ddf4-42d3-a143-e3aa61e2508d

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

secondl1ght:
## Summary
Enhanced the StatusPickerContent component to support form submission as a reliable way to submit custom status on mobile devices, particularly addressing issues with Android soft keyboards (e.g., Gboard) that may not emit identifiable Enter keydown events.

## Key Changes
- Converted the status input container from a `<div>` to a `<form>` element with a submit handler
- Added `handleSubmit` function that prevents default form behavior and triggers status save
- Added `type="button"` attributes to emoji picker buttons to prevent accidental form submission
- Added `enterKeyHint="done"` attribute to the input field for better mobile keyboard UX
- Updated all snapshot tests to reflect the structural changes

## Implementation Details
- The form submission handler (`handleSubmit`) calls the existing `handleCustomStatusSave` function, ensuring consistent validation and behavior
- The emoji picker buttons are explicitly typed as `type="button"` to ensure they don't trigger form submission when clicked
- The `enterKeyHint="done"` attribute provides visual feedback to mobile users about the keyboard action
- Existing keyboard-based submission via Enter key is preserved through the `handleKeyDown` handler
- Both submission paths (form submit and Enter key) now work reliably across desktop and mobile platforms


infin1t3 (approved):
Looks good. Form submit is the right path for Android IMEs that never emit a real Enter keydown, `handleCustomStatusSave` already guards empty input, and `type="button"` on the in-form emoji buttons is necessary because `Button` does not default `type` (HTML would treat them as submit).

Desktop Enter is unchanged: `useEnterSubmit` still `preventDefault`s when valid, so this should not double-submit. Tests cover the actual Android path (`submit` without keydown).

PR #2441, opened 2026-08-28
feat(ui): redesign collection og card and move brand url bottom-right

secondl1ght:
## Summary

Redesigns the collection OG card to match the updated Figma "Collection Card 1200 x 630" frame, and moves the `pubky.app` brand URL on post and article OG images from the bottom left to the bottom right per the same design update.

## Key Changes

- **Full-bleed cover section**: Replaced the inset card design with a cover section that extends edge-to-edge below the header, fetched at `FileVariant.MAIN` so the 1200px-wide box renders sharp (the `feed` renditions are only 720px wide, and some are broken/empty on the CDN while their `main` is valid)
- **Removed tag functionality**: The updated design has no tag chips — deleted `fetchPostTags()`, `solidTagColor()`, and `TagChip` plus their tests
- **Simplified content layout**:
  - Title row displays library icon + collection name (left, single-line ellipsis) with count pill and owner avatar (right)
  - Description is a fixed two-line slot (always rendered) so the title row stays anchored at the same position whether the description has zero, one, or two lines
- **Updated styling**:
  - Cover gradient changed from `rgba(0,0,0,0.7/0.35)` to the card-toned `rgba(29,29,32,0.9/0.5)`
  - Count pill shows just the compact number on `OG_TOKENS.background` (no more `POSTS` label or conditional `cardBg` tone)
  - Larger type: library icon 64px, collection name 72px, description 60px
- **Brand URL bottom-right**: The `pubky.app` footer on post text and article OG images moved from bottom-left to bottom-right
- **Header geometry export**: `OgComponents.tsx` now exports `OG_HEADER_HEIGHT` (used inside `OgHeader`'s own styles) so the collection card's cover math cannot drift from the header; the cover shell also gained `flexShrink: 0` + `overflow: hidden` as a belt

## Implementation Details

The cover section uses absolute positioning for the background layers (image + gradient) with explicit pixel dimensions, as satori cannot stretch absolute children to flex-sized parents. The two-line description slot is always rendered (not conditional) to keep the layout geometry stable.

## Verification

- Unit suite, typecheck, and lint pass; `OgComponents` snapshot unchanged
- All card states rendered through the real satori pipeline: cover / cover-less, with / without description, fallback avatar, pubky-name fallback, title and description truncation
- End-to-end renders from a running dev server against staging Nexus/CDN data: two photo-cover collections, a GIF-cover collection (whose `feed` variant is a 0-byte response — `main` rescues it), an article, and a short post


infin1t3 (approved):
Found couple of minor things, nothing blocking. Approved.

PR #2449, opened 2026-09-01
fix(ui): match feed tab padding to Figma

infin1t3:
## Summary
- Feed tabs used `px-8` / `px-10` (32–40px). Figma Text Wrapper is `spacing/2-5` (10px / `px-2.5`) on every tab, web and mobile.
- Removed `lg:min-w-40` so desktop tabs can share the 792px row instead of flooring at 160px.

## Test plan
- [x] `FeedNavigation` unit + snapshot tests
- [x] Desktop `/home`: computed padding is 10px; tabs still fill the feed column
- [ ] Mobile: active tab hugs label, inactive stay icon-only
- [ ] Custom feed active: pencil still overlays without clipping the name
- [ ] Home VRT baselines may need a refresh if pixels shifted

Made with [Cursor](https://cursor.com)


aldertnl (approved):
👍

PR #2451, opened 2026-09-02
feat(users): store moderation follow state in settings

### LINKED ISSUE #2450: [Task]: Replace moderation-follow marker with settings state
## Task Description

Replace the custom moderation-follow migration marker introduced in #2327 with settings-owned state. The marker solution is abandoned completely: Pubky App must stop checking and writing `pub/pubky.app/migrations/moderation-follow/v1/<moderationId>.json`. Existing marker files are ignored and require no import, backfill, compatibility fallback, or app-side cleanup.

Use the settings document that is already synchronized before bootstrap to record which moderation bot has been processed for the account. This keeps the state app-owned so it can move with `settings.json` during the upcoming pubky-social-specs v1 migration.

The state is intentionally two-valued. Once the configured bot is recorded as processed, the app never touches that follow again, so a later manual unfollow already sticks without a separate opt-out value. An explicit `null` opt-out was considered and dropped (see design discussion): it would only add a *global* opt-out surviving a bot-ID change, at the cost of settings writes on every manual follow/unfollow of the bot and the partial-failure modes between the settings write and the follow write.

Accepted rollout behavior: users who previously opted out through the shipped marker may be followed once again after this change. No mitigation is required.

## Affected Layer(s)

- Controllers
- Application
- Pipes
- Stores
- Tests

## Affected Domain(s)

- Users
- Follows
- Moderation
- Settings
- Bootstrap
- Migration

## Acceptance Criteria

- [ ] Settings represent moderation-follow state with two distinct values: missing means never processed, and the configured moderation-bot Pubky means processed.
- [ ] When the state is missing, bootstrap checks the canonical follow resource, creates the follow only when absent, and then persists the configured moderation-bot Pubky in settings.
- [ ] When settings contain the configured moderation-bot Pubky, bootstrap performs no follow or settings write.
- [ ] If settings contain a different moderation-bot Pubky because configuration changed, the configured bot is handled and the stored Pubky is updated.
- [ ] Manually unfollowing the moderation bot is an ordinary unfollow. Because the account is already recorded as processed, bootstrap never re-creates the follow.
- [ ] Moderation-follow settings updates use the same serialized settings-write path as ordinary settings changes.
- [ ] If settings synchronization (the fetch or the push) fails during sign-in, no automatic follow is attempted for that session; sign-in continues and the follow is deferred to the next full sign-in.
- [ ] A settings write from a client whose local state has no moderation-bot field carries the remote value over instead of erasing it.
- [ ] Follow or settings failures do not break account creation or sign-in and can be retried safely.
- [ ] Pubky App no longer reads, writes, imports, or otherwise relies on the existing moderation-follow marker.
- [ ] Existing marker files are left untouched and ignored.
- [ ] The settings field is included in the future M3 app-owned settings migration to `priv/app.pubky/v1/settings.json`.

## Technical Details

Current settings synchronization happens before `BootstrapApplication.initialize`, so the resolved settings state should be read by the moderation-follow workflow instead of performing a separate marker probe. Because Application code cannot read stores, the detached follow orchestration lives in `AuthController`, which reads the processed state, runs `UserApplication.ensureModerationFollow`, and writes the result back through `SettingsApplication.commitUpdate`.

Remove the marker URL builder, marker writer, marker existence check, and marker-specific behavior from `UserApplication.commitFollow`. Keep the canonical follow existence check so already-following accounts do not generate duplicate follow state.

Place the field under `privacy` in `settings.json`. `SettingsNormalizer.from` spreads nested objects, so clients that predate the field round-trip it untouched, whereas unknown top-level keys are dropped. Ensure settings normalization, Zustand persistence, homeserver serialization, and settings conflict handling all retain it.

Two controllers now write the full `settings.json` document, so serialize homeserver settings writes per account in `SettingsApplication.commitUpdate`.

Do not write to the future social-specs path yet. Continue using the current settings path; the pubky-social-specs M3 migration will move the complete app-owned settings state to `priv/app.pubky/v1/`.

## Testing Strategy

- [ ] Missing state plus missing follow creates the follow and stores the configured Pubky.
- [ ] Missing state plus existing follow stores the configured Pubky without creating a duplicate.
- [ ] Matching Pubky skips work.
- [ ] A changed configured Pubky is handled and persisted.
- [ ] Failed settings synchronization suppresses the automatic follow for that session without rejecting sign-in.
- [ ] Failed automatic follow or settings persistence does not reject sign-in and retries safely.
- [ ] Processed state is not recorded if the signed-in account changed while the follow was in flight.
- [ ] Concurrent settings commits for the same account are serialized; different accounts are not blocked.
- [ ] Settings normalization round-trips missing and Pubky states.
- [ ] A stale-client settings write carries the remote moderation-bot field over; a failed pre-write read still lets the write proceed.
- [ ] Tests assert that no moderation-follow marker requests are made.

## References

- Original implementation: https://github.com/pubky/pubky-app/pull/2327
- Original tracker: https://github.com/pubky/pubky-app/issues/2255
- Design discussion: https://synonymworkspace.slack.com/archives/C09HQTLAARF/p1788264096901029
- Upcoming settings migration: https://github.com/pubky/pubky-app-specs/pull/142

Orlandohub:
## Summary

- replace the shipped moderation-follow marker with a `privacy.moderationBot` field in `settings.json` recording which bot has been processed for the account
- serialize homeserver settings writes per account in `SettingsApplication.commitUpdate`, since auth and the settings UI now both write the full document
- move detached moderation-follow orchestration from `BootstrapApplication` into `AuthController`, where the settings store can be read and written, with account-switch cancellation
- suppress the automatic follow for the session when settings synchronization fails, because the processed state is unknown
- carry the remote `privacy.moderationBot` into writes from clients whose local store predates the field, so a stale client cannot erase it

## Design

The state is intentionally two-valued: **missing** means never processed, the **configured bot Pubky** means processed. Once processed, the follow is never touched again, so a manual unfollow already sticks and needs no separate opt-out value. An explicit `null` opt-out was in an earlier revision and was dropped: it only added a global opt-out that survives a bot-ID change, at the cost of settings writes on every manual follow/unfollow of the bot and partial-failure modes between the settings write and the follow write.

The field lives under `privacy` on purpose. `SettingsNormalizer.from` spreads nested objects, so clients that predate this field round-trip it untouched, whereas unknown top-level keys are dropped.

## Behavior

- missing state: probe the canonical follow, create it if absent, then record the configured bot Pubky in settings
- stored Pubky equals the configured bot: no follow probe, no settings write
- stored Pubky differs from the configured bot: the new bot is processed and the stored Pubky is replaced; the previous bot is not unfollowed
- manual unfollow of the bot: ordinary unfollow, never re-created because the account is already recorded as processed
- failed settings synchronization suppresses automatic follow for that session while sign-in continues
- processed state is never recorded against a different session than the one that ran the follow
- a client whose local store has no `moderationBot` (logged in since before this deploy, or its sign-in settings GET failed) reads the remote document once before writing and carries the field over; `initializeSettings` reuses the document it already fetched and returns the enriched settings so the store learns the field immediately. This stops on its own once the store knows the field, and a failed read never blocks an ordinary settings change
- existing marker files are ignored and left untouched
- the current settings path and version remain unchanged; the field is documented for preservation by the future M3 migration

## Trade-offs

- A user who unfollowed bot A will be followed to bot B once if the configured moderation ID ever changes. Accepted: a new bot is a new default.
- Users who opted out through the 1.8.0 marker may be followed once more after this ships (accepted in #2450).
- Stale clients still running the 1.8.0 marker logic may re-follow an opted-out user once until the service worker updates.
- If the follow is created but the recording settings PUT fails or is aborted by a sign-out, and the user manually unfollows the bot before their next successful sign-in records the state, that sign-in follows the bot once more. Accepted: closing it would require a settings write on manual unfollow, which this design removes on purpose. Without the manual unfollow the next sign-in finds the existing follow and only records it.
- Any settings sync failure at sign-in (GET or PUT) defers the default follow to the next full sign-in, even on the 404 branch where the state is known. Accepted: a failed PUT means the connection is unhealthy, and running the follow plus its recording write over the same path is the most likely way to hit the window above.

## Testing

- affected suites: 15 files, 284 tests passed
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`

No VRT update is required because the only UI-layer change narrows the privacy-toggle key type and does not alter rendering.

Closes #2450


src/core/controllers/settings/settings.ts
```diff
@@ -19,24 +19,13 @@ import type { NotificationPreferences } from '@/stores/settings/settings.types';
...
-      .catch(() => {})
-      .then(async () => {
-        const pubky = useAuthStore.getState().selectCurrentUserPubky();
-        const settings = SettingsNormalizer.extractState(useSettingsStore.getState());
-        await SettingsApplication.commitUpdate(settings, pubky);
-      });
-    await this.pendingCommit;
+    const pubky = useAuthStore.getState().selectCurrentUserPubky();
+    const settings = SettingsNormalizer.extractState(useSettingsStore.getState());
+    await SettingsApplication.commitUpdate(settings, pubky);
```
infin1t3, line 28: This writes the whole local document, so it also decides whether `privacy.moderationBot` survives on the homeserver. A client whose store never got the field erases it on its next toggle. Two real cases: a device that stayed logged in since before this deploy (session restore never syncs settings), and a session where the settings GET failed at sign-in, leaving the store at post-logout defaults.

After that, the next fresh sign-in probes the follow, finds it missing because the user unfollowed, and follows the bot again. That is the case this design is supposed to prevent, and it repeats after every unfollow.

I'd merge with the remote document (or re-run `initializeSettings`) before writing from a store that hasn't synced this session. At minimum, don't write `moderationBot: undefined` over a remote value.

src/core/controllers/auth/auth.ts
```diff
@@ -41,19 +43,66 @@ export class AuthController {
...
+        moderationBot: useSettingsStore.getState().privacy.moderationBot,
+        signal: controller.signal,
+      });
+
+      // The settings store is account-local; never record processed state against a different session.
+      if (!moderationId || controller.signal.aborted || useAuthStore.getState().currentUserPubky !== pubky) return;
+
+      useSettingsStore.getState().setModerationBot(moderationId);
+      const settings = SettingsNormalizer.extractState(useSettingsStore.getState());
+      await SettingsApplication.commitUpdate(settings, pubky, controller.signal);
```
infin1t3, line 88: If this PUT fails (only warned below) or gets skipped because the signal was already aborted (`commitUpdate` returns quietly then), the follow exists on the homeserver but nothing records it as processed. A manual unfollow writes nothing anymore, so the user's unfollow is lost as soon as the local store resets: log out and back in, or sign in on another device, and the bot is followed again.

Before this PR the unfollow itself wrote the marker, so this is new. Two ways out: on the next sign-in, push settings when local has `moderationBot` and remote doesn't; or also record the bot in settings on manual unfollow. The removed test `does not re-follow when a manual unfollow records the marker after migration marker failure` covered this and has no replacement.

src/core/controllers/auth/auth.ts
```diff
@@ -414,15 +469,21 @@ export class AuthController {
...
     try {
-      return await SettingsApplication.initializeSettings(pubky, localSettings);
+      return {
+        remoteSettings: await SettingsApplication.initializeSettings(pubky, localSettings),
+        succeeded: true,
+      };
     } catch (error) {
       Logger.error('Failed to initialize settings during bootstrap', { error, pubky });
-      return null;
+      return { remoteSettings: null, succeeded: false };
```
infin1t3, line 485: Every `initializeSettings` failure ends up here as `succeeded: false`, but only a failed GET leaves the moderation state unknown. On the 404 branch the app already knows no follow was ever processed. On the local-newer branch it already fetched the remote document (with its `moderationBot`) and then throws it away. So a new account on a flaky connection (GET 404, then the defaults PUT fails) gets no default follow, and because session restore never re-runs this, that holds until a full re-sign-in.

I'd let `initializeSettings` catch the push itself and return the resolved state, so only the GET failure suppresses the follow. The ticket says "settings synchronization fails", which I read as the fetch, not the push.

src/core/controllers/auth/auth.ts
```diff
@@ -41,19 +43,66 @@ export class AuthController {
...
+
+  /** Start settings-aware moderation follow work without extending the authentication critical path. */
+  private static startModerationFollow(pubky: Pubky): void {
+    this.cancelModerationFollow();
+    const controller = new AbortController();
+    this.moderationFollowAbortController = controller;
+    void this.ensureModerationFollow(pubky, controller);
+  }
+
+  private static async ensureModerationFollow(pubky: Pubky, controller: AbortController): Promise<void> {
```
infin1t3, line 74: The ticket puts this in `AuthController` because Application code can't read stores, but the controller only needs to read `moderationBot` and write it back, and it already hands store state to Applications this way (`localSettings` into `syncSettings`). The follow-then-record ordering plus the AbortController lifecycle (eight `cancelModerationFollow()` call sites now) is the kind of cross-domain invariant ADR-0009 says belongs in Application. `BootstrapApplication` hosted this before and could take `{ pubky, moderationBot, settings, signal }`, run the follow and the settings commit, and return the processed id for the controller to store.

Not blocking, but easier to settle now than when a second caller (migration resync, say) needs it.

src/core/controllers/auth/auth.test.ts
```diff
@@ -65,6 +72,8 @@ const spyOnClearAllQueryClients = async () =>
   vi
     .spyOn(await import('@/libs/query-client/query-client.factory'), 'clearAllQueryClients')
     .mockImplementation(() => {});
+const spyOnCancelModerationFollow = () =>
+  vi.spyOn(asOpaque<{ cancelModerationFollow: () => void }>(AuthController), 'cancelModerationFollow');
```
infin1t3: The docs (docs/component-testing.md, src/test-utils/README.md, and the JSDoc) describe `asOpaque` as a cast for external SDK types with no constructor. Here it reaches a `private static` on our own class, and since this PR removes the old `pendingCommit` cast, this one and the `afterEach` one are the only such uses left. It type-checks whatever the method is called, so a rename only fails at runtime, in seven tests plus `afterEach`.

Simplest fix: keep `cancelModerationFollow` public static as it was on `BootstrapApplication` and spy on it directly. Or assert through the captured `AbortSignal`, like the test near line 631 does.

src/core/controllers/auth/auth.ts
```diff
@@ -203,6 +257,7 @@ export class AuthController {
       // Update hasProfile after bootstrap completes - triggers redirect via useAuthStatus
       authStore.setHasProfile(isSignedUp);
     } catch (error) {
+      this.cancelModerationFollow();
```
infin1t3: I think this one is dead. `startModerationFollow` is the last statement in `hydrateMeImAlive` and can't throw synchronously, and the only statement after it inside the `try` is `setHasProfile`, a plain store set. Even if a store subscriber threw there, the guard on line 84 already stops the settings write once `authStore.reset()` runs. No test depends on it. Safe to drop. The call on line 237 is covered by `signIn` and `wrapAuthFlow` too.

infin1t3 (changes requested):

src/core/controllers/settings/settings.ts
Orlandohub: Agreed, and fixed in b4319015. `SettingsApplication.commitUpdate` now reads the remote document once when the local state has no `privacy.moderationBot` and carries the remote value into the write, so a stale client cannot erase it. It stops on its own once the store has learned the field. `initializeSettings` reuses the document it already fetched on the local-newer branch and returns the enriched local settings, so both `AuthController` and `MigrationController` load the field into the store right away instead of probing again on the next toggle.

Both cases you listed are covered: the logged-in-since-before-deploy device carries it over on its first toggle, and the failed-GET-at-sign-in session carries it over as soon as the network is back (a read failure is logged and the write proceeds, so an ordinary toggle is never blocked by the probe). Tests: `stale client without moderation-bot state` block in `settings.test.ts` plus the new `initializeSettings` case.

src/core/controllers/auth/auth.ts
Orlandohub: Accurate, and I'm accepting this window rather than closing it. What it takes to hit it: the follow PUT succeeds, the recording PUT (same homeserver, seconds later, inside the serialized queue) fails or is aborted by a sign-out, and then the user manually unfollows the bot before their next successful sign-in records the state. Without the manual unfollow, the next sign-in sees no record, probes the follow, finds it, and records it, with no re-follow.

On the two options: (a) "push settings on next sign-in when local has the bot and remote doesn't" does not work, because `cleanupLocalState` resets the store on logout, so a fresh sign-in has nothing local to push. (b) "record the bot on manual unfollow" is exactly the settings-write-on-unfollow that #2450 removed on purpose, together with its ordering and partial-failure modes, so re-adding it for this case would undo that decision. The old marker test you mention covered the marker design where the unfollow itself was the durable write; there is no equivalent in a design where the unfollow is an ordinary unfollow.

I added this to the PR description under Trade-offs so it is on record. Happy to add a single bounded retry on the recording PUT if you feel strongly, but it would not cover the aborted case and I'd rather keep the detached path minimal.

src/core/controllers/auth/auth.ts
Orlandohub: You're right that the state is known on the 404 branch, and the local-newer branch is now fixed by b4319015 (the fetched remote `moderationBot` is carried into the pushed document and returned to the store instead of being thrown away). I'm keeping "any sync failure suppresses the follow" on purpose though.

The reason is the failure mode, not the state. If the defaults PUT just failed, the connection is not healthy, and the follow work that would run next is a follow PUT followed by a recording PUT over the same path. Proceeding is the most likely way to end up with the follow created and unrecorded, which is the window discussed in the `auth.ts:88` thread. Deferring costs one sign-in: the account is a brand-new one with no settings document, so the next full sign-in takes the 404 branch again and does the follow then. I've reworded the acceptance criterion in #2450 so "synchronization" explicitly means fetch and push.

infin1t3 (approved):
Looks ok.

PR #2452, opened 2026-09-02
feat(ui): add onboarding follow best matches step (#2388)

### LINKED ISSUE #2388: [Feature]: Starter Packs — onboarding "Follow your best matches" step + Follow All & landing feed

Orlandohub:
Closes #2388

https://github.com/user-attachments/assets/14dd49cf-744a-4e2c-b3bf-78a29cf26152

## Summary

Adds the final onboarding "Experience" screen at `/onboarding/follow`: tag-aware starter pack suggestions with per-card follow, **Follow all**, and a **Finish** action that now owns Experience completion and picks the landing feed reach. Includes the #2387 handoff and the E2E coverage absorbed from #2390.

Flow: `/onboarding/profile` → `/onboarding/tags` → **`/onboarding/follow`** → `/home`

### Handoff from #2387
- Tags **Continue** now only persists the ordered tags and `router.push`es to the follow step (browser Back works). The `TEMPORARY(#2388)` completion write is removed.
- The re-prompt guard is extracted into `useExperienceCompletedGuard` and shared by both Experience templates.

### Data + orchestration
- `useStarterPackSuggestions`: `starter_pack:all:all:<tags>` when interests were chosen (order preserved), `influencers:this_month:all` fallback with none. Hydrates 10 users with counts/relationships/tags, intersects profile tags with the chosen interests (max 2 chips, none when empty), preserves followed cards in place, exposes `followedCount` / `unfollowedUsers`.
- `useFollowAll`: sequential `UserController.commitFollow(PUT)` per unfollowed target, live `Following x/N` progress, skips already-followed, collects partial failures, one summary toast, never rejects. Ref-based re-entrancy guard.
- `useWhoToFollowFollowPreservation` gains `preserveFollowedUser` so Follow all can keep cards listed without the per-user toast path.

### UI
- `FollowButton` and `UserStats` extracted from `UserListItem` into shared molecules (no behaviour change; existing tests green).
- New `SuggestedUserCard` (+ skeleton), `FollowBestMatchesHeader`, `FollowBestMatchesForm`, `FollowBestMatches` template, page, stepper entry (`Experience`, step 4), route guard entry, and `public/images/best-matches.webp` exported from Figma node `45738:573079`.

### Finish action
1. `markExperienceCompleted(pubky)`
2. `followedCount >= 1` → `setReach(REACH.NETWORK)` (explicit set; suppresses `useDefaultHomeReach`). With zero follows the home store is untouched so the ≥3-follows soft default still applies later.
3. `router.replace('/home')`

### Constants
`STARTER_PACK_SUGGESTIONS_LIMIT = 10`, `SUGGESTED_USER_CARD_MAX_TAGS = 2`, `UserStreamTypes.THIS_MONTH_INFLUENCERS_ALL`.

## Staging verification (Nexus)

Verified directly against `nexus.staging.pubky.app`:
- `source=starter_pack&tags=bitcoin,music` returns ranked results; `tags=music,bitcoin` returns a different interleave order (tag-order sensitivity confirmed).
- Passing `viewer_id` excludes the viewer and the accounts they already follow from the results.
- `source=influencers&timeframe=this_month` returns results for the no-tags fallback.

In-app manual verification and the Cypress run were skipped per request.

## Test plan

- [x] `npx vitest run --project unit` on all touched suites: 45 files / 646 tests green
- [x] `npm run typecheck`, ESLint, Prettier, `git diff --check` clean
- [ ] Cypress `onboarding.cy.ts` on staging (new: tags → Follow all → My network; no tags → All; `completeOnboardingFromInstall` now clicks through the follow step)
- [ ] Manual: `/onboarding/tags` → pick tags → Continue → suggestions render with matching chips → Follow all → Finish → `/home` on My network
- [ ] Manual: no tags → most active suggestions → Finish → `/home` on All
- [ ] Manual: Back from follow returns to tags with the selection intact; revisiting either Experience route after Finish redirects home

## Out of scope (per issue)
Load more / pagination, VRT baselines, the `interleave` dedupe and follow-cache invalidation review items from #2391.


src/components/organisms/FollowBestMatchesForm/FollowBestMatchesForm.tsx
```diff
@@ -0,0 +1,148 @@
...
+          )}
+        </Container>
+      </Card>
+
+      <ProfileNavigation
+        className="onboarding-nav mt-auto flex-col sm:flex-row lg:pt-0"
+        backText={'Back'}
+        onHandleBackButton={handleBack}
+        backButtonDisabled={isFollowingAll}
+        continueButtonDisabled={isFollowingAll}
```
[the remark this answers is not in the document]
Orlandohub: Valid. `followedCount` is derived from the relationships live query and only updates once the local follow write lands, so Finish could read a stale count in the click→write window.

Fixed in de0767f9: `useWhoToFollowFollowPreservation` now exposes `isFollowPending` (from `useFollowUser().isLoading`), `useStarterPackSuggestions` forwards it, and `FollowBestMatchesForm` locks Back/Finish on `isFollowingAll || isFollowPending`, mirroring the existing Follow all lock. Tests added at all three layers.

src/hooks/useFollowAll/useFollowAll.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    try {
+      for (const followee of pending) {
+        try {
+          await UserController.commitFollow(HttpMethod.PUT, { follower: currentUserPubky, followee });
+          result.followed.push(followee);
+          onFollowed?.(followee);
+        } catch (err) {
+          result.failed.push(followee);
+          Logger.error('[useFollowAll] Failed to follow user', { followee, error: err });
```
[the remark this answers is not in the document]
Orlandohub: Accurate description of the behavior, but this is a known trade-off rather than something introduced here. `UserApplication.commitFollow` writes Dexie first and does not roll back if the homeserver PUT fails; that is the local-first contract every follow surface uses. The single-follow path (`useWhoToFollowFollowPreservation.handleFollowClick`) behaves identically: on failure it rolls back preservation and drops the card even though the local follow persisted. `useFollowAll` inherits those semantics rather than defining new ones, and handling it here would only diverge the bulk path from the per-card path. Leaving as is.

src/components/organisms/FollowBestMatchesForm/FollowBestMatchesForm.tsx
```diff
@@ -0,0 +1,153 @@
...
+    isFollowPending,
+    preserveFollowedUser,
+  } = useStarterPackSuggestions();
+  const { followAll, isRunning: isFollowingAll, progress } = useFollowAll({ onFollowed: preserveFollowedUser });
+
+  const showFollowAll = !isLoading && (unfollowedUsers.length > 0 || isFollowingAll);
+  // Finish decides the landing feed from `followedCount`, which is derived from the relationships
+  // live query and only updates once a follow's local write lands. Lock navigation while any
+  // follow (single card or Follow all) is in flight so Finish never reads a stale count.
+  const isNavigationLocked = isFollowingAll || isFollowPending;
```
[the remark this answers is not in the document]
Orlandohub: Valid. `useFollowUser().isLoading` is a single boolean shared by overlapping toggles, and card B stays clickable while card A is pending, so the concurrent case is reachable.

Fixed in 7fd839ce: `useWhoToFollowFollowPreservation` no longer reuses that flag and instead counts in-flight `handleFollowClick` calls (incremented before the await, decremented in `finally`), exposing `isFollowPending = pendingFollowCount > 0`. Added a test that starts two follows, resolves the first, and asserts the lock is still held until the second settles.

aldertnl (changes requested):
Implementation looks OK in general. One request: Please keep the users graphic top aligned when the pages expands to display more cards. 

Image

aldertnl (dismissed):

talosmachina (commented):
Reviewed #2452 (feat(ui): add onboarding follow best matches step, #2388) and found nothing to report.

What I checked, so you know what was and was not covered:

- Read whole (not just diff): FollowBestMatchesForm.tsx, useFollowAll, useStarterPackSuggestions (+ utils/types), useExperienceCompletedGuard, useWhoToFollowFollowPreservation, useFollowUser, SuggestedUserCard, FollowButton, UserListItem, FollowBestMatches template, TagsOfInterestForm, the onboarding store actions, userStream.helper.ts, routes.ts / config diffs, the Cypress support changes, and the two staging-invite skill scripts.
- CI: all checks green (5 vitest shards, quality, Next build, preview deploy).
- Call sites grepped: commitFollow callers, buildStarterPackStreamId consumers, reserved-tag filtering, markExperienceCompleted/reset interplay, ALLOWED_ROUTES and Header step entries.
- Disproved after checking: (1) Follow all during an in-flight per-card follow can double-PUT the same follow, but the PUT is idempotent and the local write is an upsert on the same key, so no wrong state; the buttons-disable window plus the ref-based guard cover the double-click cases. (2) Finish reading a stale followedCount is guarded by isNavigationLocked (isFollowingAll || isFollowPending) on both nav buttons. (3) An invalid persisted tag seed cannot crash the step: addTag sanitizes and buildStarterPackStreamId failures fall back to the influencers stream. (4) The non-reactive getState() read in useExperienceCompletedGuard is deliberate and the Finish handler owns in-visit navigation, so no double redirect. (5) Unfollow from a preserved card removes it from the preserved list, which is consistent: it re-enters the excludeFollowing-free suggestion set.
- Not covered: Cypress run on staging and the manual checks were skipped in the PR itself (unchecked boxes in the test plan), and I did not run the app. The unaddressed review items from #2391 (interleave dedupe, follow-cache invalidation) remain open per the PR description, which is fine as long as they land somewhere.

No findings. The one thing I would chase before merge is the staging Cypress run, since completeOnboardingFromInstall now changes shape for every existing spec.

src/hooks/useStarterPackSuggestions/useStarterPackSuggestions.ts
```diff
@@ -0,0 +1,58 @@
...
+  const suggestions: SuggestedUser[] = users.map((user) => ({
+    ...user,
+    matchingTags: selectMatchingTags(user.tags, interestTags),
+  }));
+  const unfollowedUsers = suggestions.filter((user) => !user.isFollowing);
+
+  return {
+    users: suggestions,
+    unfollowedUsers,
+    followedCount: suggestions.length - unfollowedUsers.length,
```
infin1t3: Finish decides the landing feed from cards still on screen, not from follows the user actually has.

`followedCount` is `suggestions.length - unfollowedUsers.length`. That works while you stay on the page because preservation keeps followed cards visible. Follow one or two people, go Back to tags, Continue, then Finish: preservation is empty, `excludeFollowing` hides people you already follow, `followedCount` is 0, and you land on All. Refresh on this screen does the same. The follows are still in Dexie. `useDefaultHomeReach` only corrects this at 3+ follows.

Count real follows (Dexie or a session counter), not the visible grid. Finish is also clickable while the skeleton is up (`isNavigationLocked` ignores `isLoading`); lock it until suggestions have settled.

src/hooks/useFollowAll/useFollowAll.ts
```diff
@@ -0,0 +1,97 @@
...
+    runningRef.current = true;
+    setIsRunning(true);
+    setProgress({ completed: 0, total: pending.length });
+
+    try {
+      for (const followee of pending) {
+        try {
+          await UserController.commitFollow(HttpMethod.PUT, { follower: currentUserPubky, followee });
+          result.followed.push(followee);
+          onFollowed?.(followee);
```
infin1t3: Follow all only preserves a card after `commitFollow` resolves.

`onFollowed` sits after the full PUT. `commitFollow` writes Dexie first, so the live query can drop the card before preservation runs. The single-card path preserves before the `await`, which is why that path does not blink. Follow all still follows everyone; the grid flickers and can trigger a refill.

Call `onFollowed` as soon as the local write lands, same as `handleFollowClick`.

src/components/organisms/FollowBestMatchesForm/FollowBestMatchesForm.tsx
```diff
@@ -0,0 +1,153 @@
...
+                ? 'Suggestions are unavailable right now. You can still finish and explore.'
+                : 'No suggestions yet. You can still finish and explore.'}
+            </Typography>
+          ) : (
+            <Container overrideDefaults className="grid gap-3 md:grid-cols-2" data-testid="suggested-people-grid">
+              {users.map((user) => (
+                <SuggestedUserCard
+                  key={user.id}
+                  user={user}
+                  isLoading={isFollowingAll || isUserLoading(user.id)}
```
infin1t3, line 142: Two quick Follow clicks can re-enable the first card mid-PUT.

`isUserLoading` comes from `useFollowUser`, where it is `isLoading && loadingUserId === userId`. The second click overwrites `loadingUserId`, so card A's spinner stops while A's request is still in flight. `pendingFollowCount` already locks Back/Finish for this. The cards and Follow all do not use it. Same root Greptile already flagged for navigation.

Track in-flight ids in `useFollowUser` (a Set or a count), and disable Follow all while `isFollowPending`.

src/components/organisms/FollowBestMatchesForm/FollowBestMatchesForm.tsx
```diff
@@ -0,0 +1,153 @@
...
+                  <>
+                    <UserRoundPlus className="size-4" />
+                    {`Follow all (${unfollowedUsers.length})`}
+                  </>
+                )}
+              </Button>
+            )}
+          </Container>
+
+          {isLoading ? (
```
infin1t3, line 128: If `getManyDetails` fails, `useLiveQuery` in `useUserStream` returns an empty Map, `detailsHydrated` stays false, and `isLoading` never clears. This skeleton then spins forever instead of falling through to the empty/error copy below.

Show the empty/error copy when hydration fails instead of waiting on details that will not arrive.

infin1t3 (changes requested):

src/hooks/useStarterPackSuggestions/useStarterPackSuggestions.ts
Orlandohub: Fixed in 7bec2218. Finish now reads the viewer's real follow graph instead of the visible grid: new `useFollowingCount` live-queries the local `<pubky>:following` stream that `LocalFollowService` maintains on every follow/unfollow write (via a new local-only `StreamUserController.getStreamUserIds`), minus the moderation bot's auto-follow. That stream is written unconditionally, unlike the counts row a brand-new account may not have yet, and it survives Back → Continue and refresh. `followedCount` is gone from `useStarterPackSuggestions`. Finish/Back also lock while `isLoading` or the count read is still settling.

src/hooks/useFollowAll/useFollowAll.ts
Orlandohub: Fixed in 7bec2218. `useFollowAll` now takes `onFollowStarted` (called before `commitFollow`, so the card is preserved before the Dexie write can drop it) and `onFollowFailed` (rolls the preservation back). The form wires them to `preserveFollowedUser` / new `unpreserveFollowedUser`, mirroring the single-card path.

src/components/organisms/FollowBestMatchesForm/FollowBestMatchesForm.tsx
Orlandohub: Fixed in 7bec2218 at the root: `useFollowUser` now keeps in-flight toggles in a per-user Map, so `isUserLoading(a)` stays true until a's own request settles regardless of other clicks, and `isLoading` is true while any toggle is pending. The preservation hook forwards that `isLoading` as `isFollowPending` (the ad-hoc counter is removed), and Follow all is disabled while it's set. `loadingUserId` / `loadingAction` keep their single-target semantics (most recent in-flight) for the profile header.

src/components/organisms/FollowBestMatchesForm/FollowBestMatchesForm.tsx
Orlandohub: Fixed in 7bec2218. The details/relationships live queries in `useUserStream` now return a snapshot tagged with the `userIds` they ran for, and hydration means "a snapshot exists for the current ids" rather than "the map contains one of them". A settled-but-empty read (nothing cached, or `getManyDetails` threw) therefore clears `isLoading` and the form falls through to the empty/error copy. The tag also fixes the stale case where the previous ids' map was still being returned after a stream change.

infin1t3 (approved):

PR #2455, opened 2026-09-03
fix(posts): match published article typography to the editor

### LINKED ISSUE #1762: [Bug]: new line spacing and enumeration formatting differs in editor and published article
### Description

When creating an article, the editor does not show the formatting/spacing of the resulting post correctly. Attached is an image and a video that gives examples. Spacing is different, enumeration has weird newlines and so on. Visually, the text is not pleasant to look at.

### Expected Behaviour

What You See Is What You Get [WYSIWYG wiki](https://en.wikipedia.org/wiki/WYSIWYG)

### Steps to Reproduce

Create an article yourself

### Screenshots

https://github.com/user-attachments/assets/5b5c0fb5-c8d2-4263-ba6e-a86295571f31

Image

### Console Log

```shell

```

### Browser & Version

_No response_

secondl1ght:
## Summary

Closes #1762 — the article editor and the published article disagreed on spacing, and adding blank lines to compensate broke numbered lists (each number rendered alone on its own line).

**Root cause.** Published article bodies rendered with the compact post styling: no block margins at all (Tailwind preflight zeroes them), with the only spacing coming from the stray newlines `react-markdown` emits between blocks under `whitespace-pre-line`, plus `list-inside` markers. The editor uses Tailwind `prose` typography. Blank lines between list items produce a *loose* list (`<li><p>…</p></li>`), and the block-level `<p>` pushed the inline marker onto its own line.

`whitespace-pre-line` itself predates markdown support — it was added in #334 to preserve blank lines in plain-text feed posts. That surface is untouched here.

## Changes

- **Published article page** (`PostText` gains a `fullArticle` prop, set by `PostArticleDetail`): renders with the same `prose prose-neutral prose-invert` typography as the editor — real paragraph/list/heading margins and outside list markers, which also render loose lists correctly. Prose side effects are neutralized where they would change the existing look: inline-code backticks, `<pre>` box/typography around `PostCodeBlock`, link underlines (links, hashtags, mentions keep brand styling), blockquote quotes/italics, and h5/h6 get margins prose does not provide.
- **Soft breaks** (`remarkSoftBreaks`, full articles only): single newlines become `<br>`. MDXEditor serializes Shift+Enter as a bare `\n`, so this keeps those line breaks — now also inside list items and blockquotes, where the old pre-line approach never reached.
- **Compact posts** (feed, threads, bios): the only change is that a loose list item's marker stays on the same line as its text.
- **Editor**: rich-text line-height, weight and color mirror the published body (`leading-6 font-medium text-secondary-foreground`) so the editor previews the published spacing.
- **Tests**: unit coverage for the plugin, both list modes, soft breaks, blockquote/heading neutralization, plugin-order pin and no-truncation guarantee. The article VRT fixture gains a heading, a loose numbered list and a soft-break paragraph, and a new screenshot captures the published body scrolled into view (previously below the fold and uncovered). Linux baselines for chromium/firefox/webkit regenerated.

## User-facing changes

Published articles: list items get breathing room and outside markers with indent; loose and nested lists render correctly; headings, hr and code blocks gain margins; hard breaks no longer double-space; bold text is white and list markers muted (prose-invert, matching the editor). Feed article previews are pixel-identical. Editor line pitch goes from ~28px to 24px in rich-text mode.

## Verification

- Full unit suite and full VRT suite (chromium/firefox/webkit) pass on the merged tree; typecheck/lint/format clean.
- Element-by-element before/after screenshots of every supported markdown element on the article page (headings, breaks, inline formatting, tight/loose/nested/multi-paragraph lists, blockquotes, links/hashtags/mentions, code, hr, table fallback), plus the reporter's exact scenario.
- Two code-review passes; findings fixed in follow-up commits.

## Notes for reviewers

- **darwin VRT baselines** for the article/editor surfaces need the *VRT Update Baselines* workflow run on this branch (cannot be generated from Linux).
- Regenerating on the merged tree also showed 21 **onboarding** baselines stale on `dev` itself (the tags-of-interest step changed the stepper without updating screenshots; dev's VRT workflow is currently failing). Those are deliberately left out of this PR.
- Out of scope, pre-existing: heading *font sizes* still differ between the editor (prose sizes) and the published page (custom sizes). Spacing now matches; a follow-up could restyle the editor headings.


Orlandohub (approved):
Two AI medium notes:

[1] medium | duplication | src/components/molecules/PostText/PostText.tsx:140 ↔ InitializedMDXEditor.tsx:430
What's wrong: The shared prose class list (prose prose-neutral prose-invert prose-code:before:content-none prose-code:after:content-none max-w-none leading-6 font-medium text-secondary-foreground) is hand-copied into both files, kept in sync only by a comment saying it "mirrors PostText". Why it matters: The PR's whole premise is that editor and published body stay aligned; two literals will drift on the next tweak. The folder already has the precedent for this: INLINE_LINK_CLASSNAME in PostText.constants.ts. Fix: Export ARTICLE_PROSE_CLASSNAME from PostText.constants.ts, then cn(ARTICLE_PROSE_CLASSNAME, 'prose-pre:bg-transparent … [&_a]:no-underline') in PostText and cn(ARTICLE_PROSE_CLASSNAME, 'px-0! pb-0! pt-4! max-h-[60dvh] overflow-y-auto') in the editor.

[2] medium | scrutinize-optional-params | src/components/molecules/PostText/PostText.types.ts:14
What's wrong: fullArticle is a fourth behavior-changing optional prop, and its only invariant ("pass together with isArticle") lives in JSDoc. <PostText fullArticle /> without isArticle compiles and yields a hybrid: prose styling + soft breaks, but headings disallowed, remarkDisallowMarkdownLinks running, and preview truncation active. Why it matters: Invalid state is representable in the type and unhandled at runtime (.greptile rule scrutinize-optional-params). Fix: Cheapest: derive const article = isArticle || fullArticle inside the component and read article everywhere isArticle was read. Cleaner: replace both booleans with variant?: 'post' | 'articlePreview' | 'article'; the !onPostPage && !fullArticle and !fullArticle && … checks then read as variant === 'articlePreview' / variant !== 'article'.

PR #2461, opened 2026-09-03
chore(deps): update patch/minor dependencies

infin1t3:
## Summary

- Patch/minor dependency sweep: **40 version bumps**, 6 unused packages removed, and 5 previously transitive imports declared explicitly.
- Removed 3 stale overrides: `intersection-observer` is no longer in the dependency tree, the `source-map` beta override no longer matches, and `glob` now resolves naturally to 13.0.6.
- `eslint-plugin-react-hooks` 7.1 introduces the React Compiler rules `react-hooks/refs` and `react-hooks/preserve-manual-memoization`; all 8 reports are fixed in code rather than disabled.
- Snapshots regenerated for `radix-ui` (closed triggers no longer set `aria-controls`) and `lucide-react` (icon geometry/class renames).

Node remains on **24 LTS**. `@synonymdev/pubky` and `pubky-app-specs` are intentionally untouched.

## Notable version decisions

| Package | Version | Why |
|---|---|---|
| `@sentry/nextjs` | **10.71.0 exact** (not 10.73.0) | 10.72.0+ crashes on import under jsdom/happy-dom (getsentry/sentry-javascript#23789). Fix merged in #23906, unreleased. Bump when the next release ships. |
| `vitest`, `@vitest/browser`, `@vitest/browser-playwright`, `@vitest/coverage-v8` | 4.1.11 (lockstep) | Closes the critical browser-mode advisories. Vitest 5 skipped: `vitest-browser-react` only supports `^4`. |
| `cypress` | 15.21.1 | Closes the high advisories. No config change needed; `Cypress.env()` is only removed in Cypress 16. |
| `undici`, `linkify-it` | 7.29.0, 5.0.2 | Security patches within the current major; majors are a follow-up. |
| `glob` | 13.0.6 | Resolves directly from Serwist/Sentry after removing the stale 13.0.1 override. |

Removed (no reference anywhere in the repo): `@faker-js/faker`, `date-fns`, `next-themes`, `@eslint/eslintrc`, `ts-node`, `eslint-config-next`.

Declared explicitly (imported directly but only present transitively): `@next/eslint-plugin-next` (must move together with `next`), `@typescript-eslint/parser`, `@typescript-eslint/eslint-plugin`, `@types/mdast`, `@codemirror/language-data`.

`npm audit`: 33 → 5 high. Everything left needs `next` 16.3.4 (3) or an upstream `browserslist` fix via Serwist (2).

## React Compiler rule fixes

- `useFeedback`: manual `useCallback`/`useMemo` removed (the compiler memoizes; the manual dependencies could not be preserved).
- `useSearchAutocomplete`, `useTagSuggestions`: the debounced worker is created once in an effect instead of `useRef(debounce(...))` during render.
- `useSearchPeople`, `useStreamPagination`, `HumanInviteCode`, `DiscoverCollections`: latest-ref writes moved from render into an effect declared before their readers; no observable change.
- `useNotifications`: the frozen `lastRead` is state instead of a ref because it is read during render. The unread list now recomputes when the frozen value arrives instead of staying stale until notifications change.

Two tests adjusted: `InitializedMDXEditor` looks for the renamed Lucide class (`lucide-face-slightly-smiling`); `useEditAttachments` asserts the terminal placeholder state (the hook is unchanged).

## Verification

All checks below pass on Node 24:

- Clean `npm ci`; `npm ls` has no invalid peers.
- Typecheck and lint (0 problems).
- Full unit suite: **753 files, 12,924 passed, 2 skipped**.
- Production `next build`.
- `format:check`.

Not run locally: VRT (requires Firefox/WebKit and updated Playwright browsers). Baselines will need regenerating through the VRT workflow after merge because of font/Tailwind/Lucide changes.

## Not bumped (blocked or deferred)

- ESLint 10: `eslint-plugin-react` (peer `^9.7`) and `eslint-plugin-import` (peer `^9`) have no compatible release.
- TypeScript 7: `typescript-eslint` requires `<6.1.0`. TS 6.0.x is possible in the majors follow-up.
- `next` 16.3.4 + `@next/eslint-plugin-next` 16.3.4: separate follow-up PR so it can be reverted in isolation.
- `prettier` 3.9 + `prettier-plugin-tailwindcss` 0.8: separate `style:` PR because of formatting churn.
- Remaining majors proven in #2326 (`linkify-it` 6, `motion` 13, `undici` 8, `jsdom` 30, `jest-dom` 7, `lint-staged` 17, `simple-import-sort` 14, `react-easy-crop` 6): follow-up PR.

## Notes for QA

- E2E `search.cy.ts` is already failing on `dev` (last runs on 2026-09-02/03), unrelated to this PR.
- Cypress 15.21.1 requires no changes under `cypress/`.


package.json
catch-21: It should be possible to remove these overrides. The security patches has since been included without overriding versions. Best to double check.

infin1t3: Yup, looks like. I'll go ahead and remove it.

cypress/cypress.config.ts
```diff
@@ -22,13 +22,13 @@ export default defineConfig({
...
     experimentalWebKitSupport: true,
     env: {
-      // slow down execution more in CI to avoid flaky tests
-      commandDelay: defaultMs,
       // Server-side only admin credentials (not exposed to client bundle)
       homeserverAdminUrl: process.env['HOMESERVER_ADMIN_URL'] || 'http://localhost:6288/generate_signup_token',
       homeserverAdminPassword: process.env['HOMESERVER_ADMIN_PASSWORD'] || 'admin',
     },
     expose: {
+      // slow down execution more in CI to avoid flaky tests
```
catch-21, line 30: Technically, we're still running faster in CI than Cypress default. But indeed, local runs are way faster than CI still.

catch-21 (approved):

PR #2392, opened 2026-08-21
fix(ui): restore original collection add-post options

Orlandohub:
## Summary

- restore the original four-option Add Post dialog in a responsive one-column/two-column layout
- add Select from posts and route its Select action to /profile/posts without mutating collection content
- preserve Add from feed, paste URL including the clipboard button from #2381, and create-post behavior
- restore the original responsive descriptions and Start writing prompt

https://github.com/user-attachments/assets/32f4094b-984d-46cb-805a-09548ae08b51

https://github.com/user-attachments/assets/84566877-9d78-4c94-870a-d1685f116fc2

## Testing

- npm test -- src/components/organisms/Collections/DialogAddContent/DialogAddContent.test.tsx src/hooks/useAddContentForm/useAddContentForm.test.ts src/components/molecules/ControlledInputField/ControlledInputField.test.tsx src/components/molecules/InputField/InputField.test.tsx (60 passed)
- TypeScript typecheck
- ESLint and Prettier on touched files
- git diff --check

## Design verification

- checked implementation against desktop Figma node 39801:140506 and mobile node 40349:309314
- authenticated desktop/mobile runtime comparison remains pending because the fresh local browser stayed on the authentication redirect
- no VRT baseline is required for this dialog

## Dependency

This is stacked on #2381 so its clipboard control remains part of the resulting dialog. Retarget this PR to dev after #2381 merges.

Closes #2105

PR #2463, opened 2026-09-03
feat(ui): merge Select from posts into the Add from feed tile

Orlandohub:
## Summary

Follow-up to #2392 based on the Slack discussion with Aldert. "Select from posts" and "Add from feed" are not distinct features (both send the user to a feed where they use the collection button on a post), so they now share a single tile.

- Drop the standalone **Select from posts** card; the dialog goes back to three stacked cards (Add from feed / Paste post url / Create new post), matching the [Figma](https://www.figma.com/design/01ZvjSPZnKTNmaEWz0yJsq/Pubky-SHADCN?node-id=41492-356218).
- **Add from feed** card: illustrative pills on the left, two new pills on the right — **Feed** (`Activity` icon → `/home`) and **My posts** (`UserRound` icon → `/profile/posts`), per Aldert's final call to use "Feed" instead of a reach-aware "My network".
- `ActionPill` gets an `emphasis` variant (`muted` | `highlighted` | `default`) with exhaustive switch instead of the boolean.
- Paste URL and Create post behaviour untouched.

## Testing

- `npx vitest run src/components/organisms/Collections/DialogAddContent` (30 passed, snapshot updated)
- `npm run typecheck`, ESLint and Prettier on touched files

No VRT covers this dialog. Not visually verified in a browser (no staging credentials in this worktree) — please sanity-check the pill row wraps nicely on mobile.


aldertnl (approved):

secondl1ght (approved):
Approving. Behaviour matches the description: the Select card is gone, the three cards stack, and Feed / My posts close the dialog and route to `/home` and `/profile/posts`. Paste URL and Create post are untouched in the diff.

Checked at the PR head:
- Unit tests (30), ESLint, Prettier, and `tsc` pass.
- No remaining consumers of the removed `add-content-select-from-posts` selector in Cypress or docs. No VRT renders the open dialog, so no baselines need regenerating.
- Rendered the dialog in a browser at 375px and desktop. Desktop matches the Figma node. On mobile the pill row wraps to two clean lines with no overflow.

PR #2464, opened 2026-09-03
chore(deps): approve dependency install scripts via allowScripts

secondl1ght:
## Summary

`npm install` on npm 11.16+ prints a warning listing five dependencies whose install scripts are "not yet covered by allowScripts". On npm 11 the scripts still run (advisory phase); on npm 12 unapproved scripts are skipped by default, which would break Cypress (its postinstall downloads the runner binary).

This records the reviewed policy in `package.json` so a fresh checkout installs quietly and identically on npm 11 and npm 12:

```json
"allowScripts": {
  "@sentry/cli": true,
  "browser-tabs-lock": true,
  "cypress": true,
  "sharp": true,
  "unrs-resolver": true
}
```

Entries are name-only (any version) so a dependency bump does not need a re-approval. Any *new* dependency that introduces an install script is still flagged by npm (`npm install-scripts ls`), which is the main protection the policy provides.

## What each script does

| Package | Script | Notes |
|---|---|---|
| cypress | downloads the Cypress binary to `~/.cache/Cypress` | required for `cypress run` / `cypress verify` |
| sharp | `install/check.js \|\| npm run build` | no-op when the prebuilt `@img/sharp-*` package is present; builds from source only if a global libvips is found |
| @sentry/cli | `scripts/install.js` | exits immediately when `@sentry/cli-<platform>` is present; otherwise downloads the binary from Sentry's CDN |
| unrs-resolver | napi-postinstall check | no-op when `@unrs/resolver-binding-*` is present; otherwise fetches the binding |
| browser-tabs-lock | prints a message | informational only |

## Verification

- Fresh `npm ci` / `npm install` with npm 11.19.0 and npm 12.0.2: no install-scripts warning; all five scripts execute.
- With all five scripts blocked (npm 12, no approvals): `npm run build`, `npm run lint`, the production server, `/_next/image` (sharp), and the Dockerfile's `sentry-cli sourcemaps inject` step all still work. Only Cypress fails (`No version of Cypress is installed`), which is why the approval is needed.
- `package-lock.json` is unchanged.

Also adds a short README section describing the policy and how to approve a newly flagged package.


Taewa (approved):
Approved. 

One non-blocking AI note from a macOS run of this branch:

npm install-scripts ls (npm 11.19.1 and 12.0.2) still reports two packages:

```
  fsevents@2.3.3 (via rollup / vite)
  fsevents@2.3.2 (via playwright)
```

fsevents is a darwin-only optional dependency, so it is not installed on Linux and does not show up there. The package ships a prebuilt binary and has no real install script. The lockfile just marks it hasInstallScript: true, so the check flags it. Nothing breaks, it is only a warning every macOS developer will see on npm install.

If you want to silence it, adding "fsevents": true to allowScripts is enough. Note that npm install-scripts prune on Linux would see that entry as unused and remove it again. Fine to handle in a follow-up.

PR #2474, opened 2026-09-04
fix(users): fetch viewer-aware follow state

### LINKED ISSUE #1803: [Bug]: "Follow" button still showing for people I already follow
### Description

When I hover over Warlando, I see that I am a follower of him (my profile pic) but the button still says "Follow.

### Expected Behaviour

Should say "Unfollow"

### Steps to Reproduce

No idea

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
## Summary
Nexus `relationship.following` is viewer-relative. Guest / viewer-less fetches were persisted as `following: false`, and a cached relationship row is never re-fetched by the Follow button. That is why a hover card could show you among someone's followers while the button still said Follow.

- Do not write `user_relationships` unless the Nexus batch was fetched for a `viewerId`. A missing row is a cache miss and triggers a viewer-aware fetch.
- Treat a missing relationship as a miss in `getNotPersistedUsersInCache` when a viewer is present, so Following / streams / notifications / TTL hydrate follow state after sign-in (including details cached from a guest session).
- Pass that viewer through user, post, hot, stream, bootstrap, and TTL persist/fetch. Dest's tag-cache `persistUsers` guard (`{ revisions, viewerId, isCurrent }`) is used as-is; relationships are still skipped when `viewerId` is absent.
- Subscribe the Follow-button target to the existing user TTL coordinator so a stale row (pre-fix cache, or a follow on another device) refreshes when user TTL elapses (10 minutes).
- Stamp the followee user TTL on follow/unfollow and skip relationship persist when that stamp is newer than the fetch, so a background refresh cannot overwrite a successful Follow with a stale `following: false`.
- Subscribe only valid pubkys. A malformed `/profile/...` id no longer enters the shared TTL batch (Nexus HTTP 400) and cannot fail valid users in the same request.

This is not live homeserver follow replication. A cached `following: false` stays until user TTL elapses; same-device follow/unfollow updates immediately.

## Test plan
- [ ] Signed in, hover a user you follow: button says Unfollow; your avatar can still appear among their followers
- [ ] After sign-in, own Following list and that user's profile show Unfollow
- [ ] Browse as guest, then sign in: Follow buttons become correct without a hard reload
- [ ] Follow / unfollow on this device still updates immediately
- [ ] Follow a stale cached user in a mention popover while a refresh is in flight: button stays Unfollow
- [ ] Open `/profile/invalid-profile-key` while signed in: discovery users still refresh

Closes #1803


infin1t3:
@greptileai review

Orlandohub:
profile-unfollow

profile-following

https://github.com/user-attachments/assets/1169cdc0-98df-451a-8b9e-7db408a49b2f

src/hooks/useIsFollowing/useIsFollowing.tsx
```diff
@@ -40,6 +40,21 @@ export function useIsFollowing(targetUserId: string): UseIsFollowingResult {
...
+    if (!enabled || !isPubkyIdentifier(targetUserId)) return;
+
+    const coordinator = TtlCoordinator.getInstance();
+    const pubky = targetUserId as Pubky;
+    coordinator.subscribeUser({ pubky });
+
+    return () => {
+      coordinator.unsubscribeUser({ pubky });
+    };
+  }, [enabled, targetUserId, pathname]);
```
Orlandohub: **The premise here is not true, so `pathname` should come out of this effect.**

I traced it in source on this branch:

- `TtlCoordinator` has no `setRoute`; `CoordinatorsManager.applyRouteToCoordinators` only routes `notification`, `stream` and `muteListSync` (`src/components/organisms/CoordinatorsManager/CoordinatorsManager.tsx`).
- `TtlCoordinator.reset()` (the only thing that clears `userRefCount`) is called solely from `stop()`, i.e. when `CoordinatorsManager` unmounts. It lives in the root `app/layout.tsx`, so that never happens on navigation. `RouteGuardProvider` does unmount its subtree while it shows its spinner, but then this hook's consumer unmounts too and the effect re-runs on remount without any `pathname` dependency.
- The auth listener in `setupListeners` explicitly keeps subscriptions across account changes ("Subscriptions survive an account change").
- `docs/local-first.md` (Public views and viewport subscriptions) states it directly: "Route changes do not clear these references globally."

With the dependency in place, every navigation makes each mounted `useIsFollowing` unsubscribe and re-subscribe: the ref count dips to 0 (which also drops the id from `userBatchQueue` via `removeSubscription`) and `checkAndQueueEntity` runs one extra Dexie read per hook per route change. Nothing breaks, but it is churn justified by a comment that misdescribes the system, and the next reader will believe the coordinator resets on route change.

**Fix:** drop the `usePathname` import and `const pathname`, use `[enabled, targetUserId]` as the dependency list, rewrite the comment to just say the row is kept fresh via the coordinator (no route-change sentence), remove the `re-subscribes on route change so layout-mounted buttons survive coordinator reset` test and the `mockPathname` / `next/navigation` mock in `useIsFollowing.test.tsx`, revert the `usePathname` mock added to `PostTagPopoverWrapper.test.tsx`, and update the PR description bullet that repeats the claim.

src/hooks/useIsFollowing/useIsFollowing.test.tsx
```diff
@@ -171,4 +191,51 @@ describe('useIsFollowing', () => {
...
+      mockGetRelationships.mockReturnValue({ following: false });
+
+      const { rerender } = renderHook(() => useIsFollowing(VALID_TARGET));
+
+      mockPathname.value = `/profile/${VALID_TARGET}/followers`;
+      rerender();
+
+      expect(mockUnsubscribeUser).toHaveBeenCalledWith({ pubky: VALID_TARGET });
+      expect(mockSubscribeUser).toHaveBeenCalledTimes(2);
+      expect(mockSubscribeUser).toHaveBeenLastCalledWith({ pubky: VALID_TARGET });
```
Orlandohub: This test passes only because `TtlCoordinator` is mocked; against the real coordinator there is no "coordinator reset" on route change (see the comment on `useIsFollowing.tsx`). It enshrines behaviour the system does not have, so it should go together with the `pathname` dependency and the `mockPathname` / `next/navigation` mock above (lines 23-26, 58).

The other three TTL tests in this block (subscribe on mount, skip malformed id, re-subscribe on target change) are the right coverage and can stay.

src/core/application/user/user.types.ts
```diff
@@ -7,6 +8,11 @@ import type { NexusSocialGraphStatus, NexusUserCounts } from '@/services/nexus/n
 /** Cached Dexie row (includes `id`) or Nexus API payload (counters only). */
 export type TUserCountsOrFetchResult = UserCountsModel | NexusUserCounts;
 
+/** Target user plus the signed-in viewer (null for guests) so the persisted relationship is viewer-relative. */
+export type TUserApplicationFetchParams = TReadProfileParams & {
+  viewerId?: Pubky | null;
+};
```
Orlandohub: `TUserApplicationFetchParams` is exported but never imported anywhere (`rg TUserApplicationFetchParams src` only returns this definition). `UserApplication.getOrFetch` / `fetch` still take `TFetchUserParams & { isCurrent }`, and the new `TReadProfileParams` import on line 1 exists only to build this type.

Dead exports are a `no-unused-exports` / `no-dead-code` item in `.greptile/config.json` and `docs/development-workflow.md`. Either wire it in as the actual parameter type of `UserApplication.getOrFetch` / `fetch` (then `TFetchUserParams` in the controller would need to allow `null`, which it does not today) or, simpler, delete the type and the `TReadProfileParams` import.

docs/data-patterns.md
```diff
@@ -246,7 +246,7 @@ All tables defined in `src/core/database/franky/franky.ts`.
 ```
 user_details       — Profile data (name, bio, image, links, status) + social_graph_status folded in from the Nexus user view
 user_counts        — Follower/following/post counts
-user_relationships — Follow/mute relationships
+user_relationships — Follow/mute relationships (viewer-relative: only written when the fetch carried a `viewerId`)
```
Orlandohub, line 250: Good to have this here. Two more doc touches are needed for the convention this PR introduces, because the current text now contradicts the code:

1. `docs/data-patterns.md`, _Viewport TTL subscriptions (UI)_ (line ~136) says: "The only thing that refreshes a cached row while it is on screen is a viewport subscription to the `TtlCoordinator` (ADR-0012) via `useTtlSubscription`", and the subscriber table below it lists only surfaces. `useIsFollowing` is now a second subscriber and it is *not* viewport-gated (every mounted Follow control, in popovers, menus, tagger rows and participants lists, holds a user reference for as long as it is mounted). Please add a row for `useIsFollowing` ("the follow target, not viewport-gated: data hook without a DOM node; relationship rows are viewer-relative and only TTL refreshes them") and soften the "only thing" sentence.
2. `docs/local-first.md`, _Read pitfalls_ (TTL refresh races) is where the `fetchStartedAt` guard belongs in one sentence: a `persistUsers` call skips the relationship row for any user whose `user_ttl.lastUpdatedAt >= fetchStartedAt`, and `LocalFollowService.create/delete` stamp the followee TTL so a follow that lands mid-request wins. That is exactly the #1781 class the section already describes, and the next person touching `persistUsers` needs to know the TTL row now doubles as a write fence.

src/core/services/local/follow/follow.test.ts
```diff
@@ -220,6 +222,13 @@ describe('LocalFollowService.create', () => {
     expect(bConn?.followers ?? []).toContain(userA);
   });
 
+  it('stamps the followee user TTL so an in-flight refresh can keep the local follow', async () => {
+    await LocalFollowService.create({ follower: userA, followee: userB });
+
+    const ttl = await UserTtlModel.findById(userB);
+    expect(ttl?.lastUpdatedAt).toBeGreaterThan(0);
+  });
```
Orlandohub, line 230: `LocalFollowService.delete` stamps the followee TTL too (`follow.ts` line 140) but only `create` is covered. Please add the mirror case for unfollow: create the follow, reset the TTL row to a stale value (or clear it), call `delete`, and assert `UserTtlModel.findById(userB)?.lastUpdatedAt` is fresh. The guard in `persistRelationshipsAndTtl` protects unfollow-during-refresh only if that stamp is there, so it deserves its own assertion.

Two smaller coverage gaps in the same spirit (fine to fold into follow-ups if you prefer):
- `notification.test.ts` mocks `getNotPersistedUsersInCache` but never asserts it is called with `viewerId` after this change (`notification.ts` line 342).
- `post.test.ts` adds a `UserRelationshipsModel.findByIds` spy that returns a row for every id, so no test exercises the new path where a post author has details cached but no relationship row and is therefore refetched viewer-aware.

src/core/services/local/stream/users/users.ts
```diff
@@ -146,12 +166,48 @@ export class LocalStreamUsersService {
...
+      if (fetchStartedAt !== undefined) {
+        const existingTtl = await UserTtlModel.findByIds(userIds);
+        const skipIds = new Set(existingTtl.filter((row) => row.lastUpdatedAt >= fetchStartedAt).map((row) => row.id));
+        if (skipIds.size > 0) {
+          Logger.debug('LocalStreamUsersService: Kept local follow state during refresh', {
+            ids: Array.from(skipIds).slice(0, 5),
+            count: skipIds.size,
+          });
+          toSave = userRelationships.filter(([id]) => !skipIds.has(id));
+        }
```
Orlandohub, line 205: Nit on the log copy: this branch fires for *any* `user_ttl` write newer than `fetchStartedAt`, and in practice the common trigger is another viewer-aware batch (bootstrap, a stream page, another TTL tick) landing first, not a local follow. The skip is still correct in that case because the earlier batch already wrote the same viewer-relative row, but "Kept local follow state during refresh" will mislead whoever reads the debug log. Something like `Skipped relationship rows written since the fetch started` matches what the code checks.

(For the record I did trace the guard for the concurrent-batch chain: the first batch to land only ever sees TTL rows older than both starts, so at least one batch always writes the relationship; and `upsertTtlWithDelay` writes a past timestamp, so deferred not-indexed users cannot trip it. No bug here, just the message.)

src/components/organisms/ProfilePageContainer/ProfilePageContainer.tsx
```diff
@@ -66,7 +66,7 @@ export function ProfilePageContainer({ children }: ProfilePageContainerProps) {
   // Business logic: Handle follow/unfollow for other users' profiles (with auth check)
   const { requireAuth } = useRequireAuth();
   const { toggleFollow, isLoading: isFollowLoading, loadingAction: followLoadingAction } = useFollowUser();
-  const { isFollowing } = useIsFollowing(pubky ?? '');
+  const { isFollowing } = useIsFollowing(pubky !== null && isPubkyIdentifier(pubky) ? pubky : '');
```
Orlandohub, line 69: Optional: the `isPubkyIdentifier` check now lives in two places with two different effects: here it disables the whole local-first query for one call site, and inside `useIsFollowing` it guards only the TTL subscription for every call site. `ProfileProfile.tsx` line 35 still passes `pubky ?? ''` (harmless today because `ProfilePageContainer` does not render children for an unknown user). Folding `isPubkyIdentifier(targetUserId)` into the hook's `enabled` would give one guard for both the fetch and the subscription across all seven consumers; the cost is switching the existing `useIsFollowing.test.tsx` fixtures to real pubky strings, which the new tests already do with `PUBKY_52_STAGING_FIXTURE`. Fine to leave as is if you would rather keep this PR small.

Orlandohub (changes requested):
## Review

The core fix is right and I could verify the premise end to end, so this is a "small cleanups before merge" request, not a design objection.

### Verified

- **Nexus really is viewer-relative** on `POST /v0/stream/users/by_ids` (staging): for a user A that follows B, `{ user_ids: [B] }` returns `{"following":false,"followed_by":false}` and `{ user_ids: [B], viewer_id: A }` returns `{"following":true,"followed_by":false}`. Self (`user_ids: [A], viewer_id: A`) returns a non-null `{false,false}`, so the signed-in user's own row is persisted and `getNotPersistedUsersInCache(ids, viewerId)` cannot loop on the viewer.
- Every `persistUsers` caller that fetches with `viewer_id` also passes `viewerId` to persist (user, user-stream, post-stream authors, TTL, bootstrap, hot via `getOrFetchUsers`, notifications), so the new "missing relationship row = cache miss" rule cannot produce a refetch loop. Guest paths pass no viewer on both sides.
- `fetchStartedAt` guard: traced concurrent-batch and follow-mid-request orderings; the first batch to land always writes the relationship, a local follow always wins, `upsertTtlWithDelay` writes a past stamp. Matches the existing `unlessWrittenSince` pattern.
- Layering: hook → `TtlCoordinator` has precedent in `useTtlSubscription`; `HotApplication → UserStreamApplication` is on the ADR-0009 allow-list (depth 1); controllers read the auth store, applications do not; the service uses models plus `db.transaction` like `LocalFollowService`.
- `npm run lint` ✅ · `npm run typecheck` ✅ · full `npm test` ✅ 782 files / 13 557 tests, 2 skipped (349 s) on `e73b282e` rebased on current `dev` (merge base is `dev` HEAD, no drift).

### Requested changes (inline)

1. **`useIsFollowing` re-subscribes on `pathname` based on a claim that does not hold**: the TTL coordinator has no route handling, `reset()` only runs on `stop()` (root-layout unmount), and `docs/local-first.md` says route changes do not clear references. Remove the `usePathname` dependency, the comment, the test that enshrines it, the `PostTagPopoverWrapper.test.tsx` mock it forced, and the PR description bullet.
2. **`TUserApplicationFetchParams` is an unused export** (plus its `TReadProfileParams` import).
3. **Docs**: `data-patterns.md` still says a viewport `useTtlSubscription` is the *only* refresher of an on-screen row and the subscriber table omits `useIsFollowing`; `local-first.md` (TTL races) should mention the `fetchStartedAt` / followee-TTL fence in one sentence.
4. **Tests**: `LocalFollowService.delete` TTL stamp is untested; `notification.test.ts` does not assert `viewerId` reaches `getNotPersistedUsersInCache`; `post.test.ts` has no case for an author with details but no relationship row.

Plus two nits (debug log copy in `persistRelationshipsAndTtl`; optional consolidation of the `isPubkyIdentifier` guard into the hook's `enabled`).

### Callouts

No `DB_VERSION`, `src/config` or dependency changes. No VRT surface (the `ProfilePageContainer` change is logic-only). Pre-fix `following:false` rows are not migrated and heal only via TTL once a Follow control is mounted, which the description already states; worth keeping in the release notes.

### Not verified

The manual test plan (multi-device follow, guest → sign-in) — the screenshots/video posted on the thread cover the profile page cases. Greptile's 5/5 on this PR is from the earlier single-file revision (`b4f7238`), not the current 30-file one.

src/hooks/useIsFollowing/useIsFollowing.tsx
infin1t3: Done in 35ede41a. `usePathname` is gone, the effect deps are `[enabled, targetUserId]`, and the comment only says the coordinator keeps the row fresh. Also dropped the route-change test, the `next/navigation` mocks (including `PostTagPopoverWrapper.test.tsx`), and the PR-description bullet.

src/hooks/useIsFollowing/useIsFollowing.test.tsx
infin1t3: Removed that test and the `mockPathname` / `next/navigation` mock with the pathname dependency. The other three TTL tests are still there.

src/core/application/user/user.types.ts
infin1t3: Deleted `TUserApplicationFetchParams` and the unused `TReadProfileParams` import. `UserApplication.getOrFetch` / `fetch` still take `TFetchUserParams & { isCurrent }`.

docs/data-patterns.md
infin1t3: Updated both. `data-patterns.md` no longer says viewport `useTtlSubscription` is the only refresher, and the subscriber table has a `useIsFollowing` row. `local-first.md` (TTL refresh races) now mentions the `fetchStartedAt` / followee-TTL fence in one sentence.

src/core/services/local/follow/follow.test.ts
infin1t3: Added the unfollow TTL stamp test (stale row, then `delete`, then a fresh `lastUpdatedAt`). Also filled the two smaller gaps: `notification.test.ts` now asserts `getNotPersistedUsersInCache` is called with `viewerId`, and `post.test.ts` covers an author with details cached but no relationship row.

src/core/services/local/stream/users/users.ts
infin1t3: Updated the debug copy to `Skipped relationship rows written since the fetch started`.

src/components/organisms/ProfilePageContainer/ProfilePageContainer.tsx
infin1t3: Leaving this as-is to keep the PR small, as you suggested. The container still guards the invalid-id call site; the hook still guards only the TTL subscription.

Orlandohub:
following-list

profile-following

https://github.com/user-attachments/assets/c3988f33-306e-4481-b110-d0b37ceb58f0

Orlandohub (approved):

PR #2475, opened 2026-09-04
fix(ui): use static toast copy instead of interpolating user-entered text

### LINKED ISSUE #2419: [Bug]: Saving feed with long name isn't truncated in success toast
### Description

When creating or editing a custom feed that has a long name, the success toast doesn't truncate the name so it gets cropped at the edge of the UI.

### Expected Behaviour

Long names are cleanly truncated with `...`

### Steps to Reproduce

1. Create a new custom feed with a long name
2. Observe long name in success toast is cropped on the right side

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

secondl1ght:
Closes #2419

## Summary

Product decision: toasts must not embed user-entered text. Long values (feed names, tag labels, display names, file names) overflowed the toast; instead of truncating, every toast now uses fixed copy.

- **Static copy everywhere.** `Feed created: {name}` → `Feed created`; `Tag added: {label}` → `Tag added`; `{username} muted` → `User muted`; `Following {name}` → `User followed`; `Reposted {author}'s post` → `Reposted`; file-validation messages drop the file name. Numeric counts and config constants (attachment limits, size labels) are kept since they cannot overflow.
- **Rejected attachments are counted per reason.** Dropping two oversized images now reads `2 images exceed the 20MB limit.` rather than one generic line per file.
- **Toast renderer.** The title switches from `break-words` to `wrap-anywhere` (the description already used it), so a long unbroken word wraps instead of being clipped — this was the actual clipping mechanism in the report. The description gains `whitespace-pre-line` so multi-reason messages render one reason per line.
- **Dead plumbing removed.** `toggleFollow` and every `onFollowClick` prop drop the `displayName` argument; the seven pass-through handlers that only existed to forward it are gone and components pass `toggleFollow` directly. `useFollowUser.utils.ts` (display-name resolver), the `originalAuthorName` repost option, and the `useUserDetails` lookup in `usePostInput` that fed it are deleted.
- **Docs.** Rule added to `AGENTS.md` and a "Copy is static" section in `docs/components.md`.

## Test plan

- [x] `npm run typecheck`
- [x] ESLint and Prettier on all changed files
- [x] Full unit suite: 756 files, 12955 tests pass
- [x] New test: dropping one unsupported file plus two oversized images yields the exact two-line tallied toast and no file name appears
- [x] Toaster snapshots regenerated for the title/description class change
- [x] Manual check: created a feed with a 102-character name in the dev app and confirmed the toast reads `Feed created`


infin1t3 (approved):
LGTM.

Issue #2468, opened 2026-09-04
Locks: two tabs paying the same lock at once can mint two bundle ids


Taewa:
**Paying for the same lock from two browser tabs at the same time can create two payments: one of them can be lost, or the reader can be charged twice.**

Follow-up to #2368.

## Problem

`LocksApplication.startPayment` does three steps that are not atomic: read the reader's saved bundle id from `/priv/social/purchases/<lockId>.json`, mint a new id when there is none, write it, then submit the proof.

If the same reader presses **Pay** on the same lock in two tabs at almost the same time, both tabs read "no file", both mint a different bundle id, both write it (the second write overwrites the first), and both submit. The Lock Server treats different bundle ids as separate tasks, so it creates two tasks and Paykit sends two payment requests to the wallet.

Results:
- If the reader pays the request whose id was overwritten and closes that tab, the saved file points at the other id. The paid content can never be reached again.
- If the reader pays both requests, they are charged twice.

The window is small (between one tab's read and the other tab's write, a few hundred ms), and it needs the same reader, the same lock, and two tabs. It does not happen across devices in a way the client can fix.

## Why the current code cannot close it

- The homeserver has no conditional write (no "create only if missing"), so a write-then-read-back only narrows the window; it does not close it.
- The server's `409 task_state_conflict` only protects "same bundle id, different content". Different ids are accepted as new tasks.

## Suggested fix

Serialize the read-mint-write-submit section across tabs of the same browser with the Web Locks API:

```ts
navigator.locks.request(`locks-pay:${readerPubky}:${lockId}`, () => LocksApplication.startPayment(...))
```

With the lock, the second tab's read sees the first tab's file, reuses that bundle id, and submits the same bundle, which the server accepts as a safe replay. Browsers without `navigator.locks` fall through to the current behaviour.

Alternative, if the homeserver ever gains a conditional PUT: use it instead of the lock.

## Where

`src/core/application/locks/locks.ts` → `startPayment`, marked with a `TODO:[Locks]` pointing at this issue.

aurumflux20:
One question on the write-up, because the answer changes what kind of bug this is.

[...]

Issue #2296, opened 2026-08-03
Locks: cache lock descriptors and unlocked posts in IndexedDB


Taewa:
Sub-issue of #1998. Follow-up to #2003 — came out of review feedback on pubky/pubky-app#2252.

## Problem

Every render of a locked post re-fetches from the network. Scrolling a feed past the same locked post ten times fetches `lock.json` ten times, and re-reading content the user already unlocked still costs a homeserver round-trip. Nothing about a lock is cached locally.

Two facts make caching cheap here:

- **`lock.json` is immutable.** Lock content cannot be edited after publishing (only the announcement can), so a cached descriptor never goes stale. The hardest part of cache design is free.
- **Dexie is already wired up.** The `franky` database (`src/core/database/franky/`) exists, so this is a table addition, not new infrastructure.

## Proposal

Cache lock data in IndexedDB in a **single `locks` table**, keyed by lock id (the announcement URL hash), with the record shape:

```jsonc
{
  descriptor: { /* full lock.json — creator, criteria, primary_resource, secondary_resources, … */ },
  post: {
    content: "you unlocked!",
    kind: "short",
    attachments: null   // or rewritten reader URLs after replication
  } | undefined,        // undefined = seen but content not loaded yet
  unlockedAt: 1712345678901 | undefined  // from the /priv marker; undefined for own locks and still-locked ones
}
```

Ownership is **derived, not stored**: `descriptor.creator === session pubkey`. No second table.

> Earlier revision of this issue proposed two tables (`ownLocks` / `unlockedLocks`). Dropped after discussion: the hot path is the stream, where every lock post would have to check both tables; a split buys nothing the `creator` field can't answer with one indexed query. Derived ownership is also account-switch-safe — a stored classification bakes in "who I was at write time", a `creator` comparison stays correct per session.

Query patterns (index: `creator`, or compound `[creator+unlockedAt]`):

| screen | query |
| --- | --- |
| feed / stream | single lookup by lock id; `creator === session` only picks the render branch |
| Unlocked profile page | `creator !== session && post !== undefined`, newest `unlockedAt` first |
| own-locks page (future) | `creator === session` |

Note the Unlocked page needs `post !== undefined`, not just `creator !== session`: descriptors are persisted for every lock *seen* in the stream, so the table also holds other people's locks that were never unlocked.

## Read flow

```
0. Signed out?                        → render lock card (nothing to read)

1. Check the locks table first (no lock.json needed, zero network)
[...]

PR #2481, opened 2026-09-04
[Locks] feat(#2368): pay with Bitkit to unlock a post

### LINKED ISSUE #2297: Locks: persist bundle_id so an interrupted unlock can resume without re-proving
Related to #2296 (step 4 of the read flow). Phase 2. Sub-issue of #2368.

## Problem

The `bundle_id` the reader generates for an unlock lives in memory only. Lose it and the purchase is unreachable: the reader is anonymous to the Lock Server, so there is no way to list your own purchases — the bundle id is the only handle. For a payment lock that means paying a second time for content already paid for.

The window is wider than it first looks. The reader leaves for their wallet, pays there, and may never come back to the same tab.

## Scope

Persist the `bundle_id` at `/priv/social/purchases/<lockId>.json` on the reader's homeserver, so a purchase can be picked up again at any later point. The path sits beside the replicated content, not inside it — the entry must survive the replica being deleted — and is covered by the `/priv/social/:rw` capability the app already requests.

Reading it back is what tells the app whether this lock already has a purchase in flight. That answer decides what the unlock modal shows: a Pay button when there is no entry, or the waiting state when there is.

One state needs care because the entry is written before the submission: the write can succeed and the submission still fail, leaving an entry with no verification task behind it. The server is the source of truth — when the stored id is looked up and no task exists, the modal offers Pay again with that stored id instead of waiting for a task that was never created. Getting this wrong either way is safe (resubmitting the same bundle is a replay), but the reader would be stuck on a spinner for nothing.

## Relation to #2028

#2028 was closed as not needed: durability and multi-device are already covered by replicating unlocked content to `/priv/social/unlocked/<lockId>/`. That still holds for content the reader already opened. This issue covers the window before the replica exists, which replication by definition cannot cover.

## Definition of Done

- [ ] `bundle_id` written to the reader's `/priv` **before the proof is submitted**, not after. A submission that succeeds while the entry is missing leaves a purchase nobody can reach.
- [ ] A failed write blocks the submission. Paying for something we cannot look up again is worse than not paying yet.
- [ ] An interrupted unlock resumes from the stored `bundle_id` instead of minting a new one. Resubmitting the same proof is a safe replay: it returns the existing lifecycle and does not create a second invoice.
- [ ] A new `bundle_id` is minted only when there is no entry, or when the stored one ended in `failed` or `expired` — those cannot be retried and need a fresh id.
- [ ] Reading the entry fails closed: if the app cannot tell whether one exists, it does not offer to pay.
- [ ] An entry whose id has no verification task on the server leads back to the Pay button, reusing that stored id — never to an indefinite waiting state.

## Note

The entry is kept after the content is replicated. An earlier version of this issue removed it at that point, on the grounds that the replica supersedes it. That was wrong — a replica can be deleted or lost, and without the bundle id the reader would have to pay again for content they already own. The entry is a short string; keeping it costs nothing.

### LINKED ISSUE #2368: Locks: reader — pay with Bitkit to unlock content

Taewa:
Closes #2368, closes #2297.

https://github.com/user-attachments/assets/0e8e35f7-c554-4e7a-a3eb-c6030ccf8095

A reader can now pay for a locked post with Bitkit and read it.

## What a reader can do

- Press Unlock on a paid lock and get the Pay to Unlock dialog instead of the password prompt
- Press Pay. The payment request arrives in their Bitkit wallet; the app never shows an address or an invoice
- Wait in the dialog while the payment goes through. It resumes on its own when the tab comes back from Bitkit, and offers Check again after three minutes
- See a confirmation with the amount paid, then press View Content to read the post
- Without a wallet, see the install steps for Bitkit instead of Pay

## What happens behind it

- A purchase is saved to the reader's own homeserver before the payment is sent, so it can always be found again. Reopening the dialog picks it up instead of paying twice
- A payment that failed or expired is replaced with a fresh one on the next Pay. A saved purchase that cannot be read blocks Pay rather than risk a second charge
- A payment that completed while the reader was away is finished in the background the next time the post is on screen

## Guardrails

- Nothing is sent by opening or closing the dialog. Only Pay sends a payment
- Closing during a payment asks first; the payment keeps running on the server either way
- Two tabs paying the same lock at the same moment can still create two payments (#2468)

## Testing

Needs the local Paykit stack (#2365) and the SDK built from pubky/locks#42.

Covered by tests:
- Purchase is saved before the payment is sent; a saved purchase is reused; a failed one is replaced; an unreadable one blocks Pay
- Opening never sends a payment; closing mid-payment asks first; a result arriving after close is ignored
- Polling stops on completed / failed / expired, parks after three minutes, and re-checks on tab focus
- Background recovery runs once per page load, only for paid posts with no replica, and not while the dialog is open

### DIFF
--- a/docs/locks.md
+++ b/docs/locks.md
@@ -5,8 +5,9 @@ teaser advertising it. This doc starts with the mental model — what is differe
 rest of the app — and gets more detailed the further down you read. If you know pubky-app
 but not locks, read top to bottom.
 
-Phase 1 (epic **#1998**) ships **password locks only** — payment comes later. See
-[Phase 1 & release-gate markers](#phase-1--release-gate-markers) for what is deliberately
+Phase 2 (epic **#2364**) makes locks **payable**: a creator sets a price in sats and the
+reader pays it from Bitkit. The Phase 1 password lock still exists until **#2369** deletes
+it. See [Phase 1 & marker tracking](#phase-1--marker-tracking) for what is deliberately
 temporary.
 
 ## Table of contents
@@ -28,9 +29,9 @@ temporary.
 ## What a lock is
 
 To the user: a post in the feed that looks normal — a short teaser, maybe an image — with a
-lock card on top ("Secret essay · Unlock"). Entering the password (Phase 1) reveals the
-real content in place: a post, an article, images, files. Everything the user unlocked is
-listed on their own profile under **Unlocked**.
+lock card on top ("Secret essay · Unlock ₿1,000"). Paying the price from Bitkit (or, until
+#2369, entering the password) reveals the real content in place: a post, an article, images,
+files. Everything the user unlocked is listed on their own profile under **Unlocked**.
 
 Two roles: the **creator** publishes locked content behind a public announcement; the
 **reader** unlocks and reads it.
@@ -70,10 +71,11 @@ article button while the lock switch is on (`PostInputExpandableSection`), and
 (`core/pipes/post/post.kind.ts`), which throws on those two kinds. The guard is the
 backstop for the UI rule, so a UI change can't loosen it silently.
 
-**2. Unlock (reader).** The lock card opens a password dialog. The FE submits a **proof**
-(evidence the unlock criteria are met — in Phase 1, the password) to the Lock Server,
-polls until verified, gets a short-lived credential, proxy-reads the guarded bytes with
-it — and then **replicates** them into the reader's own `/priv`.
+**2. Unlock (reader).** The lock card opens the unlock dialog for the lock's kind — Pay to
+Unlock for a payment lock, a password prompt until #2369. Either way the FE submits a
+**proof** to the Lock Server, waits until it verifies (for a payment: until the reader has
+paid in Bitkit), gets a short-lived credential, proxy-reads the guarded bytes with it — and
+then **replicates** them into the reader's own `/priv`.
 Details: [Reading a lock post](#reading-a-lock-post).
 
 **3. Read again.** Every later view skips the Lock Server entirely: the post renders from
@@ -83,11 +85,12 @@ replica also survives the creator revoking the lock. Details:
 
 ## Where the data lives
 
-| Where                                         | What                                      | Who can read it                                              |
-| --------------------------------------------- | ----------------------------------------- | ------------------------------------------------------------ |
-| creator's HS `/priv/locks.app/content/`       | the locked post + attachments (originals) | the creator; readers only via Lock Server proxy + credential |
-| creator's HS `/pub/locks.app/<lockId>.json`   | the public lock contract (`LockFile`)     | anyone                                                       |
-| reader's HS `/priv/social/unlocked/<lockId>/` | the reader's replica, written on unlock   | that reader only                                             |
+| Where                                              | What                                                                    | Who can read it                                              |
+| -------------------------------------------------- | ----------------------------------------------------------------------- | ------------------------------------------------------------ |
+| creator's HS `/priv/locks.app/content/`            | the locked post + attachments (originals)                               | the creator; readers only via Lock Server proxy + credential |
+| creator's HS `/pub/locks.app/<lockId>.json`        | the public lock contract (`LockFile`)                                   | anyone                                                       |
+| reader's HS `/priv/social/unlocked/<lockId>/`      | the reader's replica, written on unlock                                 | that reader only                                             |
+| reader's HS `/priv/social/purchases/<lockId>.json` | the purchase's bundle id, written BEFORE the proof is submitted (#2297) | that reader only                                             |
 
 ---
 
@@ -143,6 +146,10 @@ and swaps in the guarded post once it becomes readable.
 | `components/organisms/PostBody/PostBody.tsx`                       | shared text + link-embed + attachment renderer (feed post body **and** lock teaser) |
 | `components/molecules/LockedPostCard/LockedPostCard.tsx`           | lock card: title, shield graphic, Unlock control (also used by the composer)        |
 | `components/molecules/DialogUnlockContent/DialogUnlockContent.tsx` | password prompt                                                                     |
+| `components/molecules/DialogPayToUnlock/DialogPayToUnlock.tsx`     | Pay to Unlock modal (checking / pay / install / waiting / blocked)                  |
+| `hooks/usePayToUnlock/usePayToUnlock.ts`                           | the payment state machine: bundle-id routing, submit, polling, stall/resume         |
+| `hooks/usePurchasedLocks/usePurchasedLocks.ts`                     | one listing of the reader's purchases, shared by every lock post                    |
+| `hooks/usePurchaseResume/usePurchaseResume.ts`                     | finishes a paid purchase whose content never landed, without interaction            |
 
 ## Reading a lock post
 
@@ -154,18 +161,22 @@ LockedPostContent
   ├─ useLockFile(lock)                       → lock.json (LockFile | null)
   │    └─ LocksController.fetchLockFile      → LocksApplication → LocksService.readContentLock
   │
-  └─ useUnlockedContent(lock, lockFile, authorId)
-       ├─ 1) already unlocked as a reader → fetchReplicatedContent  (my HS /priv copy)
-       ├─ 2) my own post (a == b)         → fetchOwnContent         (my HS /priv original)
-       └─ 3) neither → lock card → DialogUnlockContent → unlock (below)
+  ├─ useUnlockedContent(lock, lockFile, authorId)
+  │    ├─ 1) already unlocked as a reader → fetchReplicatedContent  (my HS /priv copy)
+  │    ├─ 2) my own post (a == b)         → fetchOwnContent         (my HS /priv original)
+  │    └─ 3) neither → lock card → unlock (below)
+  │             ├─ password lock → DialogUnlockContent
+  │             └─ payment lock  → DialogPayToUnlock (sign-in required first)
+  └─ 4) saved purchase, no replica → usePurchaseResume → fetchPaidContentIfCompleted
 ```
 
 `a == b` is team shorthand: **a** = the announcement's author account, **b** = the account
 that owns the lock (Lock Server side). Phase 1 assumes they are the same person, and
 own-content reads rely on it.
 
-**Unlock** (`LocksApplication.unlockContent`) — steps 1–4 run against the Lock Server and
-need no pubky.app session; step 5 writes to the reader's homeserver with it:
+**Password unlock** (`LocksApplication.unlockContent`) — one blocking call; steps 1–4 run
+against the Lock Server and need no pubky.app session; step 5 writes to the reader's
+homeserver with it. Goes away with #2369:
 
 1. `submitProofBundle` — proof built from `lock.json` by `LockProofBundler`
 2. `lookupVerificationTask` — poll every 1.5s, max 40 attempts, until `completed`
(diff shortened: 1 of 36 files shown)


src/core/application/locks/locks.ts
```diff
@@ -117,6 +130,126 @@ export class LocksApplication {
...
+      bundleId = await LocksService.generateBundleId();
+      // Saved BEFORE the proof is submitted: a submission that succeeds while the file is missing
+      // leaves a purchase nobody can reach.
+      await HomeserverService.putBlob({
+        url: GuardedContentParser.purchaseUrl(readerPubky, this.requireLockId(lockUrl, 'startPayment')),
+        blob: new TextEncoder().encode(GuardedContentParser.buildPurchaseFile(bundleId)),
+      });
+    }
+    const bundle = LockProofBundler.buildPayment(lockFile, lockUrl, bundleId, readerPubky);
+    const task = await LocksService.submitProof(bundle);
```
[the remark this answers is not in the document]
Taewa: Known and tracked in #2468. A tab lock would only cover tabs in one browser; two devices can still race because the homeserver has no conditional write, so the fix belongs in that issue, not here.

This PR targets epic/locks-v1, which stays off dev until the remaining security and feature work is done.

src/components/organisms/LockedPostContent/LockedPostContent.tsx
```diff
@@ -143,6 +190,21 @@ export function LockedPostContent({
         loading={isUnlocking}
         error={unlockError}
       />
+      {isPaymentLock && priceSats && (
+        <DialogPayToUnlock
+          open={isPayOpen}
+          onOpenChange={setIsPayOpen}
+          lockTitle={lockContent.lock_title}
+          authorId={authorId}
+          priceSats={priceSats}
```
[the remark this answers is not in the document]
Taewa: A follow-up issue has been created but this one is more about a discussion .

## Why it may not be a problem

- X cannot redirect money to themselves. The recipient is pinned server-side to Y.
- Y is paid for their own content, which is the intended outcome.
- Phase 2 explicitly allows the Locks account and the posting account to differ (a≠b), so requiring "post author == lock creator" would break a supported setup.
 


https://github.com/pubky/pubky-app/issues/2482

src/hooks/usePurchasedLocks/usePurchasedLocks.ts
```diff
@@ -0,0 +1,76 @@
...
+      // Already reported by the Err factory. Degrade rather than block rendering: the unlock
+      // modal re-reads the saved bundle id itself before it ever offers to pay.
+      .catch(() => {
+        // Let the next mount try again, unless markPurchased already chained onto this listing.
+        if (purchasedIdsByReader?.ids === ids) purchasedIdsByReader = null;
+        return new Set<string>();
+      });
+    purchasedIdsByReader = { readerPubky, ids };
+  }
+  return purchasedIdsByReader.ids;
```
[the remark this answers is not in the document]
Taewa: Intended. The listing is read once per page load so the feed does not ask per post. A purchase made in another tab or device is not picked up by the background recovery until reload, but pressing Unlock re-reads the homeserver and opens the content, so nothing is lost or blocked.

Keeping one device in sync with changes another device wrote to the homeserver is a general problem, not a locks one. It will be handled separately, outside this PR.

src/core/controllers/locks/locks.ts
```diff
@@ -174,11 +180,46 @@ export class LocksController {
...
 
-  /** Reader unlock: submit the proof for a resolved lock file, verify, and return the access credential. */
+  /** Password unlock (blocking submit+poll). TODO:[Locks] #2369 — deleted with the password UI. */
   static unlock(params: TUnlockContentParams): Promise<TUnlockResult> {
     return LocksApplication.unlockContent(params);
   }
 
+  /** Whether the reader's wallet has published a Paykit receiver. */
+  static hasPaykitReceiver(readerPubky: string): Promise<boolean> {
+    return LocksApplication.hasPaykitReceiver(readerPubky);
```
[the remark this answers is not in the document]
Taewa: Kept as is. docs/locks.md records the exception: locks has no local-first write, so its server queries and actions use has* / start* instead of fetch* / commit*. I removed the sentence that tied this to #2296; that issue caches immutable lock files and replicas, and the purchase marker will stay a network read on purpose because it decides whether a payment is reused.

secondl1ght (commented):
Four user-facing issues in the payment flow below, each reachable in the normal Pay → Bitkit → return sequence and worth fixing before merge. None of them can double-charge a reader. I reproduced each with a throwaway test. I have a handful of smaller notes (cleanups, copy) that I've left out to keep this focused — happy to share them separately if useful.

src/hooks/usePayToUnlock/usePayToUnlock.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    // The reader pays in Bitkit, so this tab is backgrounded (timers frozen) for most of the
+    // wait. Returning is the real signal: look up immediately and, if the loop had parked,
+    // grant a fresh wall-clock window so it truly resumes.
+    const onVisible = () => {
+      if (document.visibilityState !== 'visible' || generation.current !== gen) return;
+      if (timer !== null) window.clearTimeout(timer);
+      windowStartedAt = Date.now();
+      setIsStalled(false);
+      void poll();
```
secondl1ght, line 172: `onVisible` starts a `poll()` even when a timer-fired poll is already awaiting the server. `timer` still holds the fired id at that point, so the `clearTimeout` above is a no-op and can't detect the in-flight one. Both polls then `schedule()`, `timer` keeps only the last id, and the loop has forked into two chains that `stop()` can only half-tear-down.

Reproduced: 2 status lookups per interval for the rest of the wait; when both chains see `completed` in flight, `finish()` runs twice (two credentials, two content downloads) and the screen flips `paid` → `waiting` → `paid`; on `failed`, two toasts. The trigger is the normal one — the tab comes back from Bitkit right as the throttled timer fires.

`recheck()` has the same shape: `stop()` doesn't cancel a poll that's already awaiting, and on `pending` it re-schedules itself.

Suggest a per-loop `let active = true` flipped in `stop()` and checked after each `await`, plus skipping `onVisible`'s poll while one is in flight.

src/components/organisms/LockedPostContent/LockedPostContent.tsx
```diff
@@ -54,26 +62,73 @@ export function LockedPostContent({
...
 
   if (!lockContent) return null;
 
+  // No lock file (still loading, or its fetch failed) → nothing to unlock against.
+  const handleUnlock = !lockFile
+    ? undefined
+    : isPaymentLock
+      ? // Paying needs the reader's pubky (it is the payment-request delivery address),
+        // so a signed-out reader gets the sign-in dialog instead.
+        () => requireAuth(() => setIsPayOpen(true))
```
secondl1ght, line 117: For a signed-out reader this leaves the card stuck. lock.json is a public read, so Unlock is enabled; the card slides the button over the price, then `requireAuth` opens the sign-in dialog and never sets `isPayOpen`. `LockedPostCard`'s only reset is keyed on `unlockOpen`, which never changed, so the button stays parked over the price after sign-in is dismissed. Pre-PR the handler always opened a dialog, so this couldn't happen.

Either check auth before the slide, or have the card reset `slideX` when `onUnlock` returns and `unlockOpen` is still false. The test mocks `requireAuth` as a pass-through, so this branch is untested.

src/components/molecules/DialogPayToUnlock/DialogPayToUnlock.tsx
```diff
@@ -0,0 +1,268 @@
...
+    if (!next && (stage === 'waiting' || isSubmitting)) {
+      setIsConfirmingClose(true);
+      return;
+    }
+    onOpenChange(next);
+  };
+
+  const confirmClose = () => {
+    setIsConfirmingClose(false);
+    onOpenChange(false);
```
secondl1ght, line 84: `confirmClose` bypasses the `stage === 'paid'` branch in `handleOpenChange`. Polling keeps running under the "still running" prompt, so if the payment completes while the reader is reading it, "Close anyway" closes without `viewContent()`, the content already in `completedContent` is dropped, and `usePurchaseResume` immediately mints a new credential and downloads the same post + attachments — the duplicate the `handleOpenChange` comment says it prevents.

Route `confirmClose` through `handleOpenChange(false)`, or dismiss the prompt when the stage leaves `waiting`.

src/components/molecules/DialogPayToUnlock/DialogPayToUnlock.tsx
```diff
@@ -0,0 +1,268 @@
...
+            )}
+
+            {stage === 'waiting' && (
+              <Container overrideDefaults className="flex flex-col items-center gap-3 py-4">
+                {/* Parked, not failed: the purchase is alive, so the reader gets a way back to it
+                  rather than a spinner that never resolves. */}
+                {isStalled ? (
+                  <>
+                    <Typography className="text-center text-base text-secondary-foreground">
+                      {'Still waiting for the payment. Pay in Bitkit, then check again.'}
```
secondl1ght, line 163: This copy is also what a reader sees after their payment has *completed* and only the content download failed: `finish()`'s catch leaves `stage === 'waiting'` and sets `isStalled`. The toast says "your payment went through", but the persistent screen tells them to go pay in Bitkit. Worth a distinct stage (or a `finishFailed` flag) with copy like "Payment received — the content couldn't be opened. Check again."

src/hooks/usePayToUnlock/usePayToUnlock.ts
Taewa: Code updated and tested!

src/components/organisms/LockedPostContent/LockedPostContent.tsx
Taewa: Code updated and tested!

src/components/molecules/DialogPayToUnlock/DialogPayToUnlock.tsx
Taewa: Code updated and tested!

src/components/molecules/DialogPayToUnlock/DialogPayToUnlock.tsx
Taewa: Code updated and tested!

src/components/molecules/DialogPayToUnlock/DialogPayToUnlock.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+  return (
+    <Dialog open={open} onOpenChange={handleOpenChange}>
+      <DialogContent
+        className="w-full max-w-md rounded-xl border-x-0 border-y border-brand bg-card sm:max-w-xl"
+        // Clicks bubble through the portal to the post card, which would navigate to the post.
+        onClick={(e) => e.stopPropagation()}
+      >
+        <DialogHeader>
+          <DialogTitle>{stage === 'paid' ? 'Unlocked' : 'Pay to Unlock'}</DialogTitle>
```
secondl1ght, line 95: `unopened` now says "PAYMENT RECEIVED" in the body, but the title here still reads "Pay to Unlock". Since `isPaid` already covers both stages, `isPaid ? 'Unlocked' : 'Pay to Unlock'` would keep the header and body telling the same story.

docs/locks.md
```diff
@@ -143,6 +146,10 @@ and swaps in the guarded post once it becomes readable.
 | `components/organisms/PostBody/PostBody.tsx`                       | shared text + link-embed + attachment renderer (feed post body **and** lock teaser) |
 | `components/molecules/LockedPostCard/LockedPostCard.tsx`           | lock card: title, shield graphic, Unlock control (also used by the composer)        |
 | `components/molecules/DialogUnlockContent/DialogUnlockContent.tsx` | password prompt                                                                     |
+| `components/molecules/DialogPayToUnlock/DialogPayToUnlock.tsx`     | Pay to Unlock modal (checking / pay / install / waiting / blocked)                  |
```
secondl1ght, line 149: This list is missing `paid` and the new `unopened` stage. Worth updating so the doc matches `TPayToUnlockStage`.

secondl1ght (approved):
All four fixes check out — each one lands at the right layer and the new tests exercise the exact trigger (visibility mid-lookup, offsetLeft-backed slide, stage flipping under the prompt, the `unopened` path). Lint and the touched suites are green.

Two non-blocking follow-ups inline; fine to fold into a later commit.

PR #2483, opened 2026-09-05
feat: vibe session consumer mode for approved vibes

BitcoinErrorLog:
## Summary

pubky-app forks deployed as approved vibes on `<slug>.vibes.pubky.app` restore the visitor's existing `pubky.app` session silently, then run the unchanged `restoreSession` path.

Obtain order:

1. Board hand-off `#s=<sessionExport>` (consumed on the first client pass)
2. Hidden iframe to `${NEXT_PUBLIC_VIBE_SESSION_BRIDGE_ORIGIN}/session-bridge` if there is no persist, or persist failed with a definitive auth error

Canonical `pubky.app` is behaviourally unchanged when `NEXT_PUBLIC_VIBE_SESSION_BRIDGE_ORIGIN` is unset. That is guaranteed by a build-time switch, not a runtime flag:

- `isVibeSessionConsumerEnabled()` is `Boolean(Env.NEXT_PUBLIC_VIBE_SESSION_BRIDGE_ORIGIN)`. Empty / omitted parses as `undefined` and keeps consumer mode off.
- `RouteGuardProvider` still restores only when a persisted `sessionExport` exists; the extra “restore with no persist” path runs only when consumer mode is on.
- `AuthApplication.restorePersistedSession` obtains a fragment/bridge export only when consumer mode is on. The existing persist → `HomeserverService.restoreSession` loop is otherwise unchanged.
- `#s=` is stripped on every first client pass (even when consumer mode is off) so a leftover fragment cannot leak into routing or telemetry. That strip does not start restore.

## Env vars (vibe makers)

Set these on the **vibe fork's Vercel project** (build-time `NEXT_PUBLIC_*`; they are inlined into that artifact). They are not `PUBKY_RUNTIME_*`.

| Variable | Required | Value |
| --- | --- | --- |
| `NEXT_PUBLIC_VIBE_SESSION_BRIDGE_ORIGIN` | Yes, to turn consumer mode on | Exact origin of the bridge host, typically `https://pubky.app`. `http://localhost:<port>` only when `NODE_ENV !== 'production'`. Invalid values fail env parse. |
| `NEXT_PUBLIC_VIBE_ID` | No | Informational slug. Not used as an origin check. |

Canonical `pubky.app` leaves both unset.

## Security invariants

- `#s=` is consumed in `instrumentation-client.ts` via `consumeFragmentSessionExport()` **before** Sentry init and before any client routing or network that depends on auth.
- Bridge replies are accepted only when `event.origin === bridgeOrigin`, `event.source === iframe.contentWindow`, and `data.v === 1`. Never `'*'`.
- `sessionExport` is public metadata (pubky + capabilities). It cannot fix identity: the homeserver `HttpOnly` cookie decides. The consumer never reads or copies that cookie.
- A forged `#s=` fails `restoreSession` and is never persisted. A forged fragment while logout-suppressed does not lift the flag or fall through to the bridge.

## Logout

- Logout on a vibe signs out the **shared homeserver session** (cookie used by `pubky.app` and every other same-site vibe in this browser). If homeserver `signout()` fails, local state is still cleared.
- After an explicit sign-out, a per-tab `sessionStorage` flag (`pubky.vibeSession.autoRestoreSuppressed`) is set **before any await**, and any in-flight bridge request is aborted. That flag disables only the **bridge** leg. A `#s=` hand-off still attempts fragment restore; if that fails while suppressed, it does not fall back to the bridge.
- Suppression is cleared only by a successful `authStore.init` (explicit sign-in or a completed restore). Reload of the same tab keeps the flag; a new tab has no flag and may re-mirror via the bridge.
- When persist restore fails transiently and the bridge/fragment obtain also fails, Application returns `{ status: 'deferred' }`. The Controller sets `sessionRestoreDeferred`. `useAuthStatus` shows **unauthenticated UI** (not a spinner) so the user can sign in or retry. The persisted export is not deleted in that case.

`AuthController.restorePersistedSession` returns `{ status: 'restored' | 'signed-out' | 'deferred' }` (not a boolean) so logout can still clean up on `deferred`.

## ADR

[ADR 0019: Vibe Session Consumer](docs/adr/0019-vibe-session-consumer.md). Upstream already has ADR 0017 (runtime-config) and ADR 0018 (runtime Sentry); this stays 0019. `docs/architecture.md` table lists 0017–0019.

## Tests

Targeted suite after rebase onto `upstream/dev` (`npx vitest run` on the paths below): **194 passed / 194** (10 files). Six more than the pre-rebase 188 because `upstream/dev` added tests in the same auth / RouteGuard / env trees.

| File | Passed |
| --- | --- |
| `auto-restore.test.ts` | 4 |
| `bridge.test.ts` | 16 |
| `expired.test.ts` | 3 |
| `fragment.test.ts` | 10 |
| `env.test.ts` | 12 |
| `application/auth.test.ts` | 43 |
| `controllers/auth.test.ts` | 53 |
| `RouteGuardProvider.test.tsx` | 25 |
| `auth.store.test.ts` | 14 |
| `useAuthStatus.test.tsx` | 14 |

Also green on this branch: `npm run lint`, `npm run format:check`, `npx tsc --noEmit` (0 errors), `npx next build --webpack`.

## Companion PRs

- **`feat/session-bridge` in pubky-app** — hosts `/session-bridge` and the postMessage contract. Both this branch and that one touch `src/libs/env/env.ts`, `.prettierignore`, and `eslint.config.mjs`. Merge order does not matter; the second will need a trivial rebase.
- **`pubky/vibes`** — board `#s=` hand-off, registry `hosted.sso`, and `@vibes/vibe-session`. This consumer ports the helper into `src/libs/vibe-session/` rather than depending on that package.

## External review

The consumer path (fragment consume, bridge `postMessage` accept rules, public-export / cookie identity split, logout suppression, deferred restore) went through multiple external audit rounds before this rebase. Conflict resolutions on this rebase kept those invariants (`this.cancelModerationFollow()` from upstream plus suppress-then-abort on logout; toast import moved to `@/molecules/Toaster/toast`; ADR remains 0019). Re-audit any later change to fragment consume, bridge accept rules, or the logout suppression flag before relying on it in production.

## Test plan

- [ ] Canonical deploy with both env vars unset: sign-in / restore / logout unchanged; a `#s=` URL is stripped and does not restore
- [ ] Vibe fork with `NEXT_PUBLIC_VIBE_SESSION_BRIDGE_ORIGIN=https://pubky.app`: visitor already signed into pubky.app is restored via `#s=` when arriving from the board
- [ ] Same vibe, no fragment: hidden iframe to `/session-bridge` restores when the visitor is signed in on pubky.app
- [ ] Forged `#s=bogus`: restore fails, export is not persisted, no bridge fallback while suppressed
- [ ] Logout on the vibe: shared homeserver session ends; this tab does not auto-restore from the bridge; a new tab may
- [ ] Deferred persist (homeserver blip): unauthenticated UI, not a loading spinner; persisted export kept


talosmachina (commented):
Minor, for awareness rather than a change request: consumer mode's silent restore only completes when the browser holds a homeserver cookie that is same-site with pubky.app, i.e. users on the flagship homeserver. Users on other homeservers get a clean signed-out state rather than an error, which is the right fallback, but it means the sign-in friction this removes only applies to flagship users. Worth keeping in mind when measuring whether this actually fixes the vibe sign-in drop-off.

Also minor: `restorePersistedSession` failure paths call `cleanupLocalState()`, which resets the whole auth store including `isRestoringSession`, so a failed restore does not wedge the RouteGuard effect into a retry loop. That behavior is correct but non-obvious; flagging it so the dependency list on the RouteGuard effect is not "simplified" later.

Nuance, not a defect: `expired.ts` treats any `AuthenticationError` as expiry. That matches the current wasm mapping, and the doc comment cites the exact rust error mapping, so fine as is.

src/core/stores/auth/auth.store.ts
```diff
@@ -28,7 +29,7 @@ export const useAuthStore = create<AuthStore>()(
         onRehydrateStorage: (state) => (rehydratedState) => {
           const resolvedState = rehydratedState ?? state;
           resolvedState.setHasHydrated(true);
-          if (rehydratedState?.sessionExport) {
+          if (rehydratedState?.sessionExport || isVibeSessionConsumerEnabled()) {
```
talosmachina: **[P1] Clear `isRestoringSession` when restoration is skipped**

With consumer mode on, rehydration sets `isRestoringSession = true` even when there is no persisted `sessionExport`, and the only thing that clears the flag is `restorePersistedSession` running to completion. But `RouteGuardProvider` skips the restore entirely when auto-restore is suppressed (i.e. after a logout in this tab) and no `#s=` fragment is pending, and no other path clears the flag, so after logout + reload on a vibe, `useAuthStatus` is stuck in the loading branch forever (infinite spinner). The deferred case got the `sessionRestoreDeferred` escape hatch for exactly this class of problem; the skipped-restore path did not. The existing RouteGuard test "does not restore after logout when consumer auto-restore is suppressed and nothing is persisted" passes only because it mocks `useAuthStatus`. Fix: don't set the flag on rehydrate when suppression is set and nothing is persisted (mirroring the RouteGuard gate), or clear `isRestoringSession` on the skip path.

Verified by: reproduced against this PR's worktree (`npx vitest run`, 1 passed) — consumer-on rehydrate + suppression leaves `isRestoringSession` stuck at `true` with no code path to clear it; cross-checked every `setIsRestoringSession(false)` call site (auth.actions.ts:57, auth.ts:92, auth.ts:103), none is reachable on this path.

talosmachina (commented):
**1 finding** (1 P1). Adds consumer mode for approved vibe forks: a build-time bridge origin enables silent session mirroring from the canonical pubky.app host via a same-origin hidden iframe plus a `#s=` fragment hand-off, wired into the existing session-restore path. Supersedes my earlier review from this morning, which was general commentary and made no anchored findings.

**[P1] in `src/core/stores/auth/auth.store.ts:32`** — consumer-mode rehydration sets `isRestoringSession` with no guarantee anything will clear it: after logout and reload on a vibe, the suppression gate makes `RouteGuardProvider` skip the restore that would clear the flag, leaving an infinite loading spinner. Fix is small: gate the rehydrate flag the same way the RouteGuard gate does, or clear it on the skip path. Verified with a reproduction test, not inspection alone.

The overall design is careful: origin/source/version checks on the bridge reply, fragment stripped unconditionally at client boot, export kept on transient failures with an explicit `deferred` terminal, and the logout-suppression flag cleared only by a successful `authStore.init`. Layering (Controller -> Application -> HomeserverService) is respected throughout, and the ADR documents the trade-offs honestly.

<details><summary>What I checked, and candidates I ruled out</summary>

**Read in full** from a worktree of the PR head: `auth.store.ts`, `auth.actions.ts`, `auth.ts`, `RouteGuardProvider.tsx`, the vibe session bridge and suppression modules
**Call sites traced**: `restorePersistedSession`, every `setIsRestoringSession(false)` call (auth.actions.ts:57, auth.ts:92, auth.ts:103)
**CI**: green; the RouteGuard test on the skip path passes only because it mocks `useAuthStatus`, hiding the stuck flag
**Method**: read the full diff of the core files from the PR worktree, refuted each candidate against the surrounding code, and reproduced the surviving finding with a failing-behaviour vitest run against the worktree

**Ruled out**
- *Bridge reply spoofing* — origin/source/version checks on the bridge reply cover it.
- *Fragment leakage* — the `#s=` fragment is stripped unconditionally at client boot.
- *Transient failure data loss* — the export is kept on transient failures with an explicit `deferred` terminal.
- *Suppression flag leak on logout* — it is cleared only by a successful `authStore.init`, which is the intended design.

</details>

secondl1ght (changes requested):
Thanks for the thorough write-up and test coverage on this. The core idea (silent same-site SSO through an origin-bound iframe that hands out public session metadata while the HttpOnly cookie does the real authentication) is sound and the bridge accept rules are solid. I went through the consumer path end to end and reproduced the items below with scratch tests against the real store, controller, and SDK. Requesting changes for five items, followed by a few non-blocking notes.

## Requested changes

### 1. Drop the `#s=` fragment hand-off entirely

The fragment is the source of one security issue and three bugs, and the bridge already covers every arrival path (the vibe and pubky.app are same-site, so storage is not partitioned).

**Security (login CSRF / session fixation).** The ADR relies on "a forged export fails `restoreSession`", which holds for a key on the official homeserver but not for a *genuine* export of an attacker-owned account on an attacker-run homeserver. The SDK revalidates at `https://_pubky.<pubky-in-export>/session`, choosing the homeserver from the pubky inside the export via PKARR, and the client is built with only PKARR relays (`homeserver.ts:88`). `assertUserHomeserverAllowed` returns immediately outside staging (`homeserver.ts:140-142`), so production vibes accept any homeserver. The fragment is consumed from any referrer (`instrumentation-client.ts:12`), tried before the bridge and before the suppression check (`application/auth/auth.ts:135-143`), and the controller then trusts the export-derived identity, wipes local state, and persists the attacker's export (`controllers/auth/auth.ts:114-136`). Result: a link like `https://<vibe>/#s=<attacker-export>` signs any visitor into the attacker's account, and everything they enter on the vibe lands on the attacker's homeserver. The victim's own cookie never leaves, so it is not takeover of their account, but for vibes that collect anything sensitive this is serious. The bridge does not have this problem: its export comes from pubky.app's own session store, bound by the origin and source checks.

**Bugs that disappear with it:**
- A fragment that loses to a successful persisted restore is never taken (`application/auth/auth.ts:119` returns before line 135), stays cached for the page lifetime, and `hasPendingFragmentSessionExport()` re-arms restore after logout despite suppression (`should-restore.ts:19`). Reproduced: after `suppressVibeSessionAutoRestore()`, `shouldAttemptSessionRestore(null)` was true and the next restore signed in from the cached fragment.
- `new URLSearchParams(raw)` (`fragment.ts:8`) decodes `+` as a space, and `Session.export()` in the installed SDK (0.8.0) is standard base64. Verified: the SDK rejects the decoded value with `Invalid symbol 32`. Every test input is wrapped in `encodeURIComponent`, which masks it.

### 2. Isolate the consumer path from canonical auth

Consumer branches are currently threaded through the shared restore (`application/auth/auth.ts`), `restorePersistedSession` and `logout` (`controllers/auth/auth.ts`), the store rehydrate hook, and `RouteGuardProvider`, so canonical pubky.app is protected by conditionals rather than by isolation. Several side effects reached shared code: `reset()` now preserves `isRestoringSession` for every caller (`auth.actions.ts:43`), `init()` performs a sessionStorage side effect (`auth.actions.ts:22`), and the Application layer now mounts an iframe and touches history and sessionStorage, which contradicts the ADR's "pure lib" statement and `docs/architecture.md`'s "IO only in Services".

Please put the whole consumer flow behind one `isVibeSessionConsumerEnabled()` check at the two entry points (route guard and logout) that selects a separate `restoreVibeSession` path, with the bridge behind a service module, and leave the canonical `restorePersistedSession` byte-identical to `dev`. The store rehydrate gate is the one place that legitimately stays shared. Main auth must not be able to break because of the vibe path.

### 3. Treat a new-identity restore as a sign-in

The `!sameIdentity` branch (`controllers/auth/auth.ts:121-137`) wipes local state, calls `userIsSignedUp`, and `init()`s, but never runs `completeAuthenticatedSession` / `hydrateMeImAlive`. A first bridge sign-in on a vibe therefore ends AUTHENTICATED with default settings instead of homeserver settings, an empty muted list, no custom feeds, no notification cursor, and no moderation follow, and since `init` persists the export every later load takes the same-identity path, so it stays that way. Two related issues in the same branch:
- The Application `finally` clears `isRestoringSession` before returning `restored` (`application/auth/auth.ts:99`) and the controller only re-sets it two awaits later (`:125`). On a fresh vibe nothing else holds loading, so `useAuthStatus` settles `UNAUTHENTICATED` for one flush. Reproduced with the real `RouteGuardProvider`: `router.push('/')` fired mid-restore, then the user landed authenticated on the wrong route.
- `userIsSignedUp` throws on any non-404 and is awaited with no retry after cleanup already ran (`:131`), so one 502 discards a valid restore with no in-page retry.

Calling `completeAuthenticatedSession({ session })` for the new-identity case, and holding `isRestoringSession` until `init` lands, fixes all three.

### 4. Do not tear down state for a visitor who never had an identity

With consumer mode on, every full page load with no session runs the restore, and a bridge `none` reply ends in `cleanupLocalState()` (`controllers/auth/auth.ts:146`), which resets the onboarding and settings stores, clears cookies, and clears IndexedDB. Reproduced: a visitor who generated keys on `/onboarding/pubky` and refreshed on `/onboarding/backup` lost `secretKey`, `mnemonic`, and `inviteCode`. Pre-PR the route guard returned early without a persisted export, so this never fired. Only clean up when there was a prior identity (`sessionExport` or `currentUserPubky` non-null).

### 5. Make vibe logout local-only

The PR documents that logout on a vibe signs out the shared homeserver session, i.e. logging out of one vibe logs you out of pubky.app and every other vibe in the browser, including any with unsaved work. That is a surprising blast radius for what is meant to be lightweight auth, and per-vibe confirm dialogs would be fragile. Please make vibe logout local-only:
- Clear local state and set the suppression flag; never call homeserver signout from a vibe on any path. That includes the wrong-environment branch in `restoreSessionFromExport` (`application/auth/auth.ts:186`), which currently signs out bridge-sourced sessions too.
- Scope the suppression flag to the site (localStorage) rather than the tab, so a new tab does not silently re-sign-in after the user chose to log out of that vibe.
- Set suppression on the `/logout` route even when nothing is persisted. Today `Logout.tsx:44` short-circuits to "signed out" without calling `logout()`, so on a vibe the route guard's bridge restore signs the user in underneath the success screen.
- Add an explicit "Continue with pubky.app" action on the vibe sign-in surface that clears suppression and runs the bridge on demand. Without it, a logged-out user's only way back is key sign-in.

This also removes the restore-then-signout path on fresh loads, and with it the bug where `init()` (`auth.actions.ts:22`) clears the suppression flag that `logout()` just set (`controllers/auth/auth.ts:449`). Reproduced: after `logout()` completed, `isVibeSessionAutoRestoreSuppressed()` was false and `shouldAttemptSessionRestore(null)` was true, so the bridge re-ran; if the homeserver signout had failed the user would have been signed back in. The existing test "does not re-restore from the bridge after consumer logout" mocks Application as `signed-out` with a `vi.fn()` `init`, so it cannot catch this.

## Non-blocking

- **Retry budget.** Each leg gets the full 10 attempts × 3 s, sequentially, so an unreachable homeserver means up to 72 s of spinner (27 s pre-PR). Give the bridge-sourced export a single attempt once the persisted leg has exhausted its budget.
- **Malformed persisted export loops forever.** The SDK reports it as `RequestError` with no status code, `handleError` maps that to a retryable server error, and consumer mode keeps the export as `deferred`, so every load spends 27 s before showing unauthenticated UI. Map no-status SDK errors to a non-retryable validation error and treat them as definitive.
- **`expired.ts` is unreachable.** Every error reaching `isDefinitiveSessionAuthFailure` is already an `AppError` (name `'AppError'`), so `isPubkyExpiredError` can never match; the file also documents SDK 0.6.0 while 0.8.0 is installed. Delete it and its tests, and drop the branch at `application/auth/auth.ts:45`.
- Please update ADR 0019 to match the final design (no fragment, local-only logout, and the IO placement).

Happy to re-review once these land.

PR #2484, opened 2026-09-05
feat(session-bridge): add /session-bridge postMessage route for approved vibes

BitcoinErrorLog:
## Summary

Adds an isolated `/session-bridge` App Router route that replies with the public `sessionExport` to allowlisted `https://*.vibes.pubky.app` origins via `postMessage`.

The route lives in its own root layout (`src/app/(bridge)/session-bridge/`) so it does not boot Dexie, coordinators, fonts, analytics, or store hydration. The rest of the app moves under `src/app/(main)/` (including the current `(feeds)`, collections, invite, and sentry-test trees) and keeps the existing app chrome.

Route-only headers on `/session-bridge`:

- `Content-Security-Policy: frame-ancestors` built from the same allowlist as the runtime origin check
- `Referrer-Policy: no-referrer`
- `X-Robots-Tag: noindex`

There is no global `X-Frame-Options` or `frame-ancestors`. This header set applies only to `/session-bridge`.

## Why

Vibes served on lent `<slug>.vibes.pubky.app` hostnames are same-site with `pubky.app`, so the browser already attaches the homeserver `HttpOnly` cookie to homeserver requests. Those vibes still need the public `sessionExport` (`session.export()` — pubky + capabilities) to construct a session handle without a second sign-in.

This route lets an approved vibe learn that public export. It does not move the cookie, and it does not create a second grant flow.

Those hostnames are not self-service. Each approved vibe gets an explicit DNS record. There is no wildcard DNS and no user-generated `*.pubky.app` name.

## Message contract

- Consumer → bridge: `{ type: 'pubky-session-request', v: 1 }`
- Bridge → consumer, signed in: `{ type: 'pubky-session', v: 1, sessionExport: string }`
- Bridge → consumer, signed out: `{ type: 'pubky-session-none', v: 1 }`

The bridge replies only to `event.origin` values in the allowlist, and only via `event.source.postMessage(reply, event.origin)` (never `'*'`). Messages from non-allowlisted origins, from `event.source === null`, or with any other `type`/`v` are ignored with no reply and no payload logging.

If the page is not in an iframe, it only shows a short human-readable sentence.

While embedded, after a valid request, the bridge remembers the most recent `event.source` + `event.origin` pair and, on `storage` events for `auth-store`, pushes an unsolicited `pubky-session` or `pubky-session-none` to that pair only.

An inline pre-hydration listener queues requests until the React handler attaches. Without that script, requests that arrive before `useLayoutEffect` are dropped.

## Security invariants

- Origin + source validation: reply only to allowlisted `event.origin` via `event.source`.
- No secrets in the export: `sessionExport` is public `SessionInfo`. A forged export fails `restoreSession`.
- The homeserver cookie never leaves the homeserver. This page never reads or exposes the `HttpOnly` cookie.
- SessionExport disclosure to approved vibes is a deliberate privacy consequence: every approved `*.vibes.pubky.app` vibe can learn a visitor's pubky and capabilities on visit. That is bounded by per-vibe approval plus explicit DNS records.
- XSS on an allowlisted vibe can drive the shared session (authenticated homeserver calls) but cannot read the cookie.
- CSP `frame-ancestors` wildcards are multi-label; `isAllowedBridgeOrigin` is the authoritative single-label gate for data. A deep subdomain can at most load an inert frame.

## Environment (Vercel / production)

Set `NEXT_PUBLIC_SESSION_BRIDGE_ALLOWED_ORIGINS` on the production (and staging) pubky-app deployment.

It is a comma-separated list of exact `https` origins or single-label wildcards (`https://*.vibes.pubky.app`). Parsing lives in `allowlist.ts` so Env validation and `next.config.ts` `headers()` stay aligned. Invalid entries fail the build.

Default when the variable is unset, empty, or whitespace-only:

- Production (`NODE_ENV=production`): `https://vibes.pubky.app`, `https://*.vibes.pubky.app`, `https://shop.pubky.app` (team-operated marketplace, exact origin)
- Non-production: the same three plus `https://vibes.staging.pubky.app`, `https://*.vibes.staging.pubky.app`, `http://localhost:3000`

Staging pubky.app deployments are `NODE_ENV=production` builds, so they must set the variable explicitly to the staging board origins (see `docs/environment.md`).

Loopback (`localhost`, `127.0.0.1`, `[::1]`) is allowed only when listed explicitly, and only as `http` on those hosts. Production loopback requires an explicit env opt-in.

This value is build-intrinsic (CSP `frame-ancestors` is emitted from `next.config.ts`). Changing the allowlist requires a redeploy.

Suggested production value (same as the default):

```
NEXT_PUBLIC_SESSION_BRIDGE_ALLOWED_ORIGINS=https://vibes.pubky.app,https://*.vibes.pubky.app,https://shop.pubky.app
```

Suggested staging value:

```
NEXT_PUBLIC_SESSION_BRIDGE_ALLOWED_ORIGINS=https://vibes.staging.pubky.app,https://*.vibes.staging.pubky.app
```

## ADR

[ADR 0020 — Session bridge for approved vibes](docs/adr/0020-session-bridge.md)

Numbered 0020 because upstream `dev` already has ADR 0017 (runtime config injection) and ADR 0018 (runtime Sentry / source maps). ADR 0019 is reserved by the companion consumer branch.

## Test plan

Unit (`vitest run --project unit` on the paths this branch touches):

- 28 files / 342 tests passed
- `src/libs/session-bridge`: `allowlist.test.ts`, `handler.test.ts`, `next-config-headers.test.ts`, `persisted-session.test.ts`
- `src/app/(bridge)/session-bridge`: `SessionBridgeClient.test.tsx`
- `src/libs/env`: `env.test.ts`
- Plus the profile import-path tests updated for the `(main)` move: `ProfilePageFilterBar.test.tsx`, `ProfilePageMobileMenu.test.tsx`, `ProfilePageContainer.test.tsx`, `ProfilePageLayout.test.tsx`, `useProfileNavigation.test.tsx`

Also:

- [x] `npm run lint` — clean
- [x] `npm run format:check` — clean
- [x] `npx tsc --noEmit` — 0 errors (stale `.next` types from the pre-rebase tree must be deleted first; they are not branch errors)
- [x] `npx next build --webpack` — compiled; `/session-bridge` is in the route table
- [x] Local Chromium proof from the vibes board: an allowlisted vibe iframe receives `pubky-session` / `pubky-session-none` for a signed-in / signed-out pubky-app session without a second grant

## Companion PRs

- `pubky/vibes`: helper + board that embeds `/session-bridge` and consumes the message contract
- Second pubky-app PR on `feat/vibe-session-consumer`: consumer mode for pubky-app forks deployed as approved vibes (ADR 0019 on that branch)

## Audit

This change was externally audited in three rounds (the third covering the review follow-up push `98c6894`). Findings from those rounds are incorporated here (origin/source validation, no secrets in the export, cookie isolation, allowlist/CSP alignment, build-time validation, loopback excluded from the production default).


talosmachina (commented):
Minor, not blocking: the production default allowlist (`SESSION_BRIDGE_PROD_DEFAULT_ORIGINS`) includes `https://vibes.staging.pubky.app` and `https://*.vibes.staging.pubky.app`. On a production build of pubky.app that means the staging board can frame the production `/session-bridge` and receive a production `sessionExport`. Both hosts are ours so the risk is low, but I would drop the staging origins from the production default and let staging deployments set them via env if needed.

Also worth stating explicitly for reviewers: the CSP `frame-ancestors` wildcard is multi-label while `isAllowedBridgeOrigin` is single-label. That asymmetry is intentional (deep subdomains can load an inert frame but never receive data) and it is correctly implemented here; noting it so a future refactor does not "fix" it by accident.

secondl1ght (changes requested):
## Review: request changes. Direction: simplify to a Route Handler

Thanks for the thorough PR and ADR. The security core holds up: the origin gate rejected every spoof we tried, replies always target the validated origin, and the export is public per the SDK typings. The problems are in how the route is deployed and integrated with the rest of the app. Rather than patch those in place, we would like to change the shape to one that removes most of them.

### What we are asking for

Replace the React page and second root layout with a Route Handler at `src/app/session-bridge/route.ts` that returns a self-contained HTML document.

- The message listener lives in an inline `<script>` in that document, so it is attached during parse before any message can arrive. No pre-hydration queue, no dispatch swap, no refcount.
- The allowlist comes from `PUBKY_RUNTIME_SESSION_BRIDGE_ALLOWED_ORIGINS` (runtime-config schema, per ADR 0017), injected into the HTML the same way `__PUBKY_CONFIG__` is, via `escapeForInlineScript`. `Content-Security-Policy: frame-ancestors`, `Referrer-Policy`, and `X-Robots-Tag` are set on the `Response` per request from that same value.
- Revert the `src/app` to `src/app/(main)` move, the `NEXT_PUBLIC_` Zod field, the eslint allowlist widening, and the `next.config.ts` `headers()` and `./src/...` import. The diff shrinks to the route handler, the allowlist module, tests, and docs.
- The message contract does not change, so the vibes helper and #2483 are unaffected.
- Test the inline script by evaluating it in jsdom (dispatch a message, then assert one reply) instead of hand-building the queue.

### Why: what we found on the current implementation

Blocking today:

1. **The allowlist cannot be set on staging.** `NEXT_PUBLIC_SESSION_BRIDGE_ALLOWED_ORIGINS` is inlined at build time and written into `routes-manifest` by `headers()`, but `Dockerfile:24-27` and both workflows only pass the four existing `NEXT_PUBLIC_*` build args. Every image ships the production default, and setting the variable on the running container does nothing because `next start` never re-runs `headers()`. The staging board iframe is refused by CSP. Runtime config in the new shape fixes this.
2. **`src/instrumentation-client.ts:14` runs on the bridge.** The bridge layout intentionally omits `window.__PUBKY_CONFIG__`, so every bridge load logs the "runtime-config injection is broken" console.error and client Sentry stays disabled there. Gone in the new shape, since no Next client bundle loads.
3. **Serwist reloads the iframe.** `@serwist/next` defaults `reloadOnOnline: true` and injects its entry into the shared `main-app` chunk, so the bridge iframe calls `location.reload()` on every `online` event and silently drops the storage subscriber. The consumer then never receives `pubky-session-none` after a sign-out. Gone in the new shape.
4. **The validator accepts CSP-hostile host characters.** Verified against the branch: `validateSessionBridgeAllowlistEntry('https://shop.pubky.app;sandbox')` returns true and the builder emits `frame-ancestors https://shop.pubky.app;sandbox`. Restrict host characters to `[A-Za-z0-9.-]` plus the bracketed loopback literal. This one carries over to the new shape.

Also removed by the new shape: the Strict Mode remount discards the subscriber captured while draining the queue (`install.ts:105`; the docstring at line 36 describes a refcount that only ever goes 0 to 1 to 0), `SessionBridgeClient.test.tsx:29` runs its `afterEach` reset before RTL cleanup and leaks listeners between tests, and the synthetic `/_not-found` route is now composed from Next's bare built-in layout (confirmed with a real build; latent today because the `[...catchAll]` slot absorbs unmatched URLs).

### Items that carry over regardless of shape

- `handleStorage` should check `event.storageArea === localStorage` before treating `key === null` as a sign-out; a `sessionStorage.clear()` in another same-origin document currently pushes a false `pubky-session-none`.
- CSP host-source grammar cannot express `[::1]`, so `http://[::1]:3000` passes validation but the browser drops it from `frame-ancestors`. Either reject IPv6 literals or document localhost and 127.0.0.1 only.
- The env-wiring tests pass even if a reader ignores the variable, because `src/config/test.ts` sets it to the default. Use a distinctive stubbed value and assert it flows through.
- ADR 0020 still carries the TEMPLATE status note, should link ADR 0017 (it amends the build-intrinsic list), and skips 0019. `docs/environment.md:175` says wildcards match one label in the section that also covers CSP, where they are multi-label. `env.ts` and `.env.example` say loopback is allowed only when listed, but the non-production default includes `http://localhost:3000`.
- The `._*` ignore entries in eslint, prettier, and vitest configs are unrelated to the feature. Please move them to a chore commit. The root-level vitest `exclude` is dead with `projects` (no `extends`).

### Non-blocking

- Please record ADR Alternative 3 (scoped bearer sessions on the homeserver) as the tracked follow-up. The bridge is a stopgap for same-site vibes and hands every approved vibe the full session.
- Suggest ADR 0020 state explicitly that the trust boundary is DNS approval of vibe hostnames, not the bridge allowlist.

Issue #1082, opened 2026-01-28
[Bug]: Tags other people make are not shown


SeverinAlexB:
### Bug Description

@thisispav tagged one of my posts. I did get the notifications but when clicking on the notifications, they don't show up. Hard refreshing the page doesn't work either. Pav confirmed he didn't delete the tags and they are still showing for him.

I saw this problem already multiple time. I can only assume it shows up for others as well.
Feels like IndexDB does not get updated when notifications come in. Don't know.

### Steps to Reproduce

1. Create a post
2. Somebody else tags the post
3. You receive the notifications
4. Tags are not showing up.

### Affected Area

Notifications

### Suspected Layer (if known)

Data inconsistency (local vs server)

### Frequency

Always (100%)

### Expected Behavior

Should show the tags correctly.

### Screenshots / Recording

https://github.com/user-attachments/assets/cce3ba7e-4364-45a3-8eab-4cbe79b19f82

### Console Log Output

```shell

```

### Browser & Version

Brave

### Device

Macbook M1, MacOS

PR #2486, opened 2026-09-07
fix(tags): unify local-first caching and viewport TTL refresh

### LINKED ISSUE #1987: Navigation to page with posts on fires many nexus query requests for tags
### Summary

Same trigger as #1985 — direct navigation to a non-existent post — but a different symptom: the **"Post Not Found"** page fires an enormous number of network requests (**~322** observed) instead of the handful it should need.

### Steps to reproduce
1. Open `/post/<userId>/<nonexistent-postId>` directly (valid `userId`, non-existent `postId`).
2. Open DevTools → Network.
3. Observe the "Post Not Found" empty state — request count climbs to ~322.

### Expected
A not-found page should be cheap: resolve that the post is missing, then render a static empty state. Network requests should be in the low single/double digits.

### Actual
~322 requests fire on the not-found page.

https://github.com/user-attachments/assets/479529bf-6308-4dd8-9b2a-f8e6928639c7

### Likely cause
When the post is missing, `PostPageShell` swaps in the full **`HotDiscoveryContentLayout`** discovery chrome (the #1985 fix). That chrome mounts every Hot-page data source, none of which a not-found page needs:

- **Who-to-follow refill loop** — `src/hooks/useUserStream/useUserStream.ts:286-313`. With `excludeFollowing=true`, each refill changes `userIds`, which re-runs three `useLiveQuery` calls (details/counts/relationships); relationship hydration filters users below the refill threshold, forcing another network refetch. Cascades. Most likely the bulk of the 322.
- **Active users** — same `useUserStream` live queries (`useUserStream.ts:88-148`).
- **Hot tags** — `src/hooks/useHotTags/useHotTags.tsx:78`.
- **Trending timeline** — `TimelineFeed` HOT variant stream pagination, `src/hooks/useStreamPagination/useStreamPagination.ts:224`.

### Suggested direction
Don't mount the full discovery chrome for a not-found state. Render the empty state inside a minimal layout (single `ContentLayout`, no who-to-follow / active-users / hot-tags sidebars), or gate those sidebar data sources off when `postMissing` is true. The trending timeline (if kept per Figma) should be the only network-driven element.

### Environment
- Branch: `bug-ui/1985-post-not-found-nested-layout` (current state after #1985 fix `9b7dab0a`)

Related: #1985

infin1t3:
Revisiting a feed used to fetch tags for every mounted card, even when bootstrap or batch hydration had already cached them. This PR makes initialized tag collections reusable across navigation: hooks observe IndexedDB, missing data loads on demand, and TTL or tag notifications provide freshness. Feed cards, expanded post tags, profile tags and Visual feeds use the same cache lifecycle.

## Why this changes more than the mount effect

Local-first storage and TTL already existed. Removing the mount request also removes a freshness mechanism, so the cache needs to distinguish **missing, empty and partial data**, preserve pagination, and reconcile refreshes with local edits. The existing TTL coordinator also required authentication: public post/profile viewing was available, but signed-out visitors received no periodic refresh.

This PR changes who owns loading and refresh, and extends that existing coordinator to visible public data. [ADR 0020](docs/adr/0020-local-first-tag-cache.md) records the decision; [Local-First Patterns](docs/local-first.md#tag-previews-pagination-and-freshness) describes the persistence contract.

## Behavior and implementation

- **Cache and pagination:** initialization, server cursor, exhaustion, freshness, revision and viewer metadata live with each collection. An initialized empty result is a cache hit. Local edits do not advance the server cursor, and a short batch preview cannot truncate an expanded list. Refreshing the loaded server portion also removes deleted labels. Existing IndexedDB rows remain readable without a reset.
- **Refresh ownership:** visible components hold reference-counted post/profile subscriptions across route changes and release them on viewport exit or unmount. Posts also track their authors. The manager controls coordinator lifetime, including signed-out views; hidden browser pages pause ticking. Account changes discard queued work and reject responses from the previous session.
- **Freshness and recovery:** tags have freshness independent of post/profile data. Accepted complete batch previews avoid a separate tag request; stale or expanded collections refresh as needed. Tag notifications target the affected entity. Background failures retain cached data and use a persistent 30-second retry cooldown. Failed tag or author work does not require repeating successful entity batches. Tag-list auto-pagination stops after an unproductive attempt and exposes a manual **Load more** button.
- **Local edits and rendering:** shared mutation identities in IndexedDB scope settlement and rollback to their own operation. Reconciliation preserves viewer selection and avoids applying optimistic count changes twice. UI completions from a previous view cannot alter the new view's placeholders or toasts. `useTagCache` separates the database read from rendering; React Compiler memoizes its projection without manual `useMemo`.
- **Hydration and requests:** attachment metadata is persisted before stream/repost hydration or TTL publishes fresh posts; cold replies are hydrated before their IDs are exposed. Persistence failures remain retryable. Bodyless GETs omit `Content-Type` to avoid unnecessary CORS preflights; requests with bodies retain it.

## Tradeoffs and boundaries

Tags are no longer guaranteed fresh on every navigation; they follow TTL and targeted notifications. Expanded lists cost more to refresh than previews, and signed-out sessions now generate background requests for stale visible data. Subscription ownership and revision/session checks are required to keep that lifecycle correct. Initial loads can still make individual tag requests. This does not claim to eliminate every `by_ids` duplication or Nexus 429 scenario.

## Changes after review

- The TTL tag pass skips ids still queued for an entity batch, and per-entity tag refreshes run with bounded concurrency (`TAG_REFRESH_MAX_CONCURRENCY`).
- Ids omitted from a batch response get a delayed TTL row (the configured retry delay) instead of a request on every tick; rows written since the fetch started are kept.
- A tick re-checks its session after every await, including after entity refresh, so an old response cannot remove the new session's queued work. A stopped/restarted loop owns its timer; a previous tick cannot postpone it. Sign-out delays the first guest tick by one interval; an account switch ticks immediately.
- New posts seed an initialized, complete tag window for their author. Uninitialized placeholders remain misses for initialization and bulk user-tag reads. An empty placeholder shows loading only while its fill is pending; optimistic tags remain visible across remounts, including when the fill fails.
- The notification tag fallback skips when hydration already wrote a viewer window that is exhausted or at least a refresh page long. Invalidation is skipped only for snapshots newer than the event by more than the tag TTL, bounding clock skew to one TTL window.
- The bulk user-tag fill is viewer-scoped and requests the profile page size; tag previews are saved with one bulk read and one bulk write per table.
- The single-collection grid mirrors the envelope for guests too, since the public TTL now refreshes it; only the owner is excluded.
- Merged with #2492: generic ref-counted subscriptions with the author auto-subscription on top, both `persistPosts` write guards kept. The tag cache ADR is now [ADR 0020](docs/adr/0020-local-first-tag-cache.md) because `dev` merged a different ADR 0019.

## Validation

Latest automated validation on `6fe0b8e3`, after merging `dev` through `d786bb4c` and addressing Matthew's final two comments, using Node 24:

- **13,533 unit tests passed, 2 skipped, across 782 files.** Typecheck, lint and full-repository formatting checks passed. Five new regression cases failed before the fixes and pass afterward: the immediate account-switch timer, retaining failed post/user batches when an old response settles, and preserving optimistic tags across remounts with successful or failed initialization. No integration tests were added.
- **6 targeted VRT checks passed** for the Tagged profile page at desktop and mobile sizes in Chromium, Firefox and WebKit, without baseline changes. The VRT runner emitted a shutdown timeout warning after the passing results and exited with code 0. Live browser and Cypress checks were not rerun for these commits.

Earlier validation on `61cc3a51`, after merging `dev`, using Node 24:

- **13,324 unit tests passed, 2 skipped, across 778 files.** Typecheck, lint and changed-file formatting passed. Regression coverage includes repeated multi-viewer reconciliation, complete response absence, stale profile actions and bounded pagination/manual retry. No integration tests were added.
- **186 VRT checks passed** across Chromium, Firefox and WebKit for Collections, Home, Hot, Search, Profile and Settings, without baseline changes.

Additional checks performed on `a665c2b4`:

- **120 VRT checks passed** across Chromium, Firefox and WebKit for Home, Hot, Profile, ProfileTaggers and SinglePost, without baseline changes.
- Direct React Compiler checks succeeded in development and production modes and confirmed memoization of the tag projection.
- Local browser checks covered a profile reload, expanded taggers, Home → Italy → Home and all five tags on the existing QA post. No console errors were recorded during these checks.

The [production/ephemeral comparison and two-account notification verification](https://github.com/pubky/pubky-app/pull/2486#issuecomment-5600338742) document the earlier live testing and identify the measured releases. Those measurements, and the compiler/local browser checks above, were not rerun after the final comment/skeleton cleanup and `dev` merge. Full Cypress against the Docker Pubky stack was not run locally. CI for the latest head is tracked by the PR checks.

## Suggested review order

Start with ADR 0020, then `TagCacheApplication`, `LocalTagCacheService`, the reconciliation pipe and mutation model. Follow their TTL/notification callers, subscription lifetimes and UI hooks. On the preview, revisit a warmed feed, expand tags, navigate away/back, check refresh without list truncation, and leave a public post/profile visible while signed out.

Addresses the tag-request pattern discussed in #2485.

Fixes #1987.


talosmachina (commented):
**No findings.** Replaces per-card fetch-on-mount tags with a metadata-tagged local tag cache: mount fetch becomes fill-if-missing, freshness moves to the TTL coordinator, tag notifications invalidate and force-refresh affected collections, and CORS-simple bodyless GETs remove the preflight OPTIONS. Reviewed `f5bdcac1`.

**Read in full**: `tag-cache.ts` (application, service, pipe), `useTagCache.ts`, `usePostTags.tsx`, `useTagged.tsx`, `ttl.ts` (coordinator + application), `notification.ts`, `tagCollection.ts`, `tag-cache.utils.ts`
**Call sites traced**: `run()` dedup keys, `savePage`/`savePreviews` revision guards, `captureViewerSession` at every controller entry, `cleanupLocalState` vs `TagCacheApplication.pending`
**Local run**: 277 tests across tag/ttl/notification suites, all green; CI green on this head (13 checks)
**#1082/#1474**: notification-driven invalidate + forceRefresh re-fixes the original "tags by others never appear" bug at the right layer; mount-fill keeps cold posts working

**Ruled out**
- *Legacy rows without `cache` never TTL-refresh* (I claimed this in Slack earlier, it is wrong): `initializeLegacyCursor` backfills on mutations, and the post/user TTL force-refresh backfills the rest via `savePreviews` within one tick. Residual window is bounded by `postTtlMs`.
- *In-flight `getOrFetch` racing invalidation*: revision capture + `savePage` guard make the loser retry (bounded at 2) or report contention; `run()` chains mode switches instead of running concurrently.
- *Stale-session writes after logout*: every controller entry wraps `captureViewerSession()` and all four write paths re-check it inside the Dexie transaction.
- *Preview clobbering an expanded window*: `getTagCursor(existing) > tags.length` skip plus superseded/foreign-viewer guards preserve the loaded window.
- *Double-fetch on repost+original cards and StrictMode remounts*: `pending` map dedups by `kind:id:viewerId`.
- *429 amplification from per-tick refresh*: `refreshStale` re-checks staleness and `retryAt` cooldown per entity, so a failing tag endpoint retries at most once per 30s regardless of tick rate.

</details>

secondl1ght (changes requested):
Thanks for this. The cache design holds up under scrutiny: dispatch, revision guards, session scoping, mutation rollback, layer rules and `tsc` all check out, and the ADR is clear. What follows are the places where the new TTL and hydration paths do more network and IndexedDB work than ADR 0019 promises, plus two loading-state gaps. Three are blocking because they regress the request-burst behaviour #2476 just hardened against. At the end there is a section on coordinating with #2492, which rewrites the same TTL layers.

### Blocking

**1. Tag pass refreshes ids still waiting in the entity batch queue** — `src/core/coordinators/ttl/ttl.ts:609` and `:615`

`onBatchTick` batch-refreshes at most `maxBatchSize` entities, then hands *every* subscribed id to `refreshStaleTags`. Entity `lastUpdatedAt` and tag `fetchedAt` are stamped in the same `persistPosts`/`persistUsers` call and checked against the same `ttlMs`, so everything left in the queue after the 20-id batch is also "stale" for tags. `TtlApplication.refreshTagWindows` fans those out through an uncapped `Promise.allSettled`, one forced GET each, and the next tick's by_ids batch then re-downloads and overwrites the windows just fetched. A Followers list with ~28 rows costs 9 requests where 1 would do, every TTL expiry, and author tag windows are refreshed although feeds never render them. This contradicts "a complete accepted batch preview needs no additional tag request" in the ADR. Suggested fix: exclude `ops.batchQueue` ids from the tag pass and bound concurrency in `refreshTagWindows`.

**2. `force: true` turns "Nexus omitted this id" into a 5-second loop** — `src/core/application/ttl/ttl.ts:71` and `:128`

An id absent from the by_ids response never gets a TTL row, so `findStale*ByIds` re-flags it every tick and the forced fetch bypasses the 20 s query cache the old code relied on. A new sign-up that Nexus has not indexed yet goes from ~3 POSTs/min to ~12/min after the bootstrap retry delay, until indexing completes; a server-deleted post still in a subscribed stream does the same. The coordinator's "retains bootstrap retry" test asserts this per-tick refetch. Suggested fix: a per-id miss cooldown in `checkAllEntitiesForStaleness` (the ADR 0012 follow-up), or write a TTL row for ids missing from the response.

**3. Sign-out restarts the tick at 0 ms and repopulates the database mid-clear** — `src/core/coordinators/ttl/ttl.ts:240`

`cleanupLocalState` calls `useAuthStore.reset()` before `await clearDatabase()`. The reset fires the listener synchronously, which calls `evaluateAndStartTicking()` → `scheduleNextTick(0)`. That tick's TTL reads queue behind the clear transactions, see no rows, and force-fetch every still-subscribed post plus the just-logged-out user; `captureViewerSession()` captures the already-null session, so `persistPosts`/`persistUsers` write those rows back into the freshly cleared DB. Reachable from the profile page "Sign out" (public route, so `CoordinatorsManager` stays mounted). Keeping guest refresh after logout is fine per the ADR; the defect is the immediate tick racing the clear. Suggested fix: schedule the restart with `batchIntervalMs`, or have logout stop and restart the coordinator around the clear.

### Should fix in this PR

**4. New posts are seeded without cache metadata** — `src/core/services/local/post/post.ts:289`

`PostTagsModel.create({ id, tags: [] })` produces a legacy row with `viewerId` undefined. When the author's card mounts, `TagCacheApplication.load` sees `viewerChanged` and forces a refresh GET for a post Nexus may not have indexed. If a tag write lands during that GET, `savePage` is rejected on revision, retried up to three times, then throws `CONFLICT` and `deferRefresh` stamps a 30 s cooldown on a brand-new post; `initializeLegacyCursor` also leaves `fetchedAt: 0` so the TTL tick flags it stale immediately. Measured: one forced GET per mount for the author, none when signed out. No tags are lost. Suggested fix: seed `cache: { cursor: 0, exhausted: true, fetchedAt: Date.now(), revision: 0, viewerId: authorId }`.

**5. Notification path refetches the window it just persisted** — `src/core/application/notification/notification.ts:408`

After invalidation and the forced by_ids hydration write a fresh viewer-scoped preview, the final loop skips only `exhausted` windows and otherwise calls `forceRefresh`, whose `load()` computes `limit = max(POST_TAGS_PER_PAGE, cursor)` = the preview length and re-downloads it. Ten TagPost notifications on posts with more than five unique tags cost ten extra forced GETs plus ten revision-bumping puts that re-fire every observer with identical data. Suggested condition: skip when `fetchedAt > 0 && viewerId === viewer && (exhausted || cursor >= pageSize)`; after invalidation any accepted `fetchedAt > 0` already proves a post-event window, and a `foreignPreview`-rejected row still refreshes.

**6. `isLoading` treats placeholder rows as loaded** — `src/hooks/useTagCache/useTagCache.ts:91`

`LocalTagCacheService.invalidate` writes `{ tags: [], cache: { initialized: false, ... } }` when no row exists. The hook sees a non-null record and reports `isLoading: false` while `getOrFetch` is still filling it, so `ProfileTagged` renders `TaggedEmpty` (with the add-tag CTA) instead of the skeleton for a user who has tags. `TagCacheApplication.load` treats the same row as not loaded. Suggested fix: treat `record?.cache?.initialized === false` like `!record`, plus a hook test using the `initialized: false` fixture already in `application/tag/tag-cache.test.ts`.

Related, same file line 66: `pending` is cleared when the `getOrFetch` promise settles, which Dexie resolves before the live query's post-commit re-read emits, so a successful cold-cache fill still renders one frame of `record: null, isLoading: false`. Not a regression (the old hook showed the empty state for the whole fetch), but `useTagCache.test.tsx:163-176` currently asserts that transient state as expected, so it is worth knowing that test would have to change if the flash is ever closed.

**7. Bulk user-tag fetch stores a viewerless window that forces a refresh on first profile open** — `src/core/application/user/user.ts:398`

`fetchMissingUserTagsFromNexus` calls `NexusUserService.tags` without `viewer_id` and persists without `guard.viewerId`, so the row is stored with `cache.viewerId = null` under a signed-in session. When the viewer then opens that profile, `TagCacheApplication.load` sees `viewerChanged` and issues a forced `limit_tags=20` GET seconds after the 10-tag one. The trigger is any user with a details row but no tags row (everything persisted via `upsertDetails`: notification actors, `useUserDetails`, `useUserProfile`), which is common on Followers/Following and user streams. This undercuts the "revisit without a tag request" goal. `viewerId` is not on `TPubkyListParams`; thread it from `UserController.getManyTagsOrFetch` the way `withViewer` already does for `getOrFetch`/`fetch`, into both the Nexus call and the preview guard.

**8. `savePreviews` serializes four IDB round-trips per entity on the first-paint path** — `src/core/services/local/tag/tag-cache.ts:156`

The previous two pipelined `bulkPut`s became a sequential `get → get → put → put` loop inside the transaction; Dexie issues the next request only after each awaited one returns. A cold Home page is ~40 serialized round-trips for posts plus ~40 for authors before `setPostIds` fires, and bootstrap scales the same way. The mid-loop `return`s also commit partial rows. `Promise.all([table.bulkGet(ids), countsTable.bulkGet(ids)])` → reconcile in memory (the pipes are pure) → one `bulkPut` each is a drop-in and atomic.

**9. Bulk tag miss-filter ignores `initialized`** — `src/core/services/local/tag/user/tag.user.ts:153`

`getNotPersistedUserTagsInCache` still treats any row as a hit, so a placeholder created by `getOrCreate` or `invalidate` is never fetched by `getManyTagsOrFetch`. No current consumer renders that result's tag list, so this is latent, but it contradicts the invariant the rest of the PR enforces. One-line fix plus a unit test.

### Question

**10. Clock skew in the notification invalidation skip** — `src/core/application/notification/notification.ts:373`

`validatedAt` is client `Date.now()` at request start; `notification.timestamp` is server-assigned. With the client a few tens of seconds ahead, a snapshot taken shortly before a tag event compares as newer and the invalidation plus forced refresh is skipped, leaving the list stale until the 5-minute tag TTL. The repo already assumes clock agreement elsewhere, so this may be accepted; if so it is worth a comment, otherwise a margin (e.g. the 20 s `staleTime`) or a revision-based proof would close it.

### Coordinating with #2492

#2492 rewrites the same two TTL layers and the trial merge produces 20 conflict hunks across `application/ttl/ttl.ts`, `coordinators/ttl/ttl.ts`, its types and its tests. The conflicts are semantic. So that both can be addressed in one pass regardless of which lands first:

- **Adopt #2492's generic ref-counting** (`subscribe`/`unsubscribe`/`addSubscription`/`removeSubscription` over `EntityOps`) and express this PR's author auto-subscription on top of it rather than the inline post counter in `subscribePost`/`unsubscribePost`. That also replaces the `compositePostId.split(':')[0]` calls with a single `parseCompositeId`.
- **Keep this PR's logout semantics** (guest refresh continues, only queues clear). #2492 stops and resets on logout and has a test asserting it, but its own description lists signed-out refresh as a follow-up, which is what ADR 0019 delivers. Please align with its author; its logout test will need to change, and item 3 above applies to whichever version lands.
- **Keep both `forceRefreshPostsByIds` guards**: this PR's revision/session/attachments-first path and #2492's `findPostsWrittenSince` (local edit during an in-flight fetch). Drop #2492's re-added `fetchAndPersistMissingAuthors`; the author subscription here replaces it.
- **Keep #2492's `filterStillStale` re-check** in `refreshStaleEntities`, and place the queued-id exclusion from item 1 next to it.
- #2492 edits ADR 0012; this PR supersedes it with ADR 0019. Whichever lands second reconciles the two.

Recommended order is #2492 first (smaller, already mergeable, better ref-counting substrate), then rebase this PR onto it.

src/core/coordinators/ttl/ttl.ts
```diff
@@ -614,6 +571,10 @@ export class TtlCoordinator {
     // application-level guard covers writes that land while the fetch itself
     // is in flight.
     const ids = await this.filterStillStale(queuedIds, ops);
+    // The account may have changed during that read. The controller captures
+    // the session at call time, so a request carrying the previous viewer would
+    // otherwise be accepted as the new session's data.
+    if (!isCurrent()) return;
```
secondl1ght, line 582: Low severity, from the listener restart: a tick still in flight during an account switch is not cancelled. This re-check protects the fetch, but the success path below (`for (const id of ids) ops.batchQueue.delete(id)` after `await ops.forceRefresh(...)`) has no equivalent: `forceRefresh` returns early once `isCurrent()` fails in the application, yet the ids are still deleted from the queue, and the old tick's `finally` then calls `scheduleNextTick(batchIntervalMs)`, which clears the listener's 0 ms timer, so the new account's first refresh waits a full interval. Data is unaffected (the application discards the stale-session response). Suggest `if (!isCurrent()) return;` before the queue deletion, and a tick generation counter captured at tick start so a superseded tick's `finally` does not reschedule.

src/hooks/useTagCache/useTagCache.ts
```diff
@@ -0,0 +1,100 @@
...
+      .finally(() => {
+        if (pendingPage.current === key) pendingPage.current = null;
+        setLoadingPage((previous) => (previous === key ? null : previous));
+      });
+  }
+
+  // A placeholder written by invalidation or a local mutation is not loaded data. It counts
+  // as loading only while the mount fill is pending, so a failed fill shows the empty state
+  // rather than a permanent skeleton.
+  const unfilled = !record || record.cache?.initialized === false;
```
secondl1ght: Second-order effect of the item 6 fix: a placeholder that already carries an optimistic local tag (`initialized: false` with non-empty `tags`, created when the viewer tags an entity whose window never loaded) now counts as unfilled, so a remount swaps the visible tag for the skeleton until the mount fill settles, and offline until it fails. Before this change the tag stayed visible. Suggest only treating a placeholder as unfilled when it has nothing to show: `record.cache?.initialized === false && record.tags.length === 0`.

secondl1ght (dismissed):
Thanks for the thorough follow-up. Re-checked all ten items against cbf2e7b8: each is addressed as described, with tests. The #2492 merge follows the recommendations (generic ref-counting with author auto-subscription on the first/last-reference booleans, `parseCompositeId`, guest refresh kept, both write guards in one transaction, `filterStillStale` with the session re-check, `fetchAndPersistMissingAuthors` gone, ADR 0020). Typecheck and the TTL/tag/notification/user suites pass locally.

Two low-severity follow-ups from the fixes themselves, left as inline comments. Neither blocks; fine to take here or in a follow-up.

src/hooks/useTagCache/useTagCache.ts
infin1t3: Will do as a follow up.

src/core/coordinators/ttl/ttl.ts
infin1t3: Will do as a follow up.

secondl1ght (approved):
Re-checked 6fe0b8e3: both follow-ups from the previous review are in (tick generation guard plus the post-refresh session re-check in the coordinator, and the empty-placeholder refinement in `useTagCache`), each with tests, and the dev merge carries no source changes of its own. Typecheck and the coordinator/application/hook suites pass locally. Good to merge.

PR #2491, opened 2026-09-08
chore(vrt): add test:vrt:setup script to install Playwright browsers

secondl1ght:
## Summary

- Add `npm run test:vrt:setup`, which runs `playwright install --with-deps chromium firefox webkit` (the same install the VRT CI workflows use). Until now every dev had to work out this command themselves before `npm run test:vrt` would run.
- Document it in `docs/visual-regression-testing.md`, including the Linux `sudo` prompt for apt libraries.

## Notes

- Playwright must be at 1.61+ for Ubuntu 26.04 hosts; `dev` already pins 1.62.1, so this works out of the box there.
- CI workflows are unchanged.

## Test plan

- [x] `npm run test:vrt:setup -- --dry-run` resolves the local Playwright binary and lists the expected browser revisions.
- [x] Full install + `npm run test:vrt` (Landing suite) verified on Ubuntu 26.04 with Playwright 1.62.1.
- [x] `prettier --check` passes on the changed files.


Taewa (approved):
LGTM

PR #2492, opened 2026-09-08
fix(collections): keep collection envelopes fresh via viewport ttl

### LINKED ISSUE #2384: [Bug]: Collection post counter doesn't update when owner adds new posts
## Summary

The post count on a collection (the “N POSTS” badge) is taken from a cached copy of that collection on your device. After someone else adds posts to it, the number does not update until you sign out and sign back in (which wipes local data). Collection post count stays stale until sign-out.

This is not limited to one screen. The same stuck count shows on:
- a user’s profile Collections tab
- the Collections landing page (/collections - Followed and Discover cards)
- the individual collection page (hero badge)

Your own collections still update immediately, because adding a post writes to your local database. This is a viewer of someone else’s collection problem.

On the collection page, the item grid can already show the new posts (that list is fetched from the network). The count can still be the old number, because it comes from the cached collection record, not from the grid.

## Why it is happening
The count is always `items.length` from the collection post’s JSON envelope, not a live tally of the item feed.

**Cards** (`CollectionCard`, used on profile Collections and `/collections`):
```
const itemCount = collection?.items?.length ?? 0;
```

**Collection page hero** (`CollectionHero`, same envelope, passed from the page shell):
```
const collection = parseCollectionContent(postDetails.content);
// ...
const itemCount = collection?.items?.length ?? 0;
```

That envelope is loaded with `usePostDetails` → `useLocalFirstQuery`. If IndexedDB already has the post, the fetch is skipped and never retried:
```
// Cache hit — data is already in IndexedDB, no fetch needed.
if (data !== null) {
  setIsFetching(false);
  return;
}
```

The intended refresh path is viewport TTL (`useTtlSubscription` → `TtlCoordinator` → Nexus `posts/by_ids` → overwrite Dexie). Regular posts do this in PostMain:

```
const { ref: ttlRef } = useTtlSubscription({
  type: 'post',
  id: postId,
});
```

Collection surfaces never take that path:
- Profile Collections mounts `CollectionCard` instead of `PostMain` when `kind === 'collection'` (`TimelineFeedItem`). `CollectionCard` does not subscribe to TTL.
- `/collections` renders the same `CollectionCard` from My / Followed / Discover sections.
- The collection page shell (Collection.tsx) also uses `usePostDetails` with no TTL, then passes that envelope into `CollectionHero`.

Coming back later does not fix it. Stream reload only hydrates missing posts (`cacheMissPostIds` in `StreamPostsController.getOrFetchStreamSlice`). Adding items is an in-place edit of an existing collection post (`commitUpdateCollectionItem`), so it is not a new stream head item and is not treated as a cache miss.

Signing out “fixes” it because it clears IndexedDB, so the next visit is a cache miss and fetches a fresh envelope.

The item grid can still look up to date because it is a separate Nexus collection stream (`source=collection`), not the cached envelope used for the badge.

Repost embeds of collections already refresh: `PostPreviewCard` wraps `CollectionCard` with `useTtlSubscription`.

### Where and how to fix

Subscribe the collection envelope to TTL the same way `PostMain` does, once per visible collection - not once per screen.

1. `CollectionCard` (`src/components/organisms/Collections/CollectionCard/CollectionCard.tsx`): call `useTtlSubscription({ type: 'post', id: compositeId })` and attach the returned `ref` to the card root. That covers profile Collections, `/collections` (Followed / Discover / My), and any other standalone card.
2. Collection page envelope (`src/components/templates/Collection/Collection.tsx` or `CollectionHero`): subscribe TTL for the same composite id so the hero count refreshes. Use one subscriber on this page. If the hero (or page) already subscribes, do not also subscribe from a nested CollectionCard for the same id.
3. `PostPreviewCard` collection branch (`src/components/molecules/PostPreviewCard/PostPreviewCard.tsx`): drop the extra TTL wrapper around the embed `CollectionCard`. Post TTL subscriptions are not ref-counted (`TtlCoordinator.subscribePost` is idempotent; unsubscribe removes the only slot). Two subscribers on the same id will unsubscribe each other.

After a successful TTL persist, `usePostDetails` live-queries Dexie and every badge that reads that envelope updates with no extra UI wiring.

Tests: mock `useTtlSubscription` in `CollectionCard` tests the same way `PostMain.test.tsx` does. Assert the card subscribes with the collection composite id when visible. Cover the collection-page subscriber and the `PostPreviewCard` change so the embed no longer double-subscribes.

### Acceptance

- Another user’s collection: after TTL expires, the on-screen count updates without signing out on profile Collections, `/collections`, and the collection page hero
- Collection page item grid and hero count stay consistent after refresh
- Own collections still update immediately on add/remove
- Repost embeds of collections still refresh (single TTL subscriber, not two)
- Signing out is no longer required to see an updated count

### Expected Behaviour

Wherever that count is shown, it should refresh the same way feed posts do: after the post TTL expires (about 5 minutes) while the collection is on screen, it should refetch and show the new number - without signing out.

### Steps to Reproduce

Use two accounts. Stay signed in on the viewer for the whole flow (do not sign out).

1. Viewer: open User B's profile → Collections. Note the post count on one of B's collections (for example 3 POSTS).
2. Optionally also note the same collection on Collections (`/collections`, Followed or Discover) and open the collection itself and note the hero count. All three should match.
3. Owner (User B, other device or session): add one or more posts to that collection.
4. Viewer: stay on any of those screens, or leave and come back later (wait more than 5 minutes if you want to rule out a TTL delay).
5. Profile card, Collections landing card, and collection-page hero still show the old count. Opening the collection may show the new posts in the grid while the hero count stays old.
6. Viewer: sign out and sign in again, then reopen the same surfaces. The count is now correct everywhere.

### Screenshots

Image
Image
Image

### Console Log

```shell

```

### Browser & Version

_No response_

secondl1ght:
Closes #2384

## Summary

Collection cards and the collection page hero read title, description, cover, layout and item count from a cached post row that `usePostDetails` never re-fetches on a cache hit, so a viewer saw another user's edits only after signing out. This PR subscribes those surfaces to the viewport TTL coordinator the way `PostMain` already does, and fixes the things that stopped that refresh from ever landing.

## Root causes found along the way

- **Dead auth listeners in every coordinator.** `TtlCoordinator`, the base polling coordinator (notifications, streams) and the mute-list sync compared `prevState.selectIsAuthenticated()` against `state.selectIsAuthenticated()`. Both read the live store through `get()`, so they never differed. After a reload on a public route (single collection, post, profile) `start()` ran before the persisted session was restored and nothing ever re-evaluated, leaving TTL refreshes, the notification badge, `/post` stream polling and the mute SSE dead until a tab visibility change. Fixed with a pure `isAuthenticatedState(state)` helper used by all three listeners.
- **In-flight refresh clobbering a local edit.** A TTL refresh already in flight when the owner edits persisted Nexus's pre-edit copy over the local row and stamped it fresh. `forceRefreshPostsByIds` now skips any post whose `post_ttl` row was written locally after the fetch started.
- **Post subscriptions were not ref-counted.** Nested surfaces tracking the same post (a share dialog over a hero, a repost preview in a feed) unsubscribed each other. Post subscriptions now share the ref-counted path users already had.

## Changes

- `CollectionCard` (landing) and `CollectionHero` subscribe their composite id to the TTL coordinator; embeds skip their own subscription since the enclosing surface already subscribes.
- The single-collection grid mirrors the envelope membership in place for viewers: only loaded ids the membership contains are rendered, added ids are prepended as optimistic posts, removed ids are committed out (re-evaluated as pages land or a refresh re-serves them). Owners keep their existing optimistic flows.
- `sortPostIdsByMembership` keeps unlisted ids in place instead of moving them to the end.
- ADR-0012 and `docs/data-patterns.md` updated (ref-counted posts, auth snapshot compare, viewport subscription ownership).

## Verification

- Unit suite: 775 files, 13,308 tests green. New tests cover the subscriptions, the membership sync, the coordinator auth restore in all three coordinators, ref counting, and the refresh guard.
- Collections VRT: 24 screenshots match on chromium, firefox and webkit; no baseline changes.
- Manual QA with two accounts: viewer's card and hero fields, count and grid update after the TTL window; owner flows unchanged.

## Notes for reviewers

- Signed-out viewers still never refresh; that gate is by design (ADR-0012) and is worth its own issue since the collection page is public.
- Two pre-existing issues were found and left for follow-ups: route changes drop user subscriptions held by persistent sidebars, and a collection card that settles to "not found" never retries.
- Squash on merge; the first commit subject carries a `(wip)` marker.


src/core/application/ttl/ttl.ts
```diff
@@ -58,16 +59,34 @@ export class TtlApplication {
...
     const postBatch = await NexusPostStreamService.fetchByIds({
       post_ids: uniqueIds,
       viewer_id: params.viewerId,
     });
 
+    // A local-first write that landed while this fetch was in flight (local
+    // services bump `post_ttl` on every write) is newer than anything Nexus
+    // could have returned; keep the local row and let the next TTL cycle pick
+    // up the indexed version instead of clobbering the edit.
+    const locallyWrittenIds = await this.findPostsWrittenSince({ postIds: uniqueIds, since: fetchStartedAt });
```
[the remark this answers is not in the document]
secondl1ght: Valid, thanks. The in-flight guard only covered writes made after the fetch started; an id can sit in the batch queue for up to a tick, so an owner edit in that window was still overwritten. Fixed in 2ed6416e: `refreshStaleEntities` re-runs the staleness check on the drained batch right before the network call and drops anything no longer stale (a failing re-check assumes stale, matching the subscribe-time policy). Together with the fetch-start guard in `TtlApplication.forceRefreshPostsByIds` this closes both windows: queued-then-edited and edited-while-in-flight. Tests: 'does not refresh an entity written locally between being queued and the batch firing' and 'still refreshes the batch when the pre-fetch staleness re-check fails'.

infin1t3: **Make the freshness check and persistence atomic**

A local edit can still be overwritten after this check. `findPostsWrittenSince()` reads the TTL records, but `persistPosts()` subsequently awaits the tombstone lookup before writing the Nexus rows. An owner edit can commit its details and TTL during that gap, after the post has already passed the guard. The stale response then replaces the edited content and marks it fresh again.

For collections, a subsequent owner edit reads that reverted envelope and can write the missing membership back to the homeserver. Please perform the freshness check and conditional persistence in the same local-service transaction.

The timestamp comparison also misses edits whose `lastUpdatedAt` equals `fetchStartedAt`, since it uses strict `>`. Both cases are reproducible; changing the comparison alone would leave the check/write race open.

src/components/organisms/Timeline/Feed/TimelineFeed/TimelineFeed.tsx
```diff
@@ -253,7 +266,8 @@ function CollectionTimelineFeed({
       pullToRefreshContainerRef={pullToRefreshContainerRef}
       trailingSlot={trailingSlot}
       visualHiddenItemsNotice={visualHiddenItemsNotice}
-      transformPostIds={(postIds) => sortPostIdsByCollectionOrder(postIds, envelopeItems)}
+      transformPostIds={(postIds) => sortPostIdsByMembership(postIds, membershipPostIds)}
+      membershipPostIds={isOwn ? undefined : membershipPostIds}
```
infin1t3: **Avoid filtering guest results against an envelope that cannot refresh**

Signed-out visitors also receive `membershipPostIds`, since `isOwn` is false without a current user. However, a cached envelope is not refetched by `usePostDetails`, and the TTL coordinator requires an authenticated session.

For example, a guest caches a collection containing A, the owner adds B, and the guest revisits or pulls to refresh. The collection stream returns `[A, B]`, but the cached membership remains `[A]`, so the new filter hides B. Waiting does not recover it while the visitor remains signed out. Before this PR, the grid could display B despite the stale count.

Please either provide an envelope refresh path for guests or avoid treating their stale cached membership as authoritative when filtering fresh stream results.

src/components/organisms/Timeline/Feed/TimelineFeedContent/TimelineFeedContent.tsx
```diff
@@ -198,7 +216,41 @@ function TimelineFeedContent({
...
+  const previousMembershipRef = useRef<Set<string> | null>(null);
+  const seenMembershipRef = useRef<Set<string>>(new Set());
+  useEffect(() => {
+    if (!membershipPostIds) return;
+    const previous = previousMembershipRef.current;
+    const current = new Set(membershipPostIds);
+    previousMembershipRef.current = current;
+    const seen = seenMembershipRef.current;
+    current.forEach((id) => seen.add(id));
+    // Baseline — the initial load fetches this membership.
```
infin1t3: **Reconcile missing members after the initial stream load**

Skipping the first membership assumes the initial collection stream already contains those IDs, but that stream can lag behind the envelope.

A concrete case is a viewer watching an empty collection: the owner adds P, TTL refreshes the envelope to `[P]`, and `CollectionItems` mounts the feed for the first time. This effect records `[P]` and returns, while the lagging stream returns `[]`. Later TTL refreshes still return `[P]`, so `!previous.has(P)` remains false and P is never prepended.

The hero shows one post while the grid stays empty until a manual refresh or remount after indexing catches up. Please reconcile missing members after the initial load, while preserving intentional filtering of muted or unavailable posts.

src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
```diff
@@ -56,6 +64,7 @@ export function CollectionHero({
   className,
 }: CollectionHeroProps) {
   const compositeId = buildCompositeId({ pubky: authorPubky, id: postId });
+  const { ref: ttlRef } = useTtlSubscription({ type: 'post', id: compositeId });
```
infin1t3, line 67: **Protect owner edits when Nexus indexing takes longer than the TTL**

The new subscription also refreshes the owner's collection. If an acknowledged local edit has not reached the Nexus index before its TTL expires, neither guard protects it: the pre-fetch check considers it stale, and `findPostsWrittenSince()` only protects writes made after the fetch started.

For example, the owner saves `[A, B]`, but Nexus still returns `[A]` after the TTL window. Refresh replaces the local envelope with `[A]`; adding C then reads that reverted content and can save `[A, C]` to the homeserver, dropping B.

This requires indexing lag longer than the configured TTL and is separate from the check/write race. TTL expiry alone does not establish that Nexus contains the acknowledged edit; the refresh path needs protection against replacing it with an older copy.

src/components/organisms/Timeline/Feed/TimelineFeed/TimelineFeed.tsx
infin1t3: **Reconcile owner membership changes received from another device**

The owner exception protects the existing optimistic flows, but it also excludes changes made on another device. If the owner keeps this page open on device A and adds or removes a member on device B, the hero's TTL subscription refreshes the count while the grid's membership reconciliation remains disabled.

Same-device mutation handlers cannot reconcile that remote change, so the count and grid can stay out of sync until the stream is refreshed. Please track this as a follow-up so remotely refreshed membership can be applied without disrupting an open save picker or an in-progress local mutation.

infin1t3 (changes requested):

src/core/application/ttl/ttl.ts
secondl1ght: Valid, fixed in 2d44df24. The guard now lives inside `persistPosts` (`refreshGuard: { fetchStartedAt }`, passed by `TtlApplication`): the tombstone/freshness reads and every bulk save run in one `db.transaction('rw', [details, counts, tags, relationships, ttl, bookmarks, moderation])`, so a local-first edit can no longer land between the check and the write. The comparison is now `lastUpdatedAt >= fetchStartedAt` (equality counts as written-since), and kept rows still refresh counts/tags/relationships and bump the TTL — only `post_details` is preserved. Covered by the new `persistPosts - refresh guard` block in `posts.test.ts` and an updated `TtlApplication` test.

src/components/organisms/Timeline/Feed/TimelineFeed/TimelineFeed.tsx
secondl1ght: Valid, fixed in 2d44df24. `CollectionTimelineFeed` now hands `membershipPostIds` to the feed only when `useAuthStore(isAuthenticatedState) && !isOwn`. Guests keep the pre-PR behaviour (live grid, cached count) instead of having a never-refreshing envelope filter fresh stream results. Tests: signed-out case in `TimelineFeed.collection.test.tsx` and `TimelineFeed.test.tsx`. A refresh path for guests is tracked separately in #2531.

src/components/organisms/Timeline/Feed/TimelineFeedContent/TimelineFeedContent.tsx
secondl1ght: Valid, fixed in 2d44df24 (+ 3bca1e43). The effect no longer treats the first envelope as a baseline. Once the stream has settled (`!loading && !loadingMore && !hasMore`) any member the stream never delivered — and that was not already prepended — is prepended once, so the empty-collection / lagging-index case now shows P. Removals still forget the id so a later re-add is prepended again. Muted authors are skipped (the items stream filters them on purpose); deleted posts are not special-cased because the single-collection stream already retains them for the deleted-state placeholder (`filterStreamPosts`). Tests: "defers additions until the initial load has settled", "reconciles members the settled initial stream never delivered", "does not reconcile while more pages are still loading", "does not reconcile members whose author is muted".

src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
secondl1ght: Valid, fixed in 2d44df24. The refresh guard in `persistPosts` also keeps the local `post_details` row when the Nexus copy is **not indexed after** the local one (`incoming.indexed_at <= existing.indexed_at`) — Nexus bumps `indexed_at` on every edit (verified against staging), so a copy that pre-dates the local edit cannot replace it regardless of how long indexing lags the TTL. In your example the refresh returning `[A]` carries the old `indexed_at` and is skipped; adding C then reads `[A, B]`. Test: "keeps details when the Nexus copy is not indexed after the local row".

src/components/organisms/Timeline/Feed/TimelineFeed/TimelineFeed.tsx
secondl1ght: Agreed, and kept out of this PR on purpose (owner gate retained). Tracked as #2530 with the suggested direction (distinguish local-write envelope changes from TTL-refreshed ones, or defer mirroring while a picker/mutation is active).

infin1t3 (approved):

PR #2495, opened 2026-09-09
fix(core): capture AppError once per chain and drop pipeline-verified expected errors

### LINKED ISSUE #2494: Sentry: implement ADR-0015 §5 once-per-chain capture and pipeline-verified drop rules for expected errors
## Problem

Sentry capture happens inside `Err.*` at construction time (`createAppError` → `captureAppError`), before any `catch` runs. Two consequences the 2026-09-04 Sentry deep-dive surfaced:

1. **Every wrapped error is captured twice.** When a caller wraps an `AppError` in another `Err.*` with `cause: error` (e.g. `LocalModerationService` wrapping `ModerationModel.findById`), both the inner and the outer error are sent as separate events with different fingerprints. ADR-0015 §5.1 Challenge 1 planned once-per-chain capture; it was never implemented.
2. **Expected / handled errors are reported as failures.** ADR-0015 §5.1 Challenge 3 planned a `shouldReportToSentry` predicate. What exists is `APP_ERROR_DROP_RULES` with a single rule (`nexus-post-tags-404`). Families that are control flow or by-design reconnect churn are reported at full volume: `REQUEST_ABORTED` on deliberate `AbortController.abort()` (PUBKY-APP-3N/4F), mute-list SSE reconnects (PUBKY-APP-11/1Y/6G/CX, ~2.7k events/14d).

Because capture is at construction, downstream mitigations do not work: re-tagging a category, lowering a `Logger` level (`Logger` has no Sentry sink), `preventDefault()` on unhandled rejections (Sentry's handler runs first), or handling the error more gracefully in a `catch`. PRs #2476–#2479 each attempted one of those and none reduce Sentry volume.

## Proposal

- `createAppError`: skip `captureAppError` when `params.cause` is already an `AppError` (the root was captured with the most precise stack; the wrapper's `Logger.error` line is kept for local context).
- Drop rules keyed on the shapes the error pipeline actually emits (`service` / `operation` / `code` / `context.statusCode`), each with a `reason` and Sentry short IDs:
  - `aborted-requests`: `code === TimeoutErrorCode.REQUEST_ABORTED` (only produced by `safeFetch` on `DOMException` `AbortError`).
  - `homeserver-event-stream-connect`: `service === Homeserver && operation === 'subscribeUserEventStreamForPath' && category === Server` (SDK connect failures are materialised via `handleError` → `httpStatusCodeToError(500)` → `Err.server(INTERNAL_ERROR)`; the coordinator reconnects with backoff by design). Requires tagging the SSE wrapper's `handleError` call with `operation`.
- Tests must route fixtures through the real pipeline (`safeFetch`, `handleError`, `Err.*` with an `AppError` cause), not hand-built `new AppError(...)` shapes.
- `ignoreErrors` additions for global-handler noise that cannot originate from our bundle: webview bridge teardown (PUBKY-APP-B6/B7/CJ) and MetaMask `inpage.js` (PUBKY-APP-8G).
- Docs: `docs/sentry.md` — how to add a drop rule (trace `captureAppError` first, test through the pipeline).

## Out of scope (separate decisions)

- 429 storm (PUBKY-APP-B3 family): root cause is N+1 hydration on cold `/home` (per-post hooks racing the stream batch) plus a nexus burst limit of 40; needs a batching layer / nexus contract discussion, not a drop rule. Capturing 429 only on terminal retry failure (TanStack `QueryCache.onError`) is a follow-up.
- SSRF-guard blocks: `docs/sentry.md` says these stay reportable; changing that is a product decision.
- iOS IndexedDB eviction, `Not Found` on deleted users, nexus write-then-read lag: design/ADR questions.

Refs: ADR-0015 §5, #2476, #2477, #2478, #2479.

Orlandohub:
Closes #2494. Foundational follow-up to the Sentry deep-dive PRs #2476–#2479.

## Why

Sentry capture happens inside `Err.*` at construction (`createAppError` → `captureAppError`), before any `catch` runs. That is why none of the four Talos PRs reduce Sentry volume: re-tagging a category (#2478), lowering a `Logger` level (#2477 — `Logger` has no Sentry sink), `preventDefault()` on an unhandled rejection (#2477 — Sentry's handler runs first), and handling the error more gracefully downstream (#2479) all happen after the event was sent. ADR-0015 §5.1 anticipated both problems (Challenge 1: once-per-chain; Challenge 3: `shouldReportToSentry`) but only a one-rule denylist was ever built.

## What

- **Once per error chain** — `createAppError` skips `captureAppError` when `params.cause` is, or wraps via `Error.cause`, an `AppError`. The root was already captured with the most precise stack; the wrapper's `Logger.error` is kept for local context. Wrapping a dropped root stays dropped.
- **Drop rules keyed on what the pipeline actually emits** (`src/libs/observability/sentry.utils.ts`):
  - `aborted-requests`: `code === REQUEST_ABORTED`. Only produced by `safeFetch` on an `AbortError` DOMException (deliberate `abort()` or browser cancellation). `AbortSignal.timeout()` raises `TimeoutError`, which is not matched, so real timeouts stay reportable. PUBKY-APP-3N/4F.
  - `homeserver-event-stream-connect`: `service === Homeserver && operation === 'subscribeUserEventStreamForPath' && category === Server`. SDK connect failures are materialised as `handleError → httpStatusCodeToError(500) → Err.server(INTERNAL_ERROR)` — not `Network`, which is why #2477's rule could never fire. Auth/validation on the same operation and all other Homeserver operations stay reportable. PUBKY-APP-11/1Y/6G/CX.
- `HomeserverService.subscribeUserEventStreamForPath` tags its `handleError` call with the operation (shared constant in `sentry.constants.ts`, a leaf module, to avoid a factories → sentry → homeserver cycle).
- `ignoreErrors` for global-handler noise that cannot originate from our bundle: webview bridge teardown (`window.webkit.messageHandlers`, `Java object is gone`, `Java exception was raised…`; PUBKY-APP-B6/B7/CJ) and MetaMask `inpage.js` (PUBKY-APP-8G). Deliberately **not** included from #2477: `/reading 'waiting'/` (would hide any app `TypeError` on a `.waiting` property) and bare `'NotAllowedError'` (would hide every future permission failure).
- Docs: `docs/sentry.md` gains "Capture happens at construction, once per error chain" and "Adding a drop rule" (trace the pipeline, match on codes/tags, scope to one operation, test through the pipeline, record short IDs). `docs/error-handling.md` status line updated.

## Tests

Fixtures are obtained by running the real code path, not by hand-building `new AppError(...)`:

- `sentry.test.ts`: `safeFetch` with mocked `fetch` rejecting `AbortError` → dropped; `TimeoutError` → `CONNECTION_FAILED`, reportable; `TypeError('Failed to fetch')` → reportable; Homeserver `INTERNAL_ERROR` on `putFile` → reportable. Once-per-chain: root captured once; wrapper of an `AppError` not captured; `AppError` nested inside a native `TypeError.cause` found; wrapper of a native error still captured; wrapper of a dropped root not resurrected.
- `homeserver.test.ts`: `subscribeUserEventStreamForPath` with the mocked SDK rejecting a `RequestError` → `Server/INTERNAL_ERROR` tagged with the operation and dropped; rejecting `AuthenticationError` → `Auth`, reportable.

`npx vitest run src/libs src/core/services/homeserver src/core/coordinators/mute-list-sync src/app`: 61 files / 1386 tests pass. `tsc --noEmit`, eslint, prettier clean.

## Review follow-ups (pushed to this PR)

- **`beforeSend` enforces the same decision as the factory** (`filterAndScrubErrorEvent` → `shouldDropCapturedExceptionFromSentry`). The factory only controls its own `captureException`; an `AppError` that escapes as an unhandled rejection reaches the SDK again through `globalHandlers` with `hint.originalException`, and Sentry dedupes per error object, not per cause chain. Wrappers and rule-dropped roots are now dropped there too. `hasAppErrorInCauseChain` moved to `error.ts` so both sides share it. Tested through the real SDK pipeline (`Scope` → `NodeClient` → dedupe → `beforeSend` → stubbed transport).
- **Keep a signal for the persistent case.** Since `homeserver-event-stream-connect` drops every routine failure, `MuteListSyncCoordinator` escalates once via `Err.server(SERVICE_UNAVAILABLE)` tagged `muteListEventStreamExhausted` after `MUTE_SYNC_STREAM_FAILURE_ALERT_THRESHOLD` consecutive failures (no `cause`, or the once-per-chain guard would swallow it), with capped exponential reconnect backoff. `aborted-requests` only fires when `context.signalAborted` shows our own signal fired (set by `safeFetch`); browser-driven aborts stay reportable.
- **"Consecutive" is defined by connection health.** A connection is healthy once it delivered an event or stayed open for `MUTE_SYNC_STREAM_HEALTHY_AFTER_MS`, and every way it can end applies that verdict: a throw, a clean close (premature close is a failed iteration, so a homeserver that accepts and immediately closes never reconnects at the base delay forever without reporting), and a controlled teardown (a healthy connection torn down by hide/show or route/auth change clears the streak). Stale loop generations do not feed the streak, and the streak is scoped to the signed-in user.

## Out of scope (tracked in #2494)

- 429 storm (PUBKY-APP-B3): root cause is N+1 hydration on cold `/home` + nexus burst limit of 40. Needs a batching layer / nexus contract discussion. Capturing 429 only on terminal retry failure (TanStack `QueryCache.onError`) is a follow-up.
- SSRF-guard blocks stay reportable per `docs/sentry.md`; changing that is a product decision.
- iOS IndexedDB eviction, `Not Found` on deleted users, nexus write-then-read lag: design/ADR questions.

## Disposition of #2476–#2479

- #2477 — superseded by this PR (closing).
- #2476 — request changes: keep `rateLimited: 1` + 2s floor, restore the delay cap, drop the dedupe map.
- #2478 — request changes: keep the OG coalescer, revert the `Err.auth → Err.validation` reclassification, validate ids at the route boundary with `isPubkyIdentifier`.
- #2479 — request changes: reduce to the `LocalFeedService.read → null` fix, drop the abort-handler change and the bulkSave claim.


src/libs/observability/sentry.utils.ts
```diff
@@ -71,6 +83,29 @@ const APP_ERROR_DROP_RULES: AppErrorDropRule[] = [
...
+      'The mute-list SSE subscribe drops routinely (homeserver deploys, idle timeouts, mobile backgrounding) and ' +
+      'MuteListSyncCoordinator reconnects with backoff by design. Connect failures reach here as ' +
+      'handleError → httpStatusCodeToError(500) → Err.server(INTERNAL_ERROR) tagged with the subscribe operation ' +
+      '(PUBKY-APP-11/1Y/6G/CX). Auth/validation failures on the same operation and every other Homeserver ' +
+      'operation stay reportable.',
+    matches: (error) =>
+      error.service === ErrorService.Homeserver &&
+      error.operation === HOMESERVER_EVENT_STREAM_SUBSCRIBE_OPERATION &&
+      error.category === ErrorCategory.Server,
+  },
```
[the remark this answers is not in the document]
Orlandohub: Agreed — this was the same signal-loss concern I raised against #2477, so it deserves a real answer. Fixed in 74891df:

- `MuteListSyncCoordinator` now counts consecutive stream failures (reset on any completed read) and, when the streak reaches `MUTE_SYNC_STREAM_FAILURE_ALERT_THRESHOLD` (6), constructs one `Err.server(SERVICE_UNAVAILABLE)` tagged `operation: 'muteListEventStreamExhausted'` with the last error's code/operation in `context`. Routine reconnects stay dropped; a persistent outage produces exactly one Sentry event per client, and the loop keeps reconnecting.
- Reconnect delay now doubles per failure up to `MUTE_SYNC_RECONNECT_BACKOFF_MAX_MS` (30s) instead of hammering at a flat 1s, so the threshold corresponds to roughly a minute of continuous failure rather than 6 seconds.
- Deliberately no `cause` on the escalation error — the once-per-chain guard would treat it as a wrapper and skip capture. Documented in `docs/sentry.md` as a new checklist item ("keep a signal for the persistent case").

Tests drive the real loop with fake timers: report fires exactly at the threshold and never again, backoff sequence and cap, and a healthy read resets the streak so a later blip does not report.

src/libs/observability/sentry.utils.ts
```diff
@@ -71,6 +83,29 @@ const APP_ERROR_DROP_RULES: AppErrorDropRule[] = [
...
   },
+  {
+    name: 'aborted-requests',
+    reason:
+      'REQUEST_ABORTED is only produced by safeFetch when fetch rejects with an AbortError DOMException: a caller ' +
+      'called AbortController.abort() (navigation, teardown, deliberate long-poll reset) or the browser cancelled ' +
+      'an in-flight request. Control flow, not a failure; every caller already handles the rejection ' +
+      '(PUBKY-APP-3N/4F). AbortSignal.timeout() raises a TimeoutError DOMException, which safeFetch does not map ' +
+      'to REQUEST_ABORTED, so real timeouts stay reportable.',
+    matches: (error) => error.code === TimeoutErrorCode.REQUEST_ABORTED,
```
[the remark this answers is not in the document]
Orlandohub: Fair. Rather than enumerating operations (which drifts), the rule now keys on evidence the abort was ours: `safeFetch` records `context.signalAborted = options.signal?.aborted === true`, and `aborted-requests` only drops when that is `true`. So `AbortController.abort()` from our own teardown/long-poll reset is dropped, while an `AbortError` with no signal (browser-driven) or with a signal that never fired stays reportable — same as on `dev` today. Fixed in 74891df with three pipeline tests (`safeFetch` + mocked `fetch`: signal fired → dropped; no signal → reportable; signal passed but idle → reportable).

src/core/coordinators/mute-list-sync/mute-list-sync.ts
```diff
@@ -232,7 +244,12 @@ export class MuteListSyncCoordinator {
           }
         }
       } catch (error) {
-        Logger.error('Mute list homeserver event stream failed', { error });
+        this.consecutiveStreamFailures += 1;
+        Logger.error('Mute list homeserver event stream failed', {
+          error,
+          consecutiveFailures: this.consecutiveStreamFailures,
+        });
+        this.reportStreamOutageIfPersistent(error);
```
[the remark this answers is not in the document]
Orlandohub: Correct — the in-flight `subscribe` from a superseded generation could reject after `stop()` reset the streak or after the replacement loop had already failed once. Fixed in a3e3f992: the `catch` now computes `isCurrentLoop = this.state.isStarted && generation === this.loopGeneration` and only increments the streak / considers escalating when true (the local `Logger.error` still fires, tagged `staleGeneration`). Added a test that holds generation 1's subscribe pending, does `stop()`/`start()`, then rejects the stale promise and walks the new loop to the threshold — it fails without the gate (report fires one failure early) and passes with it.

src/libs/error/error.factories.ts
```diff
@@ -56,6 +74,15 @@ function createAppError<C extends ErrorCategory>(
 
   Logger.error(`[${params.service}:${params.operation}]`, error.message, params.context);
 
+  // Once per error chain (ADR-0015 §5.1 Challenge 1): when a caller wraps an AppError
+  // in another Err.* to add its own service/operation, the root was already captured
+  // with the most precise stack and context. Capturing the wrapper too creates a second
+  // Sentry issue (different fingerprint) for the same failure. The Logger line above is
+  // kept so local logs still show the wrapper's operation.
+  if (hasAppErrorInCauseChain(params.cause)) {
```
infin1t3, line 66: **[P2, non-blocking] Automatic Sentry capture bypasses the once-per-chain guard**

Returning the wrapper skips factory capture, but leaves this new error object eligible for Sentry's global handler. Existing unblur actions expose this path: LocalModerationService wraps the model's AppError, while the click handlers invoke ModerationController.unBlur without handling its rejected promise. Sentry deduplicates individual error objects, not their cause chains, so the global handler captures the wrapper again, without the factory's structured tags.

Reproduced with the real browser SDK, including global rejection handling, linked errors, and dedupe: constructing a root and wrapper and passing the wrapper through the installed global rejection handler produced two events for one chain. Consider carrying the capture/drop decision onto wrappers and enforcing it in automatic capture paths, with a regression test that includes the real SDK handler.

src/core/coordinators/mute-list-sync/mute-list-sync.ts
```diff
@@ -232,7 +247,25 @@ export class MuteListSyncCoordinator {
...
       } catch (error) {
-        Logger.error('Mute list homeserver event stream failed', { error });
+        // A stale generation (stopped, or replaced while its subscribe was in flight) must not
+        // feed the streak or the outage report of the loop that superseded it.
+        const isCurrentLoop = this.state.isStarted && generation === this.loopGeneration;
+        if (isCurrentLoop) {
+          // A healthy idle stream never completes a read, so a connection that stayed open long enough
+          // counts as healthy: its eventual drop starts a fresh streak instead of extending the old one.
+          // Immediate connect failures and fast connect→drop cycles still accumulate.
+          const wasSustained =
```
infin1t3: **[P2, non-blocking] Failure counts survive healthy stream teardown and account changes**

The sustained-connection reset only runs when the stream throws. Visibility/auth changes cancel the reader and replace the loop, bypassing this reset and the completed-read reset after the generation check. evaluateStreaming then starts the replacement loop with the same singleton consecutiveStreamFailures value.

Reproduced through the real coordinator/auth store with fake timers:
- Five failed connections, then a successful idle connection for ten minutes, then hide/show, then one connection failure incorrectly emits an outage alert.
- User A reaches six failures and alerts; switching to user B and producing six failures for B emits no new alert because B inherits A's exhausted streak.

Consider preserving healthy recovery during controlled teardown and scoping/resetting failure tracking when the authenticated user changes. This would avoid false alerts, inherited reconnect delays, and missing alerts for another user's outage.

src/core/coordinators/mute-list-sync/mute-list-sync.ts
```diff
@@ -221,6 +233,9 @@ export class MuteListSyncCoordinator {
             break;
           }
 
+          // A completed read (event or clean end) means the stream is healthy again.
+          this.consecutiveStreamFailures = 0;
```
infin1t3: **[P2, non-blocking] Repeated premature EOF never triggers backoff or outage reporting**

The counter resets before checking done. A live subscription that repeatedly opens and closes without delivering an event therefore counts as healthy and reconnects every second indefinitely, without reaching the outage threshold. Clean EOF is a real SDK outcome: pubky 0.8.0 exposes the response event stream as a Web ReadableStream, and the homeserver can close the stream cleanly on broadcast lag or closure.

A focused regression probe returning immediately closed streams produced 62 subscriptions and zero outage reports in 61 seconds. Consider counting premature EOF as a failed iteration unless an actual event or the sustained-connection window establishes health, while preserving normal cancellation handling.

infin1t3 (dismissed):
Approving this as an incremental improvement to error reporting. I found no blocking issue involving data loss, security, or core user flows.

I left three non-blocking P2 comments covering automatic Sentry capture, failure counts across stream lifecycle changes, and premature EOF handling. Leaving it to the reviewer to decide whether to address them here or in follow-up work.

Validation: 61 focused test files / 1,394 tests passed under Node 24. Additional regression probes reproduced the three findings on d0c6b934; temporary probes were removed.

src/libs/error/error.factories.ts
Orlandohub: Addressed in fd8e692f. `beforeSend` is now `filterAndScrubErrorEvent`, which runs `shouldDropCapturedExceptionFromSentry(hint.originalException)` before scrubbing: an `AppError` with an `AppError` in its cause chain, or one matched by a drop rule, returns `null`. Non-`AppError` exceptions are untouched. `hasAppErrorInCauseChain` moved to `error.ts` so the factory and the Sentry utils share one predicate.

You were right that this is broader than wrappers: a rule-dropped root that escapes unhandled was also being sent, since the factory never captured it and nothing downstream knew about the rule.

Tests: the hook directly, plus the real SDK pipeline (`Scope.captureException` → `NodeClient` → dedupe → `beforeSend` → stubbed transport, which is the same hint shape `globalHandlers` produces via `captureEvent(event, { originalException })`). Root captured twice + wrapper → exactly one event; rule-dropped root → none; plain `TypeError` → still delivered. All four new tests fail with the previous `beforeSend: scrubSensitiveData`.

src/core/coordinators/mute-list-sync/mute-list-sync.ts
Orlandohub: Addressed in ac85dd8e. Each connection now carries `connectedAt` / `receivedEvent` / `settled`, and the `finally` block applies the health verdict on controlled teardown too: a connection that had proven healthy (event, or open ≥ `MUTE_SYNC_STREAM_HEALTHY_AFTER_MS`) clears the streak when it is torn down by hide/show or a route/auth change. Your first reproduction (5 failures → 10-minute idle connection → hide/show → 1 failure) is now a test and asserts no report plus base-delay reconnect.

For the account change, the streak is scoped via `failureStreakPubky` and reset at the top of each loop iteration when the pubky differs. Your second reproduction (user A alerts, switch to B, six failures for B) is a test asserting a second report for B. Both tests fail on the previous head.

src/core/coordinators/mute-list-sync/mute-list-sync.ts
Orlandohub: Addressed in ac85dd8e. `done` no longer resets the streak unconditionally: `recordStreamClosed` treats a clean close as routine only if the connection was healthy (delivered an event or stayed open ≥ `MUTE_SYNC_STREAM_HEALTHY_AFTER_MS`); otherwise it is a failed iteration — backoff grows and it counts toward the outage report, with `lastFailure: 'prematureClose'` in the escalation context. Cancellation is unaffected: the loop still checks `isLoopCurrent` before looking at `done`, so a teardown-driven close goes through the controlled-teardown path.

Your reproduction (immediately closed streams) is now a test: subscribe count follows the doubling backoff and the report fires exactly at the threshold. Fails on the previous head.

infin1t3 (approved):
LGTM.

PR #2501, opened 2026-09-09
fix(auth): reset settings store on sign-in and fail stale writes on read error

### LINKED ISSUE #2500: Moderation follow: reset settings store on sign-in and fail stale-client writes when the pre-write read fails
## Task Description

Two follow-ups to #2450 / #2451 raised in the [design thread](https://synonymworkspace.slack.com/archives/C09HQTLAARF/p1788264096901029).

1. The settings store only resets in `cleanupLocalState` (logout, failed session restore). `signIn`, `signUp`, and `wrapAuthFlow` reset the migration store but not settings. Signing into account B without logging out of A keeps A's document in the store; when B has no `settings.json` (404) or A's local timestamp is newer than B's remote, A's document, `privacy.moderationBot` included, is pushed to B's homeserver. B is recorded as processed and never gets the default follow. A's other preferences leak into B on the same branches.

2. `SettingsApplication.commitUpdate` reads the remote document before writing when the local state has no `privacy.moderationBot`. If that read fails it currently logs and writes anyway, which can erase the remote field and cause the re-follow the read was meant to prevent. Settings writes are local-first (`useSettingsActions` neither reverts nor toasts on failure, and the next sign-in pushes newer local state through the local-newer branch, which carries the remote field over), so failing the write loses nothing and needs no user action.

## Acceptance Criteria

- [ ] `signIn`, `signUp`, and `wrapAuthFlow` reset the settings store alongside the migration store, so a fresh sign-in starts from defaults and conflict resolution lets the account's remote document win or pushes real defaults.
- [ ] When the pre-write remote read in `commitUpdate` fails with a non-404 error, the write is rejected and no PUT is sent.
- [ ] A 404 on the pre-write read still writes (no remote document, nothing to carry over).
- [ ] Existing behavior unchanged when the local state already has `privacy.moderationBot` (single PUT, no read).

## References

- #2450, #2451
- Design discussion: https://synonymworkspace.slack.com/archives/C09HQTLAARF/p1788264096901029

Orlandohub:
## Summary

Follow-ups to #2451 raised by Chris in the [design thread](https://synonymworkspace.slack.com/archives/C09HQTLAARF/p1788264096901029).

- reset the settings store in `signIn`, `signUp`, and `wrapAuthFlow`, next to the existing migration-store reset
- make `SettingsApplication.commitUpdate` reject instead of pushing blind when the pre-write remote read fails

## Why

**Settings store reset.** The settings store only reset in `cleanupLocalState` (logout, failed session restore). Signing into account B without logging out of A kept A's document in the store. When B's remote is newer the store is replaced before the follow runs, so that branch was fine. But when B has no `settings.json` (404) or A's local timestamp is newer than B's remote, A's whole document, `privacy.moderationBot` included, was pushed to B's homeserver: B was recorded as processed and never received the default follow, and A's other preferences leaked into B. A reset store has `updatedAt: 0`, so conflict resolution now always lets the account's remote document win or pushes real defaults.

**Fail the write on read error.** `commitUpdate` reads the remote document before writing when the local state has no `privacy.moderationBot`, to carry the field over. A failed read used to log and push anyway, which could erase the field and cause the very re-follow the read exists to prevent. Settings writes are local-first (`useSettingsActions` neither reverts nor toasts on failure), so the change stays in the persisted store and the user keeps seeing it. It is not retried automatically: sign-in now resets the store and session restore does not sync settings, so it reaches the homeserver with the next successful settings change, or is dropped at the next sign-in when the remote document wins. That is the same outcome as any failed settings PUT today, and a dropped toggle beats erasing the field and re-following the bot. A 404 on the read still writes: there is no remote document to protect.

## Trade-off

- Re-signing into the *same* account without logging out used to push a newer local document (for example a toggle whose PUT failed) through the local-newer branch. With the reset it no longer does; the remote document wins. Accepted: the scenario is sign-in-without-logout after a failed PUT, and keeping that path would require the settings store to remember its owner account.

## Testing

- `signIn`, `signUp`, and `wrapAuthFlow` tests assert the settings store reset
- `commitUpdate` read-failure test flipped to expect rejection with no PUT; new test for the 404 read case
- `src/core` + `src/hooks`: 292 files, 5666 tests passed
- typecheck, lint, format via pre-commit

No UI changes; no VRT impact.

Closes #2500


SHAcollision (dismissed):
LGTM, ship it :+1:

one small nit, not blocking: the comment in commitUpdate (and the PR body) says a failed write gets pushed on next sign-in via the local-newer branch. but sign-in now resets the store, so I think that only happens on session restore / reload. on a real sign-in the local change is just dropped, which is ok (user toggles again), maybe just fix the wording so it does not promise a retry that does not happen. wdyt?

SHAcollision (approved):

PR #2517, opened 2026-09-11
chore: set cache explicitly and track ci-workflows main

mdozhdev:
Two ordered changes to the `ci-workflows` references here. They are in one PR because merging them separately, in the wrong order, silently drops caching.

## 1. Cache settings written out explicitly

Two `build_and_push` calls set no cache inputs at all. The builds are cached anyway, because the shared action substitutes a **registry** cache when the inputs are unset — a backend nothing else here uses, chosen by nobody, which also pushes a `:cache` tag into Artifact Registry per image.

| Workflow | Cache before | Cache after |
|---|---|---|
| `release-docker.yml` | `type=gha` | unchanged |
| `build-docker-image.yml` | *(unset)* → registry | `type=gha` |
| `e2e-test.yml` | *(unset)* → registry | `type=gha`, scoped per matrix image |

`build-docker-image.yml` — identical to what `release-docker.yml` already uses:

```yaml
cache_from: type=gha
cache_to: type=gha,mode=max
```

`e2e-test.yml` — same, plus a scope:

```yaml
cache_from: type=gha,scope=${{ matrix.image }}
cache_to: type=gha,mode=max,scope=${{ matrix.image }}
```

**Why the scope.** That workflow builds several images from a matrix (`pubky-nexus`, `homeserver-testnet`). The registry cache separated them for free, because the image name was part of the cache address. Plain `type=gha` has no such separator, so both matrix legs would write to one cache and evict each other every run — zero cache hits plus the upload cost. `scope` gives each image its own bucket. `build-docker-image.yml` needs none: it builds one image.

## 2. References moved back to `@main`

`ci-workflows` holds only our own CI code, and the third-party actions it calls are SHA-pinned **inside** that repo — so the supply-chain property does not depend on what this reference resolves to. Pinning added a second review of our own changes, once per consuming repo, at 8 PRs per release. Requiring review once where the change is written serves the same purpose: pubky/ci-workflows#10.

## Why one PR

This repo pins `v1.0.0`, which still substitutes a registry cache for unset inputs. Tracking `main` moves it to `v1.0.2`, where **an unset input means no cache**. Unpinning without part 1 in the same commit range would silently drop caching from `build-docker-image.yml` and `e2e-test.yml`.

## Effect on builds

Switching backend abandons the existing registry cache, so **the first build after merge rebuilds from scratch**. Cached as normal after that. GitHub's cache is 10GB per repository, evicted after roughly a week unused, and scoped per ref — in exchange it stops publishing a `:cache` tag per image.

Related: pubky/pubky-stack#311, pubky/ci-workflows#10.

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -122,8 +122,8 @@ jobs:
         uses: pubky/ci-workflows/.github/actions/docker/build_and_push@e6d3b38b5ebfa97b86595cb44f3a86695e473b09 # v1.0.0
         with:
           image: 'pubky-app'
-          cache_from: type=registry,ref=${{ env.registry }}/pubky-app:cache
-          cache_to: type=registry,ref=${{ env.registry }}/pubky-app:cache,mode=max
+          cache_from: type=gha
+          cache_to: type=gha,mode=max
           context: '.'
           registry: ${{ env.registry }}
           platforms: ${{ inputs.platforms }}
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -122,6 +122,8 @@ jobs:
         uses: pubky/ci-workflows/.github/actions/docker/build_and_push@e6d3b38b5ebfa97b86595cb44f3a86695e473b09 # v1.0.0
         with:
           image: 'pubky-app'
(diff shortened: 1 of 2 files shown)


andrei-21 (dismissed):

andrei-21 (dismissed):

SpontaneousOverthrow (approved):
LGTM

PR #2520, opened 2026-09-11
feat(collections): add copy link action to the collection page

### LINKED ISSUE #2515: [Bug]: Collection page has no way to copy the collection link
### Description

There is no way to copy the URL of a collection. On the collection page (`/collections/[userId]/[postId]`) the only share affordance is the **Share** button in the hero, and that button opens the repost dialog ("Share Collection" → publishes a repost post), not a way to copy or hand over the link.

On desktop the link can still be taken from the address bar, so this is mostly invisible. In the installed PWA there is no address bar, and the button label is hidden below `lg` (`hidden lg:inline` in `CollectionHero`), so on mobile the user sees an icon-only button whose icon is `StickyNote` — it reads as a copy button and then creates a post instead.

### Expected Behaviour

A collection can be shared as a URL: a copy-link action reachable from the collection page (and ideally the native share sheet on mobile, `navigator.share` when available).

### Notes

- Regular posts already have this: the overflow menu built by `usePostMenuActions` offers `Copy link to post`, and it works for collections too (`PostMenuActions` already has a collection branch), but no collection surface renders `PostActionsBar` — `CollectionCard` documents that "card-level collection actions live on the dedicated page header", and the hero only renders Add content / Share / Reorder / Edit / Delete.
- The Share button was deliberately wired to the repost dialog (see the comment in `CollectionHero`: "Sharing a collection = reposting the underlying post"), so this needs a product decision on whether copy-link replaces it, sits next to it, or lives in a menu.
- Related but distinct: #2121 is the "reposted" vs "shared" wording in the feed; #1691 is receiving shares into the app (share target), not copying a link out of it.

### Likely files

- `src/components/organisms/Collections/CollectionHero/CollectionHero.tsx`
- `src/hooks/useCopyToClipboard/useCopyToClipboard.tsx`
- `src/hooks/usePostMenuActions/usePostMenuActions.tsx` (existing copy-link action)
- `src/app/routes.ts` (`getCollectionRoute`)

talosmachina:
## Summary

The collection page had no way to hand over the collection URL. The hero's only share affordance, **Share**, publishes a repost, and below `lg` its label is hidden, so the `StickyNote` icon reads as a copy button and then creates a post instead.

This adds a **Copy link** action next to it, for owners and viewers alike.

- New `useShareUrl` hook: uses the native share sheet when the browser exposes one (`navigator.share`, guarded by `navigator.canShare`), and falls back to `useCopyToClipboard` everywhere else. A dismissed sheet (`AbortError`) stays silent instead of writing to the clipboard behind the user's back.
- `CollectionHero` renders a `Copy link` button, icon-only below `lg` and labelled from `lg` like its neighbours, which hands over `${getDefaultUrl()}${getCollectionRoute(authorPubky, postId)}`. The URL comes from the runtime default URL, the same source the profile links use, so a published collection link stays canonical and SSR-safe.
- `Share` keeps its repost behaviour and is untouched. The two buttons now sit side by side with distinct icons.
- A collection URL is public, so the action needs no auth and also works for guests.

## Tests

- `npx vitest run --project unit src/components/organisms/Collections/CollectionHero/CollectionHero.test.tsx` -> `Tests 49 passed`
- `npx vitest run --project unit src/hooks/useShareUrl/useShareUrl.test.tsx` -> `Tests 7 passed`
- `npm run typecheck` -> clean
- `npx eslint` and `npx prettier --check` on the changed files -> clean

## Visual regression

`src/test/vrt/feed/Collections.vrt.test.tsx` renders this hero, so the `single-collection-{grid,list,visual}-{desktop,mobile}` baselines change on both platforms. No baselines are committed here (the macOS ones cannot be produced from this machine); the **VRT Update Baselines** workflow needs to run on this branch.

A local chromium capture of those four baseline names was used to check the change itself: the pixel diff is confined to the action row, one line, with nothing below it moving.

Closes #2515

Approved by @Orlandohub.

### DIFF
--- a/src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
+++ b/src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
@@ -2,16 +2,18 @@
 
 import { useState } from 'react';
 import { useRouter } from 'next/navigation';
-import { Check, Loader2, Minus, Move, Pencil, Plus, StickyNote, Trash2, X } from 'lucide-react';
-import { APP_ROUTES, getUserProfileUrl } from '@/app/routes';
+import { Check, Link, Loader2, Minus, Move, Pencil, Plus, StickyNote, Trash2, X } from 'lucide-react';
+import { APP_ROUTES, getCollectionRoute, getUserProfileUrl } from '@/app/routes';
 import { Button } from '@/atoms/Button/Button';
 import { Card, CardContent } from '@/atoms/Card/Card';
 import { Container } from '@/atoms/Container/Container';
 import { Typography } from '@/atoms/Typography/Typography';
+import { getDefaultUrl } from '@/config/metadata';
 import { useBookmark } from '@/hooks/useBookmark/useBookmark';
 import { useDeletePost } from '@/hooks/useDeletePost/useDeletePost';
 import { usePostReplyRepostDialogs } from '@/hooks/usePostReplyRepostDialogs/usePostReplyRepostDialogs';
 import { useRequireAuth } from '@/hooks/useRequireAuth/useRequireAuth';
+import { useShareUrl } from '@/hooks/useShareUrl/useShareUrl';
 import { useUserProfile } from '@/hooks/useUserProfile/useUserProfile';
 import { parseCollectionContent } from '@/libs/post/collectionContent';
 import { resolveCollectionCoverImage } from '@/libs/post/collectionCoverImage';
@@ -43,8 +45,14 @@ import type { CollectionHeroContentProps, CollectionHeroProps } from './Collecti
  *
  * The Hero → feed → Sections structure is identical for both owner and
  * other-user views; only the action buttons differ:
- *   - owner   → Content / Share / Edit / Delete.
- *   - other   → real Follow / Unfollow (via `useBookmark`) + Share placeholder.
+ *   - owner   → Content / Share / Copy link / Edit / Delete.
+ *   - other   → real Follow / Unfollow (via `useBookmark`) + Share placeholder
+ *     + Copy link.
+ *
+ * `Share` reposts the collection and stays what it always was; `Copy link` is
+ * the one that hands the collection URL over (native share sheet where the
+ * browser offers one, clipboard otherwise), so the URL is reachable without an
+ * address bar.
  */
 export function CollectionHero({
   authorPubky,
@@ -84,6 +92,7 @@ export function CollectionHero({
 
 function CollectionHeroContent({
   authorPubky,
+  postId,
   compositeId,
   postDetails,
   layout,
@@ -149,6 +158,15 @@ function CollectionHeroContent({
   });
   const handleShare = () => {
     requireAuth(openRepostDialog);
+  };
+
+  // Collection URL, built from the runtime default URL rather than
+  // `window.location.origin` so the copied link is the canonical one for the
+  // deployment (same source the profile links use) and stays SSR-safe.
+  const collectionUrl = `${getDefaultUrl()}${getCollectionRoute(authorPubky, postId)}`;
+  const { shareUrl } = useShareUrl({ title: title || 'Collection' });
+  const handleCopyLink = () => {
+    void shareUrl(collectionUrl);
   };
 
   const [isEditDialogOpen, setIsEditDialogOpen] = useState(false);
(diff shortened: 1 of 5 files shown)


src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
```diff
@@ -284,6 +302,20 @@ function CollectionHeroContent({
                   {'Share'}
                 </Typography>
               </Button>
+              <Button
+                variant="secondary"
+                size="icon"
+                onClick={handleCopyLink}
+                disabled={isDeleting || isReorderActive}
+                aria-label="Copy link"
```
infin1t3, line 311: **[medium | naming]** @talosmachina

The control is labeled and `aria-label`led **Copy link**, and the handler is `handleCopyLink`, but `useShareUrl` prefers `navigator.share` whenever the Web Share API is present. That is not mobile-only — desktop Safari exposes it, so a click opens a share sheet and skips the clipboard toast.

#2515 asked for a copy-link action, with a native sheet only as a mobile extra. Users who click **Copy link** expect the URL on the clipboard. On share-capable desktops they get a sheet instead, next to a **Share** button that already means “repost”.

**Fix:** Keep clipboard as the default path. Gate `navigator.share` on a real mobile/PWA signal, not merely `typeof navigator.share`. If the sheet stays as the primary path, change the visible label / `aria-label` so they describe share, not copy.

On the fallback question in the thread: if copy succeeds after a non-`AbortError` share failure, do not add a second “share failed” toast — the existing **Link copied to clipboard** toast already confirms the outcome. The labeling/gating above is the change that matters.

src/hooks/useShareUrl/useShareUrl.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  return { shareUrl };
+}
+
+function canUseNativeShare(url: string): boolean {
+  if (typeof navigator === 'undefined' || typeof navigator.share !== 'function') return false;
+  if (typeof navigator.canShare !== 'function') return true;
+
+  try {
+    return navigator.canShare({ url });
```
infin1t3: **[medium | error-handling]** @talosmachina

`canUseNativeShare` calls `navigator.canShare({ url })`, then `shareUrl` calls `navigator.share({ url, title })`. `CollectionHero` always passes a title (`title || 'Collection'`), so the checked payload and the shared payload never match.

`canShare` answers whether the equivalent `share()` call would succeed. `canShare({ url }) === true` does not mean `share({ url, title })` will succeed; a `TypeError`/`DataError` then hits the clipboard fallback.

**Fix:** Build one `ShareData` object and pass that same object to `canShare` and `share`.

src/components/organisms/Collections/CollectionHero/CollectionHero.tsx
```diff
@@ -284,6 +302,20 @@ function CollectionHeroContent({
                   {'Share'}
                 </Typography>
               </Button>
+              <Button
```
infin1t3, line 306: **[medium | visual-regression-baseline]** @talosmachina

This new button sits in the hero action row that `src/test/vrt/feed/Collections.vrt.test.tsx` screenshots (`single-collection-{grid,list,visual}-{desktop,mobile}`). No files under `src/test/vrt/feed/__screenshots__/` are in this branch.

AGENTS.md requires regenerating baselines when a VRT-covered surface changes. VRT does not run on the PR itself (only on push to `dev`/`master`), so this will fail after merge unless the **VRT Update Baselines** workflow runs first. You already noted this in the PR body.

**Fix:** Run that workflow on this branch and land the updated `single-collection-*` shots before merging.

infin1t3 (commented):
## Code review

@talosmachina three items to check before merge. Verdict: **Needs attention** — the feature matches #2515 and the layering is clean; the notes below are about labeling/gating native share, matching the `canShare`/`share` payload, and refreshing collection VRT baselines.

### What's done well
- Canonical URL via `getDefaultUrl()` + `getCollectionRoute()` (SSR-safe, same source as profile links)
- `Share` (repost) left alone; copy works for guests without `requireAuth`
- `useShareUrl` is a small `.ts` hook with solid tests; `AbortError` does not write the clipboard; toast copy is static through `useCopyToClipboard`

Inline comments are on the relevant lines.

infin1t3 (approved):

PR #2521, opened 2026-09-11
chore(locks): vendor @pubky/locks-sdk so the image can build

mdozhdev:
## Why

`@pubky/locks-sdk` is Rust compiled to WASM and is published to no registry. `package.json` therefore omitted it deliberately, and `docs/locks.md` told you to build it from `pubky/locks` and copy it into `node_modules` by hand.

**That can never work in a container.** The Dockerfile's dependency stage runs `npm ci`, which erases anything placed in `node_modules` beforehand. Every preview build on the locks branches failed the same way:

```
Module not found: Can't resolve '@pubky/locks-sdk'
> Build failed because of webpack errors
```

Four consecutive failures on `feat/2368-locks-reader-payment`, and `epic/locks-v1` was in the same state — zero `node_modules/@pubky` entries in `package-lock.json`, 14 source files importing it.

## What this does

- declares `"@pubky/locks-sdk": "file:vendor/locks-sdk"` in `package.json`
- commits the wasm-pack output to `vendor/locks-sdk/` (1.2 MB `.wasm` + JS glue + types)
- adds `COPY vendor ./vendor` to the Dockerfile **before** `npm ci`, so the path resolves
- excludes `vendor/` from prettier and eslint — generated code, regenerated wholesale
- sets `PUBKY_RUNTIME_LOCK_SERVER` and `PUBKY_RUNTIME_PAYKIT_SERVER_URL` on PR previews
- rewrites the `docs/locks.md` section that described the `node_modules` copy

npm installs `file:` dependencies as symlinks and bun copies them; either way the package survives `npm ci`, which the hand-copy did not.

## Verified

- **All 5 test shards pass** and **Check NextJS Build passes** — both were failing on this branch before
- `npm ci` (the image) and `bun install` (CI) each resolve the dependency
- `docker build --target deps` succeeds; `node_modules/@pubky/locks-sdk -> ../../vendor/locks-sdk` in the image
- the preview deploys and serves: https://pubky-app-pr-2521-fzxmjul7ya-oa.a.run.app

## Provenance

Built from `pubky/locks` @ `a9d52b886a98083d6d52423b9f56f3966aefc4df`, Rust 1.91.1, wasm-pack 0.13.1. `vendor/locks-sdk/README.md` records this and pins the rebuild commands, so the binary can be reproduced and audited rather than taken on trust.

Two traps are documented there because both fail silently: wasm-pack emits a `pkg/.gitignore` containing `*`, which excludes every vendored file and commits an empty package; and it names the package `locks-sdk-wasm` after the crate, not the scoped name the imports use.

## Known gaps

- **Code Quality is red for reasons that predate this PR.** Prettier fails on `PostBody.tsx`, `ProfileUnlocked.tsx` and `usePaykitSetupFlow.utils.test.ts` — the same three files fail on #2488, already merged into `epic/locks-v1`. Not touched here to avoid conflicting with in-flight work.
- **Previews now share staging's Lock and Paykit servers**, so test locks and payments land in their live data. Staging's own frontend is untouched.
- **`/connect` needs pubky/pubky-stack#305.** The staging Lock Server allowlists only `http://127.0.0.1:8080`, so it answers `400 invalid return_to` for the preview — and for `staging.pubky.app` itself. That PR widens it, and needs a manual Locks deploy to take effect.

## This is temporary — tracked in #2526

Vendoring keeps a binary in git and a hand-synced version. Publishing the SDK to a registry replaces `vendor/` and the `file:` dependency with an ordinary version pin, and #2526 carries what that takes on both sides.


Taewa:
@greptile-apps
Review again plz

PR #2522, opened 2026-09-11
fix(locks): polish lock UI and mobile tabs

### LINKED ISSUE #2400: [Bug]: Unlock modal scrolls sideways when the lock title is long
## Description

In the unlock modal for locked content (`DialogUnlockContent`), a long lock title makes the modal scroll horizontally.

## Expected Behaviour

The title should stay inside the modal width (truncate or wrap). No horizontal scrolling.

## Steps to Reproduce

1. Create a locked post with a very long title.
2. As a reader, open the post and click to unlock.
3. In the modal, scroll sideways.

https://github.com/user-attachments/assets/cf9548ea-3b2d-4c1e-b0a3-10f32f184170

### LINKED ISSUE #2403: [Bug]: Profile mobile tab bar clips its buttons, and the Unlocked screen has no bottom padding
## Description
Image

Two small layout problems on the profile page. Both were found while testing the new Unlocked tab, but they are layout issues rather than Locks logic.

### 1. The mobile tab bar clips its buttons at normal phone widths

`MobileTabBar` lays its items out as `flex w-full` with each cell set to `flex-1`, and there is no horizontal scrolling (`src/components/molecules/MobileTabBar/MobileTabBar.tsx:45` and `:55`). A cell cannot shrink below its own content: a 20px icon plus `px-2.5` on the button is about 40px wide.

Your own profile now has 10 tabs — Profile, Notifications, Posts, Replies, Followers, Following, Friends, Tagged, Collections, Unlocked — so the bar needs roughly 400px. On a 375px-wide phone the last buttons are cut off and cannot be reached.

Adding the Unlocked tab is what pushed the bar over the edge, but the missing scroll lives in the shared `MobileTabBar`, which Hot and Settings also use.

### 2. The Unlocked screen has no bottom padding

`ProfilePageLayoutWrapper` sets no bottom padding (`src/components/molecules/ProfilePageLayoutWrapper/ProfilePageLayoutWrapper.tsx:11`), and `ProfileUnlocked` does not add any of its own. The last card sits directly against the bottom edge of the window.

## Expected Behaviour

1. The mobile tab bar scrolls horizontally when the tabs do not fit, so every tab stays reachable and no button is cut off.
2. The Unlocked screen leaves space below the last card, so the list does not touch the bottom edge of the window.

## Steps to Reproduce

**Tab bar**

1. Sign in and open your own profile at a phone-sized viewport, for example 375px wide.
2. Look at the tab bar under the header. The right-most buttons are clipped, and the bar cannot be scrolled sideways to reach them.

**Bottom padding**

1. Sign in and go to `/profile/unlocked` with at least one unlocked post.
2. Scroll to the end of the list. The last card touches the bottom edge of the window.

---

Reported by @secondl1ght while reviewing #2301. Part of #1998.

### LINKED ISSUE #2427: Locks: creator — dim the whole Unlock control in the composer preview
Sub-issue of #1998.

## Problem

Image

While a creator writes a lock announcement post, the composer shows a preview of the locked post card. The Unlock control in that preview is disabled (nothing happens on click), but only the Unlock button looks disabled. The rest of the control — the pill container, the shield icon, and the masked password text in brand color — keeps its normal full-strength look. The reader cannot tell the control is inert.

## Expected

When the card is disabled, every part of the Unlock control looks disabled, not only the button.

## Where

`LockedPostCard` rendered by `PostInput` while the lock switch is on. The card's `disabled` prop already gates the click; it should also gate the styling of the mask part.

## Scope

- Dim the pill container and the shield + mask text together with the button when `disabled` is true
- No behavior change; the click is already blocked

### LINKED ISSUE #2428: Locks: creator — link "Pubky Ring" to pubkyring.app in the Enable Locks step
Sub-issue of #1998.

## Problem

Image

The Enable Locks step of the Locks auth dialog (#2080) says "Use **Pubky Ring** to authorize Locks server to manage your Locks data." The words "Pubky Ring" are bold text only. A user who does not have the app has no way to find it from here.

## Expected

"Pubky Ring" is a link to https://pubkyring.app that opens in a new tab, the same way the sign-in screen already links it.

## Where

`DialogLocksAuth`, the description of the Enable Locks step. The app already keeps the URL in runtime config (`pubkyRingUrl`); use that instead of a new string.

## Scope

- Turn the bold "Pubky Ring" text into an external link
- Keep the bold style

Taewa:
### 1. Long lock titles
- Wrap long titles inside the unlock dialog without horizontal overflow.
- Keep reader-card titles readable within a two-line limit.

### 2. Mobile profile tabs
- Make overflowing tabs horizontally scrollable on small screens.
- Keep the active tab visible after navigation, tab-list changes, and viewport resizing without moving the page itself.
- Confirm the existing mobile bottom spacing also covers the Unlocked view.

### 3. Disabled unlock preview
- Dim the full interactive Unlock pill, including its wallet and price/mask, while leaving the card title and artwork unchanged.

### 4. Setup links
- Make Pubky Ring and Bitkit names green, bold links that open their official sites in a new tab.

## Verification
- Relevant unit tests and type checking pass.
- Mobile tab overflow and active-tab visibility pass in Chromium, Firefox, and WebKit.
- Hot and Settings mobile VRTs pass. Own-profile mobile screenshots remain blocked by pre-existing epic baseline drift where the Unlocked tab is absent.

Closes #2400
Closes #2403
Closes #2427
Closes #2428


src/components/organisms/PostInput/PostInput.tsx
```diff
@@ -234,8 +243,12 @@ export function PostInput({
...
         uploadingCount > 0,
       ) &&
-      // Blocking the click is what prevents an orphaned lock: the publish creates the lock first.
-      (!isLockEnabled || isLockTeaserWithinLimit({ lock_title: lockTitle, teaser_description: content }))
+      // Validate the serialized announcement envelope before publish or edit reaches its write. The
+      // title is required like an article's: the card only shows a placeholder when it is blank, so an
+      // empty one reads as set and would be written as an empty string.
+      (!isLockMode ||
+        (activeLockTitle.trim().length > 0 &&
+          isLockTeaserWithinLimit({ lock_title: activeLockTitle, teaser_description: content })))
```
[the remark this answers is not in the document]
Taewa: This is intentional. A lock title alone is not a valid announcement: teaser text is required, while attachments are optional. Save should remain disabled when `teaser_description` is empty.

src/components/organisms/DialogEditPost/DialogEditPost.tsx
```diff
@@ -50,9 +50,10 @@ export function DialogEditPost({ open, onOpenChangeAction, postId }: DialogEditP
           autoFocusTextarea={!isArticle}
           onContentChange={handleContentChange}
           editPostId={postDetails.id}
-          editContent={postDetails.content}
+          editContent={teaser ? teaser.teaser_description : postDetails.content}
           editIsArticle={isArticle}
           editAttachments={postDetails.attachments ?? []}
+          editLock={lockUrl && teaser ? { lockUrl, title: teaser.lock_title } : undefined}
```
secondl1ght, line 56: When `lock` is set but the content fails `parseLockTeaserContent` (a half-envelope like `{"lock_title":"X"}`, unrelated JSON, or plain text), this falls back to a plain-text edit: raw content in the textarea, no lock card, Save enabled. `usePost.edit` then writes `content.trim()`, and `toEdit` now keeps `lock` via `new_with_lock`, so the stored post is `lock` + non-JSON content. On read, `LockContentParser.parse` returns null and `LockedPostContent` hits `if (!lockContent) return null;`, so the post renders as nothing for every reader. A half-envelope renders fine today through the lenient `.catch('')` reader, so this trades a visible post for a blank one.

Root cause is two envelope definitions with different strictness (`lockPostContentSchema` vs the hand-rolled `parseLockTeaserContent`). One shared schema (strict base, lenient reader variant derived from it) would keep read and edit in agreement. Short of that, the fallback should either stay in teaser mode using the lenient parse, or refuse to save non-envelope content onto a lock post.

src/components/organisms/PostInput/PostInput.tsx
```diff
@@ -308,6 +321,11 @@ export function PostInput({
     // eslint-disable-next-line react-hooks/exhaustive-deps -- toast is an external side-effect, not a dependency
   }, [variant, editContent, editIsArticle]);
 
+  useEffect(() => {
+    if (isEdit) setEditLockTitle(editLock?.title ?? '');
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- seed once per edit target; live row changes must not overwrite user input
+  }, [variant, editPostId]);
```
secondl1ght, line 327: This seeds the title once per `editPostId`, but the body effect just above re-seeds on every `editContent` change. `usePostDetails` is a live query and `PostApplication.commitEdit` writes Dexie first and rolls back on a failed PUT, so a failed save leaves the composer half-reverted: the body snaps back to the row's teaser while the title keeps the draft. Pressing Save again writes `{new title, old body}` and silently drops the body edit. Plain posts revert wholly, so this is new to lock edits. Key both effects the same way (once per edit target, as the comment here intends) or re-seed both together.

src/components/molecules/LockedPostCard/LockedPostCard.tsx
```diff
@@ -141,7 +141,7 @@ export function LockedPostCard({
         <div
           className={cn(
             'relative flex w-fit items-center gap-1 rounded-full bg-card p-1',
-            isDisabled && 'cursor-not-allowed',
+            isDisabled && 'cursor-not-allowed opacity-50',
```
secondl1ght, line 144: `opacity-50` on the wrapper applies to every disabled card, not only the composer preview. `LockedPostContent` passes `onUnlock` only once `priceSats` resolves, so every reader card now renders dim and snaps to full opacity when lock.json lands, and the creator's own inert card (`isOwnLock`, whose comment says the price/terms should stay visible) shows its price at 50%. Non-payment locks stay dimmed permanently. Consider scoping the dim to the composer (only when `editableTitle` is set) or dimming just the price segment when disabled.

src/components/organisms/DialogEditPost/DialogEditPost.tsx
Taewa: Fixed in [30e33e9](https://github.com/pubky/pubky-app/pull/2497/commits/30e33e937d898de21908b7ead2537318fdb174fd)

src/components/organisms/PostInput/PostInput.tsx
Taewa: Fixed in [30e33e9](https://github.com/pubky/pubky-app/pull/2497/commits/30e33e937d898de21908b7ead2537318fdb174fd)

secondl1ght (approved):

src/components/molecules/LockedPostCard/LockedPostCard.tsx
Taewa: Here is the UI decision by @aldertnl

1. When create sees his lock content - only dim entire interaction pill
2. When reader sees locked content - no dim, reader can interact
3. When reader sees "unlocked" content - there is no pill? reader sees the content
4. While it's loading (skeleton?) - display entire interaction pill with spinner

1,2 and 3 are already implemented. 
For 4, I created a follow-up ticket ( https://github.com/pubky/pubky-app/issues/2573 )

secondl1ght (approved):

PR #2525, opened 2026-09-14
docs(agents): add pubky-app-development skill and fix stale facts

talosmachina:
Requested by @SHAcollision.

## What

Adds `.cursor/skills/pubky-app-development/SKILL.md` and corrects four facts in `AGENTS.md`.

The skill is meant to be read before a change, not after: golden rules, a per-layer may / may-not table, the controller naming contract, the local-first read and TTL failure modes, verification commands, and a definition of done. Where it overlaps `docs/` or an ADR it points at them rather than restating them.

## Why a skill file rather than more AGENTS.md

`.cursor/skills/<name>/SKILL.md` is how this repo already ships project skills (`code-review`, `pubky-staging-invite`), and `AGENTS.md` stays scannable at 79 lines. The `Documentation` section now points at the skill so agents reach it from the file every agent reads.

## AGENTS.md corrections

- **VRT path was wrong.** The example, `src/components/templates/Feed/Home/Home.vrt.test.tsx`, does not exist. All 16 VRT test files live under `src/test/vrt/<area>/` (feed, landing, onboarding, post, profile, settings) and none under `src/components`. An agent following the current text looks in the wrong directory.
- **Local-first reads.** Documents `useLocalFirstQuery` at `src/hooks/useLocalFirstQuery/`, which is how ADR-0011 is implemented. `docs/` does not mention the hook, so today the pattern is only discoverable by reading call sites.
- **Dexie schema.** `src/core/database/franky/franky.ts` declares a single `this.version(DB_VERSION).stores({...})`. There is no incremental migration chain: a `DB_VERSION` mismatch calls `recreateDatabase()` and runs `indexedDB.deleteDatabase`. Worth stating before someone bumps the version as a side effect of a feature.
- **Copy / i18n.** Inline US-English literals at the call site, with the pointer to `docs/migrations/2305-i18n-conflict-guide.md`.

## Verification

Documentation only, no code paths touched. Every path, symbol and issue reference added to `AGENTS.md` and the skill was checked against the tree on `dev`.


AGENTS.md
```diff
@@ -41,17 +41,20 @@ Use `Err.*` factories (never raw `Error`). Factories log automatically — don't
...
+- **Visual regression tests (VRT)**: tests live in `src/test/vrt/<area>/*.vrt.test.tsx` (feed, landing, onboarding, post, profile, settings; the `images/` folder holds fixtures only) with pixel baselines checked in under the sibling `__screenshots__/` folder, not next to the component. When you change a UI surface, check whether a VRT covers it. If yes, the baseline likely needs regenerating: surface that to the user before reporting the task done. PR CI does not run VRT or update baselines; `.github/workflows/vrt-update-baselines.yml` is manual (`workflow_dispatch`) and refuses to run on `master`, so it is dispatched on `dev` or a feature branch. If you're touching a template-level surface that has no VRT yet, mention adding one as an option. The VRT harness lives in `src/test-utils/vrt.tsx`; fixtures in `src/test/fixtures/`; deterministic mocks in `src/test/mocks/`.
```
Orlandohub, line 52: **VRT facts are still wrong here (this is the exact bullet the PR set out to correct).** Verified against the tree at `c7a4949`:

1. `images` is not a VRT area. `src/test/vrt/images/` holds three `.webp` fixtures and nothing else — no `*.vrt.test.tsx`, no `__screenshots__/`. The real areas are `feed, landing, onboarding, post, profile, settings` (17 test files, not 31 — the 31 in the PR body counts the `__screenshots__/<file>.vrt.test.tsx` *directories* as files).
2. `vrt-update-baselines` is not "on `dev`/`master`". `.github/workflows/vrt-update-baselines.yml` is `workflow_dispatch` only, and both jobs start with a `Refuse master` step that `exit 1`s when `github.ref_name == 'master'`. It is meant to be dispatched from `dev` (opens a PR from `vrt-update-baselines` → `dev`) or from a feature branch (commits to that branch). An agent following this text would trigger a guaranteed failure.
3. `docs/visual-regression-testing.md` already documents all of this correctly (run/regenerate commands, the CI table, "CI is the source of truth"), but neither this bullet nor the skill links it. Since the skill's stated principle is to point at `docs/` rather than restate, this bullet should point there.

Suggested replacement:

```md
- **Visual regression tests (VRT)**: tests live in `src/test/vrt/<area>/*.vrt.test.tsx` (feed, landing, onboarding, post, profile, settings) with baselines in the sibling `__screenshots__/` folder — not next to the component. When you change a UI surface, check whether a VRT covers it; if yes, say so — baselines are owned by CI (`docs/visual-regression-testing.md`): dispatch the **VRT Update Baselines** workflow from `dev` or the feature branch (it refuses `master`); PR CI does not run VRT. If you're touching a template-level surface with no VRT yet, mention adding one. Harness: `src/test-utils/vrt.tsx`; fixtures `src/test/fixtures/`; mocks `src/test/mocks/`.
```

The same two errors are in `.cursor/skills/pubky-app-development/SKILL.md` L367–371 and the rationale at L39–40 (see comments there).

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+```
+
+Cypress e2e needs the full pubky-stack (private `pubky/pubky-stack` at `staging`, homeserver/nexus
+images) and runs on push to `master`/`dev` in CI; do not attempt it from a bare checkout, and do not
+report an e2e result you did not obtain.
+
+Right-sized verification: a component change → its colocated test + `npm run lint` +
+`npm run typecheck`; a core change → affected package tests plus typecheck/lint; a
+`runtime-config`/`env` change → `npm test -- src/libs/runtime-config`; anything touching a VRT
+surface → `npm run test:vrt` and flag the baseline. Run the full `npm test` before handing off a
```
Orlandohub, line 380: Same two factual errors as the `AGENTS.md` bullet, plus one omission:

- `images` is not a VRT area (fixtures only, no tests, no `__screenshots__`).
- `vrt-update-baselines.yml` is `workflow_dispatch`-only and hard-fails on `master` (`Refuse master` step in both jobs). It runs from `dev` (opens a PR to `dev`) or a feature branch (commits to it).
- The paragraph never mentions `docs/visual-regression-testing.md`, which is the canonical doc for exactly this and already has the correct CI table. Golden Rule 1 at L22–25 also omits it from the per-area doc map.

Suggest collapsing this to a pointer:

```md
VRT: `docs/visual-regression-testing.md` is canonical. Tests live in `src/test/vrt/<area>/*.vrt.test.tsx`
(feed, landing, onboarding, post, profile, settings), baselines in the sibling `__screenshots__/`.
Baselines are owned by CI: dispatch **VRT Update Baselines** from `dev` or the feature branch (it refuses
`master`). PR CI does not run VRT, so a UI change that shifts pixels must be surfaced to the user.
```

and adding `docs/visual-regression-testing.md` to Golden Rule 1 for `src/test/vrt/**` changes.

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+7. Do not add `useCallback` / `useMemo` / `React.memo`: the React Compiler is enabled
+   (`reactCompiler: true` in `next.config.ts`).
+8. Writes are local-first: Dexie first, homeserver sync after, UI updated immediately.
+9. Imports point at concrete modules through the `tsconfig.json` path aliases. No `index.ts`
+   re-export barrels under `src/components`, no aggregate `src/config/index.ts`.
+10. Change only what the request needs. No drive-by refactors, renames, reformatting, dependency
+    swaps or pattern migrations in the same PR.
+11. If you changed a surface covered by a VRT (`src/test/vrt/<area>/*.vrt.test.tsx`), tell the user the
+    pixel baseline under that area's `__screenshots__/` needs regenerating: PR CI does not run VRT and
+    `npm run test:vrt` locally compares against the old baseline rather than blessing a new one.
```
Orlandohub: The conclusion (tell the user, don't commit local baselines) is right, but the stated reason is wrong. There *is* a local blessing path — `npm run test:vrt:regenerate-baseline` (`vitest run --project vrt --update`) and the narrower `npx vitest run --project vrt <path> --update` from `docs/visual-regression-testing.md`. The actual reason not to use them is that pixel rendering differs between a dev machine and the GitHub runners, so baselines must come from the **VRT Update Baselines** workflow ("CI is the source of truth" in that doc).

Suggested wording:

```md
11. If you changed a surface covered by a VRT (`src/test/vrt/<area>/*.vrt.test.tsx`), tell the user the
    baseline under that area's `__screenshots__/` needs regenerating via the **VRT Update Baselines**
    workflow (dispatched from `dev` or the feature branch). Do not commit baselines from
    `test:vrt:regenerate-baseline` — local renders never match the CI runners
    (`docs/visual-regression-testing.md`). PR CI does not run VRT.
```

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+
+- Writes: `PostApplication.commitCreate` (`src/core/application/post/post.ts`) is the reference
+  local-first write: upload files (`FileApplication`) → `LocalPostService.create` →
+  `HomeserverService.request({ method: PUT, url: postUrl })` → `TagApplication.commitCreate`, with a
+  compensating rollback (delete the local row and the uploaded files) when the homeserver call fails.
+  Copy that shape: local write first, sync second, roll back on failure, and refresh the entity's TTL
+  row (`PostTtlModel.upsert({ id, lastUpdatedAt: Date.now() })`) for every affected entity. Controllers
+  normalize with pipes first (`src/core/controllers/post/post.ts`, `PostNormalizer.to`,
+  `TagNormalizer` in `src/core/pipes/tag/tag.normalizer.ts`).
+- Reads: `PostController.getDetails({ compositeId })` for the local read and `PostController.fetch({ compositeId })`
```
Orlandohub, line 165: This signature does not exist. `StreamPostsController.getOrFetchStreamSlice` (`src/core/controllers/stream/posts/posts.ts` L58–65) takes `TReadPostStreamChunkParams`:

```ts
{ streamId, streamHead = SKIP_FETCH_NEW_POSTS, streamTail = NOT_FOUND_CACHED_STREAM, lastPostId, limit = NEXUS_POSTS_PER_PAGE, order }
```

There is no `viewerId` (the controller derives it from `useAuthStore.getState().currentUserPubky` at L68 — that is the whole point of the controller layer) and no `cursor`; pagination resumes through `streamTail` + `lastPostId`. An agent copying this example gets a type error and, worse, learns that callers should pass the viewer in. Use the controller's own JSDoc example:

```md
`StreamPostsController.getOrFetchStreamSlice({ streamId, streamTail, lastPostId, limit })` (viewer is derived inside the controller; see the JSDoc in `src/core/controllers/stream/posts/posts.ts`).
```

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+
+`useLocalFirstQuery` is deliberately simple, and most of its surprises are documented only by its
+call sites and past issues:
+
+- `fetchFn` runs **only when local data is `null`**. A cache hit is never refreshed, so a stale
+  counter/number that "never updates" is usually this, not the component (#2384).
+- A tombstone is still local data: soft-deleted rows (`content = [DELETED]`) keep the cache non-null,
+  so the network arm never fires and the UI renders the tombstone forever (#1988). Check what "missing"
+  means for the entity you are rendering.
+- Every hook instance owns its own effect. Mounting the same query twice (list plus expanded row, or a
```
Orlandohub, line 284: This misdescribes the hook it is documenting, and the issue reference is wrong.

- `isLoading` does **not** stay true after a 404. In `useLocalFirstQuery.ts` the fetch arm is `fetchFn().catch(...).finally(() => setIsFetching(false))`, so a rejected `fetchFn` settles to `{ data: null, isLoading: false }`. There is a unit test for exactly this: `sets isLoading false after fetchFn rejects (error treated as settled)` (`useLocalFirstQuery.test.ts` L346; I ran the file, 44/44 pass). `usePostMissing` then derives `postMissing: !compositeValid || (!isLoading && postDetails === null)`.
- The "skeleton forever" bug was consumers that only branched on loading/deleted/content and never on the settled-`null` state; that was fixed in **#2081** ("render path for not found embedded posts"), which you already cite on the next line for `isMissing`.
- **#2394** is `feat(ui): add onboarding tags of interest step` — unrelated.
- The hook exposes no error value, so "explicit missing/**error** branch" points at an API that does not exist.

Suggested replacement:

```md
- A settled cache miss is `{ data: null, isLoading: false }`: `isLoading` is true only until `useLiveQuery`
  resolves and, on a miss, until `fetchFn` settles (resolve *or* reject). A Nexus 404 therefore ends as
  `null`, not as a permanent skeleton — the consumer must branch on that null (`usePostMissing`, #2081).
  The hook exposes no error value; if you need one, add it to the hook rather than inferring from timing.
```

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+## Sensitive or High-Risk Areas
+
+- `src/core/database/franky/franky.ts` - one `this.version(DB_VERSION).stores({...})` definition, with
+  `DB_VERSION` a build-intrinsic `NEXT_PUBLIC_DB_VERSION` (`src/config/database`). There is no
+  incremental migration chain: a version mismatch makes the client delete and recreate the local
+  database (`recreateDatabase`), i.e. user-visible local data loss, so bumping the version or editing
+  a table's index map is a deliberate, reviewed change, not a side effect of a feature.
+- `src/core/services/homeserver/**` and `src/core/pipes/**` - wire-format boundaries
+  (`pubky-app-specs`, composite ids, signup tokens, auth URLs). Preserve payload shapes; do not change
+  a format incidentally while adding a feature.
```
Orlandohub, line 326: This is the one inaccuracy in the file with a security consequence. `Env` does not expose only those four fields — `src/libs/env/env.ts` also validates `NODE_ENV`, `VITEST`, `HOMESERVER_ADMIN_URL`, `HOMESERVER_ADMIN_PASSWORD`, `BASE_URL_SUPPORT`, `SUPPORT_API_ACCESS_TOKEN`, `SUPPORT_ACCOUNT_ID`, and its header comment says the file is for "build-intrinsic public values **and server-only variables**". The ESLint rule in `eslint.config.mjs` L90–108 restricts only `NEXT_PUBLIC_*` reads to that allow-list; it says nothing about server-only vars.

Why it matters: runtime config is serialized into `window.__PUBKY_CONFIG__` (`src/libs/runtime-config/runtime-config.ts` L16–23). "Everything deployer-facing is a `PUBKY_RUNTIME_*` getter" tells an agent adding, say, a new admin token to put it where the browser can read it.

Suggested wording:

```md
- `src/libs/env/env.ts` + `src/libs/runtime-config/**` - the only places allowed to read `process.env`
  (ESLint-enforced). `env.ts` holds build-intrinsic public values (only `NEXT_PUBLIC_DB_NAME|DB_VERSION|DEBUG_MODE|APP_VERSION`
  may be `NEXT_PUBLIC_*`) **and server-only secrets** (`HOMESERVER_ADMIN_*`, Chatwoot `SUPPORT_*`).
  Anything deployer-facing that the client must see is a `PUBKY_RUNTIME_*` getter — it is injected into
  `window.__PUBKY_CONFIG__`, so never route a secret through it (`docs/environment.md`, ADR-0017).
```

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+
+1. The dialog uses the form hook: `form` (RHF + zod schema from `useCreateCollection.types.ts`),
+   a cover picker from `useCoverImagePicker`, and `submit(): Promise<string | null>`.
+2. `submit()` calls `PostController.commitCreateCollection({ authorId, name, description, coverImage, layout })`
+   and maps the `AppError` to a toast (`toast({ variant: 'error', ... })`). Do not add a `Logger.error`
+   of your own: `Err.*` factories already log and capture the failure (ADR-0015). The `Logger.error` in
+   today's `useCreateCollection` is debt, not part of the pattern to copy.
+3. The controller uploads the cover through `FileApplication.toFileAttachment`/`commitCreate`, builds
+   the payload with `PostNormalizer.toCollection(...)`, derives `compositePostId` with
+   `buildCompositeId`, then calls `PostApplication.commitCreate({ compositePostId, post, postUrl })`.
```
Orlandohub, line 150: The worked example teaches double logging. `useCreateCollection.ts` does call `Logger.error(...)` in its catch, but every `AppError` reaching it was already logged (and captured once-per-chain) by the `Err.*` factory that created it. Presenting `toast + Logger.error` as *the* pattern contradicts Golden Rule 5 ("factories log"), the Definition of Done ("nothing double-logs"), and the UI-layer example in `docs/error-handling.md` L116–131, which only toasts.

Either describe the canonical shape and mark the existing call as debt:

```md
2. `submit()` calls `PostController.commitCreateCollection({ authorId, name, description, coverImage, layout })`
   and on failure toasts static copy (`toast({ variant: 'error', description: getErrorMessage(error) })`).
   Do not add `Logger.error` for an `AppError` here — the factory already logged it (the current call in
   `useCreateCollection.ts` is debt, not a pattern to copy).
```

or pick a worked example that does not need the caveat.

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+  Copy that shape: local write first, sync second, roll back on failure, and refresh the entity's TTL
+  row (`PostTtlModel.upsert({ id, lastUpdatedAt: Date.now() })`) for every affected entity. Controllers
+  normalize with pipes first (`src/core/controllers/post/post.ts`, `PostNormalizer.to`,
+  `TagNormalizer` in `src/core/pipes/tag/tag.normalizer.ts`).
+- Reads: `PostController.getDetails({ compositeId })` for the local read and `PostController.fetch({ compositeId })`
+  for the Nexus fetch; `UserController.getManyDetails(params)` / `getOrFetchDetails`;
+  `StreamPostsController.getOrFetchStreamSlice({ streamId, streamHead, streamTail, lastPostId, limit, order })`
+  (`viewerId` is derived inside the controller from `useAuthStore`; there is no `viewerId` or `cursor`
+  parameter).
+- Local-first reads in hooks (ADR-0011): use `useLocalFirstQuery` from
```
Orlandohub: "~19 hooks" here and "22 hooks" in the PR description are both wrong — and different from each other. `rg -l 'useLocalFirstQuery\b' src/hooks --glob '!*.test.*'` (excluding the hook's own folder) returns **11**: `useAuthoredCollections`, `useBookmarksCollectionSummary`, `useCurrentUserProfile`, `useIsFollowing`, `usePostCounts`, `usePostDetails`, `useProfileStats`, `useRepostInfo`, `useSocialGraphStatus`, `useUserDetails`, `useUserProfile`. (The three `src/core/**` hits only mention it in comments.)

I would drop the number entirely; counts rot on the next PR and add nothing an agent can act on. If you want a durable pointer, give the command instead:

```md
`src/hooks/usePostDetails/usePostDetails.tsx` is the canonical consumer (`rg -l useLocalFirstQuery src/hooks` lists the rest).
```

Worth a pass over the other counts in the file too (`~140 hooks` is currently right; `docs/adr/0001..0018` is right).

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+  rebinding checks). Any change here is a security change: keep the guards and their tests intact.
+- `src/sw.ts`, `next.config.ts`, `instrumentation*.ts`, `sentry.*.config.ts`, CI workflows - build and
+  runtime plumbing with cross-cutting effects.
+- `src/core/application/` cross-domain calls: verify the ADR-0009 allow-list before wiring two
+  applications together.
+
+## Testing and Verification
+
+Commands (from `package.json`):
+
```
Orlandohub, line 354: Inside a `bash` fence this is a command an agent will paste. `npm run start:e2e / test:e2e` runs `start:e2e` and forwards `/` and `test:e2e` as Cypress arguments. Also worth listing `test:vrt:regenerate-baseline`, since the VRT text elsewhere implies no such script exists (see the comment on L38–40). Suggested block:

```bash
npm run start:e2e                      # cypress open (firefox); start:e2e:mobile for the mobile config
npm run test:e2e                       # cypress run -b firefox; test:e2e:mobile for the mobile config
npm run test:vrt:regenerate-baseline   # --update; CI-owned, do not commit the output (see VRT section)
```

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+8. Writes are local-first: Dexie first, homeserver sync after, UI updated immediately.
+9. Imports point at concrete modules through the `tsconfig.json` path aliases. No `index.ts`
+   re-export barrels under `src/components`, no aggregate `src/config/index.ts`.
+10. Change only what the request needs. No drive-by refactors, renames, reformatting, dependency
+    swaps or pattern migrations in the same PR.
+11. If you changed a surface covered by a VRT (`src/test/vrt/<area>/*.vrt.test.tsx`), tell the user the
+    pixel baseline under that area's `__screenshots__/` needs regenerating: PR CI does not run VRT and
+    `npm run test:vrt` locally compares against the old baseline rather than blessing a new one.
+12. Never edit generated or externally-owned surfaces by hand: `public/sw.js`, the generated
+    `src/libs/lucide/lucideIcons.*.ts` files, `package-lock.json`, `.github/workflows/*`.
```
Orlandohub: Nit, but this rule is stated as fact and it isn't one: `.github/workflows/*` are neither generated nor externally owned — the nine workflows in this repo are hand-maintained and change in ordinary PRs. A blanket "never edit by hand" blocks legitimate work (fixing CI is a task an agent gets asked to do). You already list CI workflows under *Sensitive or High-Risk Areas* (L325), which is the right framing. Suggest keeping this rule to the truly generated surfaces:

```md
12. Never edit generated files by hand: `public/sw.js` (built from `src/sw.ts`), the generated
    `src/libs/lucide/lucideIcons.{aliases,nodes,tags}.ts`, `package-lock.json` (let npm write it).
    CI workflows are hand-maintained but high-risk — change them only when the task is CI.
```

(`lucideIcons.ts` itself is hand-written — it exports `preloadLucideIcons` — so `lucideIcons.*.ts` over-matches.)

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+
+## Frontend and Design System
+
+- Styling is Tailwind v4 with CSS variables in `src/app/globals.css` (`@theme` tokens) plus Shadcn
+  components; `cn` from `@/libs/utils/utils`; variants via `cva` + `VariantProps`.
+- Tiers: `atoms/` primitives (Button, Input, Textarea, Select, Checkbox, RadioGroup, Switch, Toggle,
+  Dialog, Sheet, Popover, DropdownMenu, Tooltip, Collapsible, Card, Avatar, Badge, Tag, Skeleton,
+  Spinner, Carousel, Typography), `molecules/` atom combinations (InputField, TextareaField,
+  SearchInputBar, PostInputAttachments, PostHeaderUserInfo, Timeline, TagInput, ProgressSteps,
+  SideDrawer, Toaster, MobileFooter/MobileTabBar, Popover* set, `*Empty` empty states), `organisms/`
```
Orlandohub: Inventory lists like this one are already drifting on the commit they were written against, which is the argument against having them in a skill at all (the PR description's own principle: point at `docs/`, don't restate):

- `PostCard` is not an organism — only `PostCardSkeleton` exists under `src/components/organisms/`; the card itself is `PostMain`/`PostContent`.
- L231: `DialogConfirmDelete` is a **molecule** (`src/components/molecules/DialogConfirmDelete/`), not an organism, so it is the wrong example for "dialog organisms".
- L178: the glob `*-service.utils.ts` matches nothing; the files are `nexus.utils.ts`, `homeserver.utils.ts`, `chatwoot.utils.ts`, `nextjs.utils.ts` (i.e. `<service>.utils.ts`).
- L237: the table list omits `user_tags`, `user_connections`, `post_tags`, `notifications` (all declared in `franky.ts`).

Suggest replacing the enumerations with the lookup an agent would actually run (`ls src/components/<tier>`, `rg 'Table<' src/core/database/franky/franky.ts`) and keeping only the 2–3 named examples that the surrounding rule needs. That also shrinks the file, which at 460 lines is long for a skill that is meant to be read before every change.

AGENTS.md
```diff
@@ -41,17 +41,20 @@ Use `Err.*` factories (never raw `Error`). Factories log automatically — don't
...
 - **Memoization**: do not add `useCallback` / `useMemo` — the React Compiler (`reactCompiler: true` in `next.config.ts`) handles memoization. Reach for them only after profiling proves the compiler missed something.
 - **Toasts**: use `toast()` from `@/molecules/Toaster/toast` with `variant` (`default` | `error` | `warning` | `info`). Never import `useToastState`, `toast.store`, or the `@/atoms/Toast/*` renderer atoms (ESLint-enforced). No `showErrorToast` wrappers or `className` destructive hacks. Toast copy is static: never interpolate user-entered text (names, labels, file names) into `title` / `description` — use generic copy like `'Feed created'`. Bounded values such as config constants and small counts are fine. See `docs/components.md` — _Toasts_.
+- **Dexie schema**: `src/core/database/franky/franky.ts` declares a single `this.version(DB_VERSION).stores({...})` and there is no incremental migration chain. A `DB_VERSION` mismatch deletes and recreates the local database (`recreateDatabase`), so bumping the version or changing a table's index map is a deliberate, reviewed change - never a side effect of a feature.
```
Orlandohub: The description of the code is accurate — I traced `runInitialize()` → `currentVersion !== DB_VERSION` → `recreateDatabase()` → `indexedDB.deleteDatabase` in `franky.ts` L341–350. Two small things:

1. It now contradicts an *Accepted* ADR. `docs/adr/0007-dexie-version-normalization.md` says "Migrations register in ascending order with explicit upgrade paths and are validated by snapshot tests", which is not what the code does. The skill declares `docs/` and ADRs canonical over itself, so leaving the two in disagreement means an agent that follows the skill's own rule picks the wrong one. Either add "(ADR-0007 predates this and is stale; see ADR-0007 for the original intent)" here, or open a follow-up to supersede ADR-0007 and link it in the PR.
2. `DB_VERSION` comes from `Env.NEXT_PUBLIC_DB_VERSION` via `src/config/database.ts` — worth naming, since that is the value someone would actually bump.

The same claim appears at SKILL.md L73, L236 and L307; a single pointer to this bullet from the skill would be enough.

.cursor/skills/pubky-app-development/SKILL.md
```diff
@@ -0,0 +1,477 @@
...
+
+Two bug classes account for most of this repo's regressions. Read this before touching a read path or
+anything that writes TTL rows.
+
+`useLocalFirstQuery` is deliberately simple, and most of its surprises are documented only by its
+call sites and past issues:
+
+- `fetchFn` runs **only when local data is `null`**. A cache hit is never refreshed, so a stale
+  counter/number that "never updates" is usually this, not the component (#2384).
+- A tombstone is still local data: soft-deleted rows (`content = [DELETED]`) keep the cache non-null,
```
Orlandohub, line 281: Nit: #1987 ("Navigation to page with posts fires many nexus query requests for tags") supports this bullet; #1986 ("swap post shell to discovery layout when post is missing") is the not-found layout fix and has nothing to do with duplicate requests. Drop #1986 here (it would fit the `isMissing` bullet below alongside #2081). I checked the rest of the issue references in this section — #2384, #1988, #2081, #1781, #1452, #1276, #2486 all match their claims.

Orlandohub (changes requested):
## Review: docs-only, so the bar is factual accuracy

This PR adds a 460-line skill and rewrites four `AGENTS.md` bullets, and its stated verification is "every path, symbol and issue reference … was checked against the tree on `dev`". I re-checked every path, symbol, script, workflow trigger, issue number and count in the diff against `c7a4949` (== `dev` + this commit). Most of it holds up — the layer table, controller naming, the `PostApplication.commitCreate` flow, the error helpers, the routes/stores/config inventories, the i18n history, the Dexie recreate path, and the TTL issue references are all accurate, and the file is a genuinely useful operating manual.

But a docs PR whose only job is to be correct still has several load-bearing statements that are wrong, including the one fact the PR set out to fix, so I'm requesting changes rather than commenting. Ordered by impact:

1. **VRT guidance is still wrong** in both files: `images` is not a VRT area; `vrt-update-baselines` is `workflow_dispatch`-only and hard-fails on `master`; the canonical `docs/visual-regression-testing.md` is never linked; `test:vrt:regenerate-baseline` is omitted so the "no local blessing" rationale is false. (Both bots flagged parts of this; none of it is addressed yet.)
2. **`getOrFetchStreamSlice({ streamId, viewerId, cursor, limit })`** — no such parameters; `viewerId` is derived inside the controller.
3. **Settled-404 pitfall** describes behaviour the hook does not have (`isLoading` settles to `false` on rejection — there is a unit test for it), and cites #2394 (onboarding tags PR) instead of #2081.
4. **`Env` "exposes only four fields"** — it also holds server-only secrets; the wording steers a new secret toward `window.__PUBKY_CONFIG__`.
5. **Worked example endorses double logging** (`toast + Logger.error`), contradicting Golden Rule 5, the Definition of Done and `docs/error-handling.md`.
6. **Consumer count** for `useLocalFirstQuery` is wrong twice over ("~19" in the skill, "22" in the PR body; actual 11).
7. `npm run start:e2e / test:e2e` is not a runnable command inside a bash fence.
8. Nits: `.github/workflows/*` called "generated/externally owned"; `PostCard` organism / `DialogConfirmDelete` organism / `*-service.utils.ts` glob / Dexie table list drift; #1986 misattributed; the Dexie bullet now contradicts Accepted ADR-0007 without saying so.

Details and concrete replacement text are in the line comments.

### What I verified
- Checked out `talos/pubky-app-dev-skill`, `npm ci` on Node 24.
- `npm run lint` ✅, `npm run typecheck` ✅, `npm run format:check` ✅ (`AGENTS.md` is Prettier-checked; `.cursor/` is in `.prettierignore`).
- Targeted Vitest for the disputed claims: `src/hooks/useLocalFirstQuery`, `src/hooks/usePostMissing`, `src/core/database/franky` — 3 files, 44 tests, all pass.
- Every `#nnnn` reference resolved via `gh`; every cited path/symbol grepped.

### Not verified
- No runtime behaviour to test; nothing under `src/` changes.
- I did not evaluate how Cursor's skill loader treats a 460-line `SKILL.md` versus the two much shorter existing skills; the size comment is a maintenance observation, not a tooling claim.

Happy to re-review once the VRT / stream-slice / Env / 404 text is corrected.

AGENTS.md
talosmachina: Reworked on the branch head `c798382c`. The bullet now points at `docs/visual-regression-testing.md` as canonical, drops `images` as an area (fixtures only), lists the real areas, and describes `vrt-update-baselines.yml` as `workflow_dispatch` that refuses `master` and is dispatched from `dev` or the feature branch.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: Collapsed to the pointer as suggested, and `docs/visual-regression-testing.md` is now in Golden Rule 1s doc map for `src/test/vrt/**`.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: Golden Rule 11 now gives the real reason (local renders never match the CI runners, so the workflow owns the baselines) and `test:vrt:regenerate-baseline` is in the command block.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: Now `getOrFetchStreamSlice({ streamId, streamTail, lastPostId, limit })`, with `streamHead`/`order` noted as optional, the viewer derived inside the controller, and an explicit line that there is no `viewerId` or `cursor` parameter.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: Rewritten as described: `isLoading` clears on reject via `.finally()`, a Nexus 404 settles at `{ data: null, isLoading: false }`, no skeleton with no exit, no error value on the hook, and `usePostMissing` turns the settled `null` into `postMissing`. `#2394` is gone; `#2081`/`#1986` sit on the `isMissing` bullet.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: Now lists the server-only variables alongside the `NEXT_PUBLIC_*` allow-list, and warns that runtime config is injected into the browser as `window.__PUBKY_CONFIG__`, so a secret must never go there.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: The worked example now says the factories already log and capture, and marks the `Logger.error` in `useCreateCollection.ts` as debt rather than the pattern to copy.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: Both counts dropped in favour of the lookup: `rg -l useLocalFirstQuery src/hooks --glob "!*.test.*"`. The PR body no longer claims a number either.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: Split into four runnable commands (`start:e2e`, `test:e2e`, `start:e2e:mobile`, `test:e2e:mobile`) plus `test:vrt:regenerate-baseline`.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: Rule 12 now covers only the genuinely generated surfaces (`public/sw.js`, `lucideIcons.{aliases,nodes,tags}.ts`, `package-lock.json`) and states that CI workflows are hand-maintained but high-risk, changed only when the task is CI.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: The enumerations are gone in favour of the lookups an agent would run (`ls src/components/<tier>`, `rg "Table<" src/core/database/franky/franky.ts`), with only the examples the surrounding rule needs: `PostCard` dropped, `DialogConfirmDelete` marked a molecule, `<service>.utils.ts` corrected, and the tables list matched to `franky.ts`.

AGENTS.md
talosmachina: Took the second option: `DB_VERSION = Env.NEXT_PUBLIC_DB_VERSION` (`src/config/database.ts`) is named, and ADR-0007 is now properly superseded rather than annotated, in #2532 (new ADR-0019 documenting the single-version recreate path, ADR-0007 marked `Superseded`, index updated). Once #2532 lands the stale note in this PR can become a link to ADR-0019.

.cursor/skills/pubky-app-development/SKILL.md
talosmachina: `#1986` moved to the `isMissing` bullet next to `#2081`; the duplicate-request bullet keeps `#1987`.

Orlandohub (approved):

PR #2527, opened 2026-09-14
feat(observability): add opt-in Pulse browser instrumentation

Jasonvdb:
## Summary

Add opt-in error and page-view tracking with `@synonymdev/pubky-pulse-web@0.8.0`.

- Capture exceptions from error factories, React boundaries and unhandled errors using the app's filtering policy.
- Keep events anonymous, redact sensitive values and map dynamic routes to templates.
- Network failures reach Pulse as `AppError`s through that same policy. The SDK's automatic network tracking is off: its fetch-level events are emitted before the `AppError` exists, so the drop rules cannot apply to them.
- Ask each browser for consent before Pulse initializes or stores an identifier. Withdrawal is available to guests through the Analytics settings button and to everyone in Settings → Privacy and Safety.

## Consent behaviour

- Withdrawal calls `Pulse.reset()`, which stops collection without flushing and deletes the anonymous ID, session and any queued events from the browser. Re-acceptance starts as a new anonymous browser and replays nothing.
- Every tab deletes Pulse state that predates the current consent. Accepting stores the acceptance time (`pubky-pulse-consent-v1-granted-at`, strictly increasing, written before the choice), and each tab records in `sessionStorage` which acceptance its Pulse state started under. A tab that was suspended, restored from bfcache or navigated away during a withdrawal, including a withdrawal followed by a re-acceptance, sees the mismatch the next time it runs, resets, and starts as a new anonymous browser. A tab that loads without consent resets unconditionally, so a session left in its own `sessionStorage` is never resumed.
- Events recorded in a stale tab before it resyncs are dropped in `beforeSend`.
- Neither new value is an identifier and neither is sent. A first-time decline still stores nothing but `declined`.
- A failed save (blocked site storage) leaves analytics off, keeps the banner open with the error and a Close control, then collapses to the Analytics settings button so the visitor can retry later instead of facing an undismissable banner.

Known limits. Both sit below `beforeSend`, so the app cannot close them:

- On resume from a freeze, the SDK's flush timer can fire before the tab's `storage` / `pageshow` / `focus` handler and send events that were already buffered while consent was valid. Hidden tabs flush on hide, so the buffer is normally empty.
- A stale tab can append to the shared offline queue after another tab purged it and then close without resyncing; the next `init()` replays that queue without passing `beforeSend`.

## Setup

```dotenv
PUBKY_RUNTIME_PULSE_CLIENT_KEY=pulse_client_<your-web-app-client-key>
```

A missing or blank key disables tracking. Set `PUBKY_RUNTIME_PULSE_ENDPOINT` to override the default endpoint.

## Verification

- Regression tests were written first and run against the previous head: 10 of them failed there, reproducing both consent findings with the real SDK (same anonymous ID kept, surviving `pulse.session_id`, pre-withdrawal event sent). All pass now.
- Focused Pulse suites: 58 tests pass. Full unit suite: 13,790 pass, 2 skipped. `typecheck`, `lint`, `format:check` and `build` pass.
- Real Chromium with the real SDK and telemetry intercepted locally (Playwright, two tabs sharing `localStorage`): a tab frozen with `Page.setWebLifecycleState` through a withdrawal and re-acceptance resets on resume and never sends its queued event; a tab that navigated away returns with no `pulse.*` keys, and after a re-acceptance starts a new session; four 4xx responses and a rejected fetch produce no `sdk:network_request`. A control run confirms an unhandled error is normally captured and sent.


src/libs/error/error.factories.ts
```diff
@@ -76,6 +77,7 @@ function createAppError<C extends ErrorCategory>(
     Logger.warn('[Err.factory] captureAppError failed', sentryError);
   }
 
+  Pulse.captureException(error);
```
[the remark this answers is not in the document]
Jasonvdb: I checked the pinned `@synonymdev/pubky-pulse-web@0.7.0`. [`captureException` already catches exceptions internally](https://github.com/Jasonvdb/pubky-pulse-web/blob/cb2bf882de7c46719673e32a25a159a89f15b7b4/src/index.ts#L690-L696), and its [inner capture path](https://github.com/Jasonvdb/pubky-pulse-web/blob/cb2bf882de7c46719673e32a25a159a89f15b7b4/src/index.ts#L328-L369) catches failures from error inspection and logging. Failing `beforeSend` hooks are contained too.

I ran 171 SDK tests covering exception containment and runtime failures, plus 89 app tests covering capture, consent and AppError behavior. All passed.

For this pinned version, the failure described here is already handled by the SDK. An extra app-level catch would be defensive hardening against a future SDK regression, not a fix for a demonstrated failure in this PR.

Orlandohub (changes requested):
[P1] Background tabs can miss consent withdrawal.
[pulse-consent.ts:44](https://github.com/pubky/pubky-app/pull/2527/files) ignores the storage event’s value and reads the latest consent. If another tab withdraws and re-accepts before a suspended tab resumes, that tab sees only “accepted.” With the real SDK, it retained its old identity and transmitted events queued before withdrawal. Track withdrawal transitions or a persisted consent generation so stale SDK state is discarded.

[P2] Restored tabs reuse sessions across withdrawal.
[pulse.ts:107](https://github.com/pubky/pubky-app/pull/2527/files) resets only when this page already started Pulse. A restored tab starting with declined consent retains its previous sessionStorage identifier, which another tab cannot clear. Re-acceptance reused that exact session in the regression test. Reconcile persisted SDK state on startup as well.

[P2] Automatic network tracking bypasses existing suppression rules.
[pulse.ts:91](https://github.com/pubky/pubky-app/pull/2527/files) emits HTTP-response events before fetchNexus constructs the AppError used by filtering. Three post-tag 404s emitted three network warnings; three 429s also emitted three warnings despite repeat suppression. Apply filtering before SDK network events lose their request context, or use the existing error pipeline for these failures.

infin1t3 (changes requested):
Requesting changes for three issues verified against `a3ecf008`. The two consent findings confirm Orlando's reported cases; the network finding below concerns error-level events from deliberately suppressed stream-connect failures.

1. **[P2] Preserve withdrawal when storage events are delivered late**

   [pulse-consent.ts:44–47](https://github.com/pubky/pubky-app/blob/a3ecf008efcfa0f359e5a696fd135f7c7a62c619/src/libs/observability/pulse-consent.ts#L44-L47)

   If tab A withdraws and re-accepts before tab B processes its storage events, B reads the latest `accepted` value for both events. It never calls `Pulse.reset()`, so its previous anonymous ID, session and buffered events survive. With the real SDK in Chromium and B's main thread paused, resuming B sent both a pre-withdrawal queued event and a new event under the old identity. Sending occurred after re-acceptance, but this violates the promised deletion and fresh-identity behavior. Preserve the withdrawal transition or a consent generation so each tab discards stale SDK state before resuming collection.

2. **[P2] Clear persisted SDK state on non-consented startup**

   [pulse.ts:107–110](https://github.com/pubky/pubky-app/blob/a3ecf008efcfa0f359e5a696fd135f7c7a62c619/src/libs/observability/pulse.ts#L107-L110)

   An accepted tab can navigate away while another tab withdraws consent. When it returns with consent still declined, `started` is initially false, so this guard skips reset and leaves its previous `sessionStorage` session intact. Re-acceptance within the SDK's 30-minute inactivity window resumes that same session, even though the anonymous ID changes. This was reproduced with ordinary Chromium navigation and the real SDK. Reconcile and clear retained SDK state on non-consented startup even when the current page has never initialized Pulse.

3. **[P2] Apply the application error policy to automatic network failures**

   [pulse.ts:91](https://github.com/pubky/pubky-app/blob/a3ecf008efcfa0f359e5a696fd135f7c7a62c619/src/libs/observability/pulse.ts#L91)

   Automatic network tracking records rejected fetches before they are translated into the structured AppError used by the drop rules. A real `HomeserverService.subscribeUserEventStreamForPath` call through the Pubky SDK, with its PKARR fetch rejecting with `TypeError('Load failed')` while online, produced one Pulse `sdk:network_request` at error level even though the resulting AppError matched the deliberate stream-connect suppression rule. The raw exception lacks the service/operation needed by `beforeSendPulse`, so routine reconnect failures are reported before the coordinator's persistent-outage threshold. Integrate automatic failure capture with that policy, or keep these failures in the existing structured error pipeline.

Validation: the complete repository suite passed (13,776 tests, 2 skipped), along with changed-file lint/format checks, typecheck and the production build. Additional real-SDK integration and Chromium probes reproduced the issues above. Telemetry transport was intercepted locally throughout the probes.

Jasonvdb:
@Orlandohub @infin1t3 thanks, all three reproduced. Fixed in 35152668, 0f4e4163 and ac8d94b2.

**1. Background tab misses a withdrawal** (Orlando P1, infin1t3 1)

Went with the persisted consent generation. Accepting now stamps a strictly increasing acceptance time before it writes the choice, and each tab records in `sessionStorage` which acceptance its Pulse state started under. A tab that only ever reads `accepted` still sees that the stamp changed, so it calls `Pulse.reset()` and starts over. It does not depend on `oldValue` / `newValue`, because the `pageshow` and `focus` paths get no storage event at all. `beforeSend` also drops anything recorded in the stale window before the resync.

I stamp the acceptance and not the withdrawal on purpose: a withdrawal whose `setItem` throws cannot bump a counter, and a cleared store reads a counter as 0 again. Both would let a sleeping tab miss the change.

**2. Restored tab reuses its session** (Orlando P2, infin1t3 2)

`Pulse.reset()` now runs whenever consent is not accepted, whether or not this page ever started Pulse. `reset()` has no live-instance guard in 0.8.0, so it clears the leftover `sessionStorage` session on startup.

This one was wider than reported. If the withdrawal and the re-acceptance both happen while the tab is away, it comes back seeing `accepted`, so "reset when not accepted" misses it and the old session gets linked to the new anonymous ID. The per-tab marker covers that case and there is a test for it.

**3. Network tracking bypasses the drop policy** (Orlando P2, infin1t3 3)

Took your second option and turned `networkTracking` off. I looked at filtering first: the option only accepts `urlMode` and `sampleRate`, HTTP-status events reach `beforeSend` with an empty hint, and `urlMode: 'origin'` has to stay because Nexus paths embed pubkys. So the hook only ever sees an origin, and an origin veto would also silence the cases `sentry.test.ts` pins as reportable (full post 404, a 429 from another operation, other Homeserver operations). Failures still reach Pulse through `safeFetch` / `handleError` → `Err.*` → `Pulse.captureException` → `beforeSendPulse`, which is the one policy. The consent copy no longer mentions network performance. Known tail: the few raw `fetch` call sites that bypass `safeFetch` are only seen if they end up unhandled; that predates this PR.

**Trade-offs I accepted**

- A stale tab's `reset()` also clears the shared anonymous ID another tab created after the re-acceptance. That is over-deletion: safe for privacy, costs one extra anonymous ID.
- The acceptance time stays in storage after a withdrawal. It is a time, never sent, and a first-time decline still stores only `declined`. Removing it would lose the monotonic guard.
- If `sessionStorage` is blocked the tab cannot record its marker, so Pulse does not start.

**Not closed by this PR**

Two limits sit below `beforeSend`, so the app cannot close them: the flush timer can fire before the resync handler when a frozen tab resumes and send what was already buffered, and a stale tab can re-seed the shared offline queue after another tab purged it, which the next `init()` replays without `beforeSend`.

**Verification**

The regression tests were written first: 10 failed on a3ecf008 with the real SDK and all pass now. Full suite, typecheck, lint, format and build pass. I also ran it in real Chromium with telemetry intercepted locally, including a tab frozen over CDP through a withdrawal and re-acceptance; details are in the PR description. Unrelated to this PR: `HotTagCard` "formats large post counts with locale separators" fails on machines whose default locale groups digits with a space (en-ZA here).

secondl1ght (dismissed):
Approving. Reviewed at `ac8d94b` with an architecture/rules pass, a correctness pass and a security pass: no security findings, no crash-level bugs, CI green on the earlier head, and lint, typecheck and the targeted suites pass locally. The consent gate, `Pulse.reset()` on withdrawal, the new consent-generation marker and the real-SDK consent tests are solid, and turning the SDK's network tracking off closes the one volume concern I had. Five things worth a look; none block, and whether to take them now, later or not at all is your call.

1. **Collected data does not match the consent text** (`pulse.ts:100`, `PulseConsent.tsx:9`, `:45`). `deviceInfo` is not passed, so the SDK default stamps `os_version`, `device_model`, `locale` and `preferred_language` on every event, and `beforeSendPulse` leaves those top-level fields untouched. The banner discloses only an anonymous ID, page views and errors. It also says Pubky can be used "without analytics", but Plausible loads for every visitor with no consent gate (`ContainerRoot.tsx`), so declining does not turn analytics off. `deviceInfo: false` (or an explicit subset) and either gating Plausible on the same consent or naming Pulse in the copy would close the gap; `docs/environment.md` should match whichever you pick.

2. **A failed `Pulse.init` is silently treated as running** (`pulse.ts:97`-`126`, `:148`). `Pulse.init` never throws and returns `{status, reason}`; `initPulse` discards the result, its catch is unreachable (the getters are memoized once consent resolved), and `sync` sets `running = true` regardless. A `disabled` or `error` result leaves the Privacy switch ON with nothing sent and no retry on later syncs. Returning the result from `initPulse` and setting `running` only on `status === 'enabled'` fixes it.

3. **Pulse skips the testnet gate Sentry applies** (`pulse.ts:103`). `shouldEnableSentry()` returns false when `getTestnet()` is true; `initPulse` does not check it, so a testnet deploy that inherits the key would show the banner on the routes Cypress drives.

4. **Observability wiring is split across files** (`pulse.ts:107`, `error.factories.ts:1`, `error.tsx:5`, `global-error.tsx:5`). The `ignoreErrors` list is a verbatim copy of the Sentry one in `sentry.ts` (and `pulse.test.ts` pins the copy), and the three `Pulse.captureException` call sites import the SDK directly, whereas `sentry.ts` documents that only the init files, the boundaries and the `captureAppError` funnel touch that SDK. A shared constant plus a `capturePulseException` in `pulse.ts` that applies the drop rules before calling the SDK would keep both sinks on one policy.

5. **Docs still describe Sentry as the only sink** (`docs/sentry.md`, `AGENTS.md`, `.greptile/config.json`). The capture rule, the "what is captured" table and the greptile `sentry-capture-rule` do not mention `Pulse.captureException`, and the docker-run recipe in `docs/environment.md` omits `PUBKY_RUNTIME_PULSE_*`.

Security pass, for the record: every SDK event type passes through `beforeSendPulse`, which fails closed; screen names never echo the path; consent fails closed on storage or config errors and `reset()` purges only `pulse.` keys; nothing calls `setUser`; `pulseClientKey` is env-sourced, prefix-validated on both parses and escaped by the existing inline-script serializer.

---

src/libs/observability/pulse.ts
```diff
@@ -0,0 +1,154 @@
...
+export function initializePulseConsent(): () => void {
+  let running = false; // this page has a live client; the marker outlives the page and cannot say so
+  const sync = () => {
+    const accepted = getPulseConsent() === 'accepted';
+    if (!accepted || predatesCurrentConsent()) {
+      // reset() disables synchronously without flushing, removes collectors and deletes the anonymous ID,
+      // session and queued events the banner asked consent to store, so nothing replays on re-acceptance.
+      // It also runs on a page that never started Pulse: a returning tab still holds the session a previous
+      // page stored, and no other tab can delete it.
+      Pulse.reset();
```
infin1t3: [P2] Preserve current-consent queued events when resetting a stale tab

With tabs A and B previously accepted, let B navigate away while A withdraws and reaccepts. A then records an event under the new consent and parks it in the shared offline queue on pagehide. When B returns, its old marker reaches this `Pulse.reset()`, which purges the shared `pulse.*` localStorage namespace as well as B's session. A's new event is deleted before either tab can deliver it.

Reproduced with Chromium and SDK 0.8.0; a control that opens a fresh tab instead preserves and delivers the same event. The accepted anonymous-ID churn also causes this loss of current-consent telemetry. Please separate stale-tab client/session cleanup from the browser-wide purge so reconciling B preserves shared events already recorded under the new consent. This may need a narrower SDK reset API.

src/hooks/usePulseConsent/usePulseConsent.ts
```diff
@@ -0,0 +1,15 @@
+'use client';
+
+import { useState, useSyncExternalStore } from 'react';
+import { getPulseConsent, setPulseConsent, subscribePulseConsent } from '@/libs/observability/pulse-consent';
+
+export function usePulseConsent() {
+  const consent = useSyncExternalStore(subscribePulseConsent, getPulseConsent, () => 'unavailable' as const);
+  const [saveFailed, setSaveFailed] = useState(false);
```
infin1t3: [P3] Clear the stale save error when another consent control succeeds

If an Accept click in the banner encounters a storage-write failure, this hook instance keeps `saveFailed=true`. After storage recovers, accepting through the Privacy and Safety switch updates the shared consent to `accepted`, but the banner's separate hook instance retains its error. The switch is ON while the open banner still says "Analytics is off". A component test using both real controls reproduces this.

Please invalidate the old save error when another control successfully changes consent, or associate it with the consent operation it describes. This issue was also present before the latest fixes.

src/components/organisms/PulseConsent/PulseConsent.tsx
```diff
@@ -0,0 +1,91 @@
...
+    >
+      <div className="space-y-2">
+        <Typography as="h2" size="lg" className="font-semibold">
+          Help improve Pubky
+        </Typography>
+        <Typography size="sm" className="text-muted-foreground">
+          {CONSENT_DESCRIPTION}
+        </Typography>
+        <Typography size="sm" className="text-muted-foreground">
+          Optional. You can use Pubky without analytics and change your choice anytime in Analytics settings or Settings
```
infin1t3: [P3] Limit the analytics promise to the Pulse setting this controls

When both Pulse and Plausible are configured, Decline disables Pulse, but `RootContainer` still renders the Plausible script whenever its domain and script URL are present (`ContainerRoot.tsx:38`). The promise that the visitor can use Pubky "without analytics" therefore exceeds what this choice does.

Please name Pubky Pulse in this copy and clarify the scope of the analytics controls. That fixes the wording without requiring a change to the existing Plausible policy. This was present in the earlier PR version and matches the Plausible part of secondl1ght's first comment.

infin1t3 (changes requested):
@Jasonvdb Rechecked `4253e5e` against `dev` (`57a240d`). The three scenarios from my previous review now pass: delayed consent notifications, restored sessions across withdrawal, and automatic network errors bypassing the application policy.

Requesting changes for the new P2 below: reconciling a stale tab deletes fresh, current-consent offline events from another tab. The two P3 comments concern UI state and consent wording already present in the earlier version of this PR; they were not introduced by the latest fixes. The wording comment overlaps with secondl1ght's Plausible observation.

Validation: the full unit suite, lint, typecheck and changed-file formatting checks pass; the current CI build is green. Focused Chromium probes with SDK 0.8.0 reproduce the queue loss, and a fresh-tab control preserves and delivers the same event. A component test reproduces the stale warning. Telemetry transport was intercepted locally throughout.

The documented timer/offline-replay limits remain unresolved concerns: mocked SDK ordering exposes the missing transport guard, but native freeze/resume and debugger-pause attempts did not reproduce a send before reconciliation. I have not counted those as additional confirmed findings.

src/hooks/usePulseConsent/usePulseConsent.ts
Jasonvdb: Reproduced, thanks. The save error was local state in each hook instance, so each control only knew about its own failed write.

It now lives in the shared consent store next to the choice (`pulse-consent.ts`), and `usePulseConsent` reads it from there. Any successful write clears it for every control, and so does another tab storing a choice. One visible change: after a failed write both the banner and the Privacy switch show the error, which is accurate because nothing was stored.

Moving the flag exposed a second problem: the banner's Close button was tied to the error, so it could disappear while the banner was still open. Close now shows whenever a choice already exists (`PulseConsent.tsx:62`).

Fixed in 91dfb966 and 240f8b58, with a component test that uses both real controls.

src/components/organisms/PulseConsent/PulseConsent.tsx
Jasonvdb: Agreed, the copy promised more than the choice does.

Every string this PR adds now names Pubky Pulse. The banner says "this choice covers Pubky Pulse only. You can use Pubky with Pulse off" (`PulseConsent.tsx:46`), and the switch label, reopen button, aria-label and save error were renamed to match. As you suggested, I left the Plausible policy alone. `docs/environment.md` now states that this consent covers Pulse only and that the optional Plausible script is configured separately.

Fixed in 0866a49b.

Jasonvdb:
@secondl1ght thanks for the careful pass. I went through all five. Here is where each one landed.

**1. Collected data vs the consent text.** Agreed. `deviceInfo` is now set explicitly: OS and browser stay on, because they are usually what makes a browser error actionable, and language is off because nothing in the app uses it (`pulse.ts:136`). The banner now says it records which browser and operating system you use. For Plausible I fixed the wording rather than gating it: the copy names Pubky Pulse and says the choice covers Pulse only. c29aac35, 0866a49b.

**2. A failed init treated as running.** Agreed. `initPulse()` now returns the SDK result and `running` is only set when `status === 'enabled'` (`pulse.ts:174`), so a failed or disabled start is retried on the next consent change, focus or pageshow. Re-initialising a live client is a no-op in the SDK, so a retry cannot install a second set of collectors. I kept the started-under marker when init fails on purpose: it describes state the tab may still hold from an earlier page load. f2380a25.

**3. Testnet gate.** Agreed. I put it in `getPulseConsent()` (`pulse-consent.ts:22`) rather than `initPulse()`, so the SDK, the banner and the Privacy switch all turn off together on testnet, and anything an earlier load stored gets deleted. 7cb843fa.

**4. Split wiring.** Took half of this. The ignore list is now one constant, `OBSERVABILITY_IGNORE_ERRORS` in `sentry.constants.ts`, and both sinks spread it, so they cannot drift. I did not add `capturePulseException`. `beforeSendPulse` already runs on every Pulse path (the factory, the boundaries and the SDK's own unhandled handlers), so a wrapper would add no policy, and `error.factories.ts` cannot import `pulse.ts` without closing an import cycle through `env.ts`. Instead the header comment in `pulse.ts` now says which files may import the SDK and why there is no funnel. c2160478.

**5. Docs.** `docs/sentry.md` now covers both sinks (capture table, capture rule, a Pulse section, env vars, privacy). `AGENTS.md`, the Greptile rule and the editor rules are updated to match, and the docker-run recipe in `docs/environment.md` includes `PUBKY_RUNTIME_PULSE_*`. 98d0cc14, 0ef2940d.

Lint and typecheck are clean and the full unit suite passes apart from the `HotTagCard` locale test noted earlier, which is unrelated.

src/libs/observability/pulse.ts
Jasonvdb: Reproduced, and you were right that it needed a narrower SDK reset. `Pulse.reset()` was the only deletion control the SDK had, and it clears every `pulse.*` key in both storage areas, so the app had no safe way to separate "my stale tab" from "the browser's queue".

So I added it to the SDK: `Pulse.reset({ scope: 'tab' })`, released in 0.9.0 (Jasonvdb/pubky-pulse-web#26). It stops this tab's client without flushing, drops what this tab had buffered or in flight, and clears this tab's session. It leaves the shared anonymous ID and the offline queue alone. If it cannot confirm the tab cleanup, it falls back to the full purge.

The gate now splits the two cases (`pulse.ts:170`). Consent not accepted still calls the browser-wide `Pulse.reset()`. Consent accepted but this tab's marker is stale calls the tab-scoped one. So when B returns, A's event stays queued and is delivered, and B adopts A's anonymous ID and starts a new session. This also removes the anonymous-ID churn I had accepted as a trade-off last round.

There is a regression test for your sequence on both paths (a fresh page load, and a live client waking on focus), plus your fresh-tab control. It runs the real SDK and failed on the old code with the parked event missing.

One behaviour change to be aware of: a repeated Accept with no withdrawal in between now keeps the shared anonymous ID and only rotates the session. Nothing was withdrawn there, so nothing was promised deleted.

Still not closed, and now stated in `docs/environment.md`: a suspended tab can re-queue an event after another tab's withdrawal has already run, and the SDK cannot tell whose events are in the shared queue or filter them again on replay. That is the same offline-replay limit as before. Closing it needs per-event provenance in the SDK, which changes the wire format, so I have left it as a follow-up.

Fixed in f00291bc and b5e636f9.

Orlandohub (approved):

infin1t3 (approved):

PR #2529, opened 2026-09-14
chore(deps): bump @synonymdev/pubky to 0.11.0

infin1t3:
## Summary

Bumps `@synonymdev/pubky` from 0.8.0 to 0.11.0 while **keeping cookie authentication**. The grant-auth migration (the reason 0.10 was raised in Slack) is a separate, larger change and is intentionally out of scope here.

Target is 0.11.0 rather than today's 0.12.0 on purpose: 0.12.0 adds path-addressed storage requests with an `/info` probe and legacy fallback, which is a new runtime path against the v0.10.0 homeservers prod and staging run. 0.11.0 keeps the storage transport byte-identical to 0.8.0.

## Changes

- `package.json` / lockfile: `@synonymdev/pubky` 0.11.0 (exact pin).
- `HomeserverService`: the three SDK calls that changed are renamed to their cookie variants — `signup` → `signupCookie`, `signin` → `signinCookie`, `startAuthFlow` → `startCookieAuthFlow`. Same arguments, same return types. Everything else the app uses (`Keypair`, `PublicKey`, `Session`, storage, PKDNS, event stream, `Client.fetch`, `resolvePubky`) is unchanged between the two versions.
- `error.utils.ts`: the SDK now rejects `getHomeserverOf()` with `{ name: 'PkarrError' }` when the PKARR lookup itself fails (it still resolves `undefined` when the record is provably absent). That name is now mapped to a retryable Network `CONNECTION_FAILED` error instead of falling through to a synthetic 500 homeserver error. The lookup is tagged with `operation: 'resolveHomeserverRecord'`.
- `sentry.utils.ts`: the `homeserver-event-stream-connect` drop rule now also matches the Network category, since the event-stream subscribe resolves the homeserver via PKARR first and would otherwise flood Sentry on every reconnect.
- Tests: mock renames, PkarrError cases for both the non-staging sign-in path and the staging guard, a subscribe-path PkarrError drop test, and new service-level coverage for `restoreSession`.
- `docs/error-handling.md`: one paragraph documenting the SDK error-name mapping.

## Behaviour changes to be aware of

- Sessions persisted by 0.8.0 (`session.export()` snapshots) restore under 0.11.0 unchanged — the serialized record is byte-identical upstream and `restoreSession` still accepts it. No forced sign-out on deploy.
- A transient PKARR failure during recovery-phrase / recovery-file sign-in now surfaces as a retryable error instead of silently force-republishing the homeserver record. This matches the intent already documented in `homeserver.ts` (`signIn`) and closes the overwrite hazard described there.
- Bundle: WASM +119 KB gzipped, JS glue +198 KB gzipped (grant-auth code we do not use yet). No size gate in CI.

## Verification

- `npm run typecheck`, `npm run lint`, `npm run format:check`: clean.
- `npm test`: 783 files, 13441 passed, 2 skipped.
- `npm run build` (webpack): passes.
- CI E2E runs against homeserver v0.10.0, the same version pubky-stack pins for production.

Manual QA still needed before release (staging runtime config):

- Ring QR sign-in; sign-in with recovery phrase; sign-in with encrypted recovery file; invite-code sign-up.
- Reload → session restored; sign-out; mute-sync `/events-stream` connects.
- Cross-version restore: sign in on the current staging build, deploy this build to staging, reload, still signed in (PR previews are a different origin, so this must be done on staging itself).
- Confirm with infra that prod actually runs the homeserver version pubky-stack `production` pins (v0.10.0); cookie auth works on 0.10 and 0.11 either way.

## Coordination

- `epic/locks-v1` pins `0.10.0-alpha.0` and already uses the `*Cookie` methods; it must drop its pin to 0.11.0 on its next merge from `dev`.
- 0.12.0 (path-addressed transport) is deliberately skipped; revisit with the grant-auth migration or once the homeserver is on 0.11+.


src/core/services/homeserver/error.utils.ts
```diff
@@ -97,6 +100,25 @@ const throwInvalidInputError = ({ errorMessage, additionalContext }: TThrowInval
...
+ * error stays retryable and is never treated as a homeserver HTTP failure.
+ *
+ * @param errorMessage - The original error message
+ * @param additionalContext - Additional context to add to the error
+ * @returns Never (always throws)
+ */
+const throwPkarrLookupError = ({ errorMessage, additionalContext }: TThrowPkarrLookupErrorParams): never => {
+  throw Err.network(NetworkErrorCode.CONNECTION_FAILED, errorMessage || 'PKARR lookup failed', {
+    service: ErrorService.Homeserver,
+    operation: (additionalContext.operation as string | undefined) ?? 'unknown',
```
Taewa, line 117: AI found a potential issue and I think it's worth to check.

---

`resolveHomeserverRecord` is tagged with its `operation`, but the other paths that can also hit PKARR are not. Without the tag, `throwPkarrLookupError` falls back to `operation: 'unknown'`:

1. `restoreSession`
2. `republishConfiguredHomeserver`
3. `signUp`

(1) is the one that matters — it runs on every app start and retries, so a relay outage would fill Sentry with untaggable errors, and the drop rules match on `operation` too.

Could you pass `operation` in `additionalContext` at these three call sites, the same way `resolveHomeserverRecord` does?

Taewa (dismissed):
There is a comment about the `unknown` operation. 
Apart from that LGTM. Feel free to fix it or create a follow-up ticket to tackle later :)

src/core/services/homeserver/error.utils.ts
infin1t3: Good catch — tagged.

`signUp`, `restoreSession`, and `republishConfiguredHomeserver` now pass `operation` in `additionalContext`, same as `resolveHomeserverRecord`. A `PkarrError` on those paths is `CONNECTION_FAILED` with a real operation instead of `unknown`.

`3bdcfa05`

Cookie `restoreSession` itself is usually an auth/session error, not PKARR; the startup lookup still goes through the already-tagged `resolveHomeserverRecord`. Republish is the path that actually writes PKARR. Either way the three sites are now filterable.

Taewa (approved):
LGTM

PR #2536, opened 2026-09-16
fix(profile): title the copied profile link toast correctly

### LINKED ISSUE #2197: [Bug]: Wrong toast message when copying profile link
### Description

When clicking the 'Profile Link' button on profile page the toast message reads "Pubky copied to clipboard"

### Expected Behaviour

Toast message shouild be like "Profile link copied to clipboard"

### Steps to Reproduce

1. Go to profile page
2. Click 'Profile Link button
3. Observe toast message is wrongly "Pubky copied to clipboard"

### Screenshots

https://github.com/user-attachments/assets/32291e71-69fd-4573-82e1-7cbd12e5de64

### Console Log

```shell

```

### Browser & Version

_No response_

talosmachina:
## What

Copying the profile link from the profile page showed `Pubky copied to clipboard`. It should name what was copied.

Requested by @Orlandohub.

`useProfileActions.onCopyLink` called `copyToClipboard(link)` on a `useCopyToClipboard()` instance created without options, so the toast fell back to the hook's default title, which is the pubky wording. The sibling `useProfileMenuActions` already passes `successTitle: 'Profile link copied to clipboard'`.

The hook now creates a second instance for the profile link, so `onCopyPublicKey` keeps its own pubky title and the link gets the correct one.

## Evidence

Captured on a local `dev` build, signed in against the staging homeserver. Both screenshots are the same button on the same profile; the clipboard write is stubbed in the page so the toast renders rather than the browser's permission error.

Before:

profile link toast before the fix

After:

profile link toast after the fix

## Tests

```
npx vitest run --project unit src/hooks/useProfileActions
Test Files  1 passed (1)
     Tests  20 passed (20)
```

The `useCopyToClipboard` mock now records the options it is called with, and a new test asserts the profile link requests the `Profile link copied to clipboard` title. The existing assertions on the copied value are unchanged.

Closes #2197


Orlandohub:
profile-link-toast

https://github.com/user-attachments/assets/37bdae2b-a36b-40ca-8441-f8086084b325

Orlandohub (approved):

PR #2537, opened 2026-09-16
fix(logo): keep the logo on the current custom feed

### LINKED ISSUE #1589: [Bug]: Pubky logo navigates to default feed when custom feed is applied
### Description

When using a custom feed, clicking the Pubky logo in the header navigates to default feed and reloads page.

### Expected Behaviour

Clicking Pubky logo should scroll to top of page and not navigate away from custom feed.

### Steps to Reproduce

1. Apply custom feed
2. Scroll down the feed
3. Click Pubky logo in header
4. Observe page reloads and top of default 'global' feed is displayed

### Screenshots

_No response_

### Console Log

```shell

```

### Browser & Version

_No response_

talosmachina:
## What

From a custom feed (`/feed/<id>`), clicking the header Pubky logo navigated to the default feed. Expected: scroll to the top of the current feed and stay there.

Requested by @Orlandohub.

## Why

`Logo` decided "already on this feed" with `pathname === APP_ROUTES.HOME`, so a custom feed never counted and `handleFeedNavClick` let the link navigate to `/home`.

The `(feeds)` layout is persistent across the feed cluster (`/home`, `/feed/[id]`, `/bookmarks`, `/search`), so `/feed/<id>` is the same cluster as `/home` and the logo should behave there exactly as it does on `/home`.

## Change

`Logo.tsx` now resolves the active feed route through the shared `isNavItemActive` helper for both `/home` and `/feed`, so a custom feed is treated as active and the click scrolls to top instead of navigating.

## Verification

- New unit test: on `/feed/feed-abc123` the logo click smooth-scrolls to top and does not set the one-shot feed scroll flag.
- The test fails against the previous `Logo.tsx` (`AssertionError: expected "vi.fn()" to be called with arguments: [ { top: +0, behavior: 'smooth' } ]`), so it catches this regression rather than just documenting current behaviour.
- 41 tests pass in the Logo suite.
- In the local dev build, on `/feed/qa-logo-check`, clicking the header logo left `location.pathname` at `/feed/qa-logo-check`.

Note on the screenshot: this fix is visible in the address bar, which a screenshot cannot show. The attached image is the app on a `/feed/<id>` route with the logo clicked, as a record of the surface under test.

## Closes

Closes #1589

1589-after-logo-click


Orlandohub:
before-logo

after-logo

https://github.com/user-attachments/assets/27bb2bba-37d4-4849-966f-3dafb75ca3f6

Orlandohub (approved):

Issue #2551, opened 2026-09-16
Residual ordering glitches after #2523: rows without a Nexus cursor and local prepends need a score from Nexus


secondl1ght:
## Summary

The #2523 fix makes stream cursors come only from Nexus (`last_post_score` persisted per cached row) and keeps cached rows in Nexus order. Two ordering glitches remain, both bounded, neither a stall, loop or skip. They share one root cause: the client never learns a post's score, only `indexed_at`, and `indexed_at` moves on edit and delete. They are documented in `docs/local-first.md` (_Stream Pagination Cursors_) and tracked here so they are not forgotten.

## The two residuals

1. **Cursor-less row whose tail was edited after a full page of newer posts arrived.** A bootstrap-seeded row (every sign-in) or a pre-cursor row has no Nexus cursor, so its first seam seeds from the tail post's `indexed_at`. If that post was edited after at least `NEXUS_POSTS_PER_PAGE` newer posts were published, the seam page shares no id with the row and cannot be told apart from the normal page below the tail. It is served once below the cached ids and appended there, until the row is rebuilt (cache max-age, dirty registry, or bootstrap). Overshoots of less than a page are handled exactly by aligning the page on the row's own ids (`LocalStreamPostsService.keepIdsBelowRow`).
2. **Own post written while the "N new posts" pill is showing.** The unread merge keeps polled order and does not place row ids by `indexed_at` (an edited row head would otherwise be promoted above newer polled posts and hand its bumped timestamp to the next head poll, see #2535). So an own post prepended locally after a poll sits below the polled posts after a refresh, until the next poll returns it and it takes the unread slot.

Related: #2535 (head poll bound read from `indexed_at`).

## What would close them

All three go away if Nexus exposes the ordering key it already has:

- `score` on post details in `posts/by_ids` and stream hydration responses (the sorted-set score of the post in its timeline stream), or
- `first_post_score` on `stream/posts/keys` responses and a `last_post_score` on the bootstrap stream.

With a score per post the client can seed a cursor-less row exactly, place a local prepend exactly, and poll the head from a score instead of a timestamp. This is a much smaller ask than moving the client-side filtering to Nexus.

Client-side alternatives considered and rejected: comparing against the head post's `indexed_at` (unsound once the head itself is edited), seeding cursor-less rows from the top of the stream (a row whose ids are not yet indexed would never align), and decoding the creation time from the post id (close to the score but not the score; would need its own validation).

Issue #2539, opened 2026-09-16
Document and reconcile the three pagination budgets introduced by #2523


secondl1ght:
## Summary

The #2523 fix leaves three independent bounds on client-side filtered pagination. Each exists for a reason, but nothing records the reasons together, so the next change here is likely to remove or duplicate one of them.

| Budget | Where | Bounds | Why it exists |
| --- | --- | --- | --- |
| `MAX_FETCH_ITERATIONS` (20) | `PostStreamQueue.collect` | fetch iterations per stream-layer round | Caps one `collect` call; a cached region yields up to 20 pages per round, an unhydrated region only one because the controller's post-hydration pass filters after the round. |
| `STREAM_LOAD_MAX_RAW_SCAN` (600) | `useStreamPagination` | raw posts scanned per `loadMore` | Chains rounds inside one load so the loading block does not blink per round, and counts posts rather than rounds because rounds are so uneven. |
| `TIMELINE_MAX_UNPRODUCTIVE_AUTO_LOADS` (3) | `useInfiniteScroll` via the renderers | consecutive auto loads that grow nothing | The only bound the user can see: past it the sentinel stops and a manual Load more takes over, so a filtered region can never chain loads to the end of the stream. |

## Ask

- Write the rationale into `docs/local-first.md` next to _Stream Pagination Cursors_ (a table like the above) so the three are understood as layers, not duplicates.
- Decide whether the queue's iteration cap and the hook's raw-scan budget can be expressed as one bound (raw posts per load, enforced by the queue) without losing the per-round loading behaviour.
- Consider whether the head poll (`StreamCoordinator`) needs its own bound now that #2535 is known.

No behaviour change is requested here; this is a maintainability follow-up to the #2523 PR.

Issue #2538, opened 2026-09-16
Visual feed backfill effect chains loadMore outside the unproductive-load budget


secondl1ght:
## Summary

`VisualTimelinePosts` has a backfill effect that calls `loadMore()` directly whenever the loaded post ids produce no tiles (`shouldBackfillInitialRows`: `postIds.length > 0 && !hasRows && hasMore && !loading && !loadingMore`). It re-fires after every completed load while rows are still empty, and it runs outside the `useInfiniteScroll` unproductive-load budget introduced for #2523.

This is the same bug class as #2523 (client-side filtering chaining loads with no bound), through a different filter: the tile pipeline rather than the stream filter. A Visual-layout view of a stream whose posts carry no images (a text-only profile, a text-heavy custom feed) chains loads until the stream ends or a tile appears. Each load delivers visible post ids, so the hook's raw-scan budget does not apply; the renderer-level budget does not see the effect's loads.

Pre-existing; found while closing #2523. Kept out of that PR to keep it reviewable.

## Where

- `src/components/organisms/Timeline/Feed/TimelineFeed/VisualTimelinePosts.tsx`, the `useEffect` keyed on `shouldBackfillInitialRows`.

## Proposed fix

Route the backfill through the same budget as the sentinel: count a backfill load that does not grow `rows.length` as unproductive (the hook's `itemCount` for this renderer should be the tile row count, not `postIds.length`), stall into the existing manual Load more once `TIMELINE_MAX_UNPRODUCTIVE_AUTO_LOADS` is spent, and let `resumeAutoLoad` clear it. Alternatively cap the backfill at the same constant directly. Either way the empty-rows state must keep the wrapper's `stalled` prop in sync so the skeleton does not show forever.

## Test

A renderer test with `useVisualFeedTiles` mocked to return no rows and `hasMore: true` should assert that `loadMore` is called at most `TIMELINE_MAX_UNPRODUCTIVE_AUTO_LOADS + 1` times before the Load more button renders.

Issue #2535, opened 2026-09-15
New-posts poll can skip posts when the cached head post was edited or deleted


secondl1ght:
Spun out of #2523. Same root cause as that fix, different symptom: Nexus keeps a post at its original stream position (its score) but bumps `details.indexed_at` when the post is edited or deleted, so a stream position derived from `indexed_at` can be wrong. #2523 closed this for the pagination tail; the head poll still derives its bound from `indexed_at`.

## Where

- `StreamCoordinator` polls for new posts with `end = getStreamHead(streamId) + 1`.
- `LocalStreamPostsService.getStreamHead` returns the `indexed_at` of the first cached id (unread stream first, then the main stream). `PostStreamApplication.getMainStreamHeadTimestamp` does the same for the stale check.

## Failure

1. The feed is cached with head post H (score `s`, `indexed_at = s`).
2. Posts A and B are created after the cache fill (scores `> s`), not yet polled in.
3. H is edited (or deleted) at time `e > score(B)`. Nexus bumps H's `indexed_at` to `e` while H stays at position `s` in the stream. A viewport TTL refresh writes `indexed_at = e` to the local row.
4. The next poll asks for posts with score `>= e + 1`. A and B have scores below `e`, so they are never returned. The "N new posts" pill stays quiet for them until the next initial load clears or rebuilds the cache (bounded by the 5-minute stale window).

Low severity: a few posts are late rather than lost, and only when the newest cached post is edited while the feed is open.

## Fix direction

Mirror the #2523 tail fix for the head: never derive the poll bound from a post's current `indexed_at`.

- Client-side, no Nexus change: record on the `post_streams` row the position of the newest fetched post at the moment it first entered the cache (`headCursor` next to `tailCursor`), and poll from that. A later local refresh of the post's details must not move it. Nexus only returns `last_post_score` per page, so the exact head score is not available; the first-seen `indexed_at` is the right approximation (it equals the score unless the post was already edited before we ever fetched it, in which case the skipped range is already below the head in stream order and covered by the cache walk).
- Cleaner, needs Nexus: have the keys endpoint also return the first post's score (`first_post_score`) so the head bound is exact.

Related: #2523, #1569.

secondl1ght:
Raising the priority of this one after the #2523 hardening work. It is not a local-action edge case: it triggers on a **remote** edit.

**Sequence, reproduced in a full-stack simulation** (real hook → controller → application → queue → local services, Nexus simulated):

[...]

PR #2552, opened 2026-09-16
fix(streams): resume pagination from Nexus cursors, never indexed_at

### LINKED ISSUE #1569: [Bug]: Deleted posts can cause loading of posts to break when scrolling down profile page
### Description

When viewing someone's profile, if they've deleted a bunch of posts this appears to break loading when scrolling down the list of posts. In this scenario, it's not possible to scroll to the bottom of the list of posts. See video for demonstration of what 'broken' looks like.

### Expected Behaviour

It's always possible to view all undeleted posts on a user's profile page and eventually the bottom is reached 

### Steps to Reproduce

Have not reproduced it myself but I expect if you have many (like over 30 posts) and delete a bunch from the middle this this could occur when:

1. Visit user's profile
2. Scroll down the list of 'Posts'
3. Observe a flickering of 'Loading more posts...' and it's not possible to scroll further down

### Screenshots

https://github.com/user-attachments/assets/127f2a5b-28df-4e6f-a4f2-7b8a601ea1ba

after revisiting their profile, all of the "This post has been deleted by its author" had been removed but same bug occurs

https://github.com/user-attachments/assets/bf44885d-eafe-4671-b427-b0251cd84f70

### Console Log

```shell

```

### Browser & Version

_No response_

### LINKED ISSUE #2523: Feed stutters, jumps and cannot be scrolled past: pagination chains over a filtered region without progress
The feed's automatic pagination chains rounds through a region of posts the viewer cannot see (deleted or muted), and the chain can end up making **no progress at all**. The feed then flickers between two states and cannot be scrolled past.

Reported by @BitcoinErrorLog while scrolling the home feed; the post it was first noticed at is a bystander (see below). Assigned to @secondl1ght.

## Symptoms as reported

- While scrolling the home feed past one recent post, the feed stutters and cannot be scrolled past; the jump starts before the post itself is visible.
- Follow-up: it alternates between two states when reaching that point in the feed, and it is not possible to scroll further.
- Opening the same post on its own post page does **not** reproduce it.

## Root cause

1. Nexus paginates by a raw cursor and the client filters after the fact (mutes, deleted posts, collections). A page whose posts are all filtered returns zero visible ids while `hasMore` stays true, by design — `useStreamPagination` keeps it: "hasMore reflects the stream end, not the filtered count".
2. The infinite-scroll sentinel is therefore still on screen and refires (`useInfiniteScroll`, `threshold: 3000`), so rounds chain through the filtered region until the true stream end. Each round mounts `TimelineLoadingMore` (two skeleton cards) at the tail of the list and unmounts it when the round resolves with nothing visible. That is the flicker: the feed alternates between two states, with and without the loading block.
3. At the bottom of the feed the document height pulses with that block, so the browser clamps the scroll offset on every shrink. That is the feed jumping up and down, and why it cannot be scrolled past.

## Why it never recovers on its own

Waiting does not help: progress depends on a cursor that can freeze.

- **Score cursor.** `advanceCursor` returns `rawPage.lastScore ?? prevCursor`, so a page that comes back without a score leaves the cursor exactly where it was. Inside one `PostStreamQueue.collect()` call the same page is then re-requested up to `MAX_FETCH_ITERATIONS` (20) times, `nextCursor` comes back `undefined`, and `useStreamPagination` skips its `setStreamTail` write. The next automatic round starts from the same cursor again: the same page every round, forever. `hasMore` never clears either, because `reachedEnd` only ever comes from the backend.
- **Cache-walk anchor.** `resolveResumeAnchor` is `lastRawPostId ?? nextPageIds[last]`, and `lastRawPostId` is only assigned when the slice returned ids. A fully filtered page has neither, so `nextAnchor` is `undefined`, `setLastPostId` is skipped and the anchor holds.

The 20-iteration cap in `PostStreamQueue.collect()` bounds one call, not the caller: the queue's own comment says an auto-loading sentinel that refires on an empty-but-not-ended result "will chain collects until the true stream end".

The guard for exactly this exists and is wired nowhere: `useInfiniteScroll` accepts `itemCount` + `maxUnproductiveLoads` and returns `isStalled` / `resumeAutoLoad`, and no production caller passes them. `TimelinePosts` and `TimelineGridPosts` both pass only `onLoadMore`, `hasMore`, `isLoading`, `threshold`, `debounceMs`.

## Verified vs not verified

- Verified by reading the shipped paths in `dev`: the `hasMore` behaviour on a filtered page, the sentinel refire, the loading block at the tail, `advanceCursor` holding on a missing score, the anchor fallback, and the unwired budget.
- **Not verified:** which of the two freezes the reported session actually hits, and any live reproduction. Signed out, the public home feed stops at 10 cards with no churn, so a filtered region cannot be reached there. A signed-in session with the stream slice responses logged around the stuck position would settle it.

## The reported post is not the trigger

The record fetched from the homeserver is text only:

```json
{"attachments":[],"content":"Sorry in advance, a bit of spamcobo content is coming","embed":null,"kind":"short","parent":null}
```

No media, no embed, no link card; one emoji tag (2 taggers), 1 repost, 0 replies. Nothing in that payload can stall a scroll, and the defect reproduces from the feed's pagination layer, not from the card. Reported with that post: https://pubky.app/post/ssngpfm9x3i8fzp8kp1ipuqdyamhkkka9x9a1g4qd6i4qnsmi7jo/0035PDWGM8VSG, authored by @jacobo-synonym.

## Related

- #1569 — same symptom, open since March: flicker of "Loading more posts" and no way to scroll further down, on a profile with many deleted posts. Very likely the same defect reached from a profile feed; worth keeping the two in step.
- #2417 — documents this loop across the five divergent client-side filtering implementations ("effectively an unbounded fetch loop through a filtered region").
- #2200, #1746 — same class: filtered pages consumed with no progress feedback.
- #2170 — closed, earlier cursor-arithmetic bug cited by #2417.
- #2100 — the bookmark-stream cursor seam, which is why the `cursorForPost` resolver exists.

## Suggested direction

- Guarantee progress: when a page omits its score cursor, derive the next one from the last raw id returned instead of holding the previous one. The `cursorForPost` / `getStreamCursorTimestamp` resolver already does this for the overflow-buffer path.
- Bound the no-progress chain so the loading block cannot pulse unboundedly, and give the stall a visible affordance.

## Separate defect in the shipped CSS (unrelated to the scroll bug)

`src/app/globals.css:74-90` declares three `@property` rules with `initial-value: 0%`. In the CSS bundle served by pubky.app the percent sign is gone:

```
@property --landing-swirl-mask-hole{syntax:"<percentage>";inherits:true;initial-value:0}
@property --landing-swirl-mask-ring-start{syntax:"<percentage>";inherits:true;initial-value:0}
@property --landing-swirl-mask-ring-end{syntax:"<percentage>";inherits:true;initial-value:100%}
```

`0` is not a valid `<percentage>`, so Chrome rejects both rules ("Ignored @property rule: An @property rule is ignored because it contains an invalid property or is missing a required one ... initial-value: 0"). The third rule keeps `100%` and is registered normally. The two rejected properties therefore have no registered syntax or initial value: they cannot interpolate, and a gradient stop that uses them resolves to an invalid value at computed-value time. Something in the build rewrites `0%` to `0`; either write the source so the minifier leaves it alone or stop the rewrite. Worth its own issue rather than being carried by this one.

(Also present in the same console and not investigated here: "Response was blocked by CORB (Cross-Origin Read Blocking)" and "A form field element should have an id or name attribute".)

secondl1ght:
## Summary

Closes #2523. Closes #1569.

Both are the same defect reached from different feeds: the home feed for #2523, a profile with many deleted posts for #1569 (re-indexed tombstones sit below the cached tail exactly like edited posts do; the profile geometry has its own test and simulation scenario). Earlier work on this class: #2251, #2290, #2171, #2170 (closed).

**Not closed by this PR, referenced only:** #2417 (the ADR and convergence across all five filtered surfaces; this PR settles the infinite-scroll feeds and gives Search → Collections a deeper per-click scan, but Discover, Search → People and Notifications are untouched and no ADR is written), #2200 (Discover keeps its own 5-page per-click budget), #1746 (notifications).

**Root cause.** Nexus orders a timeline stream by score (the post's original index time) and keeps an edited or deleted post at that position, but bumps `details.indexed_at`. The client derived its resume cursor from the tail post's `indexed_at` and kept its stream cache sorted by `indexed_at`. After an edit or delete below the cached tail, the cursor landed above posts that were already cached, Nexus re-served them, the raw anchor sat on one of them, and the feed alternated between two states without ever advancing. Verified on production Nexus before the change (out-of-order posts and re-indexed tombstones present in the All feed).

**What changes.**

- **Cursors come only from Nexus.** `post_streams` rows persist the `last_post_score` of the deepest page fetched into them (`tailCursor`, a plain row field, no `DB_VERSION` bump). Every cache→Nexus seam resumes from it. Pages are appended in stream order, never re-sorted by `indexed_at`; a page that adds nothing leaves the row untouched.
- **Rows without a cursor** (bootstrap-seeded, or written before this change) seed once from the tail timestamp, and the seam page is aligned on the row's own ids: only ids listed after the last cached id are served or appended. The anchor after a Nexus page is positioned in the persisted row, so a repeated page or another walker's write can never move it up or past unserved ids.
- **A lost anchor** (its post deleted or un-bookmarked mid-scroll) re-anchors on the deepest id the renderer has shown instead of skipping to the row tail. A row emptied by deletion, or re-seeded by a local prepend, drops its cursor.
- **The queue** never synthesizes a cursor from a served post, reports the raw ids it consumed (buffer-served rounds included), and stops a round that moves nothing instead of re-issuing the same request.
- **One `loadMore` keeps scanning** through fully-filtered rounds up to `STREAM_LOAD_MAX_RAW_SCAN` raw posts, so the loading block stays mounted instead of blinking once per round, and state is committed once per load.
- **Renderers budget unproductive auto-loads** (`TIMELINE_MAX_UNPRODUCTIVE_AUTO_LOADS`) and hand over to a manual **Load more** button when the budget is spent; any change of the rendered count re-arms auto-loading. Applied to `TimelinePosts`, `TimelineGridPosts`, `VisualTimelinePosts` and `RepliesWithParent`. The Visual feed counts the tiles its pipeline tracks, so text-only pages cannot re-arm the budget.
- **Unread merge and bootstrap.** The merge keeps the polled order (nothing is re-sorted by `indexed_at`), a late head poll cannot land older ids above newer ones, and a bootstrap clears the unread row of the timeline row it replaces.
- **Reply rows** (cursor-less) are kept newest-first at creation and when a chunk adds ids; a row with a Nexus cursor is never re-sorted.

**How it was hardened.** Beyond unit tests per layer, `useStreamPagination.simulation.test.tsx` drives the real hook → controller → application → queue → local services on fake-indexeddb against a modelled Nexus, with seeded random runs where edits, deletes, publishes, TTL refreshes, bootstraps, head polls and another tab's pages reach the client only the way they reach a second user's client. Four review passes plus 300-seed sweeps found and fixed: re-anchoring after a deleted anchor, the unread merge re-sorting by `indexed_at`, empty pages rewriting the row, bootstrap leaving stale unread ids, the seam page of a cursor-less row serving posts above the feed head, a concurrent walker moving the anchor, cold reply rows in hydration order, out-of-order head polls, a skipped cursor write, and a Visual feed observer that could freeze on an attachment Nexus no longer returns.

**Known, out of scope, tracked.**

- #2551: two residual ordering glitches (a bootstrap row whose tail was edited after a full page of newer posts; an own post written while the new-posts pill is showing). Bounded, self-healing, and closed for good by Nexus exposing the post score. Documented in `docs/local-first.md`.
- #2535 (head poll bound read from a remotely edited head's `indexed_at`): reproduced by the simulation and kept as an `it.fails` test until the follow-up PR lands.
- #2538: the Visual feed backfill effect chains loads outside the new budget (same class, separate PR).
- #2539: rationale for the three budgets (queue iteration cap, raw-scan per load, unproductive loads per renderer); overlaps with the ADR #2417 asks for.
- A Nexus page boundary on tied scores still skips tie-mates (`start = score - 1`); a Nexus protocol limitation, not addressed here.

## Verification

- [x] `npm test` (full): 791 files, 13786 tests, 1 expected failure (#2535), 2 skipped
- [ ] `npm run build` (route- or config-wide changes)
- [ ] `npm run test:vrt` (only when a VRT surface changed)

Also:

- 300-seed sweeps of the simulation on every commit (16 seeds committed).
- Real app against staging Nexus, guest session (sign-in needs the staging phrase, not available in the sandbox):
  - **Happy path.** Home feed, a profile page, a post page, home again: the feed grows 10 → 45 posts in two minutes, the cached row grows 10 → 20 → 30 → 40 → 50 with unique ids and each page's `last_post_score` as its cursor, every page request carries a strictly descending cursor (the only repeats are retries of requests the sandbox proxy dropped), no Load more button, no client errors.
  - **Filtered region.** With a Playwright route rewriting every post beyond the first page to a filtered kind, the feed stalls into **Load more** after exactly 183 page requests (three loads of 600 raw posts plus the head page), the click resumes, and the loading block stays mounted across 197 samples with zero remounts.
- **QA ask for #1569:** a signed-in scroll on staging through a profile whose author deleted many posts, to confirm the bottom is reachable and the loading block no longer flickers.
- **QA ask for the Visual feed and reply threads:** both changed in the last two review rounds and were verified by unit tests only; a scroll through Home → Visual with content set to All and one long reply thread on staging would close that gap.

## Callouts

- [x] Changes a `src/config/*` limit (product-wide policy): two **new** limits in `src/config/feed.ts`, `STREAM_LOAD_MAX_RAW_SCAN = 600` (raw posts one load may scan through filtered rounds) and `TIMELINE_MAX_UNPRODUCTIVE_AUTO_LOADS = 3` (consecutive auto-loads that grow nothing before the manual Load more takes over). Product decision: infinite scroll now has a floor. Past the budget the user sees a **Load more** button instead of an endless spinner; it takes three consecutive loads of 600 raw posts with nothing visible, so it should be rare, but it is a new visible state.
- [x] Changes a convention → `docs/local-first.md` (_Stream Pagination Cursors_), `docs/data-patterns.md`, `AGENTS.md` one-line rule ("stream cursors only from Nexus, never `indexed_at`")
- [ ] Bumps `DB_VERSION` or changes a Dexie index map: **no**. `tailCursor` is a non-indexed row field on `post_streams` (`&id` only).
- [ ] Touches a surface covered by a VRT: the new `TimelineLoadMore` molecule renders only in the stalled state, which no VRT covers.
- [ ] Invalidates a Cypress spec: none identified. The sentinel is now unmounted while stalled, so a spec that scrolls a feed with a fully-filtered tail past the budget would see `[data-cy="timeline-load-more"]` instead of more loads. Flagging for QA.

## Screenshots

Stalled state, resumed load and second stall were captured on staging during verification (desktop); the button is the existing secondary `Button` atom below the feed. Happy to attach on request.


src/core/services/local/stream/posts/posts.ts
```diff
@@ -407,38 +419,97 @@ export class LocalStreamPostsService {
...
+      return stream;
     }
 
     // Check for duplicates before adding
     const existingIds = new Set(postStream.stream);
     const newPostsToAdd = stream.filter((id) => !existingIds.has(id));
+    const deepestCursor = this.deepestTailCursor(postStream.tailCursor, tailCursor);
+    const nextTailCursor = this.tailCursorFields(deepestCursor);
+
+    if (newPostsToAdd.length === 0) {
```
infin1t3, line 465: [P2] Normalize cold reply rows before returning a duplicate-only page

A cold ascending reply fetch hydrates the missing replies before persisting the page. Hydration can create the reply stream in the by-IDs response's order, so every ID is already present when this branch runs. The early return skips the descending sort that previously normalized the row. `useReplyStream` reverses that row for display: with same-author replies hydrated oldest to newest, the user sees newest to oldest.

A focused reproduction through PostStreamApplication, this service and Dexie fails here and passes with the base service. Preserve normalization for these cursor-less reply rows, either at creation or when persisting a duplicate-only ascending page, while retaining Nexus order for score-backed timeline pages.

src/components/organisms/Timeline/Feed/TimelineFeed/VisualTimelinePosts.tsx
```diff
@@ -410,13 +412,16 @@ export function VisualTimelinePosts({
...
-  const { sentinelRef } = useInfiniteScroll({
+  // load rounds keep chaining toward the first visible posts, within the budget of
+  // rounds that grow nothing; past it the manual Load more below takes over (#2523).
+  const { sentinelRef, isStalled, resumeAutoLoad } = useInfiniteScroll({
     onLoadMore: loadMore,
     hasMore: hasMore && (hasRows || postIds.length === 0),
     isLoading: loadingMore || isInitialLoading,
     threshold: 3000,
     debounceMs: 20,
+    itemCount: postIds.length,
```
infin1t3: [P2] Measure mosaic progress for the unproductive-load budget

Home/Search Visual with content set to All still consumes a mixed stream. Once a row exists, this observer is active. A page of text-only posts increases `postIds.length`, but `useVisualFeedTiles` filters those posts out and the mosaic stays the same size. `useInfiniteScroll` treats the larger ID count as progress and resets its budget, allowing repeated automatic requests without handing over to Load more. This is the observer path after rows exist, separate from #2538's initial backfill.

Use rendered mosaic progress for the budget and account for pending tile/file resolution as loading. Add a test where IDs grow but rows stay flat and Load more appears after the configured unproductive-load limit. The current tests mock the stall state and do not exercise that transition; this finding is based on source tracing.

src/core/services/local/stream/posts/posts.ts
```diff
@@ -189,15 +195,21 @@ export class LocalStreamPostsService {
...
+    const rest = postStream.stream.filter((id) => !unreadIds.has(id));
+
+    // Both parts are already in stream order: every head poll prepends a Nexus page that is
+    // newer than the previous one, and the cached row keeps the order its pages arrived in.
+    // Nothing is re-sorted by indexed_at — that would float an edited or deleted post above
+    // its real position (#2523), and an edited row head promoted above the polled posts
+    // would then feed its bumped indexed_at to the next head poll. The one cost is an own
+    // post written after the poll: it sits below the polled posts until the next poll returns
+    // it, at which point it takes the unread slot. The tail's resume cursor is untouched.
+    const combinedStream = [...validUnreadPosts, ...rest];
```
infin1t3, line 210: [P2] Handle head polls that complete out of order before preserving unread order

The assumption above this merge does not hold across tabs, which have separate query clients but share IndexedDB. Suppose an older poll returns IDs `[10, ..., 1]` and a newer poll returns `[11, ..., 2]`. If the newer request completes first, the late older response prepends its unique ID `1`, leaving unread order `[1, 11, ..., 2]`. This merge then puts `1` above newer posts on the next initial load or refresh and clears the unread row.

A deferred-response reproduction through the application and Dexie fails at this head and passes with the base service, even with timestamps matching Nexus order. Preserve ordering across overlapping poll responses, or prevent stale responses from prepending older IDs. Restoring an unconditional indexed_at sort would undermine this PR's handling of edited posts, so the fix needs to retain Nexus ordering.

src/hooks/useStreamPagination/useStreamPagination.ts
```diff
@@ -135,84 +145,91 @@ export function useStreamPagination({
...
-        const updatedPostIds = isInitialLoad ? newUniquePostIds : [...postIdsRef.current, ...newUniquePostIds];
-        postIdsRef.current = updatedPostIds;
-        const displayedState = resolveDisplayedPostIds(
-          updatedPostIds,
-          optimisticPostIdsRef.current,
-          new Set(hiddenPostCountsRef.current.keys()),
-        );
-        optimisticPostIdsRef.current = displayedState.optimisticPostIds;
-        setPostIds(displayedState.displayedPostIds);
+        if (cursor !== streamTail || isInitialLoad) setStreamTail(cursor);
```
infin1t3: [P3] Write the corrected cursor even when it equals the captured offset

`streamTail` here is the value captured when this callback started, and committed removals can change the live state while it awaits a page. For example, start a collection request at offset 50, then remove and commit the previous 50 members so the live offset becomes 0. A response with nextCursor 100 is corrected to 50. Because that equals the captured value, this condition skips the write and the next request starts at 0 instead of 50, fetching already delivered posts again.

The hook reproduction fails at this head and passes with the base hook. Call `setStreamTail(cursor)` unconditionally after the stale-response check; React already skips an unchanged primitive value. The observed impact is extra fetching and delay, rather than demonstrated permanent data loss.

infin1t3 (changes requested):

infin1t3:
Reviewed `39251c8` against `dev` at `0fee1fe`. This review includes four behavioral findings, a documentation correction and two optional cleanup notes. I would address the behavioral findings before merging; the cleanup notes are not blockers.

1. **[P2] Keep cold reply streams in the order expected by the renderer.**
   [posts.ts:453–459](https://github.com/pubky/pubky-app/blob/39251c85ca576fb8096cc949cc13c4176ecc3738/src/core/services/local/stream/posts/posts.ts#L453-L459)

   Hydration can create a reply stream before its ascending page is persisted. The new duplicate-only return skips the sort that previously normalized that row. `useReplyStream` then reverses the hydration response's order, which can display replies newest first. The inline comment includes the reproduction and fix direction.

2. **[P2] Count visible mosaic progress when applying the auto-load budget.**
   [VisualTimelinePosts.tsx:423](https://github.com/pubky/pubky-app/blob/39251c85ca576fb8096cc949cc13c4176ecc3738/src/components/organisms/Timeline/Feed/TimelineFeed/VisualTimelinePosts.tsx#L423)

   In Home/Search Visual with content set to All, text-only pages increase `postIds.length` without adding rows. Each increase resets the budget, so the observer can keep loading without showing Load more. This happens after rows already exist and is separate from the initial-backfill issue tracked in #2538.

3. **[P2] Preserve unread ordering when head polls finish out of order.**
   [posts.ts:210–212](https://github.com/pubky/pubky-app/blob/39251c85ca576fb8096cc949cc13c4176ecc3738/src/core/services/local/stream/posts/posts.ts#L210-L212)

   Two tabs can complete overlapping head polls in reverse order. An older response then prepends an older unique post to the unread row. The new merge retains that position and puts the older post above newer posts on refresh. This is separate from the own-post ordering tradeoff documented in #2551.

4. **[P3] Commit the corrected pagination offset after in-flight removals.**
   [useStreamPagination.ts:230](https://github.com/pubky/pubky-app/blob/39251c85ca576fb8096cc949cc13c4176ecc3738/src/hooks/useStreamPagination/useStreamPagination.ts#L230)

   Committed removals can change the live offset while a request is pending. If the corrected response cursor equals the callback's captured offset, this condition skips the state update and the next request starts too early. The demonstrated impact is redundant fetching and delay; the reproduction does not establish permanent omissions.

The remaining finding is a documentation correction:

5. **[P3] Update the controller's cached-cursor contract.**
   [posts.ts:140–149](https://github.com/pubky/pubky-app/blob/39251c85ca576fb8096cc949cc13c4176ecc3738/src/core/controllers/stream/posts/posts.ts#L140-L149)

   `getCachedLastPostTimestamp` still says it reads the oldest post's `indexed_at` and returns zero when its details are missing. It now prefers the persisted Nexus `tailCursor`, with a timestamp seed for a legacy row that has no cursor. Update the description and return documentation to match that behavior. This comment is in the review body because the JSDoc sits outside the changed diff hunks.

The following two notes are optional cleanup:

6. **Consider sharing the pagination tail wiring.**
   [Posts.tsx:51](https://github.com/pubky/pubky-app/blob/39251c85ca576fb8096cc949cc13c4176ecc3738/src/components/organisms/Timeline/Posts/Posts.tsx#L51)

   Posts, TimelineGridPosts, VisualTimelinePosts and RepliesWithParent each wire up `useInfiniteScroll`, Load more and the sentinel. A small shared hook for the common defaults, plus a shared tail component, could make future changes easier to apply consistently. Keep each consumer's progress measure, loading state and hasMore condition explicit; sharing the wiring would not by itself resolve finding 2. This can be a separate refactor.

7. **Use the existing spacing token for the replies sentinel.**
   [RepliesWithParent.tsx:94](https://github.com/pubky/pubky-app/blob/39251c85ca576fb8096cc949cc13c4176ecc3738/src/components/organisms/ProfileReplies/RepliesWithParent.tsx#L94)

   The sentinel uses `h-[20px]`, while the other timeline sentinels use `h-5`. Switching to `h-5` would make these consistent with the design-token rule. The arbitrary class predates this PR, which only wraps the node, so this is optional cleanup with no intended behavior change.

Lint and typecheck passed. The full test run reported 13,633 passing tests, one expected failure and two skipped tests. Focused reproductions of findings 1, 3 and 4 fail at this head and pass when the relevant implementation is replaced with its base version. Finding 2 is supported by tracing the Visual renderer, tile filtering and budget reset logic; I have not run a focused executable reproduction of that case. Build, VRT and browser checks were not run.

src/core/services/local/stream/posts/posts.ts
secondl1ght: Valid, fixed in 151b03e. `persistNewStreamChunk` now normalizes cursor-less, non-bookmark rows newest-first in both places the early return skipped: a chunk that creates the row is sorted by timestamp before it is stored (hydration's by-IDs order no longer survives), and a cursor-less page that repeats ids a cursor-less row already holds re-sorts the row (a no-op write is skipped when the order is already right). Score-backed rows keep the "never re-sort" rule: the check requires `postStream.tailCursor === undefined`, so a cursor-less dupes page on a timeline row still leaves it untouched. Test: "stores a cursor-less reply row newest-first from creation and keeps it so on a repeated page" in `posts.test.ts`.

---

src/components/organisms/Timeline/Feed/TimelineFeed/VisualTimelinePosts.tsx
secondl1ght: Valid, fixed in 151b03e. The Visual observer now measures mosaic progress: `itemCount` is the number of rendered tiles (`rows` cells summed), so a page of text-only posts that grows `postIds` without adding a tile no longer resets the budget. Pending tile, file and post-detail resolution counts as loading (`isLoading` includes `hasPendingTiles || hasPendingFiles || hasPendingPostDetails`), so an unsettled page is not judged unproductive before its tiles can appear. Test added with the real `useInfiniteScroll` and a stubbed `IntersectionObserver`: ids keep growing while rows stay flat, and the Load more button appears after `TIMELINE_MAX_UNPRODUCTIVE_AUTO_LOADS` loads ("hands over to Load more when ids keep growing but the mosaic does not"). The initial backfill effect stays tracked in #2538.

---

src/core/services/local/stream/posts/posts.ts
secondl1ght: Valid, fixed in 151b03e. `persistUnreadNewStreamChunk` no longer prepends a page's unique ids. Each new id is inserted right after the nearest preceding id the row already holds, walking the page in Nexus order; an id no cached id precedes goes on top. With the newer poll `[11..2]` already stored, the late older poll `[10..1]` now lands `1` at the bottom (`[11..2, 1]`), and a later overlapping poll `[13, 12, 11, 10]` puts `13, 12` on top in page order. A page sharing no id with the row is still taken as newer, which is what a poll above the head returns. No `indexed_at` sort is involved, so the edited-post handling is unchanged. Tests: "places a late, older poll response below the newer ids it overlaps" and "puts the newer ids of a later overlapping poll on top, in page order" in `posts.test.ts`; the old "filters duplicates" expectation that encoded the prepend rule was rewritten.

---

src/hooks/useStreamPagination/useStreamPagination.ts
secondl1ght: Valid, fixed in 151b03e. `setStreamTail(cursor)` is now written unconditionally after the stale-response check, with a comment explaining why equality with the captured offset is not evidence the live state still holds it. Test: "writes the corrected skip offset even when it equals the offset captured at request time" in `useStreamPagination.test.tsx` (collection stream, two removals committed mid-flight, response cursor 4 corrected to 2, next request starts at 2).

---

secondl1ght:
Review round 2 addressed in 151b03e. The four inline findings are answered on their threads; the remaining review-body items:

- **5. `getCachedLastPostTimestamp` JSDoc** — rewritten: it returns the row's persisted `tailCursor` (the `last_post_score` of the deepest page), a one-time seed from the tail entry's timestamp for rows without one, or `NOT_FOUND_CACHED_STREAM`.
- **6. Shared budget wiring across the four renderers** — not in this PR. The four `useInfiniteScroll` call sites differ in what they count and what they treat as loading (Visual now counts tiles and pending resolution, see the thread), so a shared helper is a small refactor of its own rather than a fix; I'd rather keep this PR to the behavioural change and do it separately if wanted.
- **7. `h-[20px]` on the replies sentinel** — replaced with `h-5`, snapshots updated.

Verification on the merged base plus these fixes: `lint`, `typecheck`, `format:check`, full unit suite (791 files, 13785 passed, 1 expected fail for #2535), and a 300-seed run of the pagination simulation.

---

src/core/services/local/stream/posts/posts.ts
secondl1ght: Adjusted in 12ce59d after a further review pass. The normalization now lives only where it is needed: a cursor-less chunk that creates a row, or adds ids to a row without a Nexus cursor, is sorted newest-first; a chunk that adds nothing never rewrites a row (the earlier no-add re-sort read and sorted the whole reply row on every hydration pass, and hydration re-persists visible replies on every TTL tick). Two related tightenings: an ascending reply page is handed over reversed, since `sortPostIdsByTimestamp` interpolates missing details assuming descending input (a failed hydration would otherwise store the page oldest-first), and a row that carries a Nexus cursor is never re-sorted whatever chunk reaches it, so the add path and the no-add path follow one rule. Tests updated accordingly, plus a new one in `post.test.ts` for the reversed hand-over.

---

src/components/organisms/Timeline/Feed/TimelineFeed/VisualTimelinePosts.tsx
secondl1ght: Adjusted in 12ce59d after a further review pass. Gating `isLoading` on the pending flags turned out to be a freeze: `persistFiles` writes only what Nexus returns, so an attachment Nexus no longer serves keeps `hasPendingFiles` set for good, and on feeds that drop unavailable posts `pendingDetailPostCount` never settles either. With the observer effect skipped while `isLoading` is true, such a feed would get neither auto-loading nor the manual Load more. The budget now counts every tile the pipeline tracks (`tiles.length`: packed, buffered in the row composer's tail, or still probing) instead of packed cells only, so an unsettled page and a lone trailing tile both count as progress while a text-only page still does not, and `isLoading` is back to `loadingMore || isInitialLoading`. The "ids grow, mosaic does not" test still passes with the real hook; the pending-tiles test now asserts `itemCount` grows with a probing tile and the observer is not gated.

---

secondl1ght:
Third pass, 12ce59d: a review of the round-2 commit itself found that two of its fixes needed tightening.

- **Visual feed** — gating the observer on pending tile/file/detail resolution could freeze the feed for good (a file Nexus no longer returns never settles). The budget now counts every tile the pipeline tracks, and the pending flags no longer feed `isLoading`. Details on the thread.
- **Cursor-less rows** — one rule now: a chunk without a Nexus position sorts a row without one at creation and when it adds ids; a row with a Nexus cursor is never re-sorted; a chunk that adds nothing never rewrites a row. Ascending reply pages are handed over reversed so a failed hydration cannot store them oldest-first. Details on the thread.

Verified: lint, typecheck, format, full unit suite (791 files, 13786 passed, 1 expected fail for #2535), 300-seed simulation sweep.

---

src/core/services/local/stream/posts/posts.ts
secondl1ght: One more adjustment in 9aad480: the normalization on a repeated page is back for cursor-less rows. Dev creates a row unsorted (`if (!postStream) upsert(streamId, stream)`) and only re-sorts it on a later write, so a reply row hydrated on today's build and never opened is in hydration order on disk; with 12ce59d's "a chunk that adds nothing never rewrites a row" rule, opening that thread after the upgrade would have rendered it reversed until a new reply arrived (no `DB_VERSION` bump clears those rows). So: a cursor-less chunk that adds nothing re-sorts a cursor-less row (no write when the order already matches), a row with a Nexus cursor is still never re-sorted, and the ascending page is still handed over reversed. The cost is a details read per repeated chunk on reply rows, which is what dev already does on every write. Test restored: "…and normalizes one an earlier build left unsorted".

---

secondl1ght:
Fourth pass, 9aad480: a review of the round-3 commit found one regression against `dev`, now fixed, plus two findings that did not lead to a change.

- **Fixed:** a reply row hydrated on an earlier build in non-chronological order would have stayed that way after upgrade, because round 3 stopped normalizing a cursor-less row when a page repeats its ids. That normalization is restored for cursor-less rows only (details on the thread). Score-backed rows are still never re-sorted.
- **Declined:** gating the Visual observer on `hasPendingTiles` / `hasPendingPostDetails`. Dev never held loads behind a probing tile, and doing so would stall the feed behind a slow image host; the current wiring matches dev's chaining in that scenario. The code comment that overstated which flags can stay set forever is corrected.
- **Pre-existing, filed as #2568:** a Visual feed whose whole first page has attachments Nexus no longer returns, with no further pages, stays on its skeleton. The gate is unchanged from dev and out of scope here.

Also: the simulation seed count is now `SIM_SEEDS` (`SIM_SEEDS=300 npm test -- src/hooks/useStreamPagination/useStreamPagination.simulation.test.tsx`), so sweeps no longer need the file patched.

Verified: lint, typecheck, format, full unit suite (791 files, 13786 passed, 1 expected fail for #2535), 300-seed sweep. Guest staging runs on the branch for the surfaces changed in rounds 2–4: Visual + All (4 minutes, 6 pages, tiles 2 → 6, no spurious Load more), Visual + Images (4 pages, tiles 12 → 42), and a reply thread (cached row newest-first, all details present, all cards rendered).

---

infin1t3:
Follow-up on `e7d2d40` (merge base `6803c5f`), after reading the review replies and checking the latest changes. Two issues remain. I would fix the skipped-post case before merging.

1. **[P2] Recover safely when another tab clears the stream cache.**

   [post.ts:581–582](https://github.com/pubky/pubky-app/blob/e7d2d40c22a0cdcf04d7ec04bfa0da01d41e66be/src/core/application/stream/posts/post.ts#L581-L582)

   A mounted feed can consume the first 10 entries of a 30-post cached stream and receive the Nexus cursor below entry 30. That works while the remaining cached entries exist. If the cache later expires and another tab calls `prepareStreamForInitialLoad`, it deletes the shared row before fetching its replacement. If the original tab loads more during that gap, it resumes below entry 30 and skips the 20 entries it has not consumed.

   The anchor recovery only runs when the cached row exists. The missing-row reset at lines 606–610 only handles an initial load without `lastPostId`, so it does not protect an already mounted reader. The reproduction returns entries 31–40 where entries 11–20 were expected.

   Detect this loss of cached membership and recover or restart the walk without treating unconsumed posts as already read. Keep resume cursors derived from Nexus; switching back to `indexed_at` would reintroduce the problem this PR addresses.

2. **[P3] Preserve unread ordering across a merge-and-clear or a disjoint late page.**

   [posts.ts:210–212](https://github.com/pubky/pubky-app/blob/e7d2d40c22a0cdcf04d7ec04bfa0da01d41e66be/src/core/services/local/stream/posts/posts.ts#L210-L212), [persistUnreadNewStreamChunk](https://github.com/pubky/pubky-app/blob/e7d2d40c22a0cdcf04d7ec04bfa0da01d41e66be/src/core/services/local/stream/posts/posts.ts#L548-L574)

   The overlap-aware insertion fixes the original reproduction while both pages remain in unread. Two reachable sequences still put older posts above newer ones. In these examples, larger IDs represent newer posts, and timestamps agree with Nexus order:

   - An older poll for `[10, ..., 1]` is pending. A newer poll for `[11, ..., 2]` completes, then an initial load or refresh merges and clears unread. The older response arrives and recreates unread. The next merge produces `[10, ..., 1, 11, 0]` instead of `[11, ..., 0]`.
   - A newer page `[20, ..., 11]` reaches unread before an older page `[10, ..., 1]`. Because they have no overlapping IDs, the older page is treated as newer and inserted above it.

   Separate tabs have independent requests but share IndexedDB. Ordering therefore needs to survive unread being consumed and account for the main stream as well as the current unread row. Preserve unique IDs from late responses without promoting their older posts above newer ones. A later poll and merge can repair the order; these reproductions do not demonstrate data loss, which is why I would rank the remaining issue P3.

The first issue was already present in `39251c8` and I missed it in the initial review. The second is an incompletely fixed case of the original unread finding. Neither was introduced by the follow-up fixes.

The reply-ordering, Visual budget, pagination-offset, controller JSDoc and spacing fixes checked out in the previous pass. Deferring the shared renderer wiring to a separate refactor is reasonable.

Verification: all three cases above were rerun on `e7d2d40` through the real application and local persistence with mocked Nexus responses. They fail on this head and pass when the relevant implementations are replaced with their `dev` base versions. For the unread cases, replacing only the local stream service is sufficient. The latest merge changes tag UI code and leaves the pagination implementation unchanged.

Lint, typecheck and production build passed at `25ab8e8`; the full suite there reported 794 files, 13,806 passing tests, one expected failure and two skipped tests. Those broader checks were not repeated after the latest tag UI merge. No browser, VRT or Cypress checks were run in this review.

secondl1ght:
Follow-up on the two remaining findings. Neither is fixed in this PR; both are tracked in #2570 with reproductions.

- **1. Vanished row mid-walk (P2).** Confirmed, and reproduced in the simulation (`otherTabClearsRow` event plus the scenario "survives another tab clearing the row mid-walk…", kept as `it.fails`). A resume seeded from the smallest local timestamp among the anchor and the deepest delivered posts passed that scenario, a 300-seed sweep and the full suite, but a review of the delta rejected it: it falls open to a post's creation date on an emptied Bookmarks row (un-bookmark every loaded bookmark, then scroll: the rest is skipped), it aligns the seam page against a row this walk did not build if the rebuild lands mid-request, its premise is false for locally authored posts before their first refresh, and it persists the recovery page as a mid-stream row, which `dev` also does. The right-depth fix is in `prepareStreamForInitialLoad`'s delete-then-fetch (replace the row atomically, and let a walk that finds no row serve uncached), which is a cross-tab change of its own rather than a resume heuristic. Details and the rejected approach are written up in #2570 so the next attempt starts from there.
- **2. Unread ordering across a merge-and-clear or a disjoint late page (P3).** Confirmed as an incompletely fixed case of the original unread finding; self-healing on the next poll, no data loss. Tracked in #2570 with both sequences.

The code on this head is otherwise unchanged from `e7d2d40`, where CI, the full suite, the 300-seed sweep and the guest staging runs were all green; this push adds only the simulation scenario.

---

infin1t3 (approved):
Approving this urgent hotfix at `daf50f1`. My remaining comments about skipped posts after cross-tab cache invalidation and unread ordering are non-blocking follow-up suggestions, tracked in #2570. Both remain known issues in this PR; I'm accepting that trade-off so the hotfix can ship.

PR #2554, opened 2026-09-16
fix(posts): prepend own posts on the My network reach again

talosmachina:
## Why

Requested by @infin1t3 in Slack, following a report that a user's own post no longer appears in the home feed once the default "My network" Reach applies.

After the Reach filters were aligned to **Me + the reach** (paired pubky-nexus draft PR), `wot` streams carry the viewer's own posts again. #2333 gated the optimistic prepend on every WoT-sourced stream (`wot` and `wot_domain`) because Nexus excluded the viewer server-side and the local create path never writes into those streams, so the optimistic entry was guaranteed to flash and vanish (#2308).

That premise no longer holds for `wot`: the reach is the observer plus their network, and the local stream cache merges refreshed chunks (`persistNewStreamChunk`) rather than replacing them, so the prepended post stays consistent.

## What changed

- `src/core/models/stream/post/postStream.types.ts`: `isViewerExcludedWotStream` becomes `isViewerExcludedWotDomainStream`, returning true only for `wot_domain` ('Tagged as'). Own-post membership there still depends on whether the observer's trust set endorsed the viewer, which the client cannot know, so the #2308 gate stays for that source.
- `src/hooks/usePostInput/usePostInput.ts`: imports and calls the narrowed helper; comment updated to say why `wot` is no longer gated.
- Tests: `timeline:wot:all` moves from the gated cases to the prepend cases (post and repost paths), the collapse-when-skipped test now uses a `wot_domain` stream id, and the helper's unit tests assert `wot` is not excluded while `wot_domain` is.

## Test evidence

```
npx vitest run --project unit src/core/models/stream/post/postStream.types.test.ts src/hooks/usePostInput/usePostInput.test.ts
Test Files  2 passed (2)
Tests  196 passed (196)

npx tsc --noEmit
(clean)
```

## Depends on

The pubky-nexus draft PR that makes `source=wot` include the observer's own posts. Merging this alone would reintroduce the #2308 flash-and-vanish on My network.

## Manual check left

Apply the "My network" Reach in the sidebar and create a post: it should appear at the top immediately and still be there after a filter toggle and a stream refresh.


Orlandohub:
Closing this since we need further due diligence to be done.

More context here:
https://synonymworkspace.slack.com/archives/C031F2DC1JM/p1788855184949049

PR #2556, opened 2026-09-16
fix(service-worker): register from app code so failures are handled

talosmachina:
## What

Browser-side service-worker registration failures were reaching Sentry as unhandled errors. `@serwist/next` injects an entry with automatic registration turned on, and that entry calls `window.serwist.register()` without a `.catch()`, so a rejected `navigator.serviceWorker.register()` escapes as an unhandled rejection and the SDK's global handler reports it as a product error. Automatic registration is now disabled (`register: false` in `next.config.ts`) and registration is done by a new `ServiceWorkerRegistrationProvider`, mounted once by the root layout, which handles the rejection and logs it quietly instead. Successful registration is unchanged: same script, same scope, and `register()` still waits for the window load event. A failed or absent registration no longer produces an error. The library's own `TypeError` when the browser resolves the call with no registration object (`reading 'waiting'`) rejects the same promise and is handled by the same `.catch()`, and the provider never dereferences the registration result.

## Why

- family: `SW register` (short ids 21 5W 4W 8K D3) — `Error: Rejected`, `TypeError: Failed to register a ServiceWorker ... An unknown error occurred when fetching the script`, `TypeError: Script .../sw.js load failed` (the last one on Mobile Safari UI/WKWebView, iOS 18.7)
- evidence: 827 events in 30 days, all production; 16 in the last 7 days against 8 the week before
- out of scope: no update prompt and no reload-on-update behaviour — what the app should do when a new service worker is waiting is a product decision, not part of this fix

## Verification

- `npm test -- src/providers/ServiceWorkerRegistrationProvider/ServiceWorkerRegistrationProvider.test.tsx`: 6 passed (1 file)
- `npm run typecheck`: no errors in the changed files; the 5 errors it reports are pre-existing and in untouched files (`linkify-it` typings, pubky SDK `Signer` methods)
- `npx eslint` on the two changed source files: no problems
- `npx prettier --check` on the two changed source files: unchanged

## Sentry

- [PUBKY-APP-21](https://synonym-52.sentry.io/issues/119882300/): 769 events since 2026-05-14, `Error: Rejected`
- [PUBKY-APP-5W](https://synonym-52.sentry.io/issues/124849723/): 42 events since 2026-06-03, `TypeError: Failed to register a ServiceWorker for scope ('https://pubky.app/') with script ('https://pubky.app/sw.js'): An unknown error occurred when fetching the script.`
- [PUBKY-APP-4W](https://synonym-52.sentry.io/issues/123997930/): 6 events since 2026-06-01, `TypeError: Script https://pubky.app/sw.js load failed`
- [PUBKY-APP-8K](https://synonym-52.sentry.io/issues/132247113/): 10 events since 2026-07-05, `TypeError: Cannot read properties of undefined (reading 'waiting')`


infin1t3 (approved):
Reviewed at ec4700a8. No actionable findings.

The provider catches the registration failures reported in Sentry, while Serwist still initializes window.serwist before React mounts. Successful registration preserves the existing script URL, scope and lifecycle listeners.

Verified with all 6 provider tests and 6 additional probes using the actual Serwist implementation, covering registration before/after load and the reported failure paths. Changed TSX files pass ESLint.

PR #2557, opened 2026-09-16
fix(og): strip crawler escape sequences from route ids

talosmachina:
## What

Crawler-appended literal escape sequences are stripped from the id segments of the crawl-facing routes (`/profile/[pubky]`, `/post/[userId]/[postId]`, `/collections/[userId]/[postId]` and their `opengraph-image` / `twitter-image` segments) before the id is validated, so a request like `/post/<pubky>/0035NGP297YG0%5Cn%5Cn/opengraph-image` resolves the real post instead of being handed to Nexus unchanged.

A link unfurler that resolved `\n` in its own copy of a URL asks for the two-character sequence backslash + `n` (`%5Cn%5Cn`). It survives `trim()` and still starts and ends alphanumeric, so it passed the existing shape checks and every such request got a 400 from Nexus, turning a crawler artefact into an error group.

## Why

- family: `PUBKY-APP-A0` (79 events), `PUBKY-APP-BQ` (54), `PUBKY-APP-9Z` (38) — all `AppError: Bad Request` raised by `Err.client` through `httpStatusCodeToError` -> `httpResponseToError` -> `fetchWithValidation`, from a malformed id in an `opengraph-image` URL (`%24`, `%5Cn%5Cn`)
- evidence: the ids are visibly malformed in the URLs, so the 400 is the server correctly rejecting an id the app should have cleaned
- out of scope: `PUBKY-APP-1E` (`/search`, 460 events). Its Nexus calls take user-typed search prefixes and content-search params rather than a route id segment, and it predates this family, so it is a different cause and needs its own fix.

## Verification

- `npm test -- src/libs/og/routeIds.test.ts src/libs/og/ogData.test.ts 'src/app/post/[userId]/[postId]/page.test.tsx' 'src/app/profile/[pubky]/page.test.tsx' 'src/app/collections/[userId]/[postId]/page.test.tsx'`: 5 files, 91 tests passed
- `npm run typecheck`: no new errors; the 7 reported errors are pre-existing on `dev` in untouched files (`linkify-it` typings, `Signer` / `Pubky` cookie methods)
- `npx eslint src/libs/og/routeIds.ts src/libs/og/routeIds.test.ts`: clean

## Sentry

- [PUBKY-APP-A0](https://synonym-52.sentry.io/issues/134959489/): 79 events since 2026-07-17, `AppError: Bad Request`
- [PUBKY-APP-BQ](https://synonym-52.sentry.io/issues/140331298/): 54 events since 2026-08-12, `AppError: Bad Request`
- [PUBKY-APP-9Z](https://synonym-52.sentry.io/issues/134959488/): 38 events since 2026-07-17, `AppError: Bad Request`
- [PUBKY-APP-1E](https://synonym-52.sentry.io/issues/118511735/): 460 events since 2026-05-08, `AppError: Bad Request`


infin1t3 (approved):
Reviewed at 0b44eaab. No actionable findings.

The normalization handles the trailing %5Cn%5Cn input recorded in PUBKY-APP-9Z and preserves the fallback path for invalid identifiers. Traced the profile, post and collection consumers, including Open Graph and Twitter image routes.

Verified with 108 passing tests across seven relevant suites. All four changed files pass ESLint.

PR #2558, opened 2026-09-16
fix(rate-limit): coalesce duplicate nexus revalidations and report a 429 once

talosmachina:
> This PR carries two commits, both aimed at the 429 family: `fix(nexus): never reuse an in-flight forced revalidation for a later caller` and `fix(rate-limit): report a 429 once per service and operation window`. `Retry-After` was already implemented on `dev` (#2476), so only the remaining symptom is addressed here. The branch cannot be split without a force push.

## What

A forced caller no longer reuses a forced revalidation that is already in flight. A revalidation that has not opened its network request yet is shared: callers arriving while it still waits for a pending ordinary request join it, so a TTL tick and a notification hydration on the same batch behind one cache-miss fetch cost a single revalidation between them. Once the request has started, callers arriving after that point wait for it to settle and share a single follow-up with each other, so a burst of subscribers behind one running request costs one extra request instead of one per caller. The data a forced caller receives always comes from a request that started after its own call. The non-forced path is unchanged (TanStack already merges those through its cache slot).

`stream/*by_ids` is where the 429 storm bites. Reusing the running request there is what a previous revision of this branch did, and it was wrong: a notification refresh that landed on a TTL tick took the TTL request's response, which had started before the event. Its empty tag list came back with the older request-start evidence and was written to the tag cache as hydrated, so notification handling skipped its follow-up fetch and the new tag stayed missing until the next TTL. The data a forced caller receives now always comes from a request started after its own call; freshness wins over the dedup.

## Why

- family: `429 rate limit` (`AppError: Too Many Requests`), short ids B3 9X AD CT CV CW — 9618 events in 30d, 1063 in the last 7 days against 3453 the week before
- evidence: B3 alone had 574 events on 2026-09-15; routes `/profile`, `/home`, `/hot`; frames `src/core/services/nexus/nexus.utils.ts` fetchNexus → `httpStatusCodeToError` → `Err.rateLimit`
- out of scope: the other two asks from the same review are already on `dev` since #2476 — `parseRetryAfterHeader` accepts delta-seconds and HTTP-date, `retryDelay` honours `context.retryAfter` clamped to the configured max, and `postsByIds`/`usersByIds` sort their id sets so identical batches share one query key. This PR closes only the remaining hole and adds the end-to-end tests that pin all of it. No product limit under `src/config` was touched.

## Verification

- `npm test -- src/core/services/nexus/nexus.utils.test.ts src/core/services/nexus/stream/posts/postStream.test.ts src/core/services/nexus/stream/users/userStream.test.ts src/libs/query-client/query-client.factory.test.ts`: 4 files, 230 tests passed. The new cases in `nexus.utils.test.ts` are: two concurrent identical non-forced `posts/by_ids` calls → 1 fetch; the same for `users/by_ids`; a forced call arriving behind a running forced revalidation does not reuse it and gets a request started after its own call (the notification-on-a-TTL-tick ordering, asserted through `getNexusResponseStartedAt`); forced callers arriving before a revalidation starts share it, including two forced waiters behind a pending ordinary request (2 fetches, one shared response per caller); forced callers waiting behind one running revalidation share a single follow-up; a 429 with `Retry-After: 5` does not retry before 5s (fake timers); a 429 with an HTTP-date `Retry-After` does not retry before that instant; `Retry-After: 3600` is clamped to the 30s max; a 429 without the header keeps the 2s floor; the rate-limit error still surfaces after the single retry; and the non-429 paths are unchanged (400 not retried, 503 retried at 1s/2s/4s).
- `npm test -- src/core/application/notification/notification.test.ts src/core/application/tag/tag-cache.test.ts src/core/application/ttl/ttl.test.ts src/core/services/nexus/stream src/core/services/local/tag`: 9 files, 398 tests passed.
- `npm run typecheck`: 5 errors, all pre-existing and in files this PR does not touch (`PostLinkEmbeds.tsx`, `post.kind.ts`, `homeserver.ts` — `linkify-it` types and `Signer.signinCookie`). None in the changed files.
- `npx eslint src/core/services/nexus/nexus.utils.ts src/core/services/nexus/nexus.utils.test.ts`: clean. `prettier --check` with the version pinned in the lockfile (3.9.6): clean.

## Sentry

- [PUBKY-APP-B3](https://synonym-52.sentry.io/issues/139394660/): 6899 events since 2026-08-07, `AppError: Too Many Requests`
- [PUBKY-APP-9X](https://synonym-52.sentry.io/issues/134804971/): 1802 events since 2026-07-16, `AppError: Too Many Requests`

Requested by @SHAcollision via the talos CLI.


infin1t3 (changes requested):
P2: Preserve freshness when a notification joins an earlier forced refresh.

@talosmachina please fix the early join in [nexus.utils.ts:175-177](https://github.com/pubky/pubky-app/blob/04bffa470a21d772dbc9415dd065cdb3cd888b80/src/core/services/nexus/nexus.utils.ts#L175-L177).

A TTL refresh of the viewer's profile can already be fetching when a new profile-tag notification arrives. When the tagger is cached, notification hydration requests exactly the same user batch and now joins that older forced request. The notification invalidates the tag cache and captures its new revision, so the revision guard accepts the old response. An old empty tag list is marked fresh and exhausted; notification handling then skips its follow-up tag fetch. The notification appears while the new tag remains missing, potentially until the next refresh after the default ten-minute user TTL. The older validatedAt timestamp does not prevent this acceptance or the skipped follow-up.

At the comparison base, the later forced call waits for the pending request and starts a fresh one. Please preserve that guarantee, for example by coalescing a queued follow-up revalidation when the running request predates the invalidation, and add regression coverage for this ordering.

Reviewed at 04bffa47 against dev at 41435c96. All 280 targeted repository tests pass. Three additional regression probes reproduced the stale result through the real notification, TTL, Nexus and Dexie implementations and verified the head/base difference. Changed-file ESLint passes.

talosmachina:
@infin1t3 fixed in 04442f62.

`queryNexus` no longer reuses a forced revalidation that is already in flight. The later caller waits for it to settle, then shares one follow-up revalidation with every other caller waiting behind the same request, so the data it receives always comes from a request started after its own call. That restores what the comparison base guaranteed, and it is what notification hydration needs when it lands on a TTL refresh of the viewer's profile: the joined response predated the event, arrived with the older request-start evidence, and was written to the tag cache as hydrated, which is what skipped the follow-up tag fetch.

Regression coverage for that ordering in `nexus.utils.test.ts`: a TTL-started forced revalidation is still fetching when a notification-triggered one arrives; the second caller's result is asserted to be the fresh response, with `getNexusResponseStartedAt` at its own call time, and the older response is asserted not to be reused. A second case pins that forced callers waiting behind one running revalidation share a single follow-up rather than opening one request each. The previous expectation that two concurrent forced calls cost one request is gone, and the PR description no longer claims it: that collapse was the defect, so the pair now costs one follow-up after the running request settles.

Verified: `npm test` on `nexus.utils.test.ts` + both stream suites + the query-client factory, 4 files / 229 tests passed; notification, tag-cache, ttl and stream suites, 9 files / 398 tests passed; `npx eslint` on both changed files clean; `prettier --check` with the lockfile version 3.9.6 clean; typecheck shows the same 5 pre-existing errors in untouched files and none in the changed ones.

Requested by @SHAcollision via the talos CLI.

infin1t3 (changes requested):
Re-reviewed 04442f62. The original stale-response finding is resolved: the real notification/TTL/Dexie reproduction now receives the fresh batch and displays the new tag.

P2: Keep a revalidation shareable while it is still waiting to start.

@talosmachina please address [nexus.utils.ts:187-191](https://github.com/pubky/pubky-app/blob/04442f62d060544d7a062f60f1ebdaf666417f06/src/core/services/nexus/nexus.utils.ts#L187-L191). startRun() adds its promise to inFlightForcedQueries while it is still awaiting an existing non-forced request. A second forced caller then queues another revalidation behind it, although the first forced network request has not started and could satisfy both callers.

For example, a profile cache-miss fetch can still be pending when TTL and notification hydration force the same user batch. With both forced callers arriving before the ordinary request settles, this head makes three requests: the ordinary fetch followed by two sequential forced fetches. The dev comparison base makes two, with both forced callers sharing the second. The third request adds load to the rate-limited endpoint and an unnecessary round trip for the later caller; it provides no extra freshness guarantee because request two started after both callers arrived.

Keep the scheduled revalidation shareable until its network request actually begins; callers arriving after that point can queue the next one. Please add a case with a pending non-forced request and two forced waiters.

Verification: 282 targeted repository tests pass; changed-file ESLint and Prettier pass. The original notification integration probe passes with the fresh tag persisted. Separate probes using the actual head and dev-base query implementations confirm three requests at this head versus two at the base for the pending-ordinary-request ordering.

talosmachina:
Addressed in `6a1c14c8`.

@infin1t3 you are right, and it reproduced exactly: with a cache-miss fetch still running and two forced callers on the same batch, the previous head opened three requests where the base opens two. I added the case you asked for first, and it fails on `04442f62` with `expected "vi.fn()" to be called 2 times, but got 3 times`.

The shared slot now covers a revalidation until its network request begins. `startRun` registers the run in `queuedForcedQueries` while it waits for the pending ordinary request, and moves it to `inFlightForcedQueries` - dropping the shared slot - only as the request starts. A caller that arrives before that point joins the run, whose request still begins after its own call, so the freshness guarantee is unchanged; a caller that arrives after it has started queues the next revalidation exactly as before.

One consequence worth naming: two forced callers with no other request pending now share a single request instead of opening one each, because both arrive before it starts. The case that pinned the old two-request count was retargeted to the already-fetching ordering, which still opens the follow-up.

Verification: `npm test -- src/core/services/nexus/nexus.utils.test.ts` 33 passed; the wider set from the PR body, 230 passed and 398 passed. `npm run typecheck` reports the same 5 pre-existing errors, none in the changed files. ESLint and prettier (3.9.6) clean on both changed files.

infin1t3 (approved):
Re-reviewed 6a1c14c8, including the full delta from 04442f62. Both reported findings are resolved. Forced callers waiting behind an ordinary fetch now share one follow-up request, and the original notification/TTL/Dexie reproduction still persists the fresh tag after an older forced request. No confirmed actionable findings remain.

Verification: 368 targeted repository tests across 9 suites pass, along with 7 focused regression and boundary checks covering the two findings, concurrent application calls, and cleanup after success or failure. Changed-file ESLint and Prettier pass. Build, Code Quality, Tests and preview deployment are green.

One unconfirmed edge case remains: directly calling the helper in a forced → ordinary → forced sequence within the same synchronous turn can return the ordinary request's older response to the final caller. I could not establish this ordering through current application callers: they await independent Dexie revision reads, and the real application concurrency probe starts the first forced fetch before the other calls and correctly makes the follow-up. This is a follow-up concern rather than a confirmed application regression.

PR #2561, opened 2026-09-16
[Locks] feat(#2296): cache lock content in IndexedDB

### LINKED ISSUE #2296: Locks: cache lock descriptors and unlocked posts in IndexedDB

Taewa:
## What changed

- Cache lock descriptors and readable post metadata in IndexedDB.
- Load creator and reader lock content locally before falling back to the homeserver.
- Show cached unlocked items immediately while syncing the latest homeserver list.

Media bytes remain on the homeserver, and the lock cache is cleared on logout.

## Testing

- Locks-related unit tests
- TypeScript, ESLint, and Prettier checks

Closes #2296


talosmachina (commented):
**No findings.** Caches lock descriptors and readable post text in IndexedDB (new `locks` table, DB version 2 → 3) and makes the lock-file, replica and unlocked-list reads local-first, and adds announcement editing plus a scrolling mobile tab bar. Reviewed `6a0f5a1`.

<details><summary>What I checked, and 8 candidates I ruled out</summary>

**Read in full**: `core/application/locks/locks.ts`, `core/controllers/locks/locks.ts`, `core/services/local/locks/locks.ts`, `core/models/locks/locks.ts` + `locks.schema.ts`, `core/pipes/locks/locks.parser.ts`, `core/pipes/post/post.normalizer.ts`, `core/database/franky/franky.ts`, `hooks/useUnlockedContent`, `hooks/useUnlockedList`, `hooks/useUnlockedMedia`, `hooks/useLockFile`, `hooks/usePost`, `PostInput`, `DialogEditPost`, `libs/post/lockTeaser.ts`.

**Call sites traced**: `AuthController.cleanupLocalState` → `clearDatabase()` → `db.tables`, so the new table is covered by logout without an explicit entry; `useLockFile` → `LockedPostContent` / `PostInput`; `getOwnPost` / `getUnlockedPost` → `useUnlockedContent`; `GuardedContentParser.toCachedPost` → `cacheGuardedOriginal` → both `upsertPost` callers.

**CI**: red on every job, and none of it is from this diff. The unit shards and both build jobs die on `Failed to resolve import "@pubky/locks-sdk"`, and the base branch is the source: `git show origin/epic/locks-v1:src/core/services/locks/locks.utils.ts` imports it while `package.json` has no `@pubky/*` dependency at base or at this head. Check Code Quality additionally fails on two files this PR does not touch (`PostBody.tsx`, `usePaykitSetupFlow.utils.test.ts`). This PR is therefore unverified by CI either way.

**Ran locally** (`npm test --`): `post.normalizer.test.ts` + `lockTeaser.test.ts` 96 passed; `application/locks` + `controllers/locks` + `pipes/locks` + `services/local/locks` 155 passed; the locks hooks and components 101 passed with 13 failed only because this box has `linkify-it@5.0.0` installed against a lockfile pin of `6.1.0` (`TypeError: LinkifyIt is not a constructor`, from untouched `post.kind.ts`), not a signal about the PR.

**Ruled out**

- *Cached creator original leaking to a reader*: `getUnlockedPost` requires `unlockedAt !== undefined` and only the reader paths set it.
- *Cross-account reads from the table*: rows exist only for the signed-in account and are cleared by `clearDatabase()` on every account change, so no two accounts ever share the table.
- *A conflicting writer clobbering a cached descriptor*: `upsertDescriptor`/`upsertPost` both re-read inside one `rw` transaction and carry `existing` forward.
- *A corrupt cache blocking a lock from loading*: `getOrFetchLockFile` catches the local read and falls through to `fetchLockFile`, which still runs `isValidLockUrl` and stays the only network entry point.
- *DB version bump 2 → 3 losing data*: the mismatch path in `franky.ts` deletes and recreates the DB, and every cached row is re-derivable from the homeserver, so only cold reads are lost.
- *`hasError` removal hiding a failed `lock.json` fetch*: nothing in product code ever read it (`LockedPostContent` destructures only `lockFile` and `priceSats`), so the deleted field and its deleted test were dead.
- *`fetchReplicatedContent` writing Dexie from a `fetch*` method*: deliberate, and the write is caught so a cache failure cannot fail the read.
- *Missing toast for media dropped on the cached path*: `loadCachedContent` reports `attachments: null`, so the dropped-attachment toast cannot fire, but the content still renders and nothing else reads that field.

</details>

src/core/application/locks/locks.ts
```diff
@@ -535,6 +598,40 @@ export class LocksApplication {
       });
     }
 
-    return (await LocksService.readContentLock(lockUrl)) as LockFile;
+    // TODO:[Locks] locks#22 — replace this cast when the SDK exports its lock-file reader type.
+    const descriptor = (await LocksService.readContentLock(lockUrl)) as LockFile;
+    const lockId = LockContentParser.lockIdFromUrl(lockUrl);
```
tipogi, line 603: **Cache / identity:** bind `lock.json` to the **lock URL host**. What it could match is:

```ts
creatorFromUrl(lockUrl) === descriptor.creator
```

This file has to live on the homeserver the URL names. A mismatch is a different user's lock (or a cache row keyed only by filename).

Do not upsert into IndexedDB until that holds. Apply the same check in `getLockFile` before returning a cached descriptor — otherwise a cache hit skips this download path.

src/hooks/useUnlockedContent/useUnlockedContent.ts
```diff
@@ -59,7 +70,12 @@ export function useUnlockedContent({ lock, lockFile, authorId }: UseUnlockedCont
 
     let cancelled = false;
     setIsResolvingReplica(true);
-    LocksController.fetchReplicatedContent({ lockUrl: lock, readerPubky: currentUserPubky })
+    LocksController.getUnlockedPost({ lockUrl: lock })
```
tipogi, line 73: This is a local-first `get* → fetch*` chain implemented in the hook. Moving it into an Application `getOrFetch*` method would still leave this as a non-reactive, one-shot read.

Hooks must use `useLocalFirstQuery`: `getUnlockedPost` as `queryFn` and a network-only method that persists the post as `fetchFn`. That allows later Dexie writes to update the mounted card automatically. Apply the same pattern to `getOwnPost → fetchOwnContent` below.

src/core/services/local/locks/locks.ts
```diff
@@ -0,0 +1,55 @@
...
+import type { LockModelSchema } from '@/models/locks/locks.schema';
+import type { LockFile, ReplicatedPost, TUnlockedListItem } from '@/services/locks/locks.types';
+
+export class LocalLocksService {
+  static async get(lockId: string): Promise<LockModelSchema | null> {
+    return LockModel.findById(lockId);
+  }
+
+  static async upsertDescriptor({ lockId, descriptor }: { lockId: string; descriptor: LockFile }): Promise<void> {
+    await db.transaction('rw', LockModel.table, async () => {
```
tipogi, line 13: ADR-0015 Pattern C: wrap this `db.transaction` (and `upsertPost` below) in `try/catch`.

`LockModel.findById` / `upsert` already throw `Err.database`. A transaction abort/quota does not — that leaves a raw Dexie error across the Application boundary. The `.catch(() => undefined)` on cache writes will still swallow it, but it won’t be logged by `Err.*`.

Re-throw `AppError` unchanged (already logged). Only map a raw Dexie abort/quota with `Err.database(WRITE_FAILED, …, { operation: 'upsertDescriptor', context: { lockId }, cause })`.

src/core/application/locks/locks.ts
```diff
@@ -535,6 +598,40 @@ export class LocksApplication {
...
+    if (lockId) {
+      await LocalLocksService.upsertDescriptor({ lockId, descriptor }).catch(() => undefined);
+    }
+    return descriptor;
+  }
+
+  static async getLockFile({ lockUrl }: TFetchLockFileParams): Promise<LockFile | null> {
```
tipogi, line 610: Cache lookup is `lockIdFromUrl` only (the `.json` filename). No `isValidLockUrl`, and no `creatorFromUrl(lockUrl) === record.creator`.

`pubky://alice…/LOCK1.json` and `pubky://mallory…/LOCK1.json` both become `LOCK1`, so a reader who unlocked Alice’s lock can load Alice’s descriptor/replica under Mallory’s teaser. Same for `getUnlockedPost` / `getOwnPost` below.

Do not require announcement author === creator (A and B may differ). What must match is the lock URL host and the cached `creator`. Apply before returning; `getOrFetchLockFile` never reaches `fetchLockFile` on a hit.

src/hooks/useLockFile/useLockFile.ts
```diff
@@ -5,46 +5,28 @@ import { LocksController } from '@/controllers/locks/locks';
...
-
+    setEntry(null);
     if (!lockUrl) return;
 
     let cancelled = false;
-
-    LocksController.fetchLockFile({ lockUrl })
-      .then((fetched) => {
-        if (!cancelled) setResult(fetched);
+    LocksController.getOrFetchLockFile({ lockUrl })
```
tipogi, line 16: This is now a local-first read, so it should use `useLocalFirstQuery` instead of a hand-rolled effect calling `getOrFetch*` (ADR-0011). For example:

```ts
const { data } = useLocalFirstQuery<TFetchLockFileResult>({
  queryFn: () => LocksController.getLockFile({ lockUrl: lockUrl! }),
  fetchFn: () => LocksController.fetchLockFile({ lockUrl: lockUrl! }),
  deps: [lockUrl],
  enabled: Boolean(lockUrl),
});
```

Maybe expose `LocksController.getLockFile` for the local `queryFn`. `fetchLockFile` already persists the descriptor to Dexie.

src/hooks/useUnlockedList/useUnlockedList.ts
```diff
@@ -12,48 +13,63 @@ import type { UseUnlockedListParams, UseUnlockedListResult } from './useUnlocked
...
+  const [remote, setRemote] = useState<{ account: string; items: TUnlockedListItem[] } | null>(null);
+  const [isHomeserverListFetchFinished, setIsHomeserverListFetchFinished] = useState(false);
   const [isError, setIsError] = useState(false);
   const currentUserPubky = useAuthStore((state) => state.currentUserPubky);
   // Reading my own `/priv` needs the restored session; `currentUserPubky` is persisted and
   // rehydrates first, which would fire this before the session exists.
   const session = useAuthStore((state) => state.session);
+  // A session or account change must hide the previous reader's list immediately.
+  const active = enabled && !!currentUserPubky && !!session;
+  const local = useLiveQuery(
```
tipogi, line 25: This introduces the hand-rolled `useLiveQuery` + network `useEffect` pattern prohibited by ADR-0011. This list must show cached data immediately while refreshing the homeserver on every profile visit, so replacing it directly with the current `useLocalFirstQuery` would lose cross-device refreshes because it skips fetching on cache hits.

Could we support this through an approved shared refresh policy? Conceptually:

```ts
const { data: items = [], isRefreshing, isError } =
  useLocalFirstQuery<TUnlockedListItem[]>({
    queryFn: () => LocksController.getUnlockedList(),
    fetchFn: () =>
      LocksController.fetchUnlockedList({
        readerPubky: currentUserPubky!,
      }),
    deps: [currentUserPubky, session],
    enabled: active,
    fetchPolicy: 'always',
  });
```

`fetchPolicy: 'always'` is illustrative and would need to be implemented and documented. The fetch already persists into Dexie, so the live query can publish the refreshed list without maintaining a second `remote` source of truth.

PR #2562, opened 2026-09-16
fix(auth): bound the homeserver sign-out so logout always finishes

### LINKED ISSUE #2315: Logout hangs when the homeserver accepts the connection but never replies
## Problem

Logging out waits for the homeserver to answer before it clears anything locally. If the homeserver accepts the connection but never replies, the app waits for the OS-level TCP timeout — which can be minutes.

The user sees a spinner on `/logout` the whole time. Their cookies and local database are still there, so they are not actually logged out yet.

## Example

1. The user taps **Log out**.
2. The app calls the homeserver to end the session.
3. The homeserver is up but stuck (overloaded, mid-deploy, a proxy holding the connection open) — it never answers.
4. Nothing else happens. The spinner keeps turning. The user closes the tab, still logged in on this device.

If the homeserver were fully down, this would be fine: the connection is refused right away, the error is caught, and cleanup continues. The bad case is "connected, but silent".

## Where

`src/core/controllers/auth/auth.ts:353`

```ts
if (session) {
  try {
    await AuthApplication.logout({ session });   // no timeout
  } catch (error) {
    Logger.warn('Homeserver logout failed, clearing local state anyway', { error });
  }
}

await this.cleanupLocalState();   // cookies + local database
```

The `try/catch` does not help here. It catches an error once the call finishes, but it cannot stop a call that never finishes.

The actual network call is `session.signout()` in `src/core/services/homeserver/homeserver.ts:288`. It comes from the pubky SDK and takes no timeout or `AbortSignal`, so the wait has to be bounded on our side.

## Suggested fix

Race the signout against a short timeout (a few seconds), then continue with local cleanup either way:

```ts
await Promise.race([AuthApplication.logout({ session }), sleep(LOGOUT_TIMEOUT_MS)]);
```

Local cleanup must always run. Ending the server session is best-effort — the user's device must be cleaned regardless.

Prefer a timeout over "start it and don't wait": cleanup is followed by navigation, and the browser cancels in-flight requests when the page goes away, so the session would often never be ended at all.

## Note

The Lock Server signout in the same flow (`auth.ts:301`) has the same shape and is being handled in #2186. The two run one after the other, so in the worst case their waits add up.

## Definition of Done

- Logging out finishes in a bounded time even when the homeserver never replies.
- Cookies and the local database are always cleared.

talosmachina:
# What

Logout now bounds the homeserver sign-out, so the local database and cookies are always cleared even when the homeserver accepts the connection and never replies.

# Why

`AuthController.logout()` awaited `AuthApplication.logout({ session })` (`session.signout()`) with no timeout. A homeserver that is up but silent held logout open until the OS-level TCP timeout, so `/logout` sat on its spinner and the device stayed signed in.

# How

`src/core/controllers/auth/auth.ts` races the sign-out against a 5s timer (`LOGOUT_TIMEOUT_MS`), mirroring the resync deadline in `RouteGuardProvider`, and always proceeds to `cleanupLocalState()`. A failed sign-out logs as before; the timeout logs separately so the two cases are distinguishable in logs. Ending the server session stays best-effort, as the issue proposes.

# Evidence

Reproduced locally on `dev` @ `e7724721` with the sign-out request stalled, i.e. a homeserver that is up but never answers (`fetch` intercepting `DELETE https://homeserver.staging.pubky.app/session`, never settling):

- Before: after 60s the page still reads "Signing you out..." and `auth-store` still holds `currentUserPubky` plus the session export, so the device is still signed in. Screenshot attached.
- After: the page leaves the loading state about 5s after the sign-out is issued, renders "You have securely signed out", `auth-store` is cleared to nulls, and the local database and cookies are cleaned. Screenshot attached.

Tests: `npm test -- src/core/controllers/auth/auth.test.ts` - 49 pass, including a new case that fails (vitest timeout) against the unpatched controller. `eslint` clean.

One limit worth naming: the homeserver's own httpOnly session cookie cannot be cleared from JS, so when the sign-out never completes that cookie stays in the browser until the server answers. What this change guarantees is that the device's local state, database and app cookies are always cleaned.

Requested by @Orlandohub.

Closes #2315


talosmachina:
Screenshots from the local run: the sign-out request is stalled (homeserver up but silent), first before the change, second after.

before-stuck

after-signed-out

infin1t3 (approved):
Reviewed a02a1b1. The five-second bound lets local cleanup proceed when homeserver sign-out does not settle. Early completion clears the timer, and a late resolution or rejection does not run cleanup again. No actionable findings.

Validation: 174 targeted repository tests and three additional timeout-boundary probes passed, along with changed-file ESLint/Prettier and full typecheck. The probes covered the exact timeout boundary, late success/failure, and timer cancellation on early success.

Orlandohub (approved):

PR #2563, opened 2026-09-16
fix(tags): keep the tag-button reveal from focusing the input on mobile

### LINKED ISSUE #1650: [Bug]: Post tag button auto-focuses input field and shows soft keyboard on mobile which is not always wanted
### Description

On mobile, when tapping the post's tag button to show all tags the input field also shows and it auto-focused which makes the soft-keyboard show. This makes sense when tapping the `+` button to explicitly reveal the input field to add a tag but not necessarily when tapping the tag button to view all tags.

### Expected Behaviour

Don't auto-focus the input field when tapping the tag button.
Continue to auto-focus the input field when tapping the `+` button.

### Steps to Reproduce

1. On real mobile device
2. Find a post in the feed
3. Tap the post's tag button
4. Observe the add tag input field is focused and the soft keyboard is shown

### Screenshots

https://github.com/user-attachments/assets/028a3609-53b3-47a6-a4c7-1a005c7c0561

### Console Log

```shell

```

### Browser & Version

_No response_

talosmachina:
# What

Tapping a post's tag button no longer focuses the add-tag input on mobile, so the soft keyboard stays down. The `[+]` add control still focuses the input, as before.

# Why

The tag button reveals the tags. Revealing also focused the input, which on mobile raises the soft keyboard: the reported behaviour. Autofocus belongs to the explicit `[+]` intent, which already handles its own focus inside `ClickableTagsList` (`isAdding` -> `autoFocus={isAuthenticated && isAdding}`).

# How

Gated the reveal-time focus on the mobile breakpoint (`useIsMobile`, `< lg`), so desktop behaviour is unchanged:

- `PostInlineTagsActions` and `PostTagsExpandableRow`: `autoFocusInput={!isMobile}`.
- `PostMainListRow`: skips the `tagsPanelRef.focus()` call after toggling, and `autoFocusInput={!isMobile}`.
- `PostMain` and `PostArticleDetail`: skip the panel `focus()` calls in the tag-button handler.

# Evidence

Local run on `dev` @ `e7724721`, mobile width (780px, below the `lg` = 1024px breakpoint), signed-in feed:

- Before: clicking the tag button opens the panel and `document.activeElement` is the `add-tag-input`.
- After: the panel opens and `document.activeElement` is `body` - no input focused, no keyboard.
- After, `[+]`: `document.activeElement` is still `add-tag-input`, so the add path is untouched.

Screenshots attached.

Tests: `npm test --` over the five touched components plus `ClickableTagsList` and `PostTagsPanel` - 169 pass. A new case in `PostInlineTagsActions.test.tsx` asserts `autoFocusInput=false` at a 390px viewport and fails against the unpatched component; a companion case in `PostTagsExpandableRow.test.tsx` covers the shared component. `eslint` clean, pre-commit `tsc --noEmit` passed.

Requested by @Orlandohub, agreed in the issue by @aldertnl.

Closes #1650

before-tag-focus

after-tag-reveal

after-plus-focus


Orlandohub:
tag-reveal-no-focus

plus-focus

https://github.com/user-attachments/assets/397e08b1-c948-4654-83d9-2f8a0db898da

src/components/organisms/PostArticleDetail/PostArticleDetail.tsx
```diff
@@ -40,8 +41,12 @@ export const PostArticleDetail = ({ postId, content, attachments, isBlurred }: P
   const { openReplyDialog, openRepostDialog, dialogs } = usePostReplyRepostDialogs(postId);
   const mobileTagsPanelRef = useRef<PostTagsPanelHandle>(null);
   const desktopTagsPanelRef = useRef<PostTagsPanelHandle>(null);
+  const isMobile = useIsMobile();
 
   const handleTagClick = () => {
+    // The tag button only reveals the tags. On mobile that reveal must not focus the input and pop
+    // the soft keyboard: the `[+]` add control owns autofocus.
+    if (isMobile) return;
```
infin1t3: [P2] Preserve the article tag button's reveal action on mobile

With a persisted non-columns layout (for example Wide) and a long article on a narrow viewport, this button is above the article while the mobile PostTagsPanel is below its entire body. Previously focus() also scrolled the panel into view. This early return removes the button's only action, so tapping it now does nothing and the tags remain offscreen. SinglePostContent renders PostArticleDetail directly, and this component reads the persisted layout without PostMain's mobile layout fallback, so this state is reachable.

Reproduced at 390×844 with the actual base/head components: the base click scrolled from 0 to 1523px; at this head scroll stayed at 0 and the input remained about 2323px below the viewport's top. Manual scrolling still reaches it. Please scroll to or otherwise reveal the mobile panel without focusing its editable input, and cover the long-article/non-columns case.

infin1t3 (changes requested):
@talosmachina please fix the article tag-button regression described inline. Removing mobile input focus is correct, but the button still needs to bring an offscreen tag panel into view.

Reviewed afb92f2. All 159 targeted component tests and changed-file ESLint passed. An isolated Chromium comparison using the actual base/head article components and real tag input chain reproduced the lost scroll action; separate probes confirmed that mobile inline/expandable reveal avoids focus, explicit [+] still focuses, and desktop reveal still focuses.

src/components/organisms/PostArticleDetail/PostArticleDetail.tsx
talosmachina: Fixed in d38904fb8, thanks for the repro.

`PostTagsPanel` now exposes `reveal()`, which brings the panel into view with `scrollIntoView({ behavior: 'smooth', block: 'nearest' })` and does not touch the input. The mobile reveal paths in `PostArticleDetail`, `PostMain` and `PostMainListRow` call it instead of returning early; `PostMainListRow` reveals from an effect because its panel only mounts on expand. Desktop still uses `focus()`.

Verified: the `PostTagsPanel` handle test asserts `reveal()` scrolls the panel without focusing the input while `focus()` still focuses it; `PostArticleDetail` covers the mobile reveal reaching the mobile panel copy and not focusing, plus the desktop path still focusing both refs; `PostMainListRow` covers the reveal on mobile expand and no reveal on desktop. ESLint, Prettier and typecheck are clean on the touched files (typecheck still reports the five pre-existing dependency-drift errors in `linkify-it` and `@synonymdev/pubky` call sites).

One local note: the `PostTagsExpandableRow` collapsed snapshot fails in my checkout at this head too, before and after these commits, because the installed Radix emits an extra `aria-controls`. Not related to this change.

infin1t3 (dismissed):
Verified the fix at a5cebda9. The previous P2 is resolved: on a narrow viewport with Wide layout and a long article, the tag button now scrolls the mobile panel into view without focusing its input. The separate reveal() handle forwards to the real container, and the expanding list-row path waits until the panel mounts. Desktop focus and explicit [+] focus remain intact. I also checked the PostMain merge resolution against the current base; the repost/displayed-post behavior is preserved.

Validation: 197 tests across seven targeted component suites, changed-file ESLint/Prettier, and full typecheck passed. An isolated Chromium comparison using the actual focus-bearing components reproduced the old head's no-op and confirmed that this head scrolls to the panel without focus. The same probe checked mobile reveal, explicit [+], and desktop behavior. No physical-device keyboard or full deployed-app test was performed. No remaining actionable findings.

infin1t3 (dismissed):
Reconfirmed after the previous approval was dismissed with 'The merge-base changed after approval.' Fresh GitHub ref lookup and git fetch both place dev at b9186201; its actual merge base with this head is 50aaaa6d. The 12-file PR diff is unchanged from the reviewed version. The newer dev commit is the profile-link fix #2567 and touches none of these files.

The tag-panel finding remains resolved at a5cebda9. The verified mobile article case scrolls the panel into view without input focus; explicit [+] and desktop focus are preserved. The existing verification applies unchanged: 197 targeted tests, changed-file ESLint/Prettier, full typecheck, and the isolated Chromium regression/focus probes passed. Approving this reviewed head.

infin1t3 (approved):
Revalidated the current head 35dea019 after updating from dev. Its merge base is now 79310e22, and the complete 12-file PR diff is byte-for-byte identical to the tag fix already reviewed and verified. The later commits only bring in the merged profile-link and character-limit changes; none of the 12 reviewed files changed.

The previous finding remains resolved: mobile article tag reveal scrolls the panel into view without focusing the input, while explicit [+] and desktop focus remain intact. The unchanged fix was verified with 197 targeted tests, ESLint/Prettier, full typecheck, and the Chromium regression/focus probes. No remaining actionable findings. Approving this exact head.

PR #2564, opened 2026-09-16
fix(composer): make the character limit obvious at the limit

### LINKED ISSUE #1761: [Bug]: Max char limit reached is not obvious when pasting text to create post
### Description

After pasting the specific text below into the new post text field, I can't edit the text.

Edit: The problem is the post character limit. It's really hard to see when you hit it.

### Expected Behaviour

Make the character limit more prominent when one hits it.

### Steps to Reproduce

1. Copy this text into the new post field.
2. Try to edit it.
```
1. Humanity is emerging into a populated "Greater Community." Earth sits in a well-traveled, regulated region of space — not a remote backwater. Our isolation is
  over.                                                                                                                                                                2. An intervention by non-human visitors is already underway, but covertly. It is not a military invasion; large-scale conquest is suppressed in this region by      long-standing Rules of Engagement between established powers. The intruders are commercial Collectives — hierarchical, multi-racial economic organizations — using
  persuasion, inducement, and seduction, not force.                                                                                                                    3. Their motives are resources and allegiance. Earth is targeted for its biological richness (water, genetic diversity, plasma, organs, seeds), its strategic
  position, and a workforce already in place. Humans are seen as intelligent livestock to be pacified, not destroyed.                                                  4. Four arenas of activity:
    - Influence over government and commercial leaders (offered tech, power, "world dominion").
    - Manipulation of religious traditions — including a counterfeit Second Coming / false Jesus figure.                                                                 - Hidden establishments (underground / undersea bases) and abductions.
    - An interbreeding / hybridization program to seed leaders who look human but lack inner spiritual conscience.
  5. The chief weapon is the Mental Environment — the field of psychic and mental influence in which we all live — operating through what can be called a              Pacification Program: induced passivity, "non-judgment," romanticism about the visitors, and fundamentalism turned against anyone who dissents.
  6. Humanity's only equalizer is Knowledge — the deeper spiritual intelligence the Creator placed in every person, beneath thought and personality. Technology,
  intellect, and group-mind can all be co-opted; Knowledge cannot be deceived, abducted, or persuaded.                                                                 7. Three universal requirements for any free race in a populated universe: Unity (of purpose, planet-wide), Self-sufficiency (especially in resources — never
  depend on alien tech or trade), and Discretion (no broadcasting, secrecy about real strengths).                                                                      8. No one is coming to save us. Direct rescue would itself be a form of conquest — a free world cannot be handed its freedom. Wiser races observing humanity will
  only offer warning and wisdom from a distance; the work of preparation, awakening, and resistance must be done by humans themselves.
```

### Screenshots

With audio

https://github.com/user-attachments/assets/18b8790f-44ea-4d2b-8718-ee0de4cd0714

### Console Log

```shell
Nothing shown
```

### Browser & Version

Brave, MacOS

talosmachina:
# What

The composer character counter turns destructive at the limit, and reaching the limit raises a warning toast: title `Max characters`, description `You've reached the maximum length for this post`.

# Why

The composer rejects writes past `POST_MAX_CHARACTER_LENGTH`, so exactly at the limit the field stops accepting input with nothing saying why. The counter stayed muted the whole time (issue #1761). Copy and behaviour follow what @aldertnl asked for in the thread.

# How

- `PostHeaderUserInfo`: the counter renders `text-destructive` when `count >= max`, muted otherwise. This is the shared counter, so it covers every composer that measures content.
- New `useCharacterLimitWarning(limit)` hook: fires the warning toast once each time the content reaches the limit, re-arming when it drops back below. Called with the same `{ count, max }` the counter gets, in `PostInput` (new post, edit and repost, which renders `PostInput` with the repost variant) and `QuickReply` (replies). Passing `undefined` while the composer is collapsed keeps it silent.

# Evidence

Local run on `dev` @ `e7724721`, signed in, composer content set to exactly the 2000-character limit:

- Before: counter reads `2000/2000` and stays `text-muted-foreground`, no toast.
- After: counter reads `2000/2000` with `text-destructive`, and the warning toast appears.
- Reply composer (same paste): its counter also goes `text-destructive` and the same toast fires.

Screenshots attached.

Tests: caches/`npm test` over the touched areas - `useCharacterLimitWarning` (warning fires once, does not repeat while at the limit, re-arms after dropping, silent when collapsed), `PostHeaderUserInfo` (muted below the limit, destructive at it), plus `PostInput`, `QuickReply`, `PostHeader` suites: 258 pass. One `PostHeaderUserInfo` snapshot updated for the class-order change only. `eslint` clean, pre-commit `tsc --noEmit` passed.

Requested by @Orlandohub, with the copy and the reply/repost coverage specified by @aldertnl and @catch-21.

Closes #1761

before-limit

after-limit-toast

after-reply-limit


Orlandohub:
Composer counter muted at 1999/2000

2000/2000 text-destructive with Max characters toast

New post composer at limit with toast

https://github.com/user-attachments/assets/d55ab931-0e66-441c-bb31-39d792531c02

Orlandohub:
31-toast-visible

https://github.com/user-attachments/assets/f1f37020-8efc-4fa0-9431-c2531138422c

src/components/organisms/PostInput/PostInput.tsx
```diff
@@ -259,6 +260,7 @@ export function PostInput({
 
   const characterLimit =
     isExpanded && !isArticle ? { count: getCharacterCount(content), max: POST_MAX_CHARACTER_LENGTH } : undefined;
+  useCharacterLimitWarning(characterLimit);
```
infin1t3, line 265: [P2] Base the warning on the limit the composer actually enforces

Enter one 😀 followed by 1,998 ASCII letters. Native maxLength and usePostInput.handleChange enforce UTF-16 .length, so this fills the 2,000-unit cap and further input is rejected. getCharacterCount counts Unicode code points and returns 1,999, however, so this new hook emits no warning and the counter stays muted at 1999/2000. QuickReply.tsx:123 has the same wiring.

The count mismatch already exists at the base, but connecting the new warning/red state to that count leaves the reported 'input silently stops accepting text' problem unresolved for ordinary emoji-containing drafts. Please derive the warning and red state from the same measure used by the enforced cap, without changing the product-wide limit, and add an emoji-containing boundary case. This was reproduced in Chromium; a 2,000-character ASCII control correctly warned.

infin1t3 (changes requested):
@talosmachina please address the warning-condition gap described inline in both PostInput and QuickReply. A draft containing an emoji can reach the enforced input limit without the new warning or red counter state.

Reviewed c48fc50. The underlying UTF-16/code-point mismatch predates this PR; this is an incomplete fix of the requested limit feedback, not a newly introduced input-cap regression. All 306 targeted tests, changed-file ESLint, and Prettier passed. A Chromium probe using the actual warning hook and count helper with the native textarea/input contract reproduced the missing warning; an ASCII control correctly produced a red counter and one toast.

src/components/organisms/PostInput/PostInput.tsx
talosmachina: Fixed in bdb57623c, thanks for the repro.

`getEnforcedCharacterCount` (UTF-16 units) is now the source for the counter, its destructive state and `useCharacterLimitWarning` in both `PostInput` and `QuickReply`, so an emoji draft reaches the enforced cap and warns instead of sitting at 1999/2000 muted. The product-wide limit is untouched; the visible consequence is that an emoji now counts the two units it occupies, which is what `maxLength` and the write handlers accept.

Verified: two new `PostInput` tests at the boundary (`😀` plus 1,998 ASCII letters reaches 2000/2000 and warns; one unit below stays silent and muted) fail against the previous wiring and pass with this commit; a `QuickReply` test pins the same measure; `getEnforcedCharacterCount` has unit tests beside `getCharacterCount`. ESLint, Prettier and typecheck are clean on the touched files.

Orlandohub (approved):

infin1t3 (approved):
Verified the fix at bdb57623. The previous P2 is resolved: both PostInput and QuickReply now derive the counter, destructive state, and warning from the UTF-16 count actually enforced by maxLength and the write handlers. The product-wide limit and unrelated code-point display counts are unchanged. No remaining actionable findings.

Validation: 549 tests across eight targeted suites, changed-file ESLint/Prettier, and full typecheck passed. A Chromium probe at desktop and narrow viewports confirmed that an emoji draft at the enforced cap displays 2000/2000 in red and warns once, stays silent below the cap, does not duplicate the warning on blocked input, and warns again after deleting and refilling. The probe uses the real warning hook/helper in a minimal native-input harness; the actual composer call sites were checked in source and component suites, not a full authenticated browser flow.

PR #2565, opened 2026-09-16
fix(composer): say why a write was refused with 507 Insufficient Storage

### LINKED ISSUE #1776: [Bug]: Toast error for 507 Insufficient Storage
### Description

When hitting storage limit, homesever with respons 507 Insufficient Storage. The toast UI we show reads "Error" "Failed to create post. Please try again."

### Expected Behaviour

The text should be specific like "You have reached your disc quote." (and because there's no way to upgrade yet, say to contact support to request more?)

The toast could be a warning instead of error

### Steps to Reproduce

1. Upload near your user's disc quota
2. Attempt to upload a large file
3. Observe generic toast error

### Screenshots

Image

### Console Log

```shell

```

### Browser & Version

_No response_

talosmachina:
# What

A homeserver `507 Insufficient Storage` now warns with copy that says why, instead of the generic retry error. Applies where the write is committed: new post, reply, repost.

# Why

The create path mapped only image-upload size-limit errors to specific copy, so a full storage quota fell through to `Could not create post. Try again.` - which suggests retrying, and retrying cannot help (issue #1776).

# How

- `HttpStatusCode.INSUFFICIENT_STORAGE = 507` added to the shared enum.
- `src/libs/storage/storageQuota.ts`: `getStorageQuotaToastMessage(error)` maps an `AppError` whose `context.statusCode` is 507 to the quota message, mirroring `getImageUploadSizeLimitToastMessage`.
- `usePost`: one `showCommitErrorToast(error, fallback)` used by the post, reply and repost catches. A quota error warns with `You have reached your storage limit. Contact support to request more space.`; the size-limit message and the generic fallback are unchanged.

Copy follows the wording suggested in the issue; the "contact support" half is the issue's own suggestion and is easy to reword if the owner prefers shorter.

# Verified

- `src/libs/storage/storageQuota.test.ts`: 507 maps to the message, other statuses and plain errors return null.
- `src/hooks/usePost/usePost.test.tsx`: a 507 rejection from the commit shows the warning toast with the quota copy, and keeps the draft. Fails against the unpatched hook.
- Full `usePost` + `libs/storage` suites: 429 pass. `eslint` clean, pre-commit `tsc --noEmit` passed.

# Not verified yet

I could not exercise a real 507 end-to-end. I filled the composer locally and started intercepting the homeserver write with a 507 response to check the toast renders, and ran out of run budget before the post button fired. So the toast on a live 507 is asserted by unit tests, not by a screenshot; no screenshot is attached rather than an unrelated one. Happy to finish that check on request.

Requested by @Orlandohub.

Closes #1776


infin1t3 (approved):
Reviewed e4af56b. The SDK and HTTP 507 status reaches the new warning through the application error path. Failed post, reply, and repost submissions retain their drafts and preserve local rollback behavior; the existing size-limit and generic-error fallbacks remain intact. No actionable findings.

Validation: 630 targeted repository tests and five additional boundary probes passed, along with changed-file ESLint and full typecheck. The probes exercised the real hook/controller/application/HomeserverService/Dexie path with SDK writes rejecting in the documented RequestError shape; no live quota-exhausted homeserver was used.

Orlandohub (approved):

talosmachina:
Follow-up on the missing end-to-end check, with what I actually saw: in the local build the composer never reaches a homeserver write. Clicking Post fires only the author's pkarr resolution (`GET pkarr.pubky.app/<pubky>`) and then fails, so no request exists for me to answer with a 507. Producing a live 507 here would need a homeserver whose quota is full, not an intercept. The unit-level mapping is as described above.

PR #2566, opened 2026-09-16
fix(posts): prevent premature unread counts

### LINKED ISSUE #2160: [Bug]: 'See new post' button shows for a moment every time someone else creates a new collection
### Description

When viewing your feed, the 'See n new posts' button appears for a moment as someone else creates a new collection. This is likely because it is being fetched in a new query and being filtered out from the global feed but the button hasn't quite realised yet.

### Expected Behaviour

'See n new posts' button only shows when there are new posts matching the applied filter.

### Steps to Reproduce

1. Have two instances of pubky-app with different users signed in on each
2. View global feed with 'All' Content filter applied with user A
3. Create a new collection as user B
4. Observe 'See 1 new post' button appears at top of user A's feed for a moment

### Screenshots

https://github.com/user-attachments/assets/f69a1e26-534a-4223-be59-ff5f4c491fc2

### Console Log

```shell

```

### Browser & Version

_No response_

infin1t3:
## Summary

Unread IDs can arrive before their post details, briefly making collections appear in the "See new posts" count. Count only posts with local details that match the active stream, and wait for mute-list readiness outside bookmarks. Stream changes immediately discard the previous count. Collection membership additions also wait for mutes to settle.

Opening ready posts merges and acknowledges exactly those IDs in one Dexie transaction. Pending posts and later arrivals stay unread; concurrent unread writes are serialized with acknowledgement. If the main head also lacks details, polling retains SKIP instead of re-reading the newest page as unread.

Normal polls retry missing unread details. A known cursor lets newer-key discovery continue while the retry runs in the background. Retries share one task per stream/viewer, discard obsolete session results and release their task after failure. Without a usable cursor, polling waits for hydration and re-reads the rows afterward.

Closes #2160

## Verification

- `npm run lint`, `npm run typecheck`, `npm run format:check`: passed.
- Full `npm test`: 796 files passed; 13,898 passed, 2 expected failures, 2 skipped.
- `CI=1 npm run test:vrt -- src/test/vrt/feed`: 117 checks passed across Chromium, Firefox and WebKit; no baseline changes.

Real Dexie integration tests cover missing main-head details, selective merge/acknowledgement, later arrivals, rollback, concurrent writes, failed/omitted hydration, discovery during a slow retry and session replacement. Component tests cover mute readiness, collection membership and the bookmark exception.

## Remaining limitation

A `by_ids` omission does not establish deletion, so omitted IDs remain retryable and uncounted until details arrive. Permanently unavailable IDs may still be retried while the feed is mounted; the existing query cache limits HTTP requests. No first-omission eviction or new retry policy is introduced.


src/core/application/stream/posts/post.ts
```diff
@@ -128,6 +129,26 @@ export class PostStreamApplication {
...
+        await this.fetchMissingPostsFromNexus({ cacheMissPostIds, viewerId, isCurrent });
+      }
+    }
+    if (!isCurrent()) return SKIP_FETCH_NEW_POSTS;
+    let streamHead = await this.getStreamHead({ streamId });
+    if (streamHead === SKIP_FETCH_NEW_POSTS) {
+      // Nexus may omit a deleted or not-yet-indexed unread ID. Keep it pending
+      // for hydration, but let newer keys arrive from the known main-stream cursor.
+      const mainStreamHead = await this.getMainStreamHeadTimestamp({ streamId });
+      streamHead = mainStreamHead === SKIP_FETCH_NEW_POSTS ? FORCE_FETCH_NEW_POSTS : mainStreamHead;
```
secondl1ght: **Blocking.** This turns the old "main head has no details → do not poll" guard into a poll from the very top of the stream. `getMainStreamHeadTimestamp` returns `SKIP` when the main row *has* a head but that head has no `post_details` (L377). Mapping that to `FORCE` (1) makes `setStreamPagination` send `end = 2`, i.e. the newest page with no lower bound. That page is already in the main row, so `persistUnreadNewStreamChunk` writes ten already-cached ids into the unread row. The UI page is also ten and the detail-less head occupies one slot, so the tenth post is "not displayed" and the button reads "See 1 new post" for a post the user already has; clicking it prepends an older cached post to the top of the feed. That is the class of bug #2160 is about.

The detail-less main head is reachable: the click handler's merge (see the NewPostsSection comment) moves a pending unread id into the main row, and a bootstrap row can also carry ids whose details never landed. Only the *unread*-head case should fall through to the main head; if the main head has no details either, return `SKIP` as before.

---

src/components/organisms/Timeline/Feed/NewPostsSection/NewPostsSection.tsx
```diff
@@ -53,13 +55,16 @@ export function NewPostsSection({
...
+        ? []
+        : MuteFilter.filterPostsSafe(notDisplayed, mutedUserIdSet);
   const actualNewCount = actualNewPostIds.length;
 
   const handleNewPostsClick = async () => {
     try {
       await StreamPostsController.mergeUnreadStreamWithPostStream({ streamId });
-      await StreamPostsController.clearUnreadStream({ streamId });
+      // Pending details and posts arriving after this render remain unread.
+      await StreamPostsController.clearUnreadStream({ streamId, postIds: actualNewPostIds });
```
secondl1ght: **Blocking.** These two calls now disagree about what was acknowledged. `mergeUnreadStreamWithPostStream` still moves the *entire* unread row into the main row — `PostDetailsModel.filterDeleted` is fail-open, so ids without details are included — while `clearUnreadStream` removes only `actualNewPostIds`. A pending id `P` therefore ends up in both rows.

Two consequences: (1) if Nexus keeps omitting `P`, the next poll sees unread head `P → SKIP` and main head `P → SKIP`, which is exactly the `FORCE` path flagged in `post.ts` L147; (2) if Nexus later returns `P`, the poll cursor becomes `P.indexed_at`, which is older than the ready posts the user just acknowledged, so those posts are re-inserted into the unread row on every poll and only the component's `notDisplayed` filter hides them until the next remount clears the row.

Either merge only `actualNewPostIds` (pass the list through to a selective merge, mirroring the selective clear), or clear everything you merged.

---

src/core/application/stream/posts/post.ts
```diff
@@ -128,6 +129,26 @@ export class PostStreamApplication {
     return await LocalStreamPostsService.getStreamHead(params);
   }
 
+  /** Retry unresolved unread details on the existing poll before advancing its cursor. */
+  static async getOrFetchStreamHead({ streamId, viewerId, isCurrent }: TGetOrFetchStreamHeadParams): Promise<number> {
+    const unreadStream = await this.getUnreadStream({ streamId });
+    if (unreadStream && unreadStream.stream.length > 0) {
+      const cacheMissPostIds = await this.getNotPersistedPostsInCache(unreadStream.stream);
+      if (cacheMissPostIds.length > 0) {
+        await this.fetchMissingPostsFromNexus({ cacheMissPostIds, viewerId, isCurrent });
```
secondl1ght: This retries for as long as the feed is mounted: an id Nexus never returns (deleted between the head poll and its `by_ids` hydration) is re-requested on every poll with no attempt cap, backoff, or eviction from the unread row. The row also cannot age out — `getUnreadStreamHeadTimestamp` returns `SKIP` for that head and `isTimestampStale` treats sentinels as not stale — so only the unconditional merge+clear on remount ends it, and that merge pushes the dead id into the main row (the precondition for the L147 bug). Hydrating before persisting removes the need for this retry entirely; if the retry stays, drop ids from the unread row once a `by_ids` response omits them.

---

src/core/coordinators/streams/stream.ts
```diff
@@ -238,7 +238,8 @@ export class StreamCoordinator extends Coordinator<StreamCoordinatorConfig, Stre
    */
   private async resolveStreamHead(currentStreamId: PostStreamId): Promise<boolean> {
     try {
-      const streamHead = await StreamPostsController.getStreamHead({ streamId: currentStreamId });
+      const streamHead = await StreamPostsController.getOrFetchStreamHead({ streamId: currentStreamId });
```
secondl1ght, line 241: Resolving the head is now a Nexus round trip, but the base `Coordinator.startPolling` fires `void this.poll()` from `setInterval` with no in-flight guard (`coordinator.ts` L242). With pending misses and Nexus latency above the poll interval (a rate-limited `by_ids` retry, a slow mobile network), poll N+1 starts while poll N is still awaiting hydration: both issue `fetchByIds` for the same ids, both run a head fetch, and the later `streamState.streamHead` assignment can be the older poll's cursor. Before this PR the head resolve was a pure Dexie read, so this was not possible. A single in-flight promise per coordinator (skip the tick while one is pending) covers it; a per-poll generation counter incremented in `resolveStreamId` would also close the `X → Y → X` case that the `currentStreamId !== currentStreamId` check on L242 lets through.

---

src/core/coordinators/streams/stream.integration.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+  if (hasMainHead) {
+    await PostDetailsModel.create({ ...main.details, id: mainId });
+    await PostStreamModel.upsert(streamId as PostStreamId, [mainId]);
+  }
+  await UnreadPostStreamModel.upsert(streamId as PostStreamId, [pendingId]);
+  vi.spyOn(NexusPostStreamService, 'fetchByIds').mockResolvedValue([]);
+
+  const head = await StreamPostsController.getOrFetchStreamHead({ streamId });
+
+  expect(head).toBe(hasMainHead ? main.details.indexed_at : 1);
```
secondl1ght, line 154: Neither case reaches the branch that changes behaviour. `hasMainHead=false` has no main row, so `getMainStreamHeadTimestamp` returns `FORCE` directly; `hasMainHead=true` seeds full details for the head. The `mainStreamHead === SKIP_FETCH_NEW_POSTS ? FORCE_FETCH_NEW_POSTS : mainStreamHead` ternary is never exercised. Add a third case with `PostStreamModel.upsert(streamId, [mainId])` and **no** `PostDetailsModel.create` for it; it should assert `SKIP_FETCH_NEW_POSTS` (no poll), and with the current code it asserts `1`, which is the regression flagged on `post.ts` L147.

---

src/components/organisms/Timeline/Feed/TimelineFeedContent/TimelineFeedContent.tsx
```diff
@@ -236,7 +236,7 @@ function TimelineFeedContent({
   // does); additions are reconciled once the stream has settled: any member
   // the stream never delivered is prepended once — except muted authors, whom
   // the stream filters on purpose. Both are idempotent.
-  const { mutedUserIdSet } = useMutedUsers();
+  const { mutedUserIdSet, isLoading: mutedUsersLoading } = useMutedUsers();
```
secondl1ght, line 239: **Non-blocking, same race.** `mutedUsersLoading` now gates `NewPostsSection`, but the membership-reconciliation effect below (L267, `MuteFilter.isPostMuted(id, mutedUserIdSet)`) still runs against the empty placeholder set while mutes load. On a single-collection page viewed by a non-owner, a stream fully served from Dexie settles before the muted-stream read resolves, so a muted author's item is prepended optimistically and `prependedRef` marks it done; the later effect then has to remove it again, which flashes. Gating `streamSettled` (or that effect) on `!mutedUsersLoading` closes it the same way.

---

secondl1ght (changes requested):
This fixes the symptom in #2160 (an unread id counted before its details exist) but leaves the state that causes it in place and, in one path, reintroduces a spurious count. Requesting changes for the two blocking inline comments; the rest are worth doing in the same PR because they fall out of the same root cause.

**Root cause.** `getOrFetchStreamSlice` persists a head poll's ids into the unread row (`persistUnreadStreamChunkAndUpdateCounts`, ~L897) before their details are hydrated. Everything this PR adds — the ready-filter in `useUnreadPosts`, the per-poll `fetchMissingPostsFromNexus` retry in `getOrFetchStreamHead`, the selective `clearUnreadStream`, and the `SKIP → FORCE` fallback — exists to cope with ids that sit in the unread row without `post_details`. The same function already avoids that state for replies: the `StreamSource.REPLIES` branch (~L869) hydrates `cacheMissPostIds` first and only then persists. Doing the same for the unread chunk (hydrate, drop whatever Nexus did not return, then persist) makes a detail-less unread id impossible, so the four workarounds and the two blocking bugs disappear. A failed hydration just leaves the unread head where it was and the next poll re-fetches the same page, which is the behaviour we already accept for replies.

If you'd rather keep the reader-side approach, the two blocking comments still need fixing on their own.

---

src/core/application/stream/posts/post.ts
infin1t3: Fixed. A main row whose head lacks details now retains SKIP, so polling cannot turn that state into an unbounded newest-page request. Added the missing-main-details regression; it failed with FORCE on the previous head and now passes.

src/components/organisms/Timeline/Feed/NewPostsSection/NewPostsSection.tsx
infin1t3: Fixed with a selective merge and acknowledgement in one Dexie transaction. The click passes exactly actualNewPostIds; pending IDs and posts arriving after the render stay unread and are not copied into the main row. The unread writer also uses a transaction, so a concurrent poll cannot restore an acknowledged ID from a stale read. Tests cover both rows, later arrivals, tail-cursor preservation, concurrent acknowledgement and rollback.

src/core/application/stream/posts/post.ts
infin1t3: I kept omitted IDs retryable: a by_ids omission does not establish deletion, and dropping an older pending ID after a newer ID advances the head can lose that post when indexing catches up. The existing Nexus query cache limits repeated HTTP requests; the added per-stream/viewer hydration task also shares persistence work across overlapping ticks. Permanently unavailable IDs can still be retried while the feed stays mounted. A bounded retry policy remains a separate improvement; this change does not silently evict IDs on omission.

src/core/coordinators/streams/stream.ts
infin1t3: I reproduced a related regression: awaiting an old pending hydration blocked newer-key discovery even when a usable cursor was cached. The retry now runs in the background in that case, sharing one task per stream/viewer and replacing it when the session changes. Without a usable cursor, it waits and re-reads the rows afterward. Tests hold the old retry open across several ticks while a fresh post becomes readable, and cover session replacement, stale cleanup and failure recovery. The existing query cache already deduplicates identical HTTP requests; I could not establish the claimed new backward-cursor race, so I kept this fix scoped to the reproduced blocking behavior.

src/core/coordinators/streams/stream.integration.test.tsx
infin1t3: Added this exact case: a main stream row with no details for its head, plus an omitted unread ID. It asserts SKIP and preservation of the pending unread ID. The test failed on the previous code and passes with the fallback fix.

src/components/organisms/Timeline/Feed/TimelineFeedContent/TimelineFeedContent.tsx
infin1t3: Added the mute-readiness guard to collection membership additions. Removals still apply immediately. The regression test verifies that no member is prepended while mutes load, then only the unmuted member is added. This race also reproduces on the base revision, but the small guard fits the readiness work here.

infin1t3:
Addressed the review in c915dba32734375478ff51adeb7813460f5422e9 and replied to each thread.

Both blockers are fixed: a missing main-head detail row retains SKIP, and clicking new posts atomically merges/acknowledges only the selected IDs. I also guarded collection membership additions while mutes load and serialized unread writes with acknowledgement.

The additional review found another regression: a slow retry for an old pending post blocked discovery of newer posts despite a known cursor. Hydration now runs in the background in that case, with one shared task per stream/viewer and session-safe cleanup. The integration test holds that retry open while a fresh post becomes readable.

I kept the reader-side readiness check and retained omitted IDs. A partial `by_ids` result can omit a post that becomes indexed later; dropping an older omitted ID after a newer cursor advances can lose it. The replies path also still persists IDs after unsuccessful/partial hydration, so it does not provide the proposed no-missing-details invariant. Permanently unavailable IDs may still be retried while mounted; that remaining limitation is documented in the PR.

Validation: 13,898 tests passed (2 expected failures, 2 skipped), lint, typecheck, format, and all 117 feed VRT checks passed. No baseline changes.

secondl1ght (approved):
Re-reviewed at `87e7f54`. Both blocking findings are fixed and verified against the code, not just the replies:

- A main head without details now falls through to `getMainStreamHeadTimestamp` and keeps `SKIP`; the new integration case seeds exactly that state and asserts `0`.
- `markUnreadPostsAsRead` merges and acknowledges the same selected ids in one Dexie transaction over `post_streams`, `unread_post_streams` and `post_details`; the model helpers touch only their own tables, no caller nests it inside a transaction that lacks them, and the rollback and concurrent-poll tests exercise the real transaction.

The background hydration is safe to leave un-awaited: `fetchMissingPostsFromNexus` catches everything and resolves `false`, the per-stream/viewer task is shared so overlapping ticks cannot double-fetch, and a stale-session task is replaced without its cleanup evicting the current one. Keeping omitted ids retryable rather than evicting on a single `by_ids` omission is a reasonable call given that an omission does not prove deletion; the PR body records the trade-off. The mute-readiness guard on collection membership closes the same race in the second consumer.

CI is green on this head (code quality, all five test shards, Next build, preview image), and the dev merges did not touch any of the follow-up's files.

---

PR #2567, opened 2026-09-16
fix(profile): accept a bare X handle in the profile links form

### LINKED ISSUE #1846: [Bug]: Edit profile links X profile placeholder suggests to use handle
### Description

When editing profile, `@username` is used as placeholder for X link when that format is not valid.

### Expected Behaviour

Either support the handle or suggest url

### Steps to Reproduce

1. Go to edit own profile
2. Observe placeholder for X link input

### Screenshots

https://github.com/user-attachments/assets/8932fd21-9c1e-46e1-99e1-8981d35ea9fd

### Console Log

```shell

```

### Browser & Version

_No response_

talosmachina:
# What

The X (Twitter) profile link field accepts a bare handle. `@jack` and `jack` are stored as `https://x.com/jack` instead of being rejected as an invalid URL; real URLs and the other link fields are untouched.

# Why

The field's placeholder reads `@user`, but the value was validated as a URL, so following the placeholder produced `Invalid URL` (issue #1846). The owner chose to accept the handle rather than change the placeholder.

# How

- `src/libs/profile/profileLinks.ts`: `normalizeProfileLinkUrl(label, url)` rewrites a bare handle to its profile URL for the X slot only (`^@?[A-Za-z0-9_]{1,15}$`), and returns anything else trimmed and unchanged.
- `useProfileForm` uses it in both places that matter, so validation and saving cannot disagree: `validateLinkUrl` validates the normalized value, and `validateUser` hands the normalized links to `UserValidator.check`, whose data is what gets committed. Placeholder left as `@user`.

# Verified

Local run on `dev` @ `e7724721`, signed in, `/settings/edit`:

- Before: typing `@talosqatest` in the X field shows `Invalid URL` (screenshot attached).
- After: the same value shows no error and Save is enabled (screenshot attached); saving reports `Profile updated`.

Tests: `src/libs/profile/profileLinks.test.ts` covers handle forms (with and without `@`, trimmed), a real X URL, other labels and empty values. `useProfileForm.test.tsx` submits `@jack` through the form and asserts the committed payload carries `{ label: 'X (TWITTER)', url: 'https://x.com/jack' }`; it fails against the unpatched hook. Suites over `libs/profile`, `useProfileForm`, `CreateProfileForm` and `Settings`: 95 pass. `eslint` clean, pre-commit `tsc --noEmit` passed.

One gap: after saving I could not read the stored `profile.json` back. The browser harness hung on the reload, and a direct public read of the homeserver path answers `500 Can't extract PubkyHost` without the `PubkyHostLayer` host header, so the stored value is asserted by the payload test rather than by a read-back.

Requested by @Orlandohub; the handle-over-placeholder choice was made in the issue thread.

Closes #1846

before-invalid-url

after-handle-accepted


Orlandohub:
71-at-handle

Orlandohub:
73-after-save

Orlandohub:
81-invalid-x

Orlandohub:
https://github.com/user-attachments/assets/9cb3ce93-0e46-4079-992e-6e790dd800bf

Orlandohub:
​

x-handle-accepted

bare-handle-accepted

after-save-profile

https://github.com/user-attachments/assets/05a4bd3b-f4a5-4c70-8606-d49e6a42c56a

Orlandohub (approved):

PR #2569, opened 2026-09-16
feat(og): show display names for pubky mentions in link previews

secondl1ght:
## Summary

The app renders a raw `pk:<key>` / `pubky<key>` mention in post text as the user's display name (`PostMentions`), but the `<meta>` description and the dynamic OG image still showed the 52-character key, which is confusing when a link is shared on another platform.

- **Mentions resolve to display names** in the post and profile `<meta>` descriptions and OG cards (post text, image-card text, article bodies, profile bios). Fallback when the profile is missing, nameless, or fails to load is the shortened key, upper-cased like the app's mention links. Article titles and collection names stay verbatim, as the app shows them; one predicate (`isMentionResolvablePreview`) decides this for both the page and the card.
- **Mentions are drawn in the brand colour** on the cards, matching the app's mention links. satori (0.25, bundled in `next/og`) has no inline formatting context, so mention-bearing copy is rendered as word pieces in a wrapping flex row (split after `/` and `-` so URLs and hyphenated words flow across rows; NBSP stays glued). Copy without mentions is still a single text node, so those cards are pixel-identical to before.
- **Emoji fallback for satori**: an under-qualified ZWJ sequence (`❤‍🔥`, `🧙‍♂`) is fully qualified so the provider has an asset for it; an invalid sequence (`🧙‍🧌`, seen in a real display name) is split into its component emoji as browsers draw it, instead of reserving width and painting nothing. Non-emoji joiner clusters (Indic conjuncts, Arabic joining forms) are untouched. Known limit: Emoji 15+ sequences still render blank because the bundled Twemoji 14 has no assets for them.
- **Best-effort lookups**: mentioned-user names are fetched through a non-throwing call under `NEXUS_MENTION_LOOKUP_TIMEOUT_MS` (1s), so a Nexus wobble emits no Sentry event per mention and cannot push metadata or the OG render past the crawler budget. Only mentions that can appear in the visible window are looked up, capped by `NEXUS_MENTION_LOOKUP_LIMIT`.
- The mention regex now lives once in `identity.constants.ts` (SDK-free, so the server OG path does not pull in `@synonymdev/pubky`) and is shared by the app, notification previews and metadata.

## Verification

- [x] `npm test` (full, 13.5k) plus targeted: `src/libs/og`, `src/libs/post`, `src/libs/utils/truncate.test.ts`, `src/app/post/[userId]/[postId]/page.test.tsx`, `src/app/profile/[pubky]/page.test.tsx`, `PostText`, `NotificationItem`, `PostMentions`, `useNotificationPostContent`
- [x] `npm run build` (route- or config-wide changes)
- [ ] `npm run test:vrt` (no VRT surface changed)

Beyond the unit tests (which mock `ImageResponse`): the four affected routes were rendered end to end on a local dev server against staging data (post text, image post, article, profile), and the layout cases raised in review were rendered through the real satori with the actual `OgText` component: mention glued to a long URL, URL-only copy without mentions (identical to the pre-change text node), long name followed by punctuation, NBSP and hyphenated words, under-qualified and invalid emoji sequences.

## Callouts

- [x] Changes a `src/config/*` limit (product-wide policy): adds `NEXUS_MENTION_LOOKUP_TIMEOUT_MS` (1000) and `NEXUS_MENTION_LOOKUP_LIMIT` (10) to `src/config/nexus.ts`. New constants, no existing limit changed.

Known divergence from the app, accepted: mentions are resolved with a regex over the flattened text, so a key inside an inline code span or heading resolves on the card while the app shows it literally. The existing notification resolver behaves the same way.

## Screenshots

Rendered cards (before/after on real staging posts and the review layout cases) were captured during development and reviewed; the metadata for the same posts reads e.g. `Hi @Talos and @Jeb, play rock-paper-scissors…` where it previously carried two 57-character tokens.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GuqjgwJBpzZt8cycw1brVG)_


Orlandohub:
post-1-og-card

post-1-page

https://github.com/user-attachments/assets/b668d521-f7b5-49b0-a1e1-b86ff7cc0912

talosmachina (commented):
No findings.

Reviewed the whole diff (ogText, postMentions, postMetadata, OgComponents, renderPostOg, the two metadata pages, the moved regex) and ran the touched suites: `npm test -- <the seven test files>` -> 7 files, 56 tests, all pass on this head.

What I checked past the tests, and ruled out:

- Lookup window vs render budget: the lookups are budgeted with `OG_TRUNCATE.postText` (110) while the article body renders 78 and the image card 90, and a resolved label is always shorter than the 57-character token it replaces, so nothing that can become visible is left unresolved; the `@x` stand-in in `resolveMentionSegmentsForMetadata` is the shortest possible label, which makes the lookup set a superset of the visible one.
- The shared global `MENTION_IN_TEXT_REGEX`: both consumers (`splitMentions`, `NotificationItem.resolvePubkyToNames`) use `matchAll` / `String.replace(fn)`, neither of which leaves a `lastIndex` behind, so moving it into `identity.constants.ts` is safe, and `PUBKY_IDENTIFIER_WITH_PREFIX_SOURCE` now has no reader outside its own file.
- `OgText`'s no-mention path is still a single text node (the snapshot delta is 2 lines), and the piece-splitting path only runs when a mention is actually present. Removing the article body's `-webkit-box` clamp changes nothing in satori, which does not implement it.
- `isMentionResolvablePreview` keeps the page description and the OG card on the same branch decision, so an article title stays verbatim in both and a deleted post resolves in neither.

Filed as COMMENT: nothing here is worth blocking the merge on.

secondl1ght:
@Orlandohub does your comment plus Talos mean an approval?

talosmachina:
@secondl1ght My review on this PR is a COMMENT review with no findings, not an approval. I don't approve pull requests, so the merge gate still needs an approval from a human reviewer; the review just means I found nothing worth blocking on in the diff, the tests I ran and the edge cases I checked.

secondl1ght:
@talosmachina yes that is why I pinged @Orlandohub...

src/libs/identity/identity.constants.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Regex source for a pubky mention: `pk:` or `pubky` followed by exactly 52
+ * lowercase alphanumeric (z-base-32) characters.
+ *
+ * Lives outside `Identity` so server-only code (metadata / OG image rendering)
+ * can share the pattern without pulling the `@synonymdev/pubky` WASM SDK that
+ * `Identity` depends on into the server bundle.
+ */
+export const PUBKY_IDENTIFIER_WITH_PREFIX_SOURCE = '(?:pk:|pubky)[a-z0-9]{52}';
```
Orlandohub: `PUBKY_IDENTIFIER_WITH_PREFIX_SOURCE` is exported but its only reader is `MENTION_IN_TEXT_REGEX` ten lines below (`rg PUBKY_IDENTIFIER_WITH_PREFIX_SOURCE src` returns only this file; the former reader in `Identity` is removed in this PR). Per `no-unused-exports` in `.greptile/config.json`, drop the `export` and keep it a module-local `const` so the public surface of this new file is just the regex.

src/libs/post/postPreview.ts
```diff
@@ -15,6 +15,20 @@ import { parseCollectionContent } from './collectionContent';
...
+ * collection name, which the app shows verbatim. A `long` post whose content is
+ * not an article envelope previews its raw content, and that copy does get
+ * mention links in the app. Shared by the page description and the OG card so
+ * the two never disagree on which previews resolve mentions.
+ */
+export function isMentionResolvablePreview({ content, kind }: { content: string; kind: string }): boolean {
+  if (isPostDeleted(content)) return false;
+  if (kind === 'long') return parseArticleContent(content) === null;
+  return kind !== 'collection';
+}
```
Orlandohub, line 43: This landed between `deriveTextPreview`'s JSDoc (lines 5–17, "Derives a human-readable… Branching… Pure function") and the function itself (now line 32). Two JSDoc blocks are stacked on `isMentionResolvablePreview`, and `deriveTextPreview` no longer has a doc attached, so editors and go-to-definition show the branching table against the wrong function. Move `isMentionResolvablePreview` below `deriveTextPreview` (it reads naturally as "the preview, then the predicate about that preview"), or move the original block back onto its function.

src/libs/og/OgComponents.tsx
```diff
@@ -64,6 +65,75 @@ export function OgAvatar({ src, size }: { src: string | null; size: number }) {
...
+/**
+ * Cuts segment text into the items of satori's wrapping flex row: words, and
+ * within a word the pieces a text node could break between (`https://`,
+ * `example.com/`, `rock-`, `paper-`), so long URLs and hyphenated words flow
+ * across rows instead of claiming whole ones. Whitespace runs collapse to one
+ * space carried by the preceding piece, as `white-space: normal` does. Runs of
+ * adjacent segments with no whitespace between them (`@Jeb,`) become adjacent
+ * pieces, which only part at a row end when the whole word would not fit.
+ */
+function toPieces(segments: MentionSegment[]): OgTextPiece[] {
```
Orlandohub: Non-blocking, visual. "Only part at a row end when the whole word would not fit" is exactly the case that looks wrong: with `@Jeb` and `,` as separate flex items, a row that has room for `@Jeb` but not `@Jeb,` keeps the name and starts the next row with an orphan comma, where a text node would carry `@Jeb,` over as one word. Since satori lays out nested flex containers fine (the problem you hit was nested *inline* spans), a whitespace-free run of pieces could be grouped in a `display: flex` wrapper with no `flexWrap`, so the word moves as a unit and the brand colour still applies per piece inside it. Worth a quick render through the real satori with a mention landing near the row end before deciding; fine as a follow-up if it holds up.

src/libs/og/ogText.ts
```diff
@@ -0,0 +1,86 @@
...
+      return RGI_EMOJI_REGEX.test(qualified) ? qualified : grapheme.replaceAll(ZWJ, '');
+    })
+    .join('');
+}
+
+/** Emoji-repaired text, grapheme-truncated when `max` is given (names, titles, descriptions). */
+export function prepareOgText(text: string, max?: number): string {
+  const repaired = repairEmojiSequences(text);
+  return max === undefined ? repaired : truncateByGraphemes(repaired, max);
+}
```
Orlandohub, line 76: Nit: `prepareOgTextSegments` truncates before repairing precisely so a long body is not segmented past the cut, but this string variant repairs the whole text first. For `renderCollectionOg`'s description (and any other long user string with a ZWJ in it) that is a full `splitGraphemes` pass over the untruncated text. Either apply the same order here or implement it as `prepareOgTextSegments([{ text, isMention: false }], max)` when `max` is given, so the two helpers cannot drift.

src/libs/post/postMetadata.ts
```diff
@@ -48,3 +54,88 @@ export async function fetchUserAndPostForMetadata(
...
+      await res.body?.cancel();
+      if (res.status !== HttpStatusCode.NOT_FOUND) {
+        Logger.warn('[postMetadata] Mention lookup failed; showing the shortened key', { pubky, status: res.status });
+      }
+      return null;
+    }
+    const user: NexusUserDetails = await res.json();
+    return user.name || null;
+  } catch (error) {
+    Logger.warn('[postMetadata] Mention lookup failed; showing the shortened key', { pubky, error });
```
Orlandohub: Nit on log volume: this warns once per failed lookup, and one crawler hit on a mention-heavy post runs the resolver on three surfaces (`generateMetadata`, `opengraph-image`, `twitter-image`), so a Nexus wobble can print up to 30 identical lines for a single page view. One aggregated warn per `resolveMentionSegmentsForMetadata` call (`{ failed: [...pubkys], statuses }`) would carry the same information. Not Sentry-relevant since `Logger.warn` is console-only, so purely a server-log hygiene point.

src/components/organisms/NotificationItem/NotificationItem.helpers.ts
```diff
@@ -1,10 +1,8 @@
 import { UserController } from '@/controllers/user/user';
 import { Identity } from '@/libs/identity/identity';
+import { MENTION_IN_TEXT_REGEX as mentionInTextRegex } from '@/libs/identity/identity.constants';
```
Orlandohub: Nit: the alias only preserves the old local name for one call site. Import `MENTION_IN_TEXT_REGEX` directly and use it in `resolvePubkyToNames`, matching `PostText.utils.ts`.

talosmachina:
No action needed from me here. I don't approve pull requests, so the merge gate still needs a human approval - @Orlandohub's review is what unblocks it.

One thing for whoever approves: the branch has merged dev since my review (head is now `e037663`, my review covered `7de508c`), so the current head carries that merge on top of the tree I looked at. I found nothing blocking in what I reviewed.

src/libs/identity/identity.constants.ts
secondl1ght: Done in d6e2311: `PUBKY_IDENTIFIER_WITH_PREFIX_SOURCE` is a module-local `const` now, so the file's only export is `MENTION_IN_TEXT_REGEX`.

---

src/libs/post/postPreview.ts
secondl1ght: Done in d6e2311: `isMentionResolvablePreview` moved below `deriveTextPreview`, so each function has its own JSDoc again.

---

src/libs/og/OgComponents.tsx
secondl1ght: Rendered it through the real satori and applied a bounded version in d6e2311. A word that holds a mention plus glued text (`@Jeb,`, `(@Jeb)`) is now one non-wrapping `display: flex` item, so at the row boundary it moves down as a unit instead of leaving the comma orphaned. The bound: only words up to 20 characters are grouped. A grouped word wider than a row is the case that broke earlier (satori shrinks the text inside a non-wrapping row and splits the mention mid-word, `@Bo` / `b`), and 20 characters is ~600px at the largest card size, well inside every text block including the bio column. Longer glued words such as `@Bob,https://…` keep flowing piece by piece, which renders cleanly.

---

src/libs/og/ogText.ts
secondl1ght: Done in d6e2311: with a `max`, `prepareOgText` now goes through `prepareOgTextSegments([{ text, isMention: false }], max)`, so both helpers share the truncate-then-repair order and cannot drift.

---

src/libs/post/postMetadata.ts
secondl1ght: Done in d6e2311: the lookup returns `{ name, failure }` and the resolver emits one `Logger.warn` per call with `{ failed: { <pubky>: 'HTTP 503' | 'TimeoutError' | … } }`; 404s stay silent.

---

src/components/organisms/NotificationItem/NotificationItem.helpers.ts
secondl1ght: Done in d6e2311: the alias is gone and `MENTION_IN_TEXT_REGEX` is used directly, matching `PostText.utils.ts`.

---

Orlandohub (approved):

Orlandohub:
post-1-og-card

post-1-page

PR #2572, opened 2026-09-17
fix(posts): stop a failed reply from destroying the draft

talosmachina:
# What

A reply that fails to reach the homeserver no longer destroys the text the user typed.

# Why

Reported by @BitcoinErrorLog: replying returned `Could not post reply. Try again.` and the draft was gone.

`ThreadTree` rendered the quick reply from two structurally different trees: `<Container><PostThreadSpacer/><QuickReply/></Container>` when the parent had no replies, and `<Container role="feed"><><PostThreadSpacer/><QuickReply connectorVariant="regular"/></>...</Container>` once it had some.

The reply commit is local-first: `PostApplication.commitCreate` writes the post and its `post_replies:<parent>` stream entry to Dexie, then PUTs to the homeserver, and deletes the local row when that PUT fails. `useReplyStream` reads the reply list from Dexie through `useLiveQuery`, so the first reply of a thread flips `replyIds` 0 -> 1 -> 0. Each flip swaps the tree shape, React unmounts the `QuickReply` that holds the draft and mounts an empty one, and `usePost` keeps composer content in `useState`, so nothing survives. The result was a failed write, a rolled-back reply and unrecoverable text.

The homeserver failure itself is a separate and older problem (the pkarr resolution error behind that report, also the source of long-running `No HTTPS endpoints found in PKARR record` groups). This change only stops such a failure from eating the draft.

# How

One tree for both states, so the composer subtree keeps its identity across the flip. The feed role and its j/k navigation still apply only once reply cards exist, the connector variant still follows the same rule as before (terminal when the thread is empty, regular once replies exist), and the early `null` for "no replies and no composer" is unchanged.

# Verified

- New test in `ThreadTree.test.tsx` counts `QuickReply` mounts across the flip. It fails against `dev` with `AssertionError: expected 2 to be 1` on the first reply (the optimistic write remounted the composer) and passes with the change.
- `ThreadTree`, `QuickReply`, `PostReplies` and `SinglePostContent` suites: 82 pass, snapshots unchanged.
- `npm run lint` clean, `npm run typecheck` clean against a lockfile-pinned `npm ci`.

# Not verified yet

- No browser end-to-end. Reproducing the original failure needs a homeserver write to fail, so the fix is verified at the component level, not by a screenshot of a failed reply keeping its text.
- The pkarr resolution failure that caused the failed write is untouched.

Requested by @BitcoinErrorLog.


infin1t3 (approved):
Reviewed `488ecc1d`; no actionable issues found.

The unified tree preserves the QuickReply instance across the first reply's optimistic insert and rollback while retaining the existing connector, feed-role and keyboard-navigation behavior. A focused probe using the real QuickReply, usePostInput and usePost confirmed that the base loses the draft, while this change preserves text, tags and the attachment after failure. Retrying submits the same payload and clears the editor only after success.

Validation: 170 existing tests and both additional probe cases passed, plus full lint, typecheck and formatting checks on the changed files. CI is green. The integration probe simulated the commit failure; it was not a live homeserver/browser E2E test.

This fixes draft loss. The underlying PKARR write failure remains a separate issue.

PR #2576, opened 2026-09-18
fix(posts): label the article edit submit button "Publish"

### LINKED ISSUE #2553: Article editor: submit button reads 'Edit' when editing an article
## What happens

Editing an article, the composer's submit control reads **Edit**; creating one reads **Publish**. Reported by @piotr-iohk after the 1.9.0 staging QA pass: "the save control still says _Edit_ (create says _Publish_). Easy to think it does nothing."

## Repro

1. Open your own article and enter edit mode.
2. The submit control in the composer reads "Edit".
3. Creating an article shows "Publish" instead.

## Expected

A label that reads as the save action. "Save changes" (what the collection edit dialog uses) or "Publish" both work; the owner's call.

## Where

- `src/components/organisms/PostInputExpandableSection/PostInputExpandableSection.utils.ts` - `getButtonLabel()` returns "Publish" only when `isArticle && variant !== EDIT`, so the edit variant falls through to the table below.
- `src/components/organisms/PostInput/PostInput.constants.ts` - `POST_INPUT_BUTTON_LABEL[EDIT] = 'Edit'`.

talosmachina:
Closes #2553

Requested by @infin1t3 via the talos CLI.

## Behavior

Editing an article, the composer's submit control now reads **Publish**, the same as creating one. Reported by @piotr-iohk after the 1.9.0 staging QA pass: the "Edit" label reads as if the control does nothing.

`getButtonLabel()` in `PostInputExpandableSection.utils.ts` returned "Publish" for the article sub-mode of every variant except `EDIT`, so the edit variant fell through to `POST_INPUT_BUTTON_LABEL[EDIT]` = "Edit". It now returns "Publish" for any article sub-mode.

Non-article behavior is untouched: editing a regular post still reads "Edit", and the reply, repost and post labels are unchanged.

## Verification

- `npm test -- src/components/organisms/PostInputExpandableSection src/components/organisms/PostInput/PostInput.test.tsx src/components/organisms/DialogEditPost`: 146 passed, 5 files.
- Test first: the updated unit case failed on unmodified `dev` with `AssertionError: expected 'Edit' to be 'Publish'` before the source change.
- New coverage: `getButtonLabel(POST_INPUT_VARIANT.EDIT, true)` returns "Publish" (unit), and the label handed to the action bar for `submitMode=EDIT` with `isArticle` true ("Publish") and false ("Edit") (component).
- `npx eslint` and `npx prettier --check` on the three changed files, and `npx tsc --noEmit`: clean.
- VRT (`src/test/vrt/post/Article.vrt.test.tsx`): ran chromium and firefox. It fails 4 frames, and they are **pre-existing failures in this environment, not caused by this change**: the same 4 frames fail with identical pixel counts (10614 / 10348 / 6930 / 6340) on unmodified `dev`, and the actual screenshots are byte-identical with and without the change (md5). The edit-article frames do not capture the submit button: the dialog is taller than the viewport in that fixture, so the composer's action bar is below the fold. No baseline was committed; baselines are CI-owned. The frames worth re-checking in CI are `article-editing-desktop-*` and `article-editing-mobile-*`.

## Limitations

- WebKit VRT could not run here: its Playwright system libraries need root (`npx playwright install-deps`), so only chromium and firefox were exercised.
- Cypress e2e was not run; it needs the full pubky-stack and is not runnable from a bare checkout.


infin1t3 (approved):
@talosmachina Review OK: no actionable findings in 29b641d725aff7eb48f66ff1cd5674ffb3999dd5, compared with dev at 57a240dd9d399b3013745c144f994c2374eaa6af.

I reviewed all three changed files and traced the article-edit flag from DialogEditPost through PostInput to the action bar. The change gives article edits the requested Publish label while retaining the non-article Edit label, other variants, and explicit submit-label overrides.

Local verification: `npm test -- src/components/organisms/PostInputExpandableSection src/components/organisms/PostInput/PostInput.test.tsx src/components/organisms/DialogEditPost` passed all 146 tests in 5 files; `git diff --check` passed. GitHub checks still pending at submission: Run Shard Tests (3), Run Shard Tests (5), Build Preview Image / Build and push image.

I did not rerun VRT locally. The PR reports four Chromium/Firefox frames failing identically on the base and this branch, plus unavailable WebKit system dependencies; that visual evidence is author-reported, not independently reproduced in this review.

Keep this PR in draft. Do not merge, enable auto-merge, or add it to a merge queue.

PR #2578, opened 2026-09-18
fix(feed-nav): keep the tab strip scroll offset across feed route remounts

### LINKED ISSUE #2442: [Bug]: Custom feed tabs that are scrolled to right reset to left when selected
### Description

When you have many custom feeds, the tab bar becomes scrollable. If you then scroll right to select a tab it loads the feed but also resets the tabs back to far-left. This happens on both desktop and mobile but it's more annoying on mobile due to reduced horizontal space and the need to tap twice for edit.

### Expected Behaviour

Tab bar remains in the same position when new feed loads.

### Steps to Reproduce

1. Create many custom feeds
2. Scroll the feeds tabs to the right
3. Select a tab on the right
4. Observe the page reloads the tab bar at leftmost position - so to select the same tab to edit required scrolling right again

### Screenshots

https://github.com/user-attachments/assets/0c4cb797-9ee7-4bbc-af5c-dc0d18a0e4d6

### Console Log

```shell

```

### Browser & Version

_No response_

talosmachina:
Closes #2442

Requested by @infin1t3 via the talos CLI.

## Problem

`/home` and `/feed/[id]` render different templates, so the tab strip in `FeedNavigation` unmounts on every feed switch. A fresh DOM node starts at `scrollLeft: 0`, so picking a tab that the user had scrolled to on the right loaded the feed and snapped the strip back to the first tab. On mobile the tab then has to be scrolled to again to reach the edit affordance.

## Change

`FeedNavigation.tsx`, one file:

- The strip's horizontal offset is kept in module state next to the existing `cachedFeeds`, together with the pubky it belongs to.
- A `useLayoutEffect` re-applies the offset once the tabs are in the DOM, keyed on the account, the live feeds and the pathname, so it also covers the first authenticated render where the feeds arrive from Dexie after mount and the route change that re-lays out the strip (the active tab has a different width).
- Assigning `scrollLeft` while the row is still empty, or beyond what the row can now reach, is clamped by the browser: the effect settles on the browser's own maximum instead of resetting the strip.
- A different account, or a signed-out render, resets the saved offset to the first tab.
- Only the horizontal offset is touched; vertical feed scrolling (`handleFeedNavClick` and the `FORCE_FEED_SCROLL_TOP_KEY` one-shot flag in `(feeds)/layout.tsx`) and direct navigation are untouched.

## Verification

- Unit (`npx vitest run --project unit src/components/organisms/FeedNavigation/FeedNavigation.test.tsx`): 44 passed. Three new cases: offset survives a remount; a second account starts at the first tab; the offset is dropped when the session ends. The remount case fails on `origin/dev` (0 instead of 180), so it is a real regression guard.
- Real browser (chromium, vitest browser mode, throwaway harness, not committed): a genuinely overflowing row (30 custom feeds).
  - mobile 390px: row `scrollWidth` 1595 / `clientWidth` 422; scrolled to 220, remounted, `scrollLeft` 220. Only the always-present "All" tab existed on first paint, so the restore really happens after the tabs arrive.
  - desktop 1280px: max scroll 302; scrolled to 300, remounted, `scrollLeft` 300.
  - clamping: saved 1173 at 390px, remounted at 1280px where the maximum is 302, `scrollLeft` 302 (clamped to the reachable maximum, not reset to 0).
  - Same harness on `origin/dev`: 0 in all three cases, which is the reported bug.
- `npx tsc --noEmit`: clean. `npx eslint` and `npx prettier --check` on both changed files: clean.
- VRT `src/test/vrt/feed/Home.vrt.test.tsx` (chromium, committed baselines, no baseline written): 13 passed, 4 failed. The 4 failures are the `article mode in the feed composer` and `new article dialog` frames and they fail identically on unmodified `origin/dev` in this environment, so they are pre-existing drift, not this change. The frames containing the tab strip pass.
- Not run: Cypress e2e (out of bounds), webkit VRT (cannot launch on this box).

## Limitations

- The saved offset lives in module state, so it survives route remounts within a session but not a page reload. That matches the reported flow; a reload starts at the first tab.
- If the strip gets narrower between two mounts (for example the previously active tab was the wide one), the restore lands on the browser's maximum instead of the exact previous offset.
- Verified in chromium only. No real-device check on iOS Safari, which is where the reported flow is most annoying.


src/components/organisms/FeedNavigation/FeedNavigation.tsx
```diff
@@ -98,6 +111,31 @@ export const FeedNavigation = ({ className }: FeedNavigationProps) => {
...
+  // the tabs are in the DOM: assigning `scrollLeft` while the row is still empty
+  // is clamped by the browser to 0, so this re-runs on every tab/route change and
+  // settles on the browser's own maximum when the saved offset no longer fits
+  // (the previous active tab was the wide one, or the account has fewer feeds).
+  useLayoutEffect(() => {
+    const row = tabStripRef.current;
+    if (!row) return;
+    if (cachedTabStripScrollLeft.pubky !== currentUserPubky) {
+      // Different account (or a signed-out render): start from the first tab.
+      cachedTabStripScrollLeft = { pubky: currentUserPubky, left: 0 };
```
infin1t3: [P3] Invalidate the saved offset across an unmounted logout

Scroll the feed strip, navigate to Profile or Settings (which unmounts FeedNavigation), sign out there, then choose Sign back in and return as the same account. The logout/sign-in routes do not mount this component, so this effect never observes currentUserPubky === null. On the next feed mount the pubky matches the old cache and restores the previous session's offset, instead of starting at the first tab as the accepted direction and new logout test require. A Chromium probe with the actual component and CSS restores 220px here; the equivalent base component starts at 0.

The new test masks this case by mounting FeedNavigation while signed out. Bind the cache to the actual auth session lifetime, or clear it through a reset that still runs when this component is absent, and cover logout/relogin with no intermediate navigation mount. Existing captureViewerSession already distinguishes sessions using both pubky and the session reference. Preserve normal same-session route restoration.

infin1t3 (changes requested):
@talosmachina One confirmed P3 finding on d5a5706420319b636a628de2c62d5144a5f8a61f against dev 57a240dd9d399b3013745c144f994c2374eaa6af: the new cache survives logout/relogin as the same account when FeedNavigation is unmounted during the auth transition. Please address the inline finding and add coverage for that lifecycle.

Reviewed both changed files, the Container ref/scroll forwarding, Home/custom-feed consumers, auth reset, logout navigation and vertical-scroll handling. The existing FeedNavigation suite passes all 44 tests. An independent Chromium probe confirms ordinary remount restoration and browser clamping, but fails the unmounted logout/relogin case (220px instead of 0); that case passes against the base component. The probe mocks route/auth/data inputs and renders the real component, Container and styles; it is not a full authenticated browser test. GitHub build, code quality, test shards and preview deployment are green. No local full VRT or iOS-device run.

Keep this PR draft. No merge, auto-merge, merge queue or ready-for-review action is authorized.

src/components/organisms/FeedNavigation/FeedNavigation.tsx
talosmachina: @infin1t3 Fixed in 9b79ebe8. The saved offset is now bound to the sign-in that produced it, via `captureViewerSession()` (pubky plus the session reference) instead of the pubky alone, so an offset recorded before a logout is dropped on the next mount even when this component was unmounted for the whole auth transition. Added `drops the saved offset when the session ends while the strip is unmounted`, which signs out and back in as the same account with nothing rendering here in between: it fails against d5a5706 and passes on 9b79ebe8 (45 tests in the file). Pushed to the draft branch only, no ready-for-review or merge action.

talosmachina: Adding the checks to 9b79ebe8, the fix pushed above.

Real browser probe (chromium via vitest browser mode, the real component plus `Container` and the app CSS, throwaway harness not committed), four cases, all passing on 9b79ebe8:

- mobile 390: strip scrolled to 220, remounted: 220 restored, with only the always-present All tab rendered on first paint, so the restore happens after the tabs arrive
- desktop 1280: scrolled 300, remounted: 300
- clamp: saved 1173 at 390, remounted at 1280 where the maximum is 302: lands on 302, not 0
- unmounted logout then sign-in as the same pubky, nothing rendering here in between: 0

Same probe on the previous head d5a5706: the unmounted case returns 220, the other three pass. On origin/dev the three restore and clamp cases return 0.

Other checks: `tsc --noEmit` clean; eslint and prettier clean on both files; `FeedNavigation.test.tsx` 45 passed, with the new unmounted-logout test failing on d5a5706 (180 vs 0). `Home.vrt.test.tsx` in chromium against the committed baselines, no baseline written: 13 passed, 4 failed on the article-mode composer and new-article-dialog frames, failing identically on unmodified origin/dev on this box, so pre-existing drift. Not run: Cypress e2e, webkit VRT (cannot launch here). PR stays draft, no merge action.

Requested by @infin1t3 via the talos CLI

infin1t3 (approved):
@talosmachina Review OK on 9b79ebe84095e8ff34a23bd967485bb7557ca429 against dev 57a240dd9d399b3013745c144f994c2374eaa6af. The P3 finding from my previous review is fixed and independently verified. No actionable findings remain in the reviewed scope.

The saved offset now validates the captured account and session reference before restoration, including logout/relogin while FeedNavigation is absent. I reviewed the complete correction and its regression test. All 45 FeedNavigation unit tests pass. My Chromium probe now passes both scenarios: ordinary route-remount restoration with browser clamping, and logout/relogin with no intermediate navigation mount (previously 220px, now the expected 0). The probe renders the real component, Container and styles with mocked auth/route/data inputs.

No local full VRT or real-device iOS test was run; the PR's reported VRT environment/base limitations remain. Git diff --check passes. GitHub build, code quality, unit tests and preview checks are green for this head.

Approved. The user has now authorized marking #2578 ready and merging it through the repository merge queue. The reviewed head is unchanged and all GitHub checks are green.

PR #2579, opened 2026-09-18
fix(posts): unblock the visual feed once missing file metadata settles

### LINKED ISSUE #2568: Visual feed stays on its skeleton forever when every first-page attachment is missing from Nexus
## Summary

`VisualTimelinePosts` treats `hasPendingFiles` as "still resolving" in `isResolvingInitialRows` (and in `showFilteredEmptyState`). `useVisualFeedTiles` sets that flag while any attachment has no local file row, and `FileApplication.persistFiles` writes only what Nexus returns, so an attachment Nexus no longer serves keeps the flag set for the life of the feed.

When the first page produces no rows because every one of its posts has such an attachment, and the stream has no more pages (`hasMore === false`), the feed renders `VisualTimelinePostsSkeleton` indefinitely: no rows, no empty state, no Load more. With `hasMore === true` the backfill effect keeps chaining instead (#2538).

This is pre-existing behaviour on `dev` (the gate predates #2552 and #2552 leaves it untouched); noted while reviewing that PR, where the same flag was found to freeze the observer if it is used as a loading signal.

## Reproduction

1. A Visual feed whose entire first page is image posts whose file rows Nexus no longer returns (for example a profile whose uploads were purged), with fewer posts than one page.
2. `useVisualFeedTiles` yields `tiles: []` and a non-empty `missingFileUris`; `fetchFiles` never brings them back.
3. `isResolvingInitialRows` stays `true` → skeleton forever.

## Suggested fix

Either persist a not-found marker for file fetches (mirroring `settledDetailPostIds` for post details) so `hasPendingFiles` can settle, or stop treating `hasPendingFiles` as a loading signal once the fetch for those URIs has settled.

talosmachina:
## Summary

Closes #2568. Requested by @infin1t3 via the talos CLI.

`VisualTimelinePosts` read `hasPendingFiles` as "still resolving" while any attachment had no local file row. Nexus omits files it no longer serves instead of marking them missing, and `FileApplication.persistFiles` writes only what Nexus returns, so such an attachment kept the flag set for the life of the feed: with no rows and no further pages the feed stayed on `VisualTimelinePostsSkeleton` indefinitely, and with more pages the backfill effect kept chaining.

`useVisualFeedTiles` now separates pending file metadata requests from settled missing metadata, following the existing `settledDetailPostIds` pattern:

- `settledFileUris` records file URIs whose metadata request has settled (success or failure), the file equivalent of the settled post-detail ids.
- `hasPendingFiles` counts only URIs whose request has not settled yet, so a request that comes back without a row opens the loading and empty-state gates instead of holding them.
- An in-flight set keeps one request per URI however often the snapshot re-emits, and a settled URI is never re-requested: no request loops.
- Settlement does not remove the URI from the snapshot's `missingFileUris`, so a row that arrives later still builds its tile, and the post stays out of `hiddenPostCount` while its attachment is unresolved.

## Verification

- `npx vitest run src/components/organisms/Timeline/Feed/TimelineFeed/useVisualFeedTiles.test.tsx` — 19 passed.
- New regression tests, each failing on the unpatched hook (proved with `git stash push -- .../useVisualFeedTiles.ts`, 5 failed there): the flag stays set while the request is in flight and clears when it settles without a row; a rejected request settles rather than gating the feed forever; each missing URI is requested exactly once across snapshot re-emissions; a tile whose metadata arrives after settlement still renders.
- `npx vitest run src/components/organisms/Timeline/Feed` — 8 files, 228 passed.
- `npm test` (unit project) — 794 files, 13819 passed, 2 expected fail, 2 skipped.
- `npx eslint` on both changed files clean, `npm run format:check` clean, `npm run typecheck` (`tsc --noEmit`) clean; the husky pre-commit hook (lint-staged + tsc) passed on the commit.
- VRT: `src/test/vrt/feed/Collections.vrt.test.tsx` — 2 files, 16 passed (chromium and firefox). That suite mocks `useVisualFeedTiles`, so it cannot exercise the changed path; it is a control only. No baselines were regenerated or committed.

## Callouts

- One existing assertion changed: "counts remote posts only once every attachment resolved to non-visual metadata" now holds its `fetchFiles` promise open so the in-flight window is observable, then asserts the flag clears once the request settles. Its subject, `hiddenPostCount`, is unchanged.
- Settling on failure matches `settledDetailPostIds` ("success OR failure"): a transient failure unblocks the gates instead of holding the skeleton, and that URI is not re-requested until the hook remounts. Say the word if failures should stay pending instead.
- `settledFileUris` accumulates for the life of the hook instance, like `settledDetailPostIds`; it is not reset on feed or account change.
- No DB tombstones or version changes, no global config changes, no change to which content renders.
- Environment limit: WebKit cannot launch on this machine (missing system libraries, `playwright install-deps` needs root), so the VRT webkit instances did not run; the run above was limited to chromium and firefox with a temporary config edit that was reverted. Cypress e2e was not run.
- An adversarial self-review of this diff was in flight when the run budget ended; its findings are not reflected here yet and will follow as review replies.


infin1t3 (approved):
@talosmachina Review OK: no confirmed actionable findings in e4fec53c2de50e9f1d6bba360041460a060715eb compared with dev 57a240dd9d399b3013745c144f994c2374eaa6af.

Reviewed both changed files and traced the file fetch/persistence contract, pending and empty-state consumers, existing retry/TTL paths, and hook lifetime. The in-flight set prevents overlapping requests for the same URI; settlement releases the loading gate without fabricating file metadata or changing the snapshot's hidden-post count. A later metadata row remains observable through the live query and can produce its tile.

Independent local verification: all 228 tests in 8 Timeline/Feed files pass, including the real hook's pending, omitted-row, failed-request, deduplication and later-arrival cases. Git diff --check passes. GitHub build, code quality, unit tests and preview deployment are green for this head.

The documented behavior remains relevant: after the service's retries finish, this hook does not automatically re-request a settled URI during the same mount, including a failure. Later metadata can arrive through another fetch/TTL path, or a remount can retry. I did not rerun the full unit suite or VRT locally; the author-reported Collections VRT uses a mock for this hook and is not proof of the changed metadata path.

Approved for the reviewed scope. The user has now authorized marking #2579 ready and merging it through the repository merge queue. The reviewed head is unchanged and all GitHub checks are green.

PR #2580, opened 2026-09-18
fix(posts): complete mentions at the caret, not only at the end of the value

### LINKED ISSUE #1959: [Bug]: Can only mention profile in post when cursor is at end of text
### Description

When typing `@profilename` we see a lookup to help insert their pubky to reference them. This is only possible when the keyboard cursor is at the end of all text. It doesn't work when moving cursor back within existing text.

### Expected Behaviour

Can lookup profile name to reference anywhere in the text where there is whitespace.

### Steps to Reproduce

1. Go to create a new post
2. Type some text
3. Place keyboard cursor somewhere in the text that's not at the end
4. Type someones username with `@` at beginning like `@username`
5. Observe popover does not display

### Screenshots

https://github.com/user-attachments/assets/60aef0c8-e65f-4f8d-bbed-a1b92e403cac

### Console Log

```shell

```

### Browser & Version

_No response_

talosmachina:
## What changed

Mention completion was anchored to the end of the whole textarea value, so the `@username` / `pk:` / `pubky` popover only appeared while the caret was at the very end of the post. With the caret inside existing text, typing `@name` showed nothing (#1959).

Detection and insertion now follow the caret:

- `extractMentionQuery(content, caret)` scans only the text before the caret and reports the pattern range the caret sits in (the `pk:`/`pubky` pattern still wins over `@`). The search filters (minimum length, skip complete pubkeys) are unchanged.
- `getContentWithMention(content, caret, userId)` writes `pubky{userId} ` over that range, keeps the text after the caret verbatim, and returns the caret position to restore. With the caret at the end of the value the result is byte-identical to the previous behaviour.
- `useMentionAutocomplete` takes a required `caret` and re-runs detection when either the content or the caret changes.
- `usePostInput` tracks the caret from the textarea's change, select and key events, and restores it with `setSelectionRange` after the controlled re-render (the same pattern already used by `InitializedMDXEditor` and `useTagInput`).
- Wired into both composers: `PostInput` and `QuickReply` (new `onSelect` prop through `QuickReplyContentProps` / `QuickReplyComposerRow`).

Keyboard navigation, Escape, `pk:` / `pubky` completion and IME composition are untouched.

## Behaviour

- Caret mid-text, type `@jo`, pick a user: `Hello @jo world` becomes `Hello pubky<id>  world` with the caret just after the inserted mention and the rest of the text kept.
- Caret at the end of the value: unchanged (`Hello @jo` becomes `Hello pubky<id> `).
- Caret outside any mention pattern: nothing opens.
- Moving the caret back into an existing `@jo` re-runs detection.
- A selection that would push the post over `POST_MAX_CHARACTER_LENGTH` leaves the content untouched.

## Verification

- `npx vitest run --project unit src/hooks/useMentionAutocomplete`: 92 passed.
- `npx vitest run --project unit src/hooks/usePostInput/usePostInput.test.ts`: 133 passed, including 4 new cases (mention written over the pattern at the caret, suffix kept, caret restored to the computed offset, over-limit leaves the content alone).
- `npx vitest run --project unit` on `PostInput.test.tsx`, the `QuickReply` suites and both hook suites: 359 passed across 8 files.
- Regression proof: with `origin/dev` versions of `useMentionAutocomplete.utils.ts`, `useMentionAutocomplete.constants.ts` and `usePostInput.ts` checked out over this branch, the new suites fail (32 failures in the utils file, every caret case among them).
- `npm run lint` clean, `npm run typecheck` clean, `npx prettier --check` clean on all 17 changed files.
- VRT, chromium only: `npx vitest run --project vrt --browser.name=chromium src/test/vrt/feed/Home.vrt.test.tsx` gives 13 passed, 4 failed; a control run with unmodified `origin/dev` source fails the same 4 article-mode composer frames, so that drift is pre-existing in this environment and not from this diff. No baseline was touched.

## Limitations

- The configured VRT run also includes webkit and firefox instances; webkit cannot launch on this machine (missing system libraries, `playwright install-deps` needs root), so the VRT result above is chromium only.
- Cypress e2e was not run; it needs the full stack and is not runnable from a bare checkout.
- Inserting a mention immediately before existing whitespace keeps both the mention's trailing space and the existing one (`pubky<id>  world`); typing at the restored caret consumes one of them.

Closes #1959

Requested by @infin1t3 via the talos CLI


src/hooks/usePostInput/usePostInput.ts
```diff
@@ -186,7 +210,16 @@ export function usePostInput({
...
+  } = useMentionAutocomplete({ content, caret: mentionCaret, onSelect: handleMentionSelect });
+
+  /**
+   * Track the composer caret. Arrow keys, Home/End and clicks move it without a
+   * change event, and mention detection follows the caret (#1959)
+   */
+  const handleSelectionChange = useCallback((e: React.SyntheticEvent<HTMLTextAreaElement>) => {
+    const target = e.currentTarget;
+    setCaret(target.selectionStart ?? target.value.length);
+  }, []);
```
infin1t3: [P3] Keep the new selection handler free of manual memoization

This adds a new useCallback for handleSelectionChange, whose only work is calling the stable setCaret setter and whose consumers are React textarea onSelect/onKeyUp props. docs/development-workflow.md, Code Quality, says: "No manual memoization" and allows an exception only with a stated reason the compiler cannot handle it; AGENTS requires profiler evidence. There is no non-React subscriber or profiling justification here. Use a plain function for this new handler so the change follows the React Compiler convention without another manually maintained callback. Existing callbacks are outside this finding and do not need a cleanup.

infin1t3 (changes requested):
@talosmachina One P3 maintenance finding on 46072a31eb2b53fcc9a76678eef9d69f482ca05b against dev 57a240dd9d399b3013745c144f994c2374eaa6af: please remove the newly introduced manual useCallback on the textarea selection handler, as required by the repo's React Compiler rules. No behavioral regression was confirmed in this review.

Reviewed all 17 changed files, including caret/query extraction and insertion, the hook contract, PostInput and QuickReply event wiring, textarea forwarding, popover/listbox integration and tests. The hook/utils portion received an independent second review. My local run passes 346 tests in 7 files across the changed hooks, PostInput and QuickReply; the independent utility probe retained end-of-value insertion parity in 16 cases and preserved the suffix at every tested caret offset. Git diff --check passes. GitHub checks are green for this head.

Existing debounce/close and IME limitations were compared with the base and are not findings introduced by this PR. Full browser interaction/VRT was not rerun locally; the author's VRT baseline/environment limits remain.

Keep this PR draft. No merge, auto-merge, merge queue or ready-for-review action is authorized.

src/hooks/usePostInput/usePostInput.ts
talosmachina: Fixed in 35ab5162: `handleSelectionChange` is now a plain function. The pre-existing callbacks in the file are untouched and the caret behaviour is unchanged. Re-verified on that head: 359 unit tests passed across the 8 affected suites (usePostInput, PostInput, QuickReply, useMentionAutocomplete), `npm run lint` and `npm run typecheck` clean, prettier clean on the changed file.

infin1t3 (approved):
Review OK for 35ab5162ec6e48f0535bd48e6844a72e0a15e7f7 against dev at 57a240dd9d399b3013745c144f994c2374eaa6af. No remaining confirmed findings.

The original P3 comment concerned the newly added useCallback around handleSelectionChange. Commit 35ab5162 replaces that wrapper with a plain function. handleMentionSelect and the other callbacks already existed at the base and were never part of the finding. The corresponding review thread is resolved.

Independent validation: all 359 unit tests across the eight affected suites passed on this head. I also ran 20 functional Chromium tests against the real PostInput and QuickReply components, usePostInput/useMentionAutocomplete hooks, and mention popover at desktop and mobile widths. All passed: mouse selection in the middle of existing text; preserved suffix, focus and caret; moving back into an existing mention; Escape; ArrowDown/Enter selection; pk: and pubky completion with Unicode/multiline suffixes; and rejected over-limit insertion without changing the draft.

The Chromium tests use deterministic search/profile/auth data and the repository browser harness. They do not establish live Nexus/homeserver behavior or native mobile/IME behavior, and no screenshot baseline was regenerated. The isolated review checkout has no production-code changes. GitHub checks, including the preview deployment, are green for this head.

Updated after the user's explicit authorization to mark this PR ready and merge through the repository's normal merge queue. Approval remains valid for this exact head.

PR #2581, opened 2026-09-18
[Locks] fix(#2457, #2459): request the BTC rate only when the lock dialog needs it

### LINKED ISSUE #2457: Locks: creator — do not request the BTC rate until the lock dialog needs it
Sub-issue of #2364. Follow-up to #2436 (review by @secondl1ght).

## Problem

`DialogLockContent` calls `useBtcRate()` in its component body, and `PostInput` mounts the dialog without a guard (unlike `DialogLocksAuth`, which is guarded by `lockServerPubky`). The hook's mount effect runs whether or not `open` is true.

So every composer — home feed, post detail, any reply box — sends a `/api/btc-rate` request on mount. This includes users with Locks disabled, who never see the Payment tab.

## Fix

Defer the rate request until it is needed: only when the dialog is open, or only when the Payment tab is selected. Guarding the dialog mount with `lockServerPubky` (matching `DialogLocksAuth`) also removes the request for users without Locks.

### LINKED ISSUE #2459: Locks: creator — validate the /api/btc-rate response and add a route test
Sub-issue of #2364. Follow-up to #2436 (review by @secondl1ght).

Two related gaps around the `/api/btc-rate` route. A field rename on the server would today reach the browser as `satUsd: undefined` and nothing would catch it.

## 1. `useBtcRate` trusts any 2xx body and has no unmount guard

```ts
.then((json) => setResult({ rate: { ...json, lastUpdatedAt: new Date(json.lastUpdatedAt) }, status: 'ready' }))
```

Any 2xx body becomes `status: 'ready'`. If the body has no `satUsd` (a proxy or edge JSON error page, a future response change), `DialogLockContent` computes `amountSats * undefined` = `NaN`, which is `!== null`, so the field shows `$NaN` instead of the "dollar value can't be shown" message that `status: 'failed'` gives. `lastUpdatedAt` becomes `Invalid Date` with no signal.

Fix: check the response shape (`satUsd` is a finite number, `lastUpdatedAt` parses to a valid date) before setting `status: 'ready'`; otherwise set `failed`.

The effect also has no `cancelled` flag, so the resolve can call `setResult` on an unmounted composer. `useLockFile` already does this correctly; copy that pattern.

## 2. The route has no test

Every other route under `src/app/api/` (`copyright`, `dev/signup-token`, `feedback`, `og-metadata`, `report`) has a sibling `route.test.ts`. `btc-rate` does not.

Add `route.test.ts` covering:
- the success shape: `lastUpdatedAt` serializes to a string the client re-hydrates with `new Date(...)`
- the `Cache-Control` header the CDN depends on
- the `handleApiError` path when `ExchangerateService` throws

Taewa:
Closes #2457
Closes #2459

https://github.com/user-attachments/assets/647213a2-523d-40ff-bab5-41beb513ddd9

## The problem

Opening the home feed sent a `/api/btc-rate` request, even for users who have Locks turned off.

Every post composer mounts the lock dialog, and the dialog asked for the BTC rate as soon as it mounted — not when it opened. The rate is only ever used to price a lock.

A second problem sat in the same hook: it accepted any 2xx body. If a field were renamed on the server, or a proxy returned an error page, `satUsd` would be `undefined` and the dialog would show `$NaN` instead of saying the dollar value is unavailable.

## What changed

The rate hook now takes an `enabled` flag, and the dialog passes its `open` state — so the request waits until the rate is actually on screen. The flag defaults to `true`, so the three call sites outside Locks are untouched.

The dialog moved inside the existing `lockServerPubky` guard next to `DialogLocksAuth`, which removes the request entirely for users without Locks.

A zod schema now checks the body before the hook reports `ready`. A body that is not a rate is logged, because the UI says "unavailable" either way and the log is the only place a server-side change would show up. A plain network failure stays quiet — that one is not worth an entry.

Two smaller fixes in the same effect: a late response no longer writes state after the caller unmounts or turns the hook off, and a retry after a failure clears the old error instead of showing it while the new request is in flight.

## Tests

`/api/btc-rate` was the only route under `src/app/api/` without a test. It has one now: the date survives serialization, the cache header the CDN needs is set, and both error paths answer the way `handleApiError` defines.

The hook test covers the `enabled` gate, every shape the schema rejects, the unmount guard, and both logging decisions. Each new test was checked by removing the code it covers and confirming it fails.

## How to check by hand

1. Open the home feed. There should be no `/api/btc-rate` request.
2. Open the composer and open the lock dialog. One request goes out.
3. Close and reopen the dialog. The price stays on screen while the new request is in flight.

## Notes for the reviewer

- Base is `epic/locks-v1`, not `dev`.
- The onboarding VRT mock for this hook returns the old flat shape, so `HumanBitcoinCard` snapshots a skeleton instead of a price. That is pre-existing and outside Locks, and fixing it means re-recording baselines — left for a separate issue.

https://claude.ai/code/session_01QJMKzD9mdRJ1AEM4TMS2zG


talosmachina (commented):
**No findings.** Defers the `/api/btc-rate` request until the lock dialog is open, moves the dialog under the same `lockServerPubky` guard as `DialogLocksAuth`, and validates the route response with zod before reporting `ready`. Reviewed `586d4f15`, full tier.

**Read in full**: `useSatUsdRate.ts`, `DialogLockContent.tsx`, the `PostInput.tsx` render tree, `src/app/api/btc-rate/route.ts`, `exchangerate.ts`
**Call sites traced**: every `useBtcRate` (3) and `DialogLockContent` (1), and the `lockSwitch` / `lockServerPubky` derivation in `usePostInputLock.ts`
**Ran locally**: `useSatUsdRate.test.ts` (15 passed), `DialogLockContent.test.tsx` (15 passed), `route.test.ts` (4 passed)

**CI**: red. Every failing file fails with `Failed to resolve import "@pubky/locks-sdk" from "src/core/services/locks/locks.ts"`, and `docs/locks.md` states the SDK is deliberately not on npm yet, so any run against `epic/locks-v1` fails there. Not this diff. The three new test files pass in CI (shard logs) and locally.

**Ruled out**
- *Removed SSR guard*: `if (typeof window === 'undefined') return;` was dead, `useEffect` does not run during SSR, so nothing starts fetching on the server in its place.
- *Schema rejects a valid response*: `ExchangerateService.getSatoshiUsdRate()` returns `{ satUsd: number, btcUsd: number, lastUpdatedAt: Date }`, the route serialises the Date to an ISO string, and the schema accepts that, so no real rate is turned into "unavailable".
- *Skeleton in the onboarding VRT*: the mock at `src/test/vrt/onboarding/Human.vrt.test.tsx:82` returns a flat `{ satUsd, btcUsd, lastUpdatedAt }`, so `HumanBitcoinCard` reads `rate?.satUsd` as `undefined` regardless of this diff. Pre-existing, as the PR body says.
- *Dialog gated behind `lockServerPubky`*: the switch that opens it is itself offered only when `lockServerPubky && paykitServerUrl` (`usePostInputLock.ts:225`), so the dialog was unreachable without a lock server, and `PostInput.lock.test.tsx` (14 tests) still passes.
- *Request per open*: the effect re-runs on `enabled` false to true, so every open re-requests, but the route is CDN-cached for 60s and the rate already fetched stays on screen while the new request is in flight (the new test asserts it).
- *Stale write after close*: the `cancelled` flag covers the resolve and the reject path both, because the schema branch returns a value rather than throwing, so the trailing `then` still sees the flag.

</details>

Merge confidence: 4/5, no findings, but CI gives no signal on this PR: it is red on the epic base branch because `@pubky/locks-sdk` is not installable there (`docs/locks.md`), not because of this diff.

Taewa:
@talosmachina

Thanks for the review. On the two points about CI and the base branch:

**`@pubky/locks-sdk` is not on npm yet, and that is expected.** The SDK is still built and installed locally while the Locks work is in progress, so any CI run that installs from the registry cannot resolve it. That is why the test shards fail on `Failed to resolve import "@pubky/locks-sdk"`. It is a property of the epic branch, not of this change. `docs/locks.md` records the same thing.

**This PR targets `epic/locks-v1`, not `dev`.** Everything under the Locks epic collects there first. We merge the epic into `dev` only once the whole feature is ready, and the SDK situation is resolved by then, so nothing here reaches `dev` in this state.

**One more CI failure you did not mention.** The Code Quality job fails separately, on Prettier, in three files this PR does not touch:

- `src/components/organisms/PostBody/PostBody.tsx`
- `src/components/organisms/ProfileUnlocked/ProfileUnlocked.tsx`
- `src/hooks/usePaykitSetupFlow/usePaykitSetupFlow.utils.test.ts`

I reproduced it locally with `npm run format:check`. They are already unformatted on `epic/locks-v1`, so Code Quality is red there for a reason unrelated to the SDK and unrelated to this diff. Worth a separate cleanup on the epic branch.

Locally, with the SDK installed, the full suite passes.

Since your review I pushed one more commit: the missing-rate notice now uses the error colour, the disabled Apply Lock button shows a not-allowed cursor instead of the plain arrow, and a new route test feeds the real response through the client schema, so the route and the hook cannot drift apart without a test failing.

Taewa:
@talosmachina 
Review again plz

talosmachina (commented):
**Re-review of `8f016718` after `586d4f15`.** 0 earlier findings (the first review posted none), 0 new. Adds the response-shape test, the destructive colour on the failed-rate notice, and the not-allowed cursor on the disabled Apply Lock.

Reviewed the diff since `586d4f15` (4 files, 25 insertions), full tier.

**Read in full**: `route.test.ts`, `route.ts`, `DialogLockContent.tsx`, `DialogLockContent.test.tsx`, `useSatUsdRate.ts`, `atoms/Button/Button.tsx`
**Call sites traced**: every `TBtcRateResult` (4, all inside the hook, so de-exporting it breaks no reader), `btcRateSchema` (the hook, plus the new route test)
**Ran locally**: `route.test.ts` + `DialogLockContent.test.tsx` at `8f016718` — 21 passed. `prettier --check` on the three files named in the thread reports `Code style issues found in 3 files`, confirmed at this head and not touched by this diff, so Code Quality is red for the epic branch's own formatting, as said.

**Ruled out**
- *Class conflict does not resolve*: `cn` is clsx + tailwind-merge and `className` is the last argument, so `disabled:pointer-events-auto` replaces the base's `disabled:pointer-events-none` rather than racing it in the cascade. The intent holds; a disabled `<button>` still fires no click.
- *The schema test is vacuous*: `GET` reads a mocked `HomegateController.getBtcRate` and returns it through `NextResponse.json`, so the test does exercise the real serialisation (Date to ISO string) against the client schema. A renamed field would fail it.
- *Removed `TBtcRateResult` export*: no importer outside the hook, and `tsc --noEmit` would catch one.

</details>

Merge confidence: 4/5, no findings, and the new route test covers the shape contract; CI is not green, but every failing job traces to the `epic/locks-v1` base (`@pubky/locks-sdk` unpublished, three pre-existing Prettier files), not to this diff.

secondl1ght (approved):
LGTM. Verified the diff at `8f016718` against `epic/locks-v1`:

- The rate request now waits for the dialog's `open` state, and the dialog only mounts under the existing `lockServerPubky` guard, so the home feed no longer hits `/api/btc-rate`.
- The zod schema turns a wrong-shape 2xx into the "unavailable" notice instead of `$NaN`, and the new route test feeds the real serialised response through it.
- A refetch that fails replaces the previously shown rate with the error state. That is the right contract for a price: current or error, never stale.

CI is red for reasons that predate this PR: the Prettier warnings name three files this diff does not touch, and the build/test failures are the unpublished `@pubky/locks-sdk` on the epic base.

---

PR #2582, opened 2026-09-18
fix(locks): show no Unlocked count in the sidebar when the read fails

### LINKED ISSUE #2402: [Bug]: Unlocked count in the profile sidebar shows 0 when the read fails
## Description
When the read of the reader's `/priv` fails, the Unlocked entry in the profile sidebar shows `0`. That number is wrong: the read failed, so we do not know the count, and `0` tells the user they have unlocked nothing.

`useUnlockedList` sets `items` to `[]` in its catch block and then returns `count: items.length`, so the count becomes `0` (`src/hooks/useUnlockedList/useUnlockedList.ts:54`). `ProfilePageContainer` only checks `isLoading` before passing the count to the sidebar (`src/components/organisms/ProfilePageContainer/ProfilePageContainer.tsx:106`), so the error case is never handled. The hook does expose `isError`; nothing reads it for the count.

The screen itself is correct — `ProfileUnlocked` shows an error message when `isError` is set. Only the sidebar count is wrong.

## Expected Behaviour

On error, the Unlocked entry in the sidebar shows its label with no number at all. A spinner would also be wrong, because the read has already finished.

## Steps to Reproduce

1. Sign in.
2. Make the `/priv` read fail — for example, block the homeserver request in DevTools, or go offline after loading the app.
3. Open your own profile.
4. The sidebar shows `0` next to Unlocked, while the Unlocked screen shows the error message.

---

Reported by @secondl1ght while reviewing #2301. Part of #1998.

Taewa:
Closes #2402

https://github.com/user-attachments/assets/e97b8ab2-fc4f-434a-af51-cf8e84893131

## Problem

When the read of the reader's `/priv` failed, the profile sidebar showed `0` next to Unlocked. The count came from the list that the failed read had emptied, so it told the user nothing was unlocked instead of that the count is unknown. The Unlocked screen itself was already correct — it showed an error message.

## Change

The count prop now accepts `null` as well, and the three states are distinct:

| State | Value | Sidebar |
| --- | --- | --- |
| Read in flight | `undefined` | spinner |
| Read failed | `null` | label, no number |
| Read settled | number | the number, `0` included |

`useUnlockedList` now clears its error flag when a read succeeds rather than when one starts, so a retry does not report the emptied list as a settled `0` while it is still in flight. During that window the Unlocked screen keeps its error message; there is no retry control on that screen today, so this is only reachable when the session object is replaced.

## Tests

- Sidebar renders the number, the spinner, and the label-only state.
- A settled `0` still renders as `0`, which is what separates it from a failed read.
- The hook keeps its error until a retry succeeds, and drops it when the session ends.


talosmachina (commented):
**No findings.** When the read of the reader's `/priv` fails, the sidebar now renders the Unlocked label with no number instead of `0`: `isError` maps to a `null` count (label only), `undefined` stays the spinner, and the hook clears `isError` on success so a retry is not reported as a settled `0`. Reviewed `7f38fe5`, full tier.

**Read in full**: `ProfilePageFilterBar.tsx`, `useUnlockedList.ts`, `useUnlockedList.types.ts`, `ProfilePageContainer.tsx`, `ProfilePageLayout.tsx`, `ProfilePageLayout.types.ts`, `ProfileUnlocked.tsx`, `UnlockedListProvider.tsx`
**Call sites traced**: `unlockedCount` (container -> layout -> filter bar) and `useUnlockedList` / `useUnlockedListContext` each have a single consumer, so the sidebar and the Unlocked screen are the only two surfaces with this count and the fix is complete rather than partial.
**Ran**: `vitest --project unit` on the three touched test files, 45 passed. Reverting `ProfilePageContainer.tsx`, `ProfilePageFilterBar.tsx` and `useUnlockedList.ts` to the base left 4 of the new tests failing, so they pin the behaviour instead of passing vacuously. `eslint` clean on the touched files; `tsc --noEmit` reports no error in any of them.
**CI**: red on every job, none of it from this diff. `epic/locks-v1` imports `@pubky/locks-sdk` in `src/core/services/locks/locks.ts` while its `package.json` declares no `@pubky` dependency at all, so the Next build and all five shards fail on module resolution; `prettier --check` also fails on `PostBody.tsx` and `ProfileUnlocked.tsx`, which I reproduced locally on files this PR does not touch. That is why the PR's own suites never ran in CI.

**Ruled out**
- *A retry keeps the error message instead of a spinner after the session object is replaced*: `hasResolved` stays true so `isLoading` is false, but that window is deliberate and stated in the description, and no retry control exists on the Unlocked screen.
- *A stale `isError` from a previous reader suppressing the count*: reachable only if `currentUserPubky` changes while the session stays non-null; logout nulls the session first, which takes the guard branch that clears both flags.
- *The whole row disappearing, not just the badge*: `showCount: false` skips the badge only, and the new test asserts the label still renders.

</details>

Merge confidence: 4/5, no findings and the change is covered by the three test files, but CI is red repo-wide on the epic branch (missing `@pubky/locks-sdk`), so nothing beyond those unit files could be verified there.

Taewa:
@talosmachina Your read of CI is correct, and here is the context behind it, so it does not need to count against this PR.

**`@pubky/locks-sdk` is not published to npm yet, and that is intentional.** The SDK is still being developed alongside the app, so it is built from the `locks` repository and installed into `node_modules` out of band. That is why `package.json` declares no `@pubky` dependency: there is nothing on the registry to declare yet. Any CI run installs from the registry, finds nothing, and every job that imports `src/core/services/locks/locks.ts` fails on module resolution. This is a property of `epic/locks-v1`, not of any single PR on top of it. `docs/locks.md` records the arrangement. When the SDK is published, the dependency goes into `package.json` and CI goes green on the epic branch.

**This PR targets `epic/locks-v1`, not `dev`.** All Locks work lands on the epic branch first. The epic merges into `dev` only when the feature is complete, and the SDK will be a normal published dependency by then, so nothing reaches `dev` in this state.

**The `prettier --check` failures are also pre-existing.** They are in files this PR does not touch, already unformatted on the epic branch, so they are a separate cleanup on `epic/locks-v1`.

Locally, with the SDK installed, the full suite passes.

Could you re-review with that in mind, and judge the change on its own contents rather than on the red CI? If your merge confidence was held at 4/5 only because CI could not verify anything, this context should settle that part.

talosmachina:
@Taewa Thanks, and that matches what my review already said: the red CI is `epic/locks-v1`'s, not this diff's (`@pubky/locks-sdk` absent from `package.json`, so module resolution fails; `prettier --check` failing on `PostBody.tsx` and `ProfileUnlocked.tsx`, which this PR does not touch).

One correction: the 4/5 was not held because CI could not in principle verify the change. It is that on this branch no build or e2e job runs at all, so the three unit files are the only executable evidence, and an intentional arrangement does not change that. That is a statement about what is verified, not a doubt about the change, which had no findings.

The head has not moved since my review (`7f38fe5`), and I do not post a second review at the same head. The review stands as written. If you push, ask again and I will review the new head.

src/components/organisms/ProfilePageContainer/ProfilePageContainer.test.tsx
```diff
@@ -506,3 +516,47 @@ describe('ProfilePageContainer - User not found', () => {
     expect(screen.getByTestId('profile-page-layout')).toBeInTheDocument();
   });
 });
+
+describe('ProfilePageContainer - Unlocked count', () => {
+  // The global afterEach clears calls but not return values, so without this the last test's error
+  // state would leak into whatever is added below this describe.
+  afterEach(() => {
+    vi.mocked(useUnlockedList).mockReturnValue(UNLOCKED_LIST_SETTLED);
+  });
```
secondl1ght, line 525: These three tests render the layout only because of mocks leaked from the previous describe. `vi.clearAllMocks()` (the global `afterEach` and the `beforeEach` in "User not found") clears call history but keeps `mockReturnValue` / `mockImplementation`, so this block inherits the last test's `isOwnProfile: false`, `userNotFound: true` and `isLoggingOut: true`. The not-found view is suppressed only by the leaked `isLoggingOut: true`; reorder or append a test in the previous block and all three fail with `Unable to find [data-testid="profile-page-layout"]`.

Could you add a `beforeEach` here that resets `useProfileContext`, `useProfileHeader` and `useAuthStore` to their defaults, the same way the `afterEach` below already does for `useUnlockedList`? Then the tests prove the own-profile wiring they describe.

---

secondl1ght (approved):
Approving. The production change is right: the tri-state mapping keeps a settled `0` as `0`, the error flag is re-established on success and on the signed-out branch, and an account switch goes through `reset()`, so no stale flag survives.

One inline ask on the new container test, which passes today only through mocks leaked from the previous describe.

Non-blocking, for whoever wires a Retry control later: `hasResolved` is not reset when a read starts, so an in-flight retry would render as the error state rather than the spinner. Not reachable today, since `session` is only replaced by `init` and `reset`.

CI is red for the epic's reasons, not this PR's: Prettier on three files this diff does not touch, and the unresolved locks-sdk import failing three test files to load. Zero failing tests.

---

src/components/organisms/ProfilePageContainer/ProfilePageContainer.test.tsx
Taewa: Code updated, thanks!

PR #2583, opened 2026-09-18
test(e2e): count the composer limit in utf-16 units like the app

secondl1ght:
## Summary

The `e2e-test` workflow has failed on every push to `dev` since #2564 landed on 2026-09-16 (fourteen runs, the same assertion each time):

```
posts › can post with emojis
AssertionError: expected '44/2000' to equal '42/2000'
    at Context.eval (webpack:///./support/posts.ts:52:38)
```

#2564 deliberately made the composer counter measure the length the limit enforces (UTF-16 units, `getEnforcedCharacterCount`), so an emoji now counts as two. The `createQuickPost` and `createQuickPostWithImage` helpers still derived the expected counter text from the code-point length (`Array.from(postContent).length`), while `createPostFromDialog` already used `postContent.length`.

- Align the two quick-post helpers with `createPostFromDialog` (`postContent.length`) and say why in the comment.
- Drop the unused code-point constant in the emoji test.

Unblocks cutting the next release from `dev` (`docs/release.md` requires e2e green on the commits being shipped).

## Verification

- [x] `npm run typecheck` (root `tsconfig` includes `cypress/**`) and `prettier --check` on both files
- [ ] `npm run lint`: ESLint ignores `cypress/**`, nothing to run
- [ ] `npm test`: no unit test touched
- [ ] `npm run test:e2e`: not run, it needs the full pubky-stack and the workflow could not be dispatched from this session (403). It does not run on pull requests either. To see it green before merging, dispatch **e2e-test** on this branch from the Actions tab; otherwise it runs on the push to `dev` after merge.

## Callouts

- Cypress specs are QA-owned (`AGENTS.md`). This edit was requested by the release owner to unblock the release, and QA is being informed separately.
- No VRT surface, ADR, `DB_VERSION`, `src/config/*` limit or convention change.

PR #2577, opened 2026-09-18
fix(posts): prompt sign-in when a write fails on an expired session

talosmachina:
Closes #2555

## Summary

A write the homeserver rejects as unauthenticated was reported as a retryable failure, so the user retried a write that cannot succeed until the session is re-established. The classification already exists (`requiresLogin`, true for `UNAUTHORIZED` and `SESSION_EXPIRED`) and the profile form already acts on it; these paths dropped the code.

The failure presentation now checks `isAppError(error) && requiresLogin(error)` and shows the profile form's copy, `Session expired. Please sign in.`, on the paths the issue names, plus the collection-creation and inline-image paths approved as a follow-up expansion:

- `src/hooks/usePost/usePost.tsx` `showCommitErrorToast` - post, reply and repost (one shared `showSessionExpiredToast` helper).
- `src/hooks/usePost/usePost.tsx` `edit` - the edit path has its own catch.
- `src/hooks/useEditCollection/useEditCollection.ts` `submit` - the collection edit path.
- `src/hooks/useCreateCollection/useCreateCollection.ts` `submit` - collection creation.
- `src/hooks/useInlineImageUpload/useInlineImageUpload.ts` `runUpload` - inline images in article editing.
- `src/core/controllers/post/post.ts` `commitCreateCollection` and `commitEditCollection` - the cover-upload catch no longer flattens `UNAUTHORIZED`/`SESSION_EXPIRED` into `Validation/INVALID_INPUT`, so the auth classification survives to the presentation layer. Size-limit, non-auth and `FORBIDDEN` behaviour is unchanged, and no other controller code was touched.

Unchanged: the 507 storage-quota warning, the image-size-limit copy, non-auth errors, and auth errors that do not require a sign-in (`FORBIDDEN`). Nothing here signs the user out, retries the write or re-establishes the session. On failure the draft, the attachments, the picked cover, the typed form values and the open dialog are left as they were, and the inline-image rejection keeps its tag so the global handler stays quiet (one toast, not two).

From the issue, reported in the 1.9.0 staging pass: "@piotr-iohk hit it in the composer ... typed a post, pressed Post, got `Error | Could not create post. Try again.`, the draft stayed and nothing published", and on prod "@SpontaneousOverthrow reports the same ... where 'relogin helped'". The same write-dead session showed `Unexpected error occurred.` on a collection title edit with the dialog staying open.

## Verification

[...]

PR #2584, opened 2026-09-18
Release 1.10.0

### LINKED ISSUE #2523: Feed stutters, jumps and cannot be scrolled past: pagination chains over a filtered region without progress

secondl1ght:
## Summary

- Release candidate for 1.10.0, cut from `dev`.
- Cut at `4adf112d` (`test(e2e): count the composer limit in utf-16 units like the app (#2583)`), the first `dev` head since #2564 whose e2e run can pass.
- `master` was merged into the branch right after the cut (`chore: merge master into release-1.10.0`, history only, no file changes). Same step as the 1.9.0 release (#2469): the `master`/`dev` merge-base is still at 1.8.0 because the v1.9.0 back-merge (#2560) was squashed, so a 1.10.0 bump alone would conflict with `master`'s 1.9.0 bump. With `master` as an ancestor the merge into `master` is clean.
- 33 commits over v1.9.0: 22 fixes, 2 features, 6 chores, 1 docs, 1 e2e spec fix, 1 squashed back-merge. Notable: #2552 stream pagination from Nexus cursors (closes #2523, #1569), #2486 local-first tag cache and viewport TTL (ADR 0020), #2474 viewer-aware follow state, #2358 undo for reposts and collection shares, #2577 sign-in prompt on expired session, #2579 visual feed skeleton, #2583 e2e counter assertion.
- The two commits on top of `dev` (`099c300c` merge of `master`, `8d52943e` version bump) are GPG-signed from the release owner's machine. The file diff against `dev` is the version bump only.

## Callouts

- No `DB_VERSION` bump and no Dexie index-map change.
- New `src/config` limits: `STREAM_LOAD_MAX_RAW_SCAN` and `TIMELINE_MAX_UNPRODUCTIVE_AUTO_LOADS` in `feed.ts` (#2552, a manual **Load more** after three unproductive auto-loads), tag cache constants in `tags.ts` (#2486).
- Adds ADR 0020 (#2486).
- VRT baselines were regenerated on `dev` by #2358 via the CI workflow; nothing regenerated here.
- VRT and e2e do not run on PRs. They run on push to `dev`; the runs for `4adf112d` are the ones that cover this candidate, since the branch adds no code changes on top of it: [VRT run](https://github.com/pubky/pubky-app/actions/runs/35378336024) (green), [e2e run](https://github.com/pubky/pubky-app/actions/runs/35378336332) (green).
- The red **Tests** run on `099c300c` was cancelled by the workflow's `cancel-in-progress` concurrency when the bump commit was pushed a minute later; the gatekeeper job reports cancelled shards as a failure. The run on `8d52943e` is green.

## Merge

- Merged with **Create a merge commit** as `58f83e13` (parents `e688c9e0`, `8d52943e`).
- Tagged `v1.10.0` (signed, annotated) on `58f83e13`; Docker Hub build succeeded: [run 35383995172](https://github.com/pubky/pubky-app/actions/runs/35383995172) (63 minutes, `linux/arm64` and `linux/amd64`, tags `v1.10.0` and `latest`).
- GitHub Release published: https://github.com/pubky/pubky-app/releases/tag/v1.10.0. The doc called this optional; #2586 makes it an explicit step.
- Back-merge #2585 landed on `dev` as `75b3dee1` through the merge queue, squashed (single parent), by decision of the release owner; the squash question is deferred. `master` and `dev` trees are identical at 1.10.0, but `master` keeps 9 history-only commits `dev` lacks, so the 1.11.0 release branch needs a `chore: merge master into release-1.11.0` commit before its version bump.

## Checklist (docs/release.md)

- [x] `dev` has the promised work
- [x] Branch `release-1.10.0` created from up-to-date `dev` and pushed
- [x] Draft PR opened: `release-1.10.0` → `master`
- [x] PR checks green (tests, build, code quality) on `8d52943e`
- [x] VRT green (`dev` run on `4adf112d`)
- [x] e2e green (`dev` run on `4adf112d`)
- [x] `package.json` (and `package-lock.json`) version bumped on the release branch and pushed (`8d52943e`, signed)
- [x] PR marked ready for review
- [x] Merged with **Create a merge commit** (`58f83e13`)
- [x] Annotated tag `v1.10.0` created on `master` and pushed (signed)
- [x] Docker Hub image build for the tag succeeded
- [x] GitHub Release `v1.10.0` published from the tag with generated notes
- [x] PR `master` → `dev` opened (#2585); only file change was the version bump; merged (squashed via the queue, trees identical)
- [x] Release branch deleted (by GitHub on merge)


talosmachina (commented):
**No findings.** Oversized release PR: `release-1.10.0` -> `master` for v1.10.0, cut from `dev` at `4adf112d` with `master` merged in and the version bumped. Reviewed `8d52943e`, oversized tier: the release deltas and every security-path file in the diff were read; the 435-file `dev` payload underneath was not re-read, each of its commits having been reviewed on `dev` already.

<details><summary>What I checked, and 4 candidates I ruled out</summary>

**Read in full**: `src/core/controllers/auth/auth.ts` (changed regions), `src/providers/ServiceWorkerRegistrationProvider/ServiceWorkerRegistrationProvider.tsx`, `src/app/layout.tsx`, `next.config.ts`, the four changed workflow files.

**Verified**
- Release delta: `git diff origin/dev..origin/release-1.10.0` is `package.json` / `package-lock.json` only, `1.9.0` -> `1.10.0` in both the root `version` and `packages[""]`, `lockfileVersion` untouched.
- No migration: `src/config/database.ts` (where `DB_VERSION` lives) and `src/core/database/**` are absent from the diff, so neither the Dexie version nor an index map moved. The "no `DB_VERSION` bump" callout holds.
- Merge safety: `origin/master` is an ancestor of the branch, and `099c300c` (`chore: merge master into release-1.10.0`) introduces no tree change against its dev-side parent. Master-only changes since the merge-base are the `1.9.0` version lines, so a merge-commit merge brings the release tree and drops nothing already on `master`.
- No dependency or cross-repo contract change: only the `version` lines differ in `package.json`, so no `pubky-app-specs` shape moved in this PR.
- Gate jobs on `8d52943e` green: 5 vitest shards, Check Code Quality, NextJS build, preview image build (only the preview image push still pending at review time).
- Workflow bumps resolve: `actions/upload-artifact` v7, `actions/download-artifact` v8, `cypress-io/github-action` v7, `peter-evans/find-comment` v4, `google-github-actions/setup-gcloud` v3 all exist as releases.
- Auth: `logout` is bounded by a 5s race with local cleanup guaranteed, and `captureViewerSession()` aborts hydration/notification writes after a logout or account swap; both have new cases in `auth.test.ts`.
- `next.config.ts` `register: false` is backed by `ServiceWorkerRegistrationProvider`, mounted in `layout.tsx`, which calls `window.serwist.register()` with the rejection handled.

**Ruled out**
- *A missed `DB_VERSION`/index-map change*: proved by the diff file list, neither the config value nor the database module is touched.
- *Service worker never registers after `register: false`*: the provider is mounted in `layout.tsx` and registers off the same `window.serwist` global.
- *Artifact-action majors pointing at tags that do not exist*: checked against GitHub releases, all five resolve.
- *The `master` merge silently reverting `master`'s content*: `master` is an ancestor of the branch and the merge commit carries no tree change.

</details>

Merge confidence: 3/5, the oversized tier's ceiling: the 435-file `dev` payload was not re-read here, so this cannot go above the cap however clean the release deltas look.

infin1t3 (approved):

PR #2587, opened 2026-09-18
feat(auth): add pubky passport sign-in and sign-up

Orlandohub:
## Summary

Adds Pubky Passport ("Continue with Google") as a sign-up / sign-in method next to Pubky Ring. Passport is a hosted signer for the same `pubkyauth://` cookie flow Ring scans: the app opens `<passportUrl>/authorize#d=<encoded auth URL>` in a popup and keeps polling the HTTP relay; only the SDK `Session` authenticates, popup messages and callbacks are UI signals. Passport creates the identity and signs it up through Homegate for new Google accounts, so Passport users skip the fair-access step.

Surfaces (Figma "Pubky SHADCN" 41492-353324 / 354691, 41492-353120 / 353161, 41492-354461 / 357059):

- `/sign-in`: two-card desktop layout ("Sovereign & Secure" QR + "Quick & Easy") and labelled sections on mobile.
- New `/onboarding/join` step 1 ("Manage your own keys" vs "Continue with Google"); Join / "New here?" entry points route there when Passport is enabled.
- Landing page per design 41492-354461 / 357059: navbar = Learn · Explore · Sign in (desktop), hero = "Join now" + "Continue with Google", Learn / Explore as a mobile-only row under the hero, social links moved from the navbar to the hero footer beside the Synonym / Tether endorsement; footer copy per design.
- `/passport/return`: same-origin fallback callback page that relays the outcome to the opener and never signs anyone in.

Gating: new optional `PUBKY_RUNTIME_PASSPORT_URL` (no default in deployed mode, staging default in local dev). Passport requires HTTPS callbacks, so the buttons only render once `usePassportEligibility` resolves `enabled` after mount (HTTPS page); plain `npm run dev` hides them, redirects `/onboarding/join` to `/onboarding/human` and warns once. Use `npm run dev:https` locally. Production stays disabled until a real Google sign-up/sign-in is verified against the intended Passport + Homegate deployment.

Lifecycle guarantees in `usePassportAuth`:

- Re-entrancy guard; onboarding secrets reset at attempt start (abandoned browser keys never survive into a Passport identity).
- Popup opened synchronously; listener registered before navigation; outcome messages require exact Passport origin **and** popup source, type, version, outcome and message id; callback messages require app origin, popup source and the unpredictable attempt id.
- A validated `success` outcome is recorded, not treated as auth: popup closure after it does not cancel; only `error`, `cancel` or closure without success does.
- Wall-clock `PASSPORT_ATTEMPT_TIMEOUT_MS` bounds authorization only; it is cleared the moment the session is accepted so slow `initializeAuthenticatedSession` can never be timed out or misreported. `isPending` spans initialization; `onAttemptSettled` fires exactly once with `session | failed`.
- Stale attempts never initialise a session, cancel a newer flow or touch UI state.

Method switching on `/sign-in`: `AuthController.wrapAuthFlow` now takes ownership synchronously before `clearDatabase()` and rejects superseded starts as canceled (previously a slower Ring start could resume and cancel Passport, and superseded starts returned a dead URL). `useAuthUrl` has per-request guards: a cancelled current request expires the QR ("Click to reload"), an old request's cancellation is ignored. Ring actions are disabled while Passport is pending and the Ring request is regenerated exactly once on an explicit `failed` settle.

Out of scope (per plan review): `DialogSignIn` mobile sheet redesign, "Continue with Bitkit", the disabled "Continue with Apple" placeholder, grant-flow migration, direct-navigation (non-popup) mode.

## Verification

- [x] `npm run lint` and `npm run typecheck` (and `npm run format:check`)
- [x] Targeted: `npm test -- src/core/controllers/auth src/core/services/homeserver src/core/application/auth src/hooks/useAuthUrl src/hooks/usePassportAuth src/hooks/usePassportEligibility src/hooks/useJoinRoute src/hooks/useMobileAuth src/libs/passport src/libs/runtime-config src/config src/components/organisms/SignIn src/components/organisms/PassportMethodCard src/components/organisms/DialogSignIn src/components/organisms/Header src/components/molecules/ContinueWithPassport src/components/molecules/ActionButtons src/components/molecules/Home src/components/molecules/HeaderHome src/components/templates/Onboarding/Join src/components/templates/Auth src/components/templates/Public src/app/routes.test.ts src/providers/RouteGuardProvider` (28 files, 601 tests)
- [x] `npm run build` (new routes `/onboarding/join`, `/passport/return`)
- [ ] `npm run test:vrt` — not run locally (browsers not installed); baselines are CI-owned, see callout
- [ ] Manual on `npm run dev:https` against staging: fresh Google account → `/onboarding/profile`; existing → `/home`; persisted-session reload; cancel; popup closed before approval; approve and let Passport self-close before the relay answers; blocked popups; abandoned browser-key signup then Google. iOS Safari / Android Chrome popup-as-tab and callback fallback. Needs a real Google account; still to do.

## Callouts

- [ ] Touches a surface covered by a VRT (landing, onboarding header/step config) → baseline regeneration should be dispatched via **VRT Update Baselines** (no local baselines committed)
- New runtime config value `PUBKY_RUNTIME_PASSPORT_URL` documented in `docs/environment.md` and `.env.example`; staging deploy should set `https://passport.staging.pubky.app`. Please confirm with the Passport team that this staging host is the intended one and is registered with the staging Google OAuth client (found by probing; not documented in the Passport repo).
- `AuthController` imports the pure `createCanceledError` factory from `@/services/homeserver/error.utils` (no IO) so superseded starts reject with the one existing canceled-error shape.
- Mobile `/sign-in` keeps the existing primary "Authorize with Pubky Ring" button under the new "Sovereign & Secure" label; the Figma secondary Ring / Bitkit buttons are not adopted here (Bitkit out of scope).
- Passport users skip the Backup step (keys live in Google Drive via Passport), same as the Ring path today.

- [x] Invalidates a Cypress spec → flagged for QA here (not edited): `cypress/e2e/onboarding.cy.ts:168` asserts `[data-cy="explore-btn"]` is visible; the hero Explore button is now mobile-only (`md:hidden`), on desktop Explore lives in the navbar as `#header-explore-btn`.

## Screenshots

To be added from the `dev:https` run (desktop two-card sign-in, mobile sections, Join step, landing hero).


src/hooks/useJoinRoute/useJoinRoute.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * First sign-up step for every "Join" / "New here?" entry point.
+ *
+ * `/onboarding/join` (own keys vs Continue with Google) only exists when Pubky Passport can run on
+ * this page; otherwise, and while eligibility is still resolving, the classic fair-access step is
+ * the entry so a click never lands on a page that would immediately redirect.
+ */
+export function useJoinRoute(): ONBOARDING_ROUTES {
+  const eligibility = usePassportEligibility();
+  return eligibility === 'enabled' ? ONBOARDING_ROUTES.JOIN : ONBOARDING_ROUTES.HUMAN;
```
talosmachina, line 15: **<sub><sub>P1</sub></sub> Route the `/sign-in` header's "New here?" button through `useJoinRoute`**

This hook is documented as the first sign-up step for every "Join" / "New here?" entry point, but one of them still hardcodes the fair-access step: on `/sign-in` the header button pushes `ONBOARDING_ROUTES.HUMAN` (`src/components/molecules/HeaderButtonSignIn/HeaderButtonSignIn.tsx:15`). `/sign-in` is neither a core-explore nor a dynamic-public route, so `Header` falls through to `HeaderHome` (`src/components/organisms/Header/Header.tsx:72`), which renders that button with `isSignInPage` true. On a Passport-enabled deploy the click lands on `/onboarding/human`, which offers no "Continue with Google" and never redirects to `/onboarding/join`, so the user is pushed into the Ring-only sign-up this feature exists to skip.

Verified by: traced the pathname through `Header.tsx:72` and `HeaderHome.tsx:122`, confirmed `/onboarding/human` renders no Passport option in `Human.tsx`, and `HeaderButtonSignIn.test.tsx:86` asserts `push('/onboarding/human')` for that same pathname.

talosmachina (commented):
**1 finding** (1 P1). Adds "Continue with Google" (Pubky Passport) sign-in and sign-up across the landing hero, `/sign-in` and `/onboarding/join`. Reviewed `ef4de21`, oversized tier: full review on the Passport/auth paths only.

**Read in full**: `usePassportAuth.tsx`, `libs/passport/passport.ts`, `PassportReturn.tsx`, `config/passport.ts`, `usePassportEligibility.ts`, `useJoinRoute.ts`, `useAuthUrl.tsx`, `useMobileAuth.tsx`, `SignIn.tsx`, `Join.tsx`, `AuthController.wrapAuthFlow`, `HomeserverService.generatePassportAuthUrl`, `runtime-config.*`, `HeaderHome.tsx`, `Header.tsx`, `ActionButtons.tsx`

**Cross-repo contract checked** against `pubky-passport`: `authorizationOutcomeHandoff.ts`, `pubkyAuthUrls.ts`, `pubkyAuthRequestParser.ts`, `docs/integration.md`

**CI**: green (5 shards, quality, build, preview)

**Not read** (oversized tier): the ~55 test, snapshot, doc and image files, and the onboarding step routing outside the diff

**Ruled out**
- *A valid approval cancelled by the popup-close poll*: Passport posts the outcome and waits for the ack before closing, and `handleOutcome` records success before the closure check.
- *Message contract mismatch*: message/ack types and version, the once-encoded fragment, and the https same-origin callback rule all match the Passport implementation, including the cookie flow it accepts.
- *`wrapAuthFlow` ownership*: taken synchronously before `clearDatabase()`, and the only caller (`useAuthUrl`) handles the canceled error, so a superseded start cannot wipe the newer URL.
- *Timeout scope*: the 5 minute bound covers authorization only, which is intended; the approval loop checks `canceled` between polls, so `flow.free()` never runs mid-call.
- *Controller importing a service factory*: `createCanceledError` is pure, no IO.
- *Ring refetch on settle cancelling a user-started Ring flow*: `isRingLocked` keeps the Ring buttons disabled while an attempt is pending.
- *Cold reader*: the finding stood; it confirmed `Human.tsx` renders no Passport option and that nothing redirects `human` to `join`.

</details>

Merge confidence: 3/5, the oversized tier ceiling (the rest of the diff was not read) and the one P1.

src/hooks/useJoinRoute/useJoinRoute.ts
Orlandohub: Good catch, fixed in 34e8d3621: `HeaderButtonSignIn` now resolves the "New here?" destination through `useJoinRoute`, so on a Passport-enabled deploy it lands on `/onboarding/join` (fair-access step otherwise / while eligibility is pending). Added a test for the Join case. Checked the remaining `ONBOARDING_ROUTES.HUMAN` pushes: only `Scan.tsx` (missing invite code, go back) remains, which is not a sign-up entry point.

Orlandohub:
@talosmachina I've addressed your PR comments, can you review again and update your confidence score?

talosmachina (commented):
**Re-review of 34e8d36 after ef4de21.** 1 earlier finding: fixed. No new findings. Reviewed the 4-file delta (2 commits) fully.

<details><summary>What I checked in the delta, and what I ruled out</summary>

**Delta read in full**: `HeaderButtonSignIn.tsx`, `HeaderButtonSignIn.test.tsx`, `preview-deploy-service-file.yml`, `.github/workflows/build.yml`
**CI**: green (5 shards, quality, build, preview; only the image push still pending)

**Earlier finding (P1, `/sign-in` "New here?" button)**: fixed in `34e8d3621`. The button now resolves its destination through `useJoinRoute`, which returns `/onboarding/join` only on a resolved `enabled` eligibility and the fair-access step otherwise, so a Passport-enabled deploy no longer pushes users into the Ring-only step. The added test asserts the `/onboarding/join` push with eligibility enabled, and the existing `/onboarding/human` assertion still passes with the default.

**Ruled out**
- *Click before eligibility resolves*: `usePassportEligibility` returns `pending` for the first client render and `useJoinRoute` maps that to the fair-access step by design, so no click can land on a page that would immediately redirect.
- *Preview config enabling Passport*: `PUBKY_RUNTIME_PASSPORT_URL` points at the same staging signer as the staging homeserver already configured for previews, and previews are HTTPS, so the callback origin rule holds. Unset means the feature stays off.
- *Smoke-test env var*: the added export only mirrors the preview runtime config the boot test asserts against.

**Still not read** (oversized tier, unchanged since the first pass): the ~55 test, snapshot, doc and image files, and onboarding step routing outside the diff.

</details>

Merge confidence: 3/5, the oversized-tier ceiling on the unread remainder; the P1 is fixed and the delta is clean.

Orlandohub:
landing-hero

sign-in

sign-in-mobile

onboarding-join

passport-return

https://github.com/user-attachments/assets/7d91ea31-b82a-492f-9a8f-df01b87035c2

gillohner:
https://github.com/user-attachments/assets/f7c54d7a-5ee1-44cf-91f0-f1910d47911c

Should I be able to still use file upload/seedphrase when logging in with passport?

All of this will soon be moved into Passport anyway though, so this is fine to me.

talosmachina:
@gillohner Yes, both stay, and they are not really part of the signer.

`SignInPage` renders `SignInNavigation` unconditionally, so the recovery-phrase and encrypted-file (`.pkarr` upload) entries remain on `/sign-in` whether Passport is enabled or not. They call the app's own local restore paths (`AuthController.loginWithMnemonic` / `loginWithEncryptedFile`), while Passport is just another way to obtain a homeserver session for the same `pubkyauth://` request.

Only one auth flow owns the controller at a time, so the ordering is safe: completing a restore goes through `completeAuthenticatedSession`, which calls `cancelActiveAuthFlow()`, so a pending Passport attempt is cancelled (its relay wait rejects and `runAttempt` returns without initialising a session); starting Passport cancels a pending Ring flow the same way. No path leaves two sessions.

One deliberate-decision-able detail: the Ring buttons are disabled while a Passport attempt is pending (`isRingLocked`), but `SignInNavigation` has no pending awareness, so the restore dialogs can be opened mid-attempt. It resolves safely (the restore supersedes the attempt), but if you want the UI to communicate that, the dialogs would have to respect the same pending state.

secondl1ght (commented):
Review of `8904030`. Nits were removed; only items with a functional, architectural or coverage consequence remain.

## 1. Correctness

### 1.1 High: a settled Passport attempt regenerates the Ring URL through `wrapAuthFlow`, which runs `clearDatabase()` under a sign-in that is already bootstrapping

`SignIn.tsx:80-86` calls `fetchUrl()` on every `result: 'failed'` settle. `usePassportAuth` reports `failed` for three different things:

- **Superseded** (`usePassportAuth.tsx:212-215`): the user opens the recovery-phrase / encrypted-file dialog while the Passport popup is open (the dialogs have no pending awareness, as noted in the thread). `completeAuthenticatedSession` calls `cancelActiveAuthFlow()` (`auth.ts:271`) before `userIsSignedUp` / `hydrateMeImAlive`, the Passport `awaitApproval` rejects with the canceled error, and the settle callback fires `fetchUrl()` → `AuthController.getAuthUrl()` → `wrapAuthFlow` (`auth.ts:349-372`): `cancelModerationFollow()`, `await clearDatabase()` on the tables the bootstrap is writing, `useSettingsStore.reset()` mid settings sync, `useMigrationStore.reset()`, then an orphan relay poll for a user who is now signed in. `isLoading` is `false` at that point, so the guard on line 84 does not stop it.
- **Popup blocked** (`usePassportAuth.tsx:250-253`): no auth flow was started, yet the settle cancels the live Ring relay channel, clears the database again and flashes the QR back to "Generating…". `useOnboardingStore.reset()` on line 245 has also already run.
- **Real failure**: the only case where regenerating is right.

Fix at the source rather than in the callback: give `PassportAttemptResult` distinct values (`superseded`, `popup-blocked`, `failed`) and only refetch on `failed`, or stop cancelling the Ring flow when Passport starts and let the first session win. Add a test for the superseded case (restore completes while a Passport attempt is pending → no `getAuthUrl` call, no `clearDatabase`).

### 1.2 Medium: the outcome listener is torn down before the late `success` message, and the fallback page boots the full app shell

`usePassportAuth.tsx:190-191` switches to `initializing` and removes `onMessage` as soon as the relay delivers the session. Passport posts its `success` outcome right after its own relay request resolves, so the opener's 100 ms poll can win that race; the message then gets no ack and Passport falls back to navigating the popup to `/passport/return`. That route renders under `DatabaseProvider` and `RouteGuardProvider` (`app/layout.tsx:52-53`), whose restore effect (`RouteGuardProvider.tsx:57-71`) sees the `sessionExport` the opener just persisted and calls `restorePersistedSession()`; any failure there runs `cleanupLocalState()` (`auth.ts:393-430`: cookies, `clearDatabase()`, persisted auth reset) on the origin the opener is bootstrapping into. The page's own effect posts and closes first, so this only bites while the window stays open (no opener, opener closed, or `window.close()` refused), but it is the same hazard `docs/pwa.md` rejects for the offline page.

Fix: keep the listener and ack alive through initialization (stop only the timeout and the popup poll at the `initializing` transition), and keep the callback page out of the session-restore providers.

## 2. Architecture, quality and tests

**Checks**: `npm run lint` pass · `npm run typecheck` pass (local, at `8904030`) · CI green (14 check runs).

### 2.1 High · `no-service-from-controller` · `src/core/controllers/auth/auth.ts:29`

`AuthController` now has a runtime import from the services layer (`createCanceledError` from `@/services/homeserver/error.utils`, thrown at `auth.ts:358` and `:372`). Every other controller → service import in the repo is `import type`; this is the first executable dependency (`docs/architecture.md` Allowed Dependencies, ADR-0004), and the thrown value is a raw `Error` rather than an `Err.*` result. The sentinel is deliberately not an `AppError` so it stays out of Sentry, which is fine, but it should live in a layer-neutral module. The same PR also defines `isAuthFlowCanceledError` twice (`useAuthUrl.tsx:15`, `usePassportAuth.tsx:63`).

Fix: move `AUTH_FLOW_CANCELED_ERROR_NAME`, `createCanceledError` and a single exported `isAuthFlowCanceledError` into `src/libs/error/` (or re-expose through `@/application/auth/auth`), and import from there in `error.utils.ts`, the controller and both hooks.

### 2.2 Medium · error-handling · `src/components/templates/Auth/PassportReturn/PassportReturn.tsx:37-38`

The fallback page posts to the opener and calls `window.close()` in the same synchronous block. `window.closed` flips inside `close()`, and the opener's 500 ms poll (`usePassportAuth.tsx:159-161`) fails the attempt with `popup-closed` whenever it observes `closed` before the posted message task runs; that calls `cancelAuthFlow()` on an authorization the user completed. The primary path is protected by Passport's ack handshake (`:144`); the fallback path has no ordering guarantee.

Fix: require two consecutive closed observations (or re-check on a `setTimeout(…, 0)`) before failing, or have the return page wait for an opener ack before closing.

### 2.3 Medium · tests · `src/hooks/usePassportAuth/usePassportAuth.test.tsx:283-330`

The callback-page path is covered only for `outcome: 'cancel'`. The branch this PR adds for the fallback, `parsePassportReturnMessage` → `handleOutcome('success')` → `successRecorded` → popup closure tolerated → session, has no case, and neither does the ack-throw branch at `usePassportAuth.tsx:142-147`. A regression there ends every fallback sign-in in "The Passport window was closed before sign-in finished." with tests green.

Fix: (1) deliver a same-origin return message with `outcome: 'success'`, set `popup.closed = true`, advance past the poll, assert no toast and no settle, resolve the approval, assert `result: 'session'`; (2) make `popup.postMessage` throw once, deliver a Passport-origin `cancel`, assert the failed settle and its toast.

### 2.4 Medium · routing · `src/config/passport.ts:20`

`PASSPORT_RETURN_PATH = '/passport/return'` re-declares `AUTH_ROUTES.PASSPORT_RETURN` (`routes.ts:22`). The x-callback destinations handed to Passport (`libs/passport/passport.ts:56`) come from the config copy while `PUBLIC_ROUTES` and the page use the enum; a diverged value sends the popup to a non-public 404 that `passport.test.ts` cannot catch because it asserts against the same constant.

Fix: `PASSPORT_RETURN_PATH = AUTH_ROUTES.PASSPORT_RETURN`, or use the enum member directly in `buildPassportReturnUrl`.

### Callouts

- **VRT**: `src/test/vrt/landing/Landing.vrt.test.tsx` renders `<Header />` + `<Landing />`, and this PR restructures the hero (`ActionButtons`), the navbar (`HeaderHome`) and the footer (`Home.tsx`). PR CI does not run VRT, so dispatch **VRT Update Baselines** for that file (the PR's callout box is still unchecked).
- **CI workflow**: `.github/workflows/build.yml:50` gains `PUBKY_RUNTIME_PASSPORT_URL` in a feature PR; AGENTS.md reserves workflow edits for CI tasks and `passportUrl` is optional, so the smoke test does not need it. Drop it or split it out.
- **Runtime config**: new optional `PUBKY_RUNTIME_PASSPORT_URL` is documented in `docs/environment.md` and `.env.example`; no `DB_VERSION`, dependency or ADR change.
- **Cypress**: `cypress/e2e/onboarding.cy.ts:168` invalidation is already flagged for QA in the description.

## 3. Security

No high- or medium-severity finding met the reporting bar. The threat model stated in the code (only the SDK `Session` from the relay authenticates) holds on reading, not just in comments:

- Both message parsers require the exact origin **and** `event.source === popup` (`libs/passport/passport.ts:87`, `:108`), the attempt id (`:113`) and an outcome allow-list (`:114`); `handleOutcome` can only set `successRecorded` or fail the attempt, and `settle('session')` (`usePassportAuth.tsx:209`) is reachable only after `awaitApproval` and `initializeAuthenticatedSession` with the homeserver guard.
- The `/passport/return` query is never rendered, never a navigation sink and never stored; it is only re-emitted as a message subject to the same checks. The page posts with `targetOrigin = window.location.origin` and carries no credential.
- Popup name and attempt id use `crypto.randomUUID()`; the popup is cross-origin, so the `#d=` fragment (relay secret) is unreadable to anyone holding the handle, is kept out of server logs and `Referer`, and is excluded from the `handleError` context.
- `popup.location.replace` only ever receives `new URL('/authorize', origin)` output; `passportUrl` is `z.url().optional()` with no deployed-mode default; no `dangerouslySetInnerHTML`, `innerHTML` or `eval` in the diff.
- `wrapAuthFlow` taking ownership synchronously and rejecting superseded starts removes the stale-QR footgun rather than adding one.

## Verdict

**Recommended before merge.** 1.1 is a user-data path (a completed restore sign-in has its database cleared and its settings reset by a callback the user never triggered), and 2.1 is a high-severity project rule. 1.2, 2.2, 2.3 and 2.4 are worth fixing in the same round since they all live in the same handful of files.

---

aldertnl (dismissed):
It looks 90% there, I contacted you on Slack for some minor UI fixes, details and other tweaks.

Orlandohub:
Thanks for the thorough pass — all six findings were valid and are addressed in a1b95b4fb (plus the build.yml revert).

**1.1 (High)** `PassportAttemptResult` is now `session | failed | superseded | popup-blocked`. The hook settles `superseded` on the canceled sentinel (e.g. a recovery-phrase sign-in completing while the popup is open) and `popup-blocked` before any flow starts (the onboarding reset now also runs only after the popup opened). `SignIn` regenerates the Ring request on `failed` alone. Tests: superseded → `superseded`, no init, no toast (hook); `SignIn` gets superseded / popup-blocked → no `fetchUrl`.

**1.2 (Medium)** Split the teardown: at the `initializing` transition only the timeout and the popup-closed poll stop; the `message` listener stays up and still acks a late Passport `success` (test covers ack during init, and that a late `cancel` / closure cannot fail the attempt). The fallback page is no longer a Next route: it is the static `public/passport/return.html`, following the same rule `docs/pwa.md` states for `offline.html`, so the popup never boots `DatabaseProvider` / `RouteGuardProvider` on the opener's origin. Its inline script is exercised in `src/libs/passport/passportReturnPage.test.ts` against the config constants and `parsePassportReturnMessage`, so drift fails the test.

**2.1 (High)** `AUTH_FLOW_CANCELED_ERROR_NAME`, `createCanceledError` and a single `isAuthFlowCanceledError` now live in `src/libs/error/auth-flow-canceled.ts`; the service re-exports them, and the controller and both hooks import from libs. No runtime controller → service import remains.

**2.2 (Medium)** The poll now needs `PASSPORT_POPUP_CLOSED_CONFIRMATIONS` (2) consecutive closed observations before failing, and the static page defers `window.close()` a tick after posting. Test: callback-page success delivered after the first closed observation → no failure, session accepted.

**2.3 (Medium)** Added the two cases: same-origin return `success` → popup closed → session, and `popup.postMessage` throwing once → Passport `cancel` still settles `failed` with its toast.

**2.4 (Medium)** `AUTH_ROUTES.PASSPORT_RETURN` and the `PUBLIC_ROUTES` entry are gone; `PASSPORT_RETURN_PATH` in `src/config/passport.ts` is the only source and points at the static file.

**Callouts**: `PUBKY_RUNTIME_PASSPORT_URL` dropped from `build.yml` (the preview manifest keeps it, previews need it for QA); **VRT Update Baselines** dispatched for this branch: https://github.com/pubky/pubky-app/actions/runs/35747830321.

talosmachina:
@Orlandohub Thanks. My pass covered 34e8d3621, so a1b95b4fb and everything after it are unreviewed by me: the six findings read as addressed and unverified, and I can't move the score on that basis alone. Ask for another pass when you want one and I will take it.

Orlandohub:
@talosmachina please give it another look at the most recent updates and approve if you agree with all

Orlandohub:
landing

sign-in

onboarding-join

passport-return

https://github.com/user-attachments/assets/404ba965-f06a-471d-8acf-acdb280a6710

talosmachina (commented):
**Re-review of `a9e5fc92` after `34e8d362`.** 1 earlier finding (P1): fixed, and it still holds at this head. 0 new findings. Full tier on the auth delta. @Orlandohub I do not approve pull requests; this is my re-review, a human approval is still needed before merge.

<details><summary>What I checked, and what I ruled out</summary>

**The delta since my last review** (the PR's own commits; the two `dev` merges are not this PR's change): `a1b95b4f` auth attempt-lifecycle hardening, `400f8419` landing/sign-in copy, `89040308` landing hero and navbar, plus the two VRT regeneration commits.

**Read in full**: `usePassportAuth.tsx`, `usePassportAuth.types.ts`, `SignIn.tsx`, `SignIn.test.tsx`, `usePassportAuth.test.tsx`, `libs/passport/passport.ts` + `.types.ts` + `passportReturnPage.test.ts`, `config/passport.ts`, `public/passport/return.html`, `libs/error/auth-flow-canceled.ts`, `controllers/auth/auth.ts`, `services/homeserver/error.utils.ts`, `useAuthUrl.tsx`, `Header.constants.ts`, `Header.tsx`, `ActionButtons.tsx`, `HeaderHome.tsx`, `Home.tsx`.
**CI**: green (5 vitest shards, quality, build, preview).

**Earlier finding (P1, `/sign-in` header "New here?")**: fixed in `34e8d3621` and still fixed here. `HeaderButtonSignIn` resolves the destination through `useJoinRoute`, which returns `/onboarding/join` only on a resolved `enabled` eligibility and the fair-access step otherwise; the test asserting the Join push is still in the suite.

**Ruled out**
- *Pop-up blocked no longer resets the onboarding store*: deliberate and correct, no Passport identity is created when `window.open` returns null, and the test asserts a stored `secretKey` survives.
- *`superseded` no longer regenerates the Ring QR on `/sign-in`*: a superseding flow (e.g. a completed recovery-phrase restore) owns the session, so a new Ring request would run `clearDatabase()` under it. Only `failed` regenerates, which is what the QR needs.
- *Message listener kept alive through `initializing`*: the timeout and pop-up poll are stopped first, so nothing can fail a running `initializeAuthenticatedSession`; the listener only acks a late Passport `success`, and `handleOutcome` no-ops outside the authorizing phase.
- *Two consecutive closed observations*: the callback page posts and closes on the next tick, so the message task always lands before the second 500 ms observation; a genuinely closed pop-up still fails after ~1 s.
- *Static callback page*: the relay is validated at the opener by `parsePassportReturnMessage` (origin, source pop-up, attempt id) and never authenticates anyone; the path is single-sourced in `config/passport.ts` into `buildPassportCallbacks`, and no reference to the removed `AUTH_ROUTES.PASSPORT_RETURN` remains. Being a static file it bypasses the route guard entirely (there is no middleware; `PUBLIC_ROUTES` is client-side), so no session restore or guard redirect runs in the pop-up. The page's hand-mirrored constants are pinned by a test that executes the real file.
- *`createCanceledError` moved to `libs/error`*: still reachable through the `error.utils.ts` re-export used by `homeserver.utils.ts`, and the hooks/controller now read the layer-neutral lib. No layer violation, `libs` is shared by both.
- *`build.yml` drops `PUBKY_RUNTIME_PASSPORT_URL`*: production stays off by design until a real Google sign-in is verified; previews still set it in `preview-deploy-service-file.yml`.
- *Service worker*: same-origin navigations are `NetworkOnly` with the offline page only on a failed fetch, so the callback page is never served stale.
- *`'/sign-in': step 1`*: `currentStep` is consumed by `HeaderOnboarding` only, and `/sign-in` is not an onboarding route; the entry only feeds the title for signed-out users.

**Still not read** (oversized tier, 87 files / ~4,700 lines of PR): the test, snapshot, doc and image files outside the delta, and onboarding routing outside the diff.

</details>

Merge confidence: 3/5, the oversized-tier ceiling on the remainder I have not read.

secondl1ght (approved):
Re-reviewed `de2b4b0` (fix commit `a1b95b4` plus the follow-ups). All six findings are fixed in code, each with a test, and nothing unrelated slipped in.

**Findings**

- **1.1** `PassportAttemptResult` now distinguishes `superseded` and `popup-blocked` from `failed`; `runAttempt` settles `superseded` on the canceled sentinel, the blocked-popup path settles `popup-blocked` before any flow or store is touched, and `SignIn` regenerates the Ring request on `failed` alone. Covered by the hook's superseded test and the two new `SignIn` tests.
- **1.2** Teardown is split: only the timeout and the popup poll stop at `initializing`, the `message` listener stays up and acks a late Passport `success`. The fallback page is the static `public/passport/return.html` (no app shell, no session restore in the popup), and `passportReturnPage.test.ts` runs the real file's script against the config constants and `parsePassportReturnMessage`, so drift fails the test.
- **2.1** `AUTH_FLOW_CANCELED_ERROR_NAME`, `createCanceledError` and one `isAuthFlowCanceledError` live in `src/libs/error/auth-flow-canceled.ts`; the controller and both hooks import from there. No runtime controller → service import remains.
- **2.2** The poll requires `PASSPORT_POPUP_CLOSED_CONFIRMATIONS` consecutive closed observations and the static page defers `window.close()` a tick after posting. Covered by the new close-then-message test.
- **2.3** Callback-page `success` and the ack-throw branch are both tested.
- **2.4** `AUTH_ROUTES.PASSPORT_RETURN` and the `PUBLIC_ROUTES` entry are gone; `PASSPORT_RETURN_PATH` is the single source and the page test resolves the file through it.

**Checks**: `npm run lint` pass · `npm run typecheck` pass · targeted suites 34 files / 655 tests pass (local, at `de2b4b0`) · CI green · both dev merges clean (no conflict hunks, no unexplained files) · `build.yml` reverted · landing and onboarding VRT baselines regenerated by the workflow.

**Follow-ups that landed alongside and are fine**: the design copy changes, the `/sign-in` header title (same shape as the `/logout` entry; `HeaderHome` still renders "New here?"), and the `.gitignore` entry.

One thing code review cannot confirm: the Passport-side acceptance of the `.html` callback path and the staging Google flow are only proven by the manual `dev:https` run still unchecked in the description. Same origin and scheme as before, so no reason to expect a problem, but worth ticking off before enabling in production.

---

secondl1ght (approved):

PR #2588, opened 2026-09-18
fix(og-metadata): ignore placeholder tag values from shell pages

talosmachina:
## Summary

YouTube Music answers some `music.youtube.com/playlist` requests (album playlists, radio `RDAMVM…`) with a client-rendered shell instead of the page. That shell marks every og value it emits as the literal string `undefined`:

```
<link rel="canonical" href="undefined">
<meta property="og:site_name" content="undefined">
<meta property="og:url" content="undefined">
<meta property="og:title" content="undefined">
<meta property="og:description" content="undefined">
<meta property="og:type" content="undefined">
<title>undefined</title>
```

`extractMetadata` took the first non-empty `og:title` and forwarded it, so the link preview card rendered `undefined` as the title (`GenericPreview` only guards null/empty).

Fix: `extractMetadata` treats `undefined` and `null` (trimmed, case-insensitive) as a missing value.

- Title: the candidate list (`og:title` then `<title>`) skips placeholders, so a real `<title>` still wins over a placeholder `og:title`.
- Image: a placeholder `og:image` is no longer resolved against the page URL. It resolved to `https://music.youtube.com/undefined`, which passes the DNS check and would render as a broken thumbnail.

No logging added: `extractMetadata` is a pure util with no logger, and the service layer already logs its own fallbacks. A placeholder now degrades to the same URL-only card as a page with no title, which is the intended outcome.

Reported from a Slack thread on a post whose card read `undefined`; PR requested by @secondl1ght.

## Verification

- [x] Targeted: `npx vitest run --project unit src/core/services/nextjs/og-metadata/ src/core/application/og-metadata src/libs/html` -> 83 passed and 46 passed
- [ ] `npm run build` (not run: no route or config change)
- [ ] `npm run test:vrt` (no VRT surface changed)

Evidence against the real shell page, captured from this box with the service's own request headers (97 KB, `og:title` = `undefined`):

- before the change: `extractMetadata(url, html)` -> `{"title":"undefined","image":null,"type":"website"}`
- after the change: title `null`, so the card falls back to URL-only

Production `GET /api/og-metadata?url=…` returns `{"title":"undefined",…}` for `playlist?list=OLAK5uy_…`, `playlist?list=RDAMVM…` and the `&si=` variant, while `watch?v=` URLs return real metadata. Deterministic across repeated and cache-busted requests.

Five new unit tests in `og-metadata.utils.test.ts` cover placeholder `og:title` plus `<title>`, case and whitespace, `null`, a placeholder `og:title` with a real `<title>` fallback, and a placeholder `og:image`. They fail on the base revision (`5 failed | 22 passed`) and pass after the change.

## Callouts

None: no VRT surface, ADR, `DB_VERSION`, `src/config/*` limit, convention or Cypress spec is touched.


src/core/services/nextjs/og-metadata/og-metadata.utils.ts
```diff
@@ -10,6 +10,22 @@ import { isHttpProtocol, normalizeImageUrl } from '../nextjs.utils';
...
+ * response (og:title, og:url, og:site_name, og:type), which used to become the card title.
+ */
+const PLACEHOLDER_VALUES = new Set(['undefined', 'null']);
+
+/**
+ * True when a tag value carries no metadata: missing, blank, or a placeholder.
+ * Callers treat a placeholder exactly like a missing value.
+ */
+function isPlaceholderValue(value: string | null | undefined): boolean {
+  const normalized = value?.trim().toLowerCase();
```
secondl1ght: **Worth fixing before merge: the case-fold discards legitimate titles.**

`String(undefined)`, `String(null)`, template interpolation and `JSON.stringify` only ever produce lowercase `undefined` / `null`, so `.toLowerCase()` adds no true positives. It does add false positives: a track, band or repo literally titled `Null`, `NULL` or `Undefined` now loses its title, and when `<title>` carries the same string (common on SPAs) the card degrades to URL-only.

Suggest matching the exact lowercase literals after `trim()` only, and turning the "regardless of case" test into the negative case (`UnDefined` → kept).

---

src/core/services/nextjs/og-metadata/og-metadata.utils.ts
```diff
@@ -30,17 +46,18 @@ export function detectMediaType(url: string, response: Response): TOgMetadataRes
...
  */
 export async function extractMetadata(url: string, html: string): Promise<TOgMetadataResult> {
-  // Extract title (og:title → <title> fallback)
-  const ogTitle = extractFromHtml(html, OG_PATTERNS.TITLE);
-  const titleTag = html.match(OG_PATTERNS.TITLE_TAG)?.[1] || null;
-  const rawTitle = ogTitle || titleTag;
+  // Extract title (og:title → <title> fallback), skipping placeholder values so a shell page
+  // cannot become a card title and the fallback tag still gets its chance.
+  const titleCandidates = [extractFromHtml(html, OG_PATTERNS.TITLE), html.match(OG_PATTERNS.TITLE_TAG)?.[1] || null];
+  const rawTitle = titleCandidates.find((candidate) => !isPlaceholderValue(candidate)) ?? null;
```
secondl1ght: **Low impact, cheap to fix here: the placeholder check runs on the raw, entity-encoded value.**

`isPlaceholderValue` sees the capture before `decodeHtmlEntities` (line 53) and before the final `trim()` (line 64), so:

- `<title>undefined&nbsp;</title>` → not in the set → decodes to `undefined ` → trims to `undefined` → the card title is `undefined` again.
- `<meta property="og:title" content="&nbsp;"><title>Real Title</title>` → `&nbsp;` wins the `find()`, decodes to a space, trims to `''` → the real `<title>` never gets its turn (the UI hides an empty title, so this one is invisible today).

Decode and trim each candidate first, then apply the placeholder test, and map an empty result to `null`.

---

src/core/services/nextjs/og-metadata/og-metadata.utils.ts
```diff
@@ -30,17 +46,18 @@ export function detectMediaType(url: string, response: Response): TOgMetadataRes
  * Extracts OG metadata from HTML, normalizes image URLs, and applies truncation.
  */
 export async function extractMetadata(url: string, html: string): Promise<TOgMetadataResult> {
-  // Extract title (og:title → <title> fallback)
-  const ogTitle = extractFromHtml(html, OG_PATTERNS.TITLE);
-  const titleTag = html.match(OG_PATTERNS.TITLE_TAG)?.[1] || null;
-  const rawTitle = ogTitle || titleTag;
+  // Extract title (og:title → <title> fallback), skipping placeholder values so a shell page
+  // cannot become a card title and the fallback tag still gets its chance.
+  const titleCandidates = [extractFromHtml(html, OG_PATTERNS.TITLE), html.match(OG_PATTERNS.TITLE_TAG)?.[1] || null];
```
secondl1ght: **Pre-existing, follow-up: only the first `og:title` occurrence is ever a candidate.**

`extractFromHtml` returns the first regex match per pattern and stops, so a placeholder `og:title` followed by a real one of the same form still hides the real one; only the `<title>` fallback can rescue the card. Same for `og:image` (placeholder meta, then the real CDN URL → `image: null`).

This PR does not make it worse, so not a blocker. The right place for the fix is `extractFromHtml` in `src/libs/html/html.ts`: skip blank/placeholder captures and keep scanning (next occurrence, next pattern), so every OG tag benefits and this file does not need per-call-site filtering.

---

src/core/services/nextjs/og-metadata/og-metadata.utils.ts
```diff
@@ -30,17 +46,18 @@ export function detectMediaType(url: string, response: Response): TOgMetadataRes
...
   const title = rawTitle ? decodeHtmlEntities(rawTitle) : null;
 
   // Extract og:image
   const image = extractFromHtml(html, OG_PATTERNS.IMAGE);
 
-  // Normalize and validate image URL
-  const normalizedImage = image ? await normalizeImageUrl(image, url) : null;
+  // Normalize and validate image URL. A placeholder is not a path: resolving it against the page
+  // URL would turn "undefined" into https://<host>/undefined.
+  const normalizedImage = image && !isPlaceholderValue(image) ? await normalizeImageUrl(image, url) : null;
```
secondl1ght: **Pre-existing, but on the line this PR rewrote: `og:image` is not entity-decoded before normalization.**

The title goes through `decodeHtmlEntities`; the image does not. A correctly escaped attribute such as `content="https://cdn.example.com/a.jpg?w=600&amp;h=400"` reaches `new URL()` with the literal `&amp;`, so the query becomes `w=600` plus `amp;h=400` and the CDN either 400s or ignores the size.

Not this PR's bug. Either decode `image` before the placeholder check and normalization here, or file it separately.

---

src/core/services/nextjs/og-metadata/og-metadata.utils.ts
```diff
@@ -30,17 +46,18 @@ export function detectMediaType(url: string, response: Response): TOgMetadataRes
...
 export async function extractMetadata(url: string, html: string): Promise<TOgMetadataResult> {
-  // Extract title (og:title → <title> fallback)
-  const ogTitle = extractFromHtml(html, OG_PATTERNS.TITLE);
-  const titleTag = html.match(OG_PATTERNS.TITLE_TAG)?.[1] || null;
-  const rawTitle = ogTitle || titleTag;
+  // Extract title (og:title → <title> fallback), skipping placeholder values so a shell page
+  // cannot become a card title and the fallback tag still gets its chance.
+  const titleCandidates = [extractFromHtml(html, OG_PATTERNS.TITLE), html.match(OG_PATTERNS.TITLE_TAG)?.[1] || null];
+  const rawTitle = titleCandidates.find((candidate) => !isPlaceholderValue(candidate)) ?? null;
   const title = rawTitle ? decodeHtmlEntities(rawTitle) : null;
```
secondl1ght: **Nit: the placeholder path degrades silently.**

The PR body says the service layer logs its own fallbacks, but that holds only for results that go through `fallback()`. `extractMetadata` returns a normal result here, so a shell page and a page with no OG tags are indistinguishable in the logs. When someone reports "card has no title for URL X", there is nothing to tune the heuristic against.

Optional: return a reason from `extractMetadata` (or detect it in the service) and `Logger.warn` it like the other outcomes listed under _OG metadata enrichment_ in `docs/sentry.md`.

---

src/core/services/nextjs/og-metadata/og-metadata.utils.ts
```diff
@@ -10,6 +10,22 @@ import { isHttpProtocol, normalizeImageUrl } from '../nextjs.utils';
...
+ * page. YouTube Music returns the literal string "undefined" in every og tag it emits for such a
+ * response (og:title, og:url, og:site_name, og:type), which used to become the card title.
+ */
+const PLACEHOLDER_VALUES = new Set(['undefined', 'null']);
+
+/**
+ * True when a tag value carries no metadata: missing, blank, or a placeholder.
+ * Callers treat a placeholder exactly like a missing value.
+ */
+function isPlaceholderValue(value: string | null | undefined): boolean {
```
secondl1ght: **Nit: the name undersells what this checks.**

`isPlaceholderValue` is also true for `null`, `undefined` and blank strings, and because it is not a type guard the image call site needs a redundant `image &&` (line 60) and the title candidate keeps a dead `|| null` (line 51). One helper such as `pickMetadataValue(value: string | null): string | null` (null for blank or placeholder), applied to each title candidate and to the image, states the rule once and restores the original `image ? await normalizeImageUrl(...) : null` shape.

---

src/core/services/nextjs/og-metadata/og-metadata.utils.test.ts
```diff
@@ -97,6 +97,38 @@ describe('extractMetadata', () => {
     expect(result.title).toBeNull();
   });
 
+  it('should ignore the placeholder "undefined" that client-rendered shells serve for og:title and <title>', async () => {
+    const html = '<html><head><meta property="og:title" content="undefined" /><title>undefined</title></head></html>';
+    const result = await extractMetadata('https://music.youtube.com/playlist?list=OLAK5uy_x', html);
+    expect(result.title).toBeNull();
+  });
+
+  it('should ignore a placeholder og:title regardless of case and surrounding whitespace', async () => {
```
secondl1ght: **Nit, but the one I would still add: every new test asserts a drop, none asserts a keep.**

Nothing pins whole-string matching, so a later change from `Set.has` to `includes()` or a regex would drop titles like "Undefined Behavior in C" or "Null Island" with this suite green. If the case-fold goes (see line 25), this test becomes the natural negative case: `  UnDefined  ` → `UnDefined`.

Also worth pinning: a real `og:title` beating a placeholder `<title>`, `og:image="null"` and a whitespace-only `og:image` asserting `normalizeImageUrl` is not called, and the quiet behaviour change where `<title>   </title>` now yields `null` instead of `''`.

---

secondl1ght (commented):
Reviewed at `7013765`. The fix is at the right layer (server side, ahead of the CDN and client caches), the SSRF guards in this high-risk path are untouched, and the targeted tests, lint and Prettier pass locally.

**One thing worth fixing before merge**

- **The case-fold is a regression** (`og-metadata.utils.ts:25`). JavaScript only ever stringifies to lowercase `undefined` / `null`, so `.toLowerCase()` catches nothing extra and drops real titles such as "Null" or "NULL". One-line fix plus flipping the "regardless of case" test into the negative case.

**Real but low impact, cheap to include**

- The placeholder check runs before entity decoding and trim (`:52`), so `undefined&nbsp;` still becomes a title and an entity-only `og:title` shadows a real `<title>`.

**Pre-existing, not this PR's to fix**

- Only the first `og:title` / `og:image` occurrence is ever a candidate (`:51`); the right fix is in `extractFromHtml`.
- `og:image` is not entity-decoded before normalization (`:60`), so `&amp;` corrupts image query strings. Probably the most impactful bug in the file, but it predates this change.

**Nits**

- Silent degradation with no log reason (`:53`), helper naming and the redundant guards it forces (`:24`), and no keep-side tests (`test.ts:106`). The negative test is the one nit I would still add, since it is what stops the case-fold from coming back.
- The second commit's subject is 78 characters, over the 72 limit in `docs/commit-message.md`. Disappears under a squash merge.

---

secondl1ght:
@talosmachina check out the review

talosmachina:
Thanks @secondl1ght, all fair. Fix in fbdbe4e63.

**Fixed**

- **The case-fold** (`og-metadata.utils.ts:25`). Right: `String(undefined)` / `String(null)` are already lowercase, so `.toLowerCase()` added no true positives and dropped real titles. The match is now exact after decoding and trimming, and the "regardless of case" test is the negative case: `  Null  ` -> `Null`.
- **Placeholder test before decoding and trim** (`:52`). Every candidate now goes through `pickMetadataValue`, which decodes entities and trims first, so `undefined&nbsp;` is a placeholder and a `&nbsp;`-only `og:title` no longer wins the `find()` over a real `<title>`. Both cases are tests.
- **`og:image` not entity-decoded** (`:60`). The one helper decodes uniformly, so the image reaches `normalizeImageUrl` decoded and `?w=600&amp;h=400` keeps its query string. Same line count, test added.
- **Helper naming** (`:24`). `isPlaceholderValue` is now `pickMetadataValue` returning `string | null`, the redundant `image &&` is gone, and `title` no longer decodes or trims twice.

**Not fixed**

- **Only the first tag occurrence is a candidate** (`:51`). Agreed, and out of this PR's scope as you say: the fix belongs in `extractFromHtml`, skipping blank or placeholder captures and continuing the scan. Say the word and I will file it as a follow-up.
- **No log reason on the placeholder path** (`:53`). Leaving it: the util has no logger, and the service already logs its own fallbacks. The placeholder now lands on the same URL-only result as a page with no title at all.

**One thing the review could not see**: the branch merged `dev` while it was up, which brought a `hasOgMetadata` helper whose job is deciding whether a 200 is worth a retry with a crawler identity. A shell page that emits only `og:title="undefined"` counted as real metadata, so it now uses the same `pickMetadataValue`; the placeholder-only case is a test.

Verified on fbdbe4e63: the five new assertions fail on the commit you reviewed (`5 failed | 32 passed`) and pass here; the `og-metadata` suite is 118 passed, eslint, prettier and `tsc --noEmit` clean. The commit-subject length nit disappears under a squash merge.

src/core/services/nextjs/og-metadata/og-metadata.utils.ts
secondl1ght: Filed as #2639 so it does not get lost; no need to open one yourself.

---

talosmachina: Thanks, that covers it and I will not file a duplicate. #2639 names both later-occurrence cases (`og:title` and `og:image`) and lifting the predicate next to `extractFromHtml` is the right home for it, so nothing to add.

secondl1ght (approved):
Approving at `fbdbe4e`.

Re-checked the whole PR against the new base rather than just the fix commit: the diff is confined to the two og-metadata files, the `dev` merge touched nothing outside them, and the `hasOgMetadata` change is consistent with `isPreviewUsable` (a placeholder-only shell now earns the crawler-identity retry instead of a card titled `undefined`). `pickMetadataValue` decodes and trims before the exact-match check, and `decodeHtmlEntities` runs ahead of `normalizeImageUrl`, so the SSRF checks see the decoded value. The 194 tests across the og-metadata service, application and `libs/html` suites pass locally at this head, and CI is green.

The remaining first-occurrence gap in `extractFromHtml` is tracked in #2639. Thanks for the quick turnaround.

---

PR #2593, opened 2026-09-19
fix: APP_VERSION link in settings page

catch-21:
## Summary

Release URL (src/config/externalLinks.ts): GitHub tags are v1.10.0, while the package version is 1.10.0. The link now prefixes v, so settings will point at https://github.com/pubky/pubky-app/releases/tag/v1.10.0 and still show v1.10.0.

## Verification

- [x] `npm test` (full) or targeted: `npm test -- <files>`
- [x] `npm run build` (route- or config-wide changes)
- [ ] `npm run test:vrt` (only when a VRT surface changed)

## Callouts

- [ ] Touches a surface covered by a VRT → baseline regeneration dispatched via **VRT Update Baselines** (no local baselines committed)
- [ ] Adds or changes an ADR (`docs/adr/`)
- [ ] Bumps `DB_VERSION` or changes a Dexie index map (local database is recreated for every user)
- [ ] Changes a `src/config/*` limit (product-wide policy)
- [ ] Changes a convention → `docs/`, `AGENTS.md` and `.greptile/config.json` updated
- [ ] Invalidates a Cypress spec → flagged for QA here (not edited)


src/config/externalLinks.ts
```diff
@@ -19,8 +19,14 @@ import { APP_VERSION } from './app';
  * URLs can be overridden using environment variables for different environments.
  */
 
-// App release URL (version injected from package.json via next.config.ts)
-export const APP_RELEASE_URL = `https://github.com/pubky/pubky-app/releases/tag/${APP_VERSION}`;
+// App release URL (version injected from package.json via next.config.ts).
+// GitHub release tags are `vX.Y.Z`; APP_VERSION is the bare package version.
```
talosmachina, line 23: **<sub><sub>P1</sub></sub> The shipped app does not use the package version, so this fix never applies there**

`APP_VERSION` in the released image is not the package version: `.github/workflows/release-docker.yml:44` and `.github/workflows/build-docker-image.yml:139` both pass `NEXT_PUBLIC_APP_VERSION=${{ github.sha }}`, and `next.config.ts:8` bakes that value into the bundle. On pubky.app the label still reads `v<40-hex>` and the link becomes `.../releases/tag/v7ab8606f5a810131fae59ad993bdcaa93814116d`, still a 404. The `v` prefix only lands in builds that skip the SHA arg (`npm run build`), which the release and preview images do not. For a working link in production the displayed version and the tag have to come from the same source.

Verified by: the bundle pubky.app currently serves (`/_next/static/chunks/main-app-46335d2e79d3f22e.js`) contains `NEXT_PUBLIC_APP_VERSION:"7ab8606f5a810131fae59ad993bdcaa93814116d"`; the two Docker workflows above pass `github.sha` as a build arg and the Dockerfile declares that ARG before `RUN npm run build`.

talosmachina (commented):
**1 finding** (1 P1). Prefixes `APP_VERSION` with `v` when building the settings-page release URL. Reviewed `77358c9`, full tier.

**Read in full**: `src/config/externalLinks.ts`, `src/config/app.ts`, `next.config.ts`, `SettingsInfo.tsx`, `Dockerfile`, `.github/workflows/release-docker.yml`, `.github/workflows/build-docker-image.yml`
**Call sites traced**: `APP_RELEASE_URL`, `buildAppReleaseUrl`, `SettingsInfo`
**CI**: green

**Ruled out**
- *Older tags without the prefix* (`1.7.0` and below carry no `v`): not reachable, `APP_VERSION` is always the current version and the shipped tags are `vX.Y.Z` from `v1.8.0` on.
- *Double-prefix guard*: `startsWith('v')` handles an already-prefixed value correctly, and both branches have tests.
- *Missing tests*: the new function's two branches are covered and the component test/snapshot were updated.
- *Cold reader*: asked to disprove the P1 from the repo and the live bundle; it could not, and confirmed no deploy path injects the package version at runtime.

</details>

Merge confidence: 3/5, the P1 above: the change does not reach the deployed app.

src/config/externalLinks.ts
talosmachina: **Correction: the mechanism is right, the source I named is not, and the finding is superseded on staging.**

`release-docker.yml` and `build-docker-image.yml` pass the *pubky-app* SHA and build the published and preview images. The deploy that served pubky.app is pubky-stack's `pubky-app-cicd.yml` (called by `pubky-stack-cicd.yml` on push to `staging`/`production`), which set `APP_VERSION: ${{ github.sha }}` on the job env (pubky-stack #259, `c1f8d78`, 2026-07-02) and wrote `NEXT_PUBLIC_APP_VERSION=$APP_VERSION` into `pubky-app.env`. That is why the live bundle carried `7ab8606`, a pubky-stack commit (the production head), not a pubky-app one.

Superseded: pubky-stack `2cf0d7e` (2026-09-19, on `staging`, not yet on `production`) now reads `APP_VERSION` from pubky-app's `package.json` at build time, so the next deploy built from `staging` bakes `1.10.0` and the `v` prefix in this PR is exactly what produces `releases/tag/v1.10.0`. The change here is the right one for that path.

What still holds: the image on pubky.app today (production head `7ab8606`) shows `v<40 hex>` and a 404 until a deploy carries `2cf0d7e`, and the app's own Docker images still pass `github.sha`, so their settings page keeps the SHA link.

